PR #62, opened 2026-06-30
feat: windows support -- installer and cross-platform tui transport

arelchan:
## Change description

Full Windows support for the one-line install + TUI, so Windows gets a command
equivalent to the macOS/Linux `curl -fsSL .../install.sh | sh`:

```powershell
iex (irm https://raven.evermind.ai/install.ps1)
```

Two layers:

### 1. Install layer
- **`install.ps1`** (new): PowerShell counterpart of `install.sh` — ensures uv,
  a private Node >= 22 (downloads `node-vXX-win-<arch>.zip` to
  `%RAVEN_HOME%\runtime`, SHA256-verified, self-tested), then `uv tool install`
  raven (latest release wheel for remote mode; editable for a local checkout).
- **`find_node()`**: Windows-aware — globs `%RAVEN_HOME%\runtime\node-*\node.exe`
  (Windows zip has node.exe at the package root, no `bin/`). POSIX behavior
  unchanged (`os.name == "nt"` guard).
- **`install.sh`**: unsupported-OS message now prints the real PowerShell command.

### 2. Cross-platform TUI transport (the part that lets `raven tui` actually run on Windows)
The Python<->Node RPC transport used an AF_UNIX unix-domain socket with
`os.dup`'d fds + `connect_read/write_pipe` — all unsupported on Windows. Replaced with:
- **TCP loopback** `127.0.0.1:<ephemeral>` + a **256-bit random auth token**.
  The accepted socket *object* is handed to `RpcServer` (no socket-fd dup) and
  wired via `loop.connect_accepted_socket` (works on both selector and proactor
  loops). The parent validates the token as the first line **before any
  dispatch** — loopback is reachable by any local process, so the token
  restores the trust boundary the old `0600` socket file provided.
- Node client (`ui-tui/src/rpc/client.ts`) parses `RAVEN_RPC_SOCKET` as
  `host:port`, connects over TCP, sends `RAVEN_RPC_TOKEN` first.
- The legacy pipe/fd path in `RpcServer` is retained for `--check`/unit tests.

## Verification
- 306 tui/rpc tests pass; real Node bundle exercised under a PTY (TCP + token +
  `system.hello`); wrong/missing token rejected (exit 3), correct token accepted;
  listener bound to `127.0.0.1` only.
- **Zero regressions**: full-suite failure set is identical to a pristine `main`
  baseline (all pre-existing environmental/ordering failures).

## ⚠️ Not verified — needs real Windows QA
Developed/verified on macOS. The cross-platform logic is exercised here and the
Windows-specific paths were statically reviewed, but **`install.ps1` execution
and the TUI on a real Windows console have NOT been run on Windows**. Please
smoke on a Windows box before relying on it: run `install.ps1`, then `raven`
(confirm Ink TUI renders + interacts).

## Type of change
- [x] New feature

### Development
- [x] Lint rules pass locally (ruff)
- [ ] Application changes tested thoroughly (Windows paths NOT run on Windows — see above)
- [x] Automated tests covering modified code pass (CI)
### Security
- [x] Security impact considered (256-bit token, loopback-only, validated pre-dispatch; Node SHA256-verified)
### Code review
- [x] Descriptive title and context provided

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Raven treats those problems as the product, not edge cases.
 **macOS / Linux:**
 
 ```bash
-curl -fsSL http://raven.evermind.ai/install.sh | bash
+curl -fsSL https://raven.evermind.ai/install.sh | sh
 ```
 
 **Windows (PowerShell):**
--- a/README.md
+++ b/README.md
@@ -96,16 +96,26 @@ Raven treats those problems as the product, not edge cases.
 
 ## Quick Install
 
+**macOS / Linux:**
+
 ```bash
 curl -fsSL http://raven.evermind.ai/install.sh | bash
 ```
 
-After installation, reload your shell and run the setup wizard:
+**Windows (PowerShell):**
+
+```powershell
+iex (irm https://raven.evermind.ai/install.ps1)
+```
+
+After installation, start a fresh shell so `PATH` refreshes, then run the setup wizard:
 
 ```bash
-source ~/.bashrc    # or: source ~/.zshrc
+source ~/.bashrc    # macOS / Linux (or: source ~/.zshrc)
 raven onboard
 ```
+
+On Windows, just open a new PowerShell window, then run `raven onboard`.
 
 Raven supports OpenRouter, OpenAI, Anthropic, Gemini, DeepSeek, GitHub Copilot,
 OpenAI Codex OAuth, and custom OpenAI-compatible endpoints.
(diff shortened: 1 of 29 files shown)


arelchan:
Superseded by #65, which cherry-picked this branch's TCP-loopback TUI transport and all the Windows compatibility fixes onto the latest main and merged there (squash 255e2cb). Closing to avoid a duplicate/conflicting merge.

PR #67, opened 2026-07-01
fix(agent): route subagent results to the originating conversation

arelchan:
## Change description

**Bug**: in the TUI, a spawned subagent's result never showed up. On completion
the subagent re-injects a `SUBAGENT`-origin turn, and its `conversation` was
hardcoded to `f"{channel}:{chat_id}"` -> `"tui:default"`. But the TUI front-end
subscribes on the **minted session key** (`tui:<sid>`, from `session.create`).
The reply was emitted to `"tui:default"`, which has no subscriber, so it was
silently dropped and the user never saw the subagent result.

**Fix**: carry the originating turn's `conversation` through the spawn path and
use it as the re-injected turn's conversation:
- `_SpawnOrigin` gains a `conversation` field; `session_key` prefers it.
- `AgentLoop._set_tool_context` passes the turn's `conversation` (`msg_session_key`) to the spawn tool.
- `SubagentManager.spawn` records it in `origin`; `_announce_result` uses `origin.get("conversation") or f"{channel}:{chat_id}"`.

Only `conversation` changes. `source` is untouched, so **CLI** (renders directly,
ignores conversation) and **gateway** (routes by `source.chat_id`) behavior is
unchanged. Side benefit: for CLI / Telegram-topic paths the re-injection now
lands on the originating turn's lane instead of a divergent `channel:chat_id`
lane, which is more correct.

## Verification
- New regression tests (`tests/test_subagent_manager.py`): minted-key path
  asserts the announce conversation == the minted key (would fail without the
  fix), plus a channel fallback test for non-TUI paths.
- Three independent reviewers (functional / three-path-no-regression / cross-platform+e2e) all PASS:
  - functional: relevant suites 211 passed, stable across 2 runs, ruff clean;
  - no-regression: `source` unchanged -> CLI/gateway delivery identical; new params optional/back-compat;
  - cross-platform: diff touches **zero** platform/transport/install code; real Node PTY transport probe still PASS; end-to-end proof that the reply now reaches the minted-key subscriber (and previously reached 0 subscribers on `tui:default`).
- Pre-existing unrelated failure `test_find_node_discovers_windows_private_runtime` is NOT touched by this PR.

## Type of change
- [x] Bug fix

### Development
- [x] Lint rules pass locally (ruff)
- [x] Application changes tested (unit + e2e proof; see above)
- [x] Automated tests covering modified code pass
### Security
- [x] Security impact considered (subagent result still fenced via wrap_untrusted; no change to trust boundary)
### Code review
- [x] Descriptive title and context provided

### DIFF
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -1069,12 +1069,25 @@ class AgentLoop:
                 self._mcp_stack = None
             raise
 
-    def _set_tool_context(self, channel: str, chat_id: str, message_id: str | None = None) -> None:
+    def _set_tool_context(
+        self,
+        channel: str,
+        chat_id: str,
+        message_id: str | None = None,
+        conversation: str | None = None,
+    ) -> None:
         """Update context for all tools that need routing info."""
         for name in ("message", "spawn", "cron"):
             if tool := self.tools.get(name):
                 if hasattr(tool, "set_context"):
-                    tool.set_context(channel, chat_id, *([message_id] if name == "message" else []))
+                    if name == "message":
+                        tool.set_context(channel, chat_id, message_id)
+                    elif name == "spawn":
+                        # spawn carries the conversation so subagent results
+                        # re-inject on the originating turn's key (TUI mints one).
+                        tool.set_context(channel, chat_id, conversation)
+                    else:
+                        tool.set_context(channel, chat_id)
 
     @staticmethod
     def _strip_think(text: str | None) -> str | None:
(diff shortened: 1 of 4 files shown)


JasonJarvan:
Thanks for this — it independently confirms the same root cause (a spawned subagent's result was re-injected on `tui:default` instead of the originating session's conversation key, so the live TUI never received it).

We already have a fix on branch `feat/subagent_tui_visibility` (EVE59-2) — implemented, reviewed, and verified live. It takes a slightly stronger form, and since both branches edit the same three functions (`_set_tool_context`, `SubagentManager.spawn`/`_announce_result`, `_SpawnOrigin`/`SpawnTool.set_context`), landing both isn't viable. Closing this in favor of ours:

- Enforcement: rather than an optional `conversation` field that falls back to `f"{channel}:{chat_id}"`, we make `_SpawnOrigin.session_key` a **required** field and remove the reconstruction path entirely — so a future caller can't silently regress to the old behavior.
- We also corrected the now-misleading `_announce_result` debug log (it still printed `channel:chat_id` rather than the conversation the reply was actually routed on); this PR leaves that line unchanged.

Closing as duplicate. The corroboration on the root cause was genuinely useful — thank you.

Issue #113, opened 2026-07-13
Missing 'orjson' dependency causes APIConnectionError on fresh install


jonah791:
## Description

On a fresh `uv tool install raven` on Windows (PowerShell 5.1), running `raven agent -m "hi"` fails with:

```
Error calling LLM: litellm.APIConnectionError: APIConnectionError:
OpenAIException - No module named 'orjson'
```

The TUI also shows `error: turn_failed (code=-32099)` when trying to send a message.

## Root Cause

`orjson` is not listed in `pyproject.toml` dependencies, but LiteLLM (via the OpenAI SDK) attempts to import it at runtime. The wheel installs `litellm` without `orjson` as a transitive dependency, making it unavailable on first install.

## Temporary Workaround

```bash
# Install orjson into the raven tool environment
uv tool run --from raven python -m pip install orjson
```

## Environment

- Raven v0.1.3
- Windows 11, PowerShell 5.1
- Installed via `uv tool install raven[channels]`
- Python 3.12.10

## Suggested Fix

Add `orjson>=3.9.0` to the `[project.dependencies]` list in `pyproject.toml`, or ensure the wheel's transitive dependency resolution includes it.

PR #117, opened 2026-07-13
build: pin fresh-install deps to the release lockfile

arelchan:
## Change description

Make a fresh `curl | sh` / `uv tool install` install resolve the exact dependency
set that CI locked and tested, instead of whatever upstream published latest.

**Why:** the wheel only carries loose ranges (e.g. `litellm>=1.82.1,<2.0.0`), and
neither the wheel nor `uv tool install` consults `uv.lock`. So a fresh install
resolves the newest in-range versions at install time. That is exactly how the
missing-`orjson` breakage (issue #113) reached users: litellm silently moved
`1.85 -> 1.92`, and 1.92's request path needs `orjson`, which the older locked
1.85 did not.

**What:**
- `release.yml` exports the locked set from `uv.lock` (`uv export --frozen
  --no-emit-project --no-dev --all-extras --no-hashes`) and ships it as a
  `requirements.txt` release asset alongside the wheel.
- `install.sh` (remote mode) resolves that asset next to the wheel and installs
  with `--constraint`, pinning transitive deps to the tested versions.
- Backward compatible: releases without the asset install unpinned (current
  behavior). `RAVEN_CONSTRAINT_URL` overrides for a custom wheel.

**Verified locally** with a freshly built wheel:

| install | resolved litellm |
| --- | --- |
| unpinned (current) | 1.92.0 (drifts to latest) |
| `--constraint requirements.txt` (this PR) | 1.85.0 (matches uv.lock) |

Note: this is the distribution-side guard. It complements #116 (declare orjson)
and does not pin litellm's range in `pyproject.toml` (that stays flexible for
`uv sync` / dev).

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Document
- [x] Others (build / release + install reproducibility)

### Development
- [x] Lint rules pass locally
- [x] Application changes have been tested thoroughly
- [ ] Automated tests covering modified code pass

### Security
- [x] Security impact of change has been considered
- [x] Code follows security best practices and guidelines

### Code review
- [x] Pull request has a descriptive title and context useful to a reviewer

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,6 +57,16 @@ jobs:
           uv build --wheel
           uv build --sdist
 
+      # Export the exact dependency set from uv.lock so install.sh can pin a
+      # fresh install to it (--constraint). The wheel only carries loose ranges,
+      # so without this a curl|sh install resolves whatever upstream published
+      # latest, drifting off the versions CI actually tested.
+      - name: Export locked dependency constraints
+        run: |
+          uv export --frozen --no-emit-project --no-dev --all-extras --no-hashes \
+            --format requirements-txt -o dist/requirements.txt
+          test -s dist/requirements.txt  # fail fast if the export is empty
+
       # 4) Sanity gate: the wheel MUST contain the bundle and MUST NOT carry
       #    node_modules or ui-tui sources (keeps the wheel lean and the smoke honest).
       - name: Verify wheel contents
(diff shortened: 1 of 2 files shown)


arelchan:
Closed as landed by other means, not as stale.

The mechanism this PR adds is on main already, in a more complete form. `release.yml` exports the locked set (`uv export --locked --all-extras --no-hashes --no-emit-project -o dist/raven-constraints.txt`), `install.sh` installs local mode with `-c "$constraints"` and derives the constraints URL from the wheel URL for remote mode. The asset name settled on `raven-constraints.txt` rather than `requirements.txt`, which is the only reason this branch still looks unmerged.

The root cause behind issue #113 was also fixed independently: `orjson>=3.11.6,<4.0` is a direct dependency in `pyproject.toml` now, so a fresh install no longer relies on litellm to drag it in.

Nothing here is lost by closing, and the branch conflicts with main on both files it touches.

PR #163, opened 2026-07-21
feat(routing): add opt-in knn task-level model routing backend

ypflll:
## Summary

Adds an optional task-level KNN model-routing backend alongside the existing EcoClaw router. At the start of each turn the router embeds the task, retrieves the nearest training tasks from a prebuilt memory, and picks the model with the best expected value (reward - lambda_cost * cost). Routing is opt-in and off by default (routing.enabled = false); when disabled the agent loop is unchanged.

Commit by commit:
- config: RoutingConfig gains a backend switch (ecoclaw | knn) plus knn fields (models, memory_path, k, lambda_cost, embedding_endpoint) and safety-gate fields; all additive with defaults.
- providers: PerModelProvider dispatches each call to a per-model endpoint by model name; unknown models fall back to the wrapped provider unchanged.
- routing: KNNModelRouter exposes the same select_model_chain interface as the EcoClaw router; any failure yields (None, []) so the caller keeps its model.
- cli: the gateway builds the router by backend and, for knn, wraps the provider in PerModelProvider.
- routing (gates): leave the default model only with enough evidence - a memory-size floor, a count of similar neighbours (cosine >= min_similarity), and a margin over the default model's score; scoring uses only the similar neighbours. Defaults keep routing safe out of the box.
- test: a real-LLM e2e drives AgentLoop._process_message over two OpenRouter models (self-skips without credentials).

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Commands and results:
- pytest tests/test_knn_router.py tests/test_config_routing.py tests/test_per_model_provider.py tests/test_cli_gateway_commands.py -> 70 passed
- pytest tests/integration/test_knn_routing_real_llm.py -> 1 passed with credentials, skips cleanly without
- ruff check on the changed files -> All checks passed
- Full suite: 4047 passed; the remaining failures are pre-existing environment gaps (sandbox / TUI pty / Windows node runtime / channel SDKs) that reproduce on origin/main without this branch.

Offline routing reproduction against a prebuilt memory (the algorithm makes the live decision; the routed model's outcome is looked up from stored per-model trajectories):
- ScienceWorld: KNN 58.47 vs Pure 27B 57.99 at ~13% lower cost.
- AppWorld: KNN 55.3% pass vs Pure 27b 50.9%.

## Risk

Off by default: with routing.enabled = false, build_model_routing returns the provider unchanged and the loop's routing block is skipped, so behaviour is identical to before. The new provider and router classes are only imported when the knn backend is enabled. Rollback is to disable routing in config or revert the branch.

## Related Issues

N/A


0xKT:
LGTM overall — CI is green and I ran the unit suite locally
(`test_knn_router`, `test_config_routing`, `test_per_model_provider`,
`test_cli_gateway_commands`), all pass.

One non-blocking nit before merge: `raven/routing/knn_router.py` imports `numpy`
at module top level, but numpy isn't a declared direct dependency — it's only
transitively locked (via `everos -> lancedb`), and this is the first shipped
`raven/` source file to import it directly. Low risk since that chain is all-hard,
but it's inconsistent with the repo's own convention: `pyproject.toml` already
pins `tiktoken` directly for exactly this case ("... so the everos extractor keeps
working if upstream ever drops it"). Suggest `uv add numpy` with a similar
one-line comment.

0xKT (approved):

PR #170, opened 2026-07-21
feat(importer): cold-start import from other AI tools with EverOS HTTP backend

Kendrick-Song:
## Change description

Cold-start import: scan conversation history and memory files from other AI tools (starting with Claude Code), feed them into EverOS long-term memory so Raven knows the user from day one.

Key deliverables:

- **Importer framework** (`raven/importer/`): Scanner protocol, ClaudeCodeScanner (JSONL conversations + memory/*.md files), orchestrator with batched store (100 msgs / 30K chars), idempotent state tracker with resume, cancel-flag stop mechanism
- **CLI commands** (`raven import scan/run/status/stop`): interactive platform/tier selection, Rich progress bar, background execution support, cancel via `raven import stop`
- **Onboard integration** (step 6): platform/tier/mode selection, foreground or background execution, wired after EverOS memory config
- **EverOS HTTP-only backend**: removed embedded mode entirely, EverOS runs as an independent server on port 18791 with auto-start/health-check lifecycle; unlocks multi-process concurrency (TUI + gateway + background import)
- **Data alignment fixes**: sender_id, app_id, project_id now omitted from import metadata so EverOS defaults match daily recall queries
- **Architecture cleanup**: domain logic (scan_all, filter_by_tier) moved from CLI to importer layer; concurrent scanner support via asyncio.gather; state path uses config/paths.get_data_dir()

## Type of change
- [x] New feature
- [x] Bug fix
- [ ] Document
- [ ] Others

## Related issues (if there is)

> Closes #104
> Related to #176 (native-Windows EverOS memory: this PR adds the interim
> win32 guard; the upstream EverOS fcntl->portalocker fix is tracked there)

### Development

- [x] Lint rules pass locally
- [x] Application changes have been tested thoroughly
- [x] Automated tests covering modified code pass

### Security

- [x] Security impact of change has been considered
- [x] Code follows security best practices and guidelines

### Code review

- [x] Pull request has a descriptive title and context useful to a reviewer. Screenshots or screencasts are attached as necessary


raven/plugin/memory/everos/backend.py
```diff
@@ -535,31 +270,25 @@ def _make_http_adapter(self) -> _Adapter:
...
             type(self._adapter).__name__,
         )
-        # Embedded real adapter: bring up the in-process everos runtime
-        # (schema + OME engine) so store / recall actually work. HTTP and
-        # no-op adapters need no local everos lifespan.
-        if isinstance(self._adapter, _RealEverosAdapter):
-            await _acquire_embedded_everos(self._logger)
-            self._embedded_started = True
+        if isinstance(self._adapter, _HttpEverosAdapter):
+            from raven.cli._everos_server import ensure_everos_server
```
0xKT: Layering: this is the only plugin-layer module that imports `raven.cli`. It is not a circular-import workaround (`raven/cli/__init__.py` is empty and `_everos_server.py` never imports back into the plugin), so the deferred import does not buy anything structurally -- the dependency direction is just backwards. `ensure_everos_server` is a server-lifecycle helper shared by the backend, onboard, and import; consider moving it to a lower layer (e.g. `raven/memory_engine/` or under this plugin package) so nothing under `raven/plugin/` reaches up into `raven/cli/`.

raven/plugin/memory/everos/backend.py
```diff
@@ -362,8 +115,8 @@ def _jsonify(obj: Any) -> Any:
     return obj
 
 
-# Default timeout — HTTP mode is per-turn, so we keep it tight.
-_DEFAULT_HTTP_TIMEOUT_S: float = 10.0
+# Default timeout — per-turn, so we keep it tight.
+_DEFAULT_HTTP_TIMEOUT_S: float = 360.0
```
0xKT: The comment still says "keep it tight" while the value is now 36x larger (10s -> 360s). This timeout flows into the single `httpx.AsyncClient` used by both cold-start import and the live per-turn store path (`AgentLoop._dispatch_backend_store` -> `backend.store` at `raven/agent/loop/main.py:965`, which is awaited). `raven-plugin.toml` sets no `timeout_s`, so 360s is the effective per-turn default -- a hung EverOS server can block turn completion for up to 360s per httpx phase. Suggest a separate longer timeout for the bulk-import path, keeping the per-turn client tight, and fixing the now-contradictory comment.

raven/cli/tui_commands.py
```diff
@@ -595,101 +595,77 @@ def _agent_loop_factory():
...
-    # this function ran), so this dup2 does not affect the child's terminal.
-    async def _start_memory_backend() -> None:
-        # Bring up the memory backend off the render path: its everos/lancedb
-        # import (~2-3s) + lifespan must not block the handshake / first render.
-        # recall/store degrade to empty until it is ready.
+    # Start the embedded backend before serving so the EverOS runtime is up
+    # and its index lock is held for the session. No-op for http/no-op backends.
+    if agent_loop is not None and agent_loop.backend is not None:
         try:
             await agent_loop.backend.start()
```
0xKT: This awaits `backend.start()` before `serve_task = create_task(server.serve_forever())` and before the handshake wait. On main this was deliberately deferred to a background task created only after the handshake ("must not block the handshake / first render"). Since `start()` now calls `ensure_everos_server()` which can spawn the server and poll for up to 30s, and the Node RPC client has no handshake timeout, a cold start (server not yet up) leaves the terminal unrendered until `start()` returns. Note the retained comment "No-op for http/no-op backends" no longer holds -- HTTP is the only adapter now and its `start()` does real work. Consider restoring the post-handshake background start, or documenting the cold-start block as intentional.

raven/importer/types.py
```diff
@@ -0,0 +1,104 @@
...
+    sender_id: str
+    tool_calls: tuple[dict[str, Any], ...] | None = None
+    tool_call_id: str | None = None
+
+
+@dataclass(frozen=True)
+class ImportSession:
+    """A complete importable unit ready for store()."""
+
+    app_id: str
```
0xKT: `app_id` / `project_id` are set by the scanner (`scanners/claude_code.py:365-366` etc.) but never read: `orchestrator._flush` sends only `metadata={"is_final": ...}`, so the params threaded through `store()` -> `memorize()` -> `/add` + `/flush` body (`backend.py:391-392`) are always None on every runtime path (the live `AgentLoop.store` call passes no metadata). If imports are meant to land in EverOS's `default/default` partition (as the orchestrator test comment states), drop the unused fields and backend params; otherwise wire these into the metadata. Carrying both halves disconnected reads as accidental.

raven/importer/orchestrator.py
```diff
@@ -0,0 +1,240 @@
...
+                        total=total,
+                    )
+                )
+            continue
+
+        logger.info("[{}/{}] importing {}/{}", i + 1, total, platform, key)
+        try:
+            session = await scanner.read(result)
+            await _feed_session(backend, session)
+            state.mark_submitted(platform, key)
```
0xKT, line 113: Resume granularity is per source unit, not per batch: `mark_submitted` runs only after the whole session feeds, and `state.is_submitted` treats a "failed" entry as not-done. A multi-batch conversation (>100 msgs or >30k chars) that fails mid-way re-reads and re-sends already-accepted batches on retry. Whether that duplicates depends on EverOS-side dedup (not visible from this repo). Fine for memory files; worth a per-batch checkpoint or at least a note if large conversations are common.

raven/cli/_everos_server.py
```diff
@@ -0,0 +1,79 @@
...
+    base_url: str = "http://localhost:18791",
+    *,
+    timeout: float = 30.0,
+) -> None:
+    if await asyncio.to_thread(_probe_health, base_url):
+        logger.info("everos server already running at {}", base_url)
+        return
+
+    port = _extract_port(base_url)
+    await asyncio.to_thread(_start_server, port)
```
0xKT: Unlocked probe-then-spawn: concurrent processes (tui + gateway + import) starting while the server is down each pass the health probe and each spawn `everos server start` on the same port; the losers die on port-in-use. The poll loop still converges so correctness holds, but the child exit status is never checked, so this leaves stray failed children plus misleading "started everos server" log lines. A pidfile/lock would make it clean.

raven/importer/state.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def _read_from_disk(self) -> dict[str, Any]:
+        if not self._path.exists():
+            return {}
+        try:
+            raw = json.loads(self._path.read_text(encoding="utf-8"))
+        except (json.JSONDecodeError, ValueError):
+            backup = self._path.with_suffix(".json.corrupt")
+            logger.warning(
+                "Corrupt import state at %s -- backing up to %s",
```
0xKT: loguru uses `{}`-style formatting, not printf. This `%s` call emits a literal "%s" and drops both paths (loguru silently ignores the extra positional args). Switch to `{}`.

raven/cli/onboard_commands.py
```diff
@@ -12,6 +12,7 @@
   4. EverOS long-term memory (optional; llm/embedding required once enabled,
      rerank/multimodal optional)
   5. deep_research tool (optional; MiroThinker key + model)
+  6. Cold-start import from other AI tools (optional)
```
0xKT, line 15: Off-by-one in the docstring: the next line is also numbered `6.` (`6. Done` -> should be `7.`), and line 1 still says "Six-step" while the list now has seven steps after Welcome.

raven/cli/import_commands.py
```diff
@@ -0,0 +1,496 @@
...
+@import_app.command("status")
+def status_cmd(
+    output_json: bool = typer.Option(False, "--json", help="Output raw JSON"),
+) -> None:
+    """Show cold-start import progress."""
+    import time
+    from collections import Counter
+
+    from rich.progress_bar import ProgressBar
+    from rich.table import Table
```
0xKT: Redundant: `Table` is already imported at module top (line 15). (The local `time` / `Counter` imports here are fine -- not imported at module level.)

0xKT (commented):
Reviewed the diff; CI is green and test coverage is good. A few findings below, anchored inline and ordered roughly by impact. Nothing blocking, but items 1-4 are worth a look before merge.

raven/plugin/memory/everos/backend.py
Kendrick-Song: Fixed in ecd02a9. Moved `_everos_server.py` from `raven/cli/` to `raven/plugin/memory/everos/_server.py`. All import sites (backend, onboard, tests) updated.

raven/plugin/memory/everos/backend.py
Kendrick-Song: Fixed in ecd02a9. Client default timeout set to 60s (recall, health). Memorize add/flush uses per-request `timeout=360.0` override via httpx. Stale comment removed.

raven/cli/tui_commands.py
Kendrick-Song: Fixed in ecd02a9. `backend.start()` moved to after handshake, runs as a background `create_task` so first render is not blocked.

raven/importer/types.py
Kendrick-Song: Fixed in ecd02a9. Removed `app_id`/`project_id` from `ImportSession` dataclass, scanner, and all test sites. Dead helper `_project_dir_from_path` also removed.

raven/importer/orchestrator.py
Kendrick-Song: Acknowledged. Added a NOTE comment at the checkpoint site explaining the per-source-unit granularity, EverOS session_id dedup safety, and that per-batch checkpoint is deferred until full-conversation import is common. (ecd02a9)

raven/cli/_everos_server.py
Kendrick-Song: Fixed in ecd02a9. Added `fcntl.flock` non-blocking try-lock in `_start_server_if_unlocked()`. Only the process that acquires the lock spawns the server; others skip spawn and go straight to health poll. Lock file at `~/.raven/everos-server.lock`.

raven/importer/state.py
Kendrick-Song: Fixed in ecd02a9. Changed `%s` to `{}` in the loguru warning call.

raven/cli/onboard_commands.py
Kendrick-Song: Fixed in ecd02a9. `6. Done` changed to `7. Done`.

raven/cli/import_commands.py
Kendrick-Song: Fixed in ecd02a9. Removed the local `from rich.table import Table` in `status_cmd`; module-level import is sufficient.

arelchan (approved):

PR #208, opened 2026-07-24
fix: inherit agent model in Skill Forge LLM calls

### LINKED ISSUE #155: Skill Forge rewriter/gate ignore agents.defaults.model, fall back to region-blocked provider default
## Summary

Skill Forge's auxiliary LLM calls (the query **rewriter** and the LLM **gate**)
do not use the configured agent model (`agents.defaults.model`). When no
Skill-Forge-specific model is set, they fall through to the provider's
hardcoded `default_model`, which for the `openrouter` provider is
`openrouter/anthropic/claude-sonnet-4-5`. If that default model is unavailable
(e.g. region-blocked), the rewrite/gate step hard-fails even though the agent is
configured with a working model -- so the main turn succeeds while
`skill.rewrite` / `skill.gate` error out.

## Evidence (from tracing)

Config: `agents.defaults.model = "minimax/minimax-m3"`, provider `openrouter`.

In one turn, the trace shows:

- `llm.call` with `invocation_source = session.turn` (main reasoning) ->
  `llm.model = minimax/minimax-m3`, status OK.
- `llm.call` with `invocation_source = skill.rewrite` ->
  `llm.model = openrouter/anthropic/claude-sonnet-4-5`, status ERROR:
  `OpenrouterException - {"error":{"message":"This model is not available in your region.","code":403}}`.

So the rewriter used a different model than the configured `minimax/minimax-m3`,
and that model is region-blocked. The rewriter then falls back to
`need_retrieval = True` with an empty rewritten query.

## Root cause

- `raven/context_engine/factory.py`: `QueryRewriter(provider, max_tokens=...)`
  is constructed with **no model**; `LLMGateFilter(..., model=skill_forge.llm_gate_model or None, ...)`
  is `None` unless `llm_gate_model` is set.
- `QueryRewriter.analyze` / `LLMGateFilter.filter` call
  `provider.chat_with_retry(...)` without a `model`, so they use
  `provider.default_model`.
- `raven/providers/registry.py`: the `openrouter` provider spec hardcodes
  `default_model = "openrouter/anthropic/claude-sonnet-4-5"`.

The main agent loop passes `model=agents.defaults.model` explicitly on every
call, so it is unaffected; only the auxiliary Skill-Forge calls hit the
hardcoded default.

## Impact

- Query rewrite and skill gate fail whenever the provider's hardcoded default
  model is unavailable/region-blocked, regardless of the configured agent model.
- The rewriter has **no config override** (unlike the gate's `llm_gate_model`),
  so this cannot be worked around by configuration alone.

## Proposed fix

Make the Skill-Forge auxiliary calls inherit `agents.defaults.model` when no
Skill-Forge-specific model is configured:

- Thread the configured agent model into `QueryRewriter` / `LLMGateFilter` in
  `factory.py` (add a `model` parameter to `QueryRewriter`).
- Pass that model on the `chat_with_retry` calls, so rewrite/gate use the same
  working model as the main loop by default; a Skill-Forge override (e.g.
  `llm_gate_model`) still wins when set.

## Workarounds (until fixed)

- Gate: set `skill_forge.llm_gate_model` to a model available in your region
  (e.g. `minimax/minimax-m3`).
- Rewriter: no config knob; either disable it (`skill_forge.rewrite_enabled = false`)
  or apply the code fix above.

guix4ever:
## Summary

- pass the active agent model to the Skill Forge query rewriter
- make the LLM gate inherit the active agent model when `llm_gate_model` is unset
- preserve the explicit gate model override and document the fallback behavior
- add regression coverage for model forwarding, inheritance, and override priority

## Root cause

The context-engine factory did not pass the active agent model into either Skill Forge LLM stage. `QueryRewriter` had no model parameter, while `LLMGateFilter` received `None` when no gate-specific override was configured. Both calls could therefore fall back to a provider-level default that differed from `agents.defaults.model`, including models unavailable in the user's region.

## Impact

The rewriter and gate now use the same configured model as the agent by default. Deployments that set `skillForge.llmGateModel` continue to use that explicit model for gate calls.

## Follow-up consideration

A future change could add an optional `rewrite_model` override, with `agents.defaults.model` as its fallback. That would let operators pin a cheaper or lower-latency model for query rewriting without relying on an implicit provider default. This PR intentionally does not add a new configuration field so the fix remains focused on model inheritance.

## Validation

- `uv run pytest tests/test_skill_forge_rewriter.py tests/test_skill_forge_gate.py tests/test_phase_a_default_engine.py -q` — 33 passed
- `uv run --extra dev ruff check` on all changed Python files
- `uv run --extra dev ruff format --check` on all changed Python files
- `git diff --check`

Closes #155.


0xKT:
Heads-up on a naming change that will collide with this branch.

A test-debt cleanup (#223) renames `tests/test_phase_a_default_engine.py` to `tests/test_context_engine_factory.py` -- the old name carried a phase code that AGENTS.md section 5.1 disallows. This PR is the only open one touching that file, so once the rename lands you will need to rebase and move your changes onto the new filename; the content is otherwise untouched by the rename.

Sorry for the churn. Nothing to do until the rename merges.

guix4ever:
Got it, thanks for the heads-up. I'll rebase onto the latest main and move my test changes to `tests/test_context_engine_factory.py` once #223 lands.

gloryfromca:

```bash
git fetch upstream main
git push --force-with-lease
```

PR #209, opened 2026-07-24
feat(providers): add MiniMax Global and CN OAuth

jellyjelly814:
## Summary

Add first-class MiniMax Token Plan OAuth support for both Global and CN regions.

This change adds:

- `minimax_global` and `minimax_cn` provider entries with curated MiniMax-M3 and M2.7 models.
- OAuth 2.0 device authorization with PKCE S256 and strict state validation.
- The shared MiniMax public client ID `coding-plan-cli` with the `openid profile coding_plan` scope.
- Region-specific OAuth and Anthropic-compatible API endpoints.
- Proactive access-token refresh with a five-minute buffer, cross-process locking, transient retry handling, and atomic refresh-token persistence.
- Separate Global and CN credential files, isolated from existing Codex OAuth storage.
- Runtime token refresh and MiniMax authentication headers for both normal and streaming requests.
- CLI login commands, onboarding, provider status/test/reset support, TUI setup detection, and model-picker entries.
- English and Chinese documentation updates.

Public model references remain under `minimax-global/` and `minimax-cn/`. The runtime translates them to the Anthropic-compatible LiteLLM route only when sending a request, so provider selection continues to use the correct MiniMax OAuth credential.

The implementation accepts MiniMax expiry values expressed as relative seconds, Unix epoch seconds, or Unix epoch milliseconds. OAuth-returned verification and resource URLs are restricted to the expected HTTPS hosts before use.

## Type

- [ ] Fix
- [x] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_minimax_oauth.py tests/test_cli_provider_commands.py tests/test_provider_catalog.py tests/test_config_update_providers.py tests/test_tui_rpc_model.py tests/test_tui_rpc_setup.py tests/test_cli_status_commands.py -q -x` - 143 passed.
- Focused MiniMax onboarding tests - 3 passed.
- `uv run ruff check raven tests` - passed.
- `uv run ruff format --check raven tests` - 761 files already formatted.
- `make check-large-files` - passed.
- `make check-commits` - passed.
- Two independent read-only reviews completed; the final review found no blocking issues.
- The full test collection was not run because the local environment does not include the optional `dingtalk_stream` channel dependency. All provider, OAuth, onboarding, TUI setup, status, and configuration tests relevant to this change passed.

## Risk

- OAuth credentials are stored in region-specific files with restrictive permissions and atomic replacement.
- Refresh operations use a cross-process lock and persist rotated refresh tokens before exposing them to the runtime.
- OAuth verification and API resource URLs are validated against region-specific host allowlists.
- Existing `minimax` API-key configuration remains unchanged; the new OAuth providers use separate slugs and storage.
- Rollback consists of reverting this change. Existing API-key provider configuration and credentials are not migrated or modified.

## Related Issues

N/A

- Live OAuth smoke test: MiniMax Global MiniMax-M3 returned "MiniMax OAuth works" successfully after normalizing the Anthropic-compatible resource URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ raven
 ```
 
 Raven supports OpenRouter, OpenAI, Anthropic, Gemini, DeepSeek, GitHub Copilot,
-OpenAI Codex OAuth, and custom OpenAI-compatible endpoints.
+OpenAI Codex OAuth, MiniMax Global/CN OAuth, and custom OpenAI-compatible endpoints.
 
 If setup fails or a provider is not ready, run:
 
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -101,7 +101,7 @@ raven
 ```
 
 Raven 支持 OpenRouter、OpenAI、Anthropic、Gemini、DeepSeek、GitHub Copilot、
-OpenAI Codex OAuth，以及自定义 OpenAI-compatible endpoints。
+OpenAI Codex OAuth、MiniMax Global/CN OAuth，以及自定义 OpenAI-compatible endpoints。
 
 如果配置失败，或者 provider 还没有准备好，运行：
 
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -59,7 +59,7 @@ Creates `~/.raven/config.json` and the workspace directory. Edit the config to a
 | `raven gateway` | Start full server (all channels + heartbeat + cron) |
 | `raven status` | Show config path, workspace, and API key status |
 | `raven channels status` | Show which messaging channels are enabled |
-| `raven provider login <name>` | Authenticate with an OAuth provider (e.g. `openai-codex`) |
+| `raven provider login <name>` | Authenticate with an OAuth provider (for example `openai-codex`, `minimax-global`, or `minimax-cn`) |
 
 ### 6. Run tests
 
--- a/raven/cli/_helpers.py
+++ b/raven/cli/_helpers.py
@@ -103,6 +103,13 @@ def make_provider(config: Config):
 
     if provider_name == "openai_codex" or model.startswith("openai-codex/"):
         provider = OpenAICodexProvider(default_model=model)
+    elif provider_name in {"minimax_global", "minimax_cn"}:
+        from raven.providers.minimax_oauth_provider import MiniMaxOAuthProvider
+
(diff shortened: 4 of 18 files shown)


0xKT:
Overall implementation looks solid. Two test issues to address before merge, plus two optional items.

**Blocking**

**1. `test_exactly_six_concrete_backend_classes` fails under full-suite collection**
`_concrete_provider_subclasses()` relies on `__subclasses__()` but only imports the 4 backend modules explicitly, not `lazy.py`. So `LazyProvider` (a proxy, not a real backend) only shows up when another test imports it, pushing the set to 7 and failing the assertion when collected in the same process:

```
uv run pytest tests/test_provider_catalog.py tests/test_lazy_provider.py -q -k "concrete or lazy"
# FAILED: Extra items in the left set: LazyProvider
```

Suggest excluding `LazyProvider` in `_concrete_provider_subclasses()` so the result is independent of import order.

**2. `_configure_existing_provider_model()` has no test coverage**
This function and the "Choose default model" branch added in onboard step 1 (`68dabc8d`) are untested; `tests/test_cli_onboard_commands.py` does not touch them. A CLI command change should update the matching test file (AGENTS.md 5.4).

**Non-blocking**

3. `minimax_oauth.py` imports `platformdirs` at module level, but it is not declared in `pyproject.toml` (it arrives transitively via `oauth-cli-kit`). Suggest declaring it explicitly with `uv add platformdirs`, following the repo's existing tiktoken/numpy pins.
4. `minimax_oauth.py:279` treats the device-flow `interval` as milliseconds; RFC 8628 defines it in seconds. Please confirm the unit MiniMax returns.

0xKT:
Thanks for this - the implementation is solid. To keep it moving, I'm pushing a small commit to your branch that addresses the two test items from my review: excluding `LazyProvider` in `_concrete_provider_subclasses` (deterministic backend-class set), and adding coverage for `_configure_existing_provider_model` and the step-1 "Choose default model" branch. Feel free to adjust.

The two non-blocking notes (declare `platformdirs`, confirm the device-flow `interval` unit) are left for you.

0xKT (approved):

PR #211, opened 2026-07-24
fix(weixin): reject nonzero send ret

### LINKED ISSUE #210: WeChat send silently succeeds when iLink returns non-zero ret with errcode=0
## Summary

The WeChat adapter treats a send as successful when the iLink `sendmessage` endpoint returns a non-zero `ret` while `errcode` is zero. This can produce a false-positive delivery receipt even though the message was not delivered.

## Version

- Reproduced against Raven 0.1.4 installed package
- Confirmed the same logic in the v0.1.9 tag and current `main`

## Minimal reproduction

```python
import asyncio
from raven.channels.adapters.weixin.channel import WeixinChannel

async def main():
    ch = object.__new__(WeixinChannel)

    async def fake_post(endpoint, body=None, *, auth=True):
        return {"ret": -2, "errcode": 0, "errmsg": "prepare failed"}

    ch._post = fake_post
    await ch._send_text("sanitized-user", "test", "sanitized-context")
    print("false success")

asyncio.run(main())
```

Actual result: `_send_text()` returns normally and the caller can report success.

Observed real response, sanitized:

```json
{"ret": -2, "errcode": 0, "errmsg": "prepare failed"}
```

## Expected behavior

`_send_text()` should raise when either `ret` or `errcode` is non-zero, consistent with `_poll_once()`. The exception should preserve `ret`, `errcode`, and the server message so automation callers cannot claim delivery.

## Actual behavior

The implementation only checks `errcode`:

```python
if errcode := data.get("errcode", 0):
    raise RuntimeError(...)
```

A non-zero `ret` with `errcode=0` is silently ignored.

## Impact

- Automated jobs can report successful WeChat delivery when the server rejected message preparation.
- Operators may miss time-sensitive notifications.
- Delivery audits become misleading.

## Suggested fix

Validate both fields and add regression tests for non-zero `ret`, non-zero `errcode`, and the success case.

## Privacy

All user identifiers, tokens, local paths, and message contents have been removed from this report.

louiseboo:
## Summary

- Treat a non-zero iLink `ret` as a failed WeChat text send even when `errcode` is zero.
- Include `ret`, `errcode`, and the server message in the raised error.
- Add a regression test for the observed sanitized response `{ret: -2, errcode: 0, errmsg: "prepare failed"}`.

Fixes #210

## Why

`_send_text()` only checked `errcode`. iLink can reject message preparation with a non-zero `ret` and `errcode=0`, causing Raven callers to report a false successful delivery. `_poll_once()` already treats either field as authoritative failure status.

## Verification

- Red: the new regression test failed with `DID NOT RAISE` before the implementation change.
- Green: `uv run pytest tests/test_channels_weixin.py::test_send_text_rejects_nonzero_ret_even_when_errcode_is_zero -q` → 1 passed.
- Relevant suite: `uv run pytest tests/test_channels_weixin.py tests/test_channels_manager.py tests/test_channels_outlet.py -q` → 71 passed.
- `uv run ruff check raven/channels/adapters/weixin/channel.py tests/test_channels_weixin.py` → clean.
- `uv run ruff format --check raven/channels/adapters/weixin/channel.py tests/test_channels_weixin.py` → clean.

## Risk

Low. The success path (`ret=0`, `errcode=0`, or absent zero-default fields) is unchanged. The only behavior change is that a server-declared failure now propagates to the caller instead of being silently treated as success.

## Rollback

Revert commit `2d8cd2f`; no data migration or configuration change is involved.

## Privacy

The issue and test use sanitized identifiers and content. No user tokens, chat IDs, local paths, or private messages are included.


gloryfromca:

```bash
git fetch upstream main
git push --force-with-lease
```

PR #214, opened 2026-07-25
fix(agent): clamp context_percent to 100 when the window is a guess

### LINKED ISSUE #124: TUI context usage can exceed 100% (percent falls back to a hardcoded 64k window when the model window can't be resolved)
### Summary

The TUI context-usage indicator can read **over 100%** (e.g. ~150%), and messages still send fine after it does. Root cause: the percentage is computed against a denominator that silently falls back to a hardcoded 64k default whenever the model's real context window can't be resolved — so once usage passes 64k but stays under the model's true (larger) window, the bar exceeds 100% while requests keep succeeding.

### Where

`raven/agent/loop/main.py:1447-1456`:
```python
context_max = resolve_context_window(call_model) or self.context_window_tokens
context_used = prompt_tokens + completion_tokens
usage_sink["context_percent"] = round(100 * context_used / context_max) if context_max else 0
```
Same fallback pattern in `raven/tui_rpc/methods/session.py:111-116` (`context_max = config...context_window_tokens`, replaced by the live window only if resolvable).

`self.context_window_tokens` defaults to **65_536** (`raven/config/schema.py:252`). So when `resolve_context_window(model)` returns `None`, `context_max` becomes 64k. If the model's real window is larger (128k / 200k / 1M), a conversation between 64k and the real window renders as **>100%** yet is still well within the model's limit — hence "150% but still sends".

### Why it's intermittent (and how to trigger)

`resolve_context_window` -> `_try_litellm_context_window` reads litellm's model map (`raven/token_wise/pricing.py:166-201`). It returns `None` when:
- the model isn't in litellm's cost/window map (version-dependent — e.g. `deepseek/deepseek-v4-flash` was "isn't mapped yet" in the #119 environment, though it resolves to 1M here), or
- litellm's remote cost-map fetch fails and the local backup doesn't cover the model (the #119 log shows `Failed to fetch remote model cost map ... SSL handshake timed out. Falling back to local backup.`).

On this machine resolution currently succeeds (`claude-sonnet-4-5` -> 200000), so it does **not** repro here — but the fallback path makes the percentage meaningless whenever resolution misses.

Note: the per-turn emit uses `call_model` (the model of that specific call), so a sub-call on a cheaper/unmapped model can momentarily set a wrong `context_max`.

### Expected

A usage percentage should never exceed 100%, and should not be computed against a guessed denominator presented as authoritative.

### Suggested fix

- When `resolve_context_window` returns `None`, don't render a percentage against the 64k guess. Instead show used-tokens with the window marked unknown/estimated (or hide the bar), so the number never implies "over budget" when it isn't.
- Optionally clamp/annotate when `context_used > context_max`, and allow a per-model `context_window_tokens` override for models litellm can't map.

### Environment

macOS TUI. Reported: bar once showed ~150% while sending continued to work.

AmirF194:
Root cause: `resolve_context_window()` returns `None` whenever a model isn't in litellm's static map or OpenRouter's live `/models` table, and the per-turn usage handler in `raven/agent/loop/main.py` falls back to `self.context_window_tokens` (65536 by default) as the denominator for `context_percent`. That fallback is a guess, not the model's real window. Real usage can legitimately pass 65536 tokens while staying well under a model's true window (128k/200k/1M), so the computed percent renders past 100% even though the request keeps succeeding.

Fix: clamp `context_percent` to 100. `context_max` and `context_used` are unchanged (still the real numbers behind the used/max label); only the derived percent is bounded, so the gauge never implies an over-budget state that isn't real.

`raven/tui_rpc/methods/session.py`'s `_baseline_usage` uses the same fallback-window pattern but always pairs it with `context_used=0` (the docstring: "Usage starts at zero for a fresh session by design... counters refresh on the next turn"), so it can't itself produce a percent over 100 and needed no change.

Verified in a clean `python:3.12-slim` container (uv-managed venv, `uv sync --frozen --extra dev --dev`):
- Added `test_usage_sink_context_percent_clamps_when_window_is_a_guess`, which drives the real `_process_message` usage-sink path with a `stub` model (never resolves via `resolve_context_window`) and a configured window smaller than actual usage.
- Confirmed it fails on unpatched `main` (`800 == 100`, i.e. the raw unclamped ratio) and passes with the fix.
- `uv run --extra dev pytest tests/test_agent_loop_usage_sink.py tests/test_tui_rpc_session_init_bundle.py -q` -> 17 passed.
- `ruff check` / `ruff format --check` on both changed files, and `pre-commit run --files raven/agent/loop/main.py tests/test_agent_loop_usage_sink.py` (ruff, trailing-whitespace, end-of-file-fixer, large-files, merge-conflict, private-key) all pass.

Not changed: the reporter's other suggestion (mark the window as estimated/unknown to the UI instead of clamping) would need a new wire field in `raven/tui_rpc/models.py` plus TUI/bridge changes to consume it, which is a separate, larger concern; happy to follow up if that's wanted.

Fixes #124

Co-authored-by: Claude (claude-sonnet-5) <noreply@anthropic.com>

### DIFF
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -2012,7 +2012,12 @@ class AgentLoop:
                 usage_sink["cost_usd"] = usage_snapshot.estimated_cost_usd
                 usage_sink["context_max"] = context_max
                 usage_sink["context_used"] = context_used
-                usage_sink["context_percent"] = round(100 * context_used / context_max) if context_max else 0
+                # context_max is a real provider window when resolve_context_window()
+                # succeeds, but otherwise a guessed default (self.context_window_tokens)
+                # that can be smaller than the model's true window. Clamp so the gauge
+                # never reports past full: an unresolved model must not render as
+                # "over budget" while the request is still well within its real limit.
+                usage_sink["context_percent"] = min(100, round(100 * context_used / context_max)) if context_max else 0
 
             # Context-window overflow recovery: the structured classifier flags
             # should_compress (a smaller window won't help, but eliding the bulk
--- a/tests/test_agent_loop_usage_sink.py
+++ b/tests/test_agent_loop_usage_sink.py
@@ -336,3 +336,31 @@ def test_refresh_context_window_on_an_openrouter_model_never_touches_the_network
 
     assert agent.context_window_tokens == rates.DEFAULT_CONTEXT_WINDOW_TOKENS
     assert counter["calls"] == 0
+
+
+@pytest.mark.asyncio
+async def test_usage_sink_context_percent_clamps_when_window_is_a_guess(workspace):
+    """Usage past the configured fallback window must not render past 100%.
+
+    ``stub`` never resolves via resolve_context_window(), so context_max here is
+    the configured (guessed) default. Real usage can legitimately exceed that
+    guess while staying under the model's true, unresolved window, so the
+    reported percent must clamp rather than imply an impossible over-budget bar.
+    """
+    provider = UsageProvider("stub", prompt_tokens=6000, completion_tokens=2000)
+    agent = _make_agent(workspace, provider, model="stub", window=1000)
+    sink: dict = {}
+
+    await agent._process_message(
+        TurnRequest(
+            origin=Origin.USER,
+            source=Source(channel="test", chat_id="c1", sender_id="user", chat_type=ChatType.DM),
+            text="hi",
+        ),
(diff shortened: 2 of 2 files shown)


AmirF194:
Checking in after a week with no review yet. Happy to switch to marking the window as estimated in the UI instead of clamping, per the alternative noted in the PR body, if that's preferred; no rush.

AmirF194:

PR #216, opened 2026-07-26
fix: stop pointing the agent at an empty episodic log under a memory backend

### LINKED ISSUE #122: The episodic log has no writer: the prompt advertises it and five Sentinel producers read it
### Summary

The episodic-memory path advertised to the agent is **hardcoded** and does not follow the active memory backend. It always points at `{workspace}/user_memory/episodic/episodes.md`, which is only written by the legacy `memory_engine` backend. Under the **EverOS** backend that file stays empty and episodes live elsewhere, so any prompt/guidance that references it points the agent at an empty file.

> Split from the original report: this issue is the **path / backend-blindness** half. The separate "the agent isn't told where raw session transcripts live and conflates them with episodic memory" concern is tracked in its own issue.

### Details

The system prompt renders this line unconditionally, regardless of `memory.backend`:

- `raven/context_engine/segments/render.py:87`
- `raven/agent/context/builder.py:211`

```
- Episodic log: {workspace_path}/user_memory/episodic/episodes.md (grep-searchable). Each entry starts with [YYYY-MM-DD HH:MM].
```

That workspace file is only written by the legacy consolidator (`raven/memory_engine/consolidate/consolidator.py:700`, `history_file = workspace/user_memory/episodic/episodes.md`). When `memory.backend = "everos"`, EverOS writes episodes to `~/.everos/<instance>/.../episodes/*.md` instead, and the workspace file is never populated.

### Evidence (this machine, `memory.backend = everos`)

- `~/.raven/workspace/user_memory/episodic/episodes.md` -> **0 bytes**
- `~/.everos/raven/default_app/default_project/users/default/episodes/*.md` -> **8 episode files**

### Suggested fix

- Make the episodic-memory guidance **backend-aware**: under EverOS, point at the EverOS episode store (or, better, tell the agent to use the `recall` / memory tool rather than grepping a file); under the legacy backend, keep the `episodes.md` path.
- The prompt line is duplicated in `render.py:87` and `builder.py:211` — unify so the two cannot drift.

### Environment

macOS, `memory.backend = everos`.

AmirF194:
## Summary

The Workspace section of the system prompt always advertises `{workspace}/user_memory/episodic/episodes.md` as the episodic log to grep, in both `render.py`'s `identity_text` (the live per-turn prompt) and `builder.py`'s hand-duplicated copy (token-size estimation only). That file is written only by the host's legacy `MemoryConsolidator`. When a `MemoryBackend` plugin is wired (e.g. EverOS), episodes live in the plugin's own store instead and are surfaced through per-turn recall into the `# Memory` section (`MemorySegmentBuilder`), so the workspace file stays empty and the guidance sends the agent to grep nothing, exactly as the issue's own evidence shows (0-byte `episodes.md` next to 8 populated EverOS episode files).

`identity_text` now takes a `has_memory_backend` flag. `IdentitySegmentBuilder` and `build_context_engine` already receive `backend`, so wiring it through is one added parameter on each. With a backend, the `Episodic log` line is replaced with guidance pointing at the automatic `# Memory` recall instead of a file path that stays empty; without one (the default), the rendered text is unchanged. `ContextBuilder._get_identity` now delegates to the shared `render.identity_text` instead of keeping its own copy of the same text, which is the "unify" half of the issue: the two had already drifted once (`builder.py` was missing the language directive `render.py` has).

## Type

- [x] Fix

## Verification

Commands run (this container, `python:3.12-slim` + `uv sync --frozen --extra dev --dev`):

- `uv run --extra dev pytest tests/test_segments.py tests/test_context_invariants.py tests/test_default_context_engine.py tests/test_security_untrusted_context.py -v` -> 42 passed.
- New regression tests (`test_identity_with_backend_omits_stale_episodic_path`, `test_identity_with_backend_matches_legacy_builder` in `tests/test_segments.py`) fail on `main` with `TypeError: IdentitySegmentBuilder.__init__() takes 2 positional arguments but 3 were given` (confirmed by reverting only the source files and re-running against the new tests) and pass on this branch.
- Broader regression sweep: every test file that constructs `ContextBuilder`/`AgentLoop`/`build_context_engine` (29 files, `git grep` list), minus `tests/integration/` and the messaging-channel adapter tests that fail to import for missing optional deps unrelated to this change (`dingtalk_stream`, `lark_oapi`, `nh3`, `botpy`, `slackify_markdown`, `python-telegram-bot`, `wecom_aibot_sdk`, none installed by `--extra dev`) -> 402 passed, 22 failed. All 22 failures are in `tests/test_runtime_checkpoint_bug2*.py` and are pre-existing environment gaps (`FileNotFoundError: [Errno 2] No such file or directory: 'git'`, the `python:3.12-slim` image has no `git` binary), unrelated to this diff.
- `uv run --extra dev ruff check` / `ruff format --check` on every changed file: clean.
- `uv run --extra dev pre-commit run --files <changed files>`: all hooks pass (ruff, ruff-format, trailing-whitespace, end-of-file-fixer, large-files, merge-conflict, private-key).

Not verified: an end-to-end run against a live EverOS instance (the issue's own repro environment). The fix is scoped to the prompt-text guidance the issue describes and is covered by the unit tests above; I did not stand up EverOS to confirm the rendered prompt reaches a running agent unchanged elsewhere.

## Risk

- [x] Security impact considered: none, prompt text only.
- [x] Backward compatibility considered: `has_memory_backend` defaults to `False` and the new `backend` parameters default to `None`, so every existing call site (all of which construct without a backend) renders byte-identical text to before.
- [x] Rollback path is clear for risky changes: revert the single commit.

## Related Issues

Fixes #122

### DIFF
--- a/raven/agent/context/builder.py
+++ b/raven/agent/context/builder.py
@@ -12,6 +12,7 @@ from raven.security.trust import wrap_untrusted, wrap_untrusted_blocks
 from raven.utils.helpers import build_assistant_message
 
 if TYPE_CHECKING:
+    from raven.memory_engine.backend import MemoryBackend
     from raven.providers.base import LLMProvider
 
 
@@ -36,8 +37,10 @@ class ContextBuilder:
         now_fn: Callable[[], datetime] | None = None,
         *,
         start_watcher: bool = True,
+        backend: "MemoryBackend | None" = None,
     ):
         self.workspace = workspace
+        self._backend = backend
         self.memory = MemoryStore(workspace)
         self.skills = LocalSkillCatalog(
             workspace,
@@ -177,14 +180,15 @@ Skills with available="false" need dependencies installed first - you can try in
     def _get_identity(self) -> str:
         """Get the core identity section.
 
-        Delegates to the request path's renderer so the estimation prompt
-        (this class) and the real per-turn prompt can never drift apart.
-        Imported lazily: ``context_engine`` imports this module back via
-        its factory, so a module-level import would be circular.
+        Delegates to the shared :func:`render.identity_text` (also used by
+        the live :class:`IdentitySegmentBuilder` path) so the estimation
+        prompt (this class) and the real per-turn prompt can never drift
+        apart. Imported lazily: ``context_engine`` imports this module back
+        via its factory, so a module-level import would be circular.
         """
         from raven.context_engine.segments import render
 
-        return render.identity_text(self.workspace)
+        return render.identity_text(self.workspace, has_memory_backend=self._backend is not None)
 
     def _build_runtime_context(self, channel: str | None, chat_id: str | None) -> str:
         """Build untrusted runtime metadata block for injection before the user message."""
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -445,6 +445,7 @@ class AgentLoop:
             skill_forge_config=skill_forge_config,
             llm_provider=provider,
             now_fn=now_fn,
+            backend=backend,
         )
         self.sessions = session_manager or SessionManager(workspace)
         # Tool names to omit from the registry — applied after default-tool
--- a/raven/context_engine/factory.py
+++ b/raven/context_engine/factory.py
@@ -114,7 +114,7 @@ def build_context_engine(
     )
 
     builders = [
-        IdentitySegmentBuilder(workspace),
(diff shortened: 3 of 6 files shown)


AmirF194:
No rush, just checking in whenever you get a chance. CI hasn't run yet, still waiting on approval for the fork PR.

AmirF194:

Issue #229, opened 2026-07-29
Remaining test gaps in the episodes feature


0xKT:
Auditing the suite for #223 included a pass over #217 (collapsible episodes) and #220 (upgrade nudge). Most of what those added is solid, and the small gaps -- an empty test body, the `session.py` wiring, `clipToWidth`, `hasMeaningfulReasoning`, `_upgrade_command_works` -- are closed in the #223 cleanup. Three items were left out of that PR because they need real work or a product decision rather than a quick assertion.

Each finding below was confirmed by reading assertions and, where noted, by mutating the code under test and watching the suite stay green. None of this breaks a user today; the risk is that a regression would pass silently.

## Table width fitting has no tests

`ui-tui/src/components/markdown.tsx:217-270` -- roughly 90 lines of `clipWidth` / `fitWidths` deciding how a markdown table is squeezed into the available columns. Boundary-heavy code (single wide column, more columns than cells, zero budget, CJK glyphs) and none of it is exercised.

## Episode height estimation has no tests

`ui-tui/src/lib/virtualHeights.ts:91-138` -- pure arithmetic over collapsed and expanded episodes, easy to pin with a table of cases, currently unpinned. Flagged as follow-up R8 during the #217 review.

## The middle of the episode.start path is untested

The event has coverage at both ends -- the agent loop emits it, `turnController` consumes it -- but the three hops between have none:

- `raven/tui_rpc/spine.py:169-172` (outbound)
- `ui-tui/src/app/chatStream.ts:113-115` (inbound)
- `ui-tui/src/app/createGatewayEventHandler.ts:304-307` (dispatch)

Renaming the event on either side turns nothing red, so the wire contract is only guarded by convention.

## A field nothing reads, credited by a test

`ui-tui/src/domain/episodeSummary.ts:11-14` defines `VerbRule.style`, which is only ever written -- no reader anywhere in the tree. `episodeSummary.test.ts:43` attributes the `'ran 2 commands'` output to it, but that string comes from `unit`; setting every `style` value to `'count'` leaves the test green.

This one is a product decision rather than a test gap: either something should read the field, or it should go along with the test's claim about it.

0xKT:
Closing as not planned, with the findings recorded so picking this up later is cheap.

This is four separate pieces of work rather than one, and they differ a lot in cost:

- `virtualHeights.ts` is the easy one -- everything it exposes is already an exported pure function, so a table of cases pins it with no production change.
[...]

Issue #228, opened 2026-07-29
test_dogfood_whitelist is flaky when the whole file runs


0xKT:
## Symptom

`tests/tui/autotest/tests/test_dogfood_whitelist.py` parametrises 14 slash commands, each spawning its own real `raven tui` through the tui-use harness. Run the file as a whole and a handful fail; which ones changes between runs.

Measured on macOS, `origin/main` (16719d0), with tui-use 0.1.20 and a built `ui-tui/dist`:

```
run 1: 5 failed, 2 passed, 6 xfailed, 1 xpassed
run 2: 3 failed, 4 passed, 5 xfailed, 2 xpassed
```

Different slash commands fail each time (`status`, `skill_list`, `sentinel_nudges`, `sandbox_list` all appeared). Run a single case on its own and it passes -- `skill_list` alone takes 5.9s and is green.

The failure is always the same shape: `expect_exit(0)` times out after the escape/ctrl+c sequence, with the screen still showing the welcome frame.

## What has been ruled out

While cleaning up test debt (#223) three hypotheses were tested and rejected:

1. **Readiness signal.** The file keys off the banner (`wait(r"Raven")`), which paints about 5s before the status bar reports ready at about 12s. Fixing that in the sibling chat e2e was a real bug fix there, but applying the same change here did not stabilise this file -- and it made `skill_list` fail deterministically, so it was reverted.
2. **Composer confirmation.** Asserting the composer shows the text before pressing enter -- no effect here.
3. **Autocomplete popup timing.** Sleeping after the text lands so the slash menu settles -- no effect.

The file is unchanged from `origin/main` today; the experiments above were all reverted.

## Where the evidence points

Individual cases pass; the file as a whole does not. That points at interference between the 14 sequential harness sessions rather than at any one slash command -- leaked tui-use sessions, port or socket reuse, or shared temp state between spawns. The file's own xfail reasons blame "Raven TUI overlay exit UX inconsistent (Esc+Ctrl+C doesn't always terminate this command's overlay)", which is consistent with an exit-path race rather than a routing bug.

Note that two cases are currently marked `xfail(strict=False)` and sometimes XPASS, so the marker set does not describe the real behaviour either.

## Impact

Low today: the file is `@pytest.mark.e2e`, and no CI job runs the e2e tier, so this blocks nobody. It does mean the dogfood coverage of 14 slash commands is unreliable when someone runs it locally, which is exactly when it would be useful.

## Suggested next step

[...]

PR #230, opened 2026-07-29
test: clear the audit debt in the test suite

### LINKED ISSUE #223: Test-suite audit: naming violations, one zero-coverage e2e file, and assorted weak assertions
## Context

With the whole unit suite finally running in CI (see the CI-coverage issue), the next question is whether the tests that now run actually test anything. All ~298 test files (265 top-level, 23 integration, 10 tui) were audited file by file.

**The good news: quality is high.** Assertions are generally specific, drive real product code paths, and isolate well (global state snapshot/restore, mocking at network / subprocess / LLM boundaries, injected clocks and fake providers). Exactly one file was found to be a genuine fake pass.

What follows is the remaining debt, none of it urgent, all of it mechanical to clear.

## 1. One zero-coverage file pretending to be an e2e suite

`tests/integration/test_chat_e2e.py` carries a module-level unconditional `pytest.mark.skip`, and all three of its test functions are `raise NotImplementedError` stubs. It contributes zero coverage while its name suggests three chat e2e scenarios exist. Proposal: delete it.

## 2. Naming violations (AGENTS.md section 5)

Thirteen files carry phase / ticket / bug suffixes that section 5.1 explicitly forbids:

| File | Proposed name |
|---|---|
| test_ag1_backend_dispatch.py | test_agent_loop_backend_dispatch.py |
| test_cl1_plugin_stack.py | test_cli_plugin_stack.py |
| test_config_cfg1.py | test_config_raven_sections.py |
| test_em1_skeleton.py | test_everos_plugin_discovery.py |
| test_em2_backend.py | test_everos_backend.py |
| test_em3_http.py | test_everos_http_adapter.py |
| test_fb1_feedback_dispatch.py | test_backend_feedback_dispatch.py |
| test_phase_a_default_engine.py | test_context_engine_factory.py |
| test_skill_forge_phase_a.py | test_skill_forge_catalog.py |
| test_skill_router_sr1.py | test_skill_forge_local_source.py |
| test_skill_router_sr2.py | test_skill_router_fusion.py |
| test_skill_router_sr4.py | test_skill_router_everos_source.py |
| test_tier1_skeleton.py | test_package_skeleton.py |
| test_runtime_checkpoint_bug2.py | test_runtime_checkpoint.py |
| test_runtime_checkpoint_bug2_deep.py | test_runtime_checkpoint_deep.py |

One file has the CLI naming reversed (section 5.3): `test_sandbox_cli.py` covers `cli/sandbox_commands.py` and should be `test_cli_sandbox_commands.py`.

Two integration files use a `<kind>` outside the allowed set (section 5.2): `test_sandbox_cli_integration.py` (real socket, mocked boxlite -> `test_sandbox_cli_smoke.py`) and `test_sandbox_integration.py` (requires KVM, real VM -> `test_sandbox_real_vm.py`).

## 3. Assertions worth tightening

- Three tests may assert vacuously or never run in a clean environment and need verification: `test_everos_backend_e2e.py` (both assertions live inside a `for h in hits` loop, so an empty result set passes), `test_tui_exit_e2e.py` (skips when the TUI build is unavailable), `test_sentinel_e2e.py` (reimplements the dispatch routing inside the test, so the production entry point is unguarded).
- About eight `pytest.raises(Exception)` should be narrowed to the specific exception type (auth_allowlist, decision_consumer, sandbox_unit, and four in the tui_rpc turn tests).
- Two concurrency tests assert on wall-clock windows with little margin and may flake in CI (`test_event_wake.py`, `test_subscription_emitter.py`).
- Minor cleanups: dead assignments in `test_routine_learner_decay.py` and `test_task_discoverer.py`, a never-exercised command registration in `slash_routing.py`, redundant `importorskip` guards on hard dependencies, and hardcoded `/tmp` paths in `test_decision_consumer.py` that should use `tmp_path`.

## Plan

- [ ] Delete the zero-coverage e2e file
- [ ] Rename the 18 files above
- [ ] Verify and fix the three possibly-vacuous tests
- [ ] Narrow the over-broad exception assertions and clear the minor cleanups

Renames are reported rather than applied unilaterally, per AGENTS.md section 5.4.

0xKT:
## Summary

A read-only audit of all 298 test files found the suite in good shape -- exactly one genuinely fake test -- with the real debt sitting in file names and a scattering of assertions that passed for the wrong reason. This clears that, and fixes the bugs found while clearing it.

**Three dead tests dealt with.** Two removed because neither could ever assert anything:

- `test_chat_e2e.py` was a module-level skip over three `NotImplementedError` stubs. The wire-up it waited for has shipped (`tui_commands` instantiates `SubscriptionEmitter`) and the proposal doc it cited is gone.
- `test_tui_exit_e2e.py` asserted that building the agent loop leaves lancedb's thread live. It does not: the loop is constructed but never started, and the memory plugin talks to everos over HTTP, so lancedb is not even imported. The assertion could not hold in any configuration.

The third, `slashParity.test.ts`, shelled out to python for a `raven_cli.commands.COMMAND_REGISTRY` that no longer exists, and skipped all three of its cases when the import failed -- three skips on every run. The parity it wanted needs no python: `createSlashHandler.ts:45-51` routes a slash locally when `findSlashCommand` resolves it and otherwise hands it to `slash.exec`, a CLI subprocess a mutation cannot reach the live session from. It now asserts that predicate directly. `NATIVE_MUTATING_COMMANDS` and the three-way `classifyRoute` went with it: all six names in that set are also in the local registry, so the native branch could not change any outcome.

**19 files renamed** off phase, ticket and bug codes (AGENTS.md section 5.1), including one reversed CLI name and two integration files moved onto a legal `<kind>`. Reference updates ride along: `.pre-commit-config.yaml` names one of these files in a `detect-private-key` exclude, and three docs pages point at them.

**Assertions that passed for the wrong reason, tightened**: eight over-broad `pytest.raises(Exception)` narrowed to the type each site actually raises; an `except` clause that swallowed the `AssertionError` its own fakes use as a tripwire; a `hasattr` probe replaced by binding a broker and checking it stuck; and a recall assertion that read as a pass on an empty result set.

**Dead weight removed**: ten unused locals, three dead code blocks, six `importorskip` guards on hard dependencies, two hardcoded `/tmp` paths, and ticket codes left in docstrings after the renames.

**Two coverage gaps the deletions exposed are now closed**:

- `tests/test_cli_exit.py` pins the hard-exit guard, which had no tests at all.
- The TUI chat e2e gains a multi-turn case. It reads the persisted session rather than the screen, because a prompt is echoed into the transcript the moment it is typed -- screen-scraping for a planted word passes even when history is broken.

**Three timing bugs fixed in the TUI e2e tests.** They keyed off the banner, which paints about seven seconds before the app accepts input, and submitted the moment text was typed. Dropped prompts made the pipeline look dead in one test, and made a negative assertion (`assert not leak_detected`) pass vacuously in another -- no prompt means no streaming means no leak.

**Gaps in the two features that landed during review**, small enough to close here: the `session.py` wiring that fills `update_available` / `update_command` (removing it left the suite green), `_upgrade_command_works` (a fixture stubbed it to `True` everywhere), an `episodeSummary` case with an empty test body, and `clipToWidth` / `hasMeaningfulReasoning`, which had no coverage at all. The larger items from that audit are tracked in #229.

**Upstream review attributions dropped from ui-tui comments.** 36 mentions of another project's review thread (#19835 as "Copilot round-N review"), plus #18994, #19194 and #14045 -- none resolve to anything here, and AGENTS.md section 1.1 rules out comments pointing at information only visible elsewhere. The rationale each carried is kept; four `describe()` titles lose a trailing `(#18994)`. Three went further and documented a python counterpart that does not exist: `platform.ts` credited key-spelling normalization to `raven_cli/voice.py` (no such module -- `raven/tui_rpc/methods/_stubs.py` answers `voice.toggle` with "voice not supported in Raven v0.1"), and `useConfigSync.ts` plus its test credited an `interrupt` default to `raven_cli/config.py` and `tui_gateway/server.py::_load_busy_input_mode` (`busy_input_mode` appears nowhere in python; neither module exists).

**One flaky test fixed, caught by this PR's own CI run.** `test_stdio_no_executor_does_not_raise` set `command="true"` and let `connect_mcp_servers` really spawn it, so the result depended on how a live child process raced the MCP handshake -- it failed once on CI in the full-suite run while passing 30 out of 30 standalone runs locally. It also asserted only that whatever came back was not a `SandboxInitError`. The transport now raises instead of running (the idiom its sibling test already uses) and the test asserts the stdio branch was reached.

Review follow-ups, folded in: the hard-exit test now skips rather than asserting no lancedb thread exists process-wide, `tests/conftest.py` no longer credits the pytest segfault to a lancedb thread the suite never starts (raven imports lancedb nowhere), and the status-bar patterns the two TUI e2e tests wait on live in one module instead of a hand-mirrored copy per file.

Every new or changed assertion was checked by mutating the code under test and confirming the assertion fails.

## Type

- [x] Others

Test files, test configuration, and comments. The only non-test files touched are ui-tui comments and four `describe()` titles -- no behaviour change.

## Verification

```
uv run --all-extras pytest -q
# 4654 passed, 25 deselected

cd ui-tui && npx vitest run
# 936 passed, no skips

uv run --all-extras pytest tests/tui/autotest -m e2e -q
# passes locally with tui-use installed and ui-tui built

bash .claude/scripts/preflight_ci.sh
# ALL PREFLIGHT CHECKS PASSED (including the TUI lane)
```

Mutation checks, each confirming the new assertion fails when the code under test is broken: the `session.py` nudge wiring removed; `clipToWidth` replaced with identity; `hasMeaningfulReasoning` forced to `true`; a never-sent prompt added to the multi-turn session assertion.

## Risk

No behaviour changes: the non-test diff is comments only. The renames touch git history for 19 test files; #208 is the only open PR affected and has been notified.

Two known items are deliberately left alone: `test_dogfood_whitelist.py` is flaky when the whole file runs (tracked in #228, unchanged here), and the `lancedb_finalization_hazard` gate in `commands.py` never fires today -- its test docstring says so explicitly, so the coverage does not read as an argument for keeping it.

## Related Issues

Closes #223


0xKT (commented):
Reviewed the whole diff (merge-base..head, 51 files, renames resolved with -M) in a
separate worktree. **Verdict: approve / mergeable.** No before-merge findings. Four
items below, all follow-up or nit; R4 from my first pass is withdrawn.

### Verified, so no need to re-check

- **Renames are complete.** Grepped all 20 old basenames across the whole repo
  (docs/, .github/, Makefile, scripts/, benchmarks/): zero hits. The
  `detect-private-key` exclude in `.pre-commit-config.yaml:25` and the three docs
  paths were updated in the same commit.
- **Both deletions are justified.** `test_chat_e2e.py` was a module-level skip over
  three `NotImplementedError` stubs; `tui_commands.py:520` does instantiate
  `SubscriptionEmitter`, and `docs/openspec/changes/tui-chat/` no longer exists.
  `test_tui_exit_e2e.py` asserted `lancedb_finalization_hazard()` is true, and the
  probe keys on a live thread name that nothing opens today, so the test could only
  fail (rc 43) or skip. `tests/test_cli_exit.py` is a strictly better replacement.
- **The six dropped `importorskip` guards were all on hard dependencies.** `httpx`
  and `everos[multimodal]==1.1.3` are in `[project.dependencies]`;
  `raven.plugin.memory.everos` is bundled in-repo.
- **New API references all exist.** `SessionManager.find_most_recent_chat_id` /
  `peek` / `get_history` (`raven/session/manager.py:192,402,88`), `clipToWidth`
  (`ui-tui/src/lib/text.ts:65`), `hasMeaningfulReasoning`
  (`ui-tui/src/lib/reasoning.ts:51`).
- **`_WORKING_RE` matches the source of truth.** All 15 verbs are identical to
  `VERBS` in `ui-tui/src/content/verbs.ts:22-38`.
- **The removed locals and dead blocks are genuinely dead** (the `response` block in
  `test_task_discoverer.py` was shadowed by the reassignment below it).
- Branch name, commit type, header length, `Co-authored-by` trailer, and ASCII-only
  in both the commit message and this description: all compliant.

---

### R1 - the hazard probe test asserts process-global state

- **Tier:** follow-up
- **Where:** `tests/test_cli_exit.py:28`
- **Claim:** `assert not any(t.name == _LANCEDB_THREAD for t in threading.enumerate())`
  asserts a property of the whole process, not of the unit under test.
- **Why it matters:** it holds today (full suite is green), but any future test that
  opens a lancedb connection earlier in collection order flips it red, and the
  failure message will read as "the exit guard broke" rather than "something opened
  lancedb". `tests/conftest.py:15` also still documents the opposite ("native
  runtimes pulled in by the suite (lancedb's Rust/tokio thread ...)"), so the two
  docstrings now contradict each other.
- **Fix:** turn the precondition into a guard rather than an assertion, and reconcile
  the conftest comment with what is actually true today:

```python
def test_hazard_false_without_the_lancedb_thread() -> None:
    if any(t.name == _LANCEDB_THREAD for t in threading.enumerate()):
        pytest.skip("a lancedb connection is already open in this process")
    assert lancedb_finalization_hazard() is False
```

- **Verify:** `uv run pytest tests/test_cli_exit.py -q` standalone and inside
  `uv run --all-extras pytest -q`; both still pass, and the test skips rather than
  fails when a thread of that name is alive at collection time.

### R2 - the coverage pins a gate that never fires

- **Tier:** follow-up (separate PR - out of scope here, this one is test-only)
- **Where:** `raven/cli/commands.py:177-178`
- **Claim:** `if lancedb_finalization_hazard(): flush_and_hard_exit(code)` is dead.
  The probe returns False in every configuration today (this PR's own docstring says
  so), and the only live caller of `flush_and_hard_exit` is `tests/conftest.py:32`.
- **Why it matters:** not a defect in this PR - the scoping call to leave production
  code alone is right. Flagging it so the new coverage does not later become the
  argument for keeping the branch: an audit that proves a gate is dormant should end
  in deleting the gate, not in pinning it.
- **Fix:** open an issue to delete the branch in `commands.py` and keep
  `flush_and_hard_exit` for the pytest hook. If it is kept deliberately (e.g. a
  planned embedded-lancedb mode), say so in `_exit.py`'s module docstring, which
  currently describes the gate as load-bearing.
- **Verify:** n/a for this PR.

### R3 - the working-verb regex is now duplicated verbatim

- **Tier:** nit
- **Where:** `tests/tui/autotest/tests/test_e2e_streaming_no_log_overlay.py:40-48`
  and `tests/tui/autotest/tests/test_e2e_raven_tui_chat.py:43-51`
- **Claim:** both files now carry the same hand-mirrored `_READY_RE` / `_WORKING_RE`
  pair, each with a "keep in sync with verbs.ts" comment.
- **Why it matters:** two hand-maintained copies of one source of truth; a third
  arrives with the next e2e file, and a verb-pool edit in `verbs.ts` has to find all
  of them.
- **Fix:** lift both regexes into one module under `tests/tui/autotest/` and import
  them in the two test files.
- **Verify:** `uv run --all-extras pytest tests/tui/autotest -m e2e -q` locally
  (tui-use installed, `ui-tui/dist` built).

### R4 - withdrawn

I first read `_run_turn`'s `harness.screen()` polling
(`tests/tui/autotest/tests/test_e2e_raven_tui_chat.py:130-139`) as a flake vector
versus `harness.wait()`. It is not, on two counts: the same loop already exists on
main at `test_tui_chat_round_trip` (so the helper follows the in-file idiom rather
than introducing one), and the loop has to race two patterns at once -
`model_not_available` routes to skip, the working verb routes to continue - which a
single-pattern `wait()` cannot express. No action.

### R5 - commit body and description numbers disagree

- **Tier:** no action
- **Claim:** the commit body says "Renamed 18 files" / "4615 passed"; this
  description says 19 / 4654. The rename count is 19.
- **Why it matters:** it does not. The repo squash-merges with
  `squash_merge_commit_message=PR_BODY`, so the correct numbers are what land on
  main. Noting it only in case the commit is ever amended.

0xKT (commented):
Re-reviewed the full diff at `9c03caa` (not just the delta). R1 and R3 are fixed as
suggested, R4 stays withdrawn, R5 needs nothing. **One new finding from the third
commit, cosmetic. Approve-ready either way** - it is a comment, so merging with it is
also fine.

### R7 - stray character left by a comment rewrap

- **Tier:** nit
- **Where:** `ui-tui/src/lib/platform.ts:225`
- **Claim:** joining the last two lines of the `parseVoiceRecordKey` doc comment kept
  the continuation line's leading marker as `/`:

```
 * back to the documented Ctrl+B default so a typo never silently
 * disables the shortcut. / */
```

- **Why it matters:** harmless to tsc, prettier and vitest (it is comment text, which
  is why CI is green), but it is visible garbage in the one commit dedicated to
  cleaning comments up, and it lands on main until somebody next edits that block.
- **Fix:** drop the ` / ` so the block ends `disables the shortcut. */`.
- **Verify:** `cd ui-tui && npx tsc --noEmit && npx prettier --check "src/**/*.{ts,tsx}"`.
  I grepped the same pattern across all 11 files this commit touches; this is the only
  occurrence.

### Verified on this round, so no need to re-check

- **`cd591a6` (R1/R3).** `test_cli_exit.py:28` now skips instead of asserting
  process-global thread state, and `tests/conftest.py` no longer credits the pytest
  segfault to a lancedb thread the suite never starts. The status-bar patterns live in
  `tests/tui/autotest/statusbar.py` and both e2e files import them; `re` is still used
  in both, so nothing was left dangling.
- **`d8f3c86` (slash routing).** Every claim holds: `raven_cli` is gone repo-wide (no
  `COMMAND_REGISTRY` anywhere), `createSlashHandler.ts:45-51` is the right anchor,
  `findSlashCommand` is `byName.get(name.toLowerCase())` over names plus aliases, and
  all six `NATIVE_MUTATING_COMMANDS` names are in the local registry - so the dropped
  native branch could not change an outcome, and "native" was never a route the
  implementation had. The negative and alias fixtures are real: no local `channels`
  command, `bg` is an alias of `background` (`session.ts:46-48`), `model` exists
  (`session.ts:70`).
- **`9c03caa` really is comments only.** Read every hunk of `platform.test.ts` (45
  lines), `useConfigSync.test.ts` (31), `createSlashHandler.test.ts` (19) and
  `session.ts` (22): no assertion added, changed or removed, and `session.ts`'s
  `const parsed = r.record_key ? ... : undefined` is untouched.
- **The sweep is complete.** Zero repo-wide hits for `19835`, `18994`, `19194` or
  `14045`, and zero `raven_cli` / `tui_gateway` references left in `ui-tui/src`. The
  remaining `Copilot` matches are all GitHub Copilot the provider.
- **The three deleted python back-references were genuinely fictional.** No
  `raven_cli` or `tui_gateway` module exists, `busy_input_mode` appears nowhere in
  python, and `raven/tui_rpc/methods/_stubs.py` answers `voice.toggle` with "voice not
  supported in Raven v0.1".
- **`R2` unchanged and still correct.** `raven/cli/commands.py:177-178` cannot fire:
  nothing under `raven/` imports lancedb (all hits are prose), and everos runs
  out-of-process via `subprocess.Popen` (`_server.py:56`), so its lancedb thread never
  lives in the CLI process. Still worth a separate PR, still out of scope here.

### Residual risk, stated for the record

CI does not run `tests/integration/*` (deselected) or `tests/tui/autotest` (needs
`tui-use` and a reachable model). So the two deletions, the `xfail` change in
`test_everos_backend_e2e.py`, the new multi-turn e2e and the three timing fixes rest
on the local runs reported in the description rather than on a machine. That is the
repo's existing CI gap, not something this PR introduced.

arelchan (approved):

PR #251, opened 2026-07-31
feat(tools): read images with read_file, and fix four silent type-check bugs

gloryfromca:
## Summary

`read_file` could not read images: it called `read_text(encoding="utf-8")`, so a PNG raised `UnicodeDecodeError` on byte 0. Raven has supported image *understanding* since its first commit, but only on the inbound path (an IM attachment goes into a `user` message, which every provider accepts). The path an agent needs -- looking at a file on disk -- puts the image in a **tool result**, and that is where it gets hard: OpenAI's Chat Completions types `role:"tool"` content as `string | ChatCompletionContentPartText[]`, so an image has nowhere to go at the schema level.

This branch adds that capability, and fixes four pre-existing defects found on the way there. Nineteen commits, each independently reviewable; they read best oldest-first.

### Pre-existing bugs (independent of the feature)

| Commit | Defect |
|---|---|
| `fix(utils)` | Image tokens were counted as the length of the base64 data URI. Measured on a 2.9MB PNG: **2,813,577 estimated vs ~1,300 real**, over by ~2168x, and the factor grows linearly with file size. This starves the history budget and can trip emergency shrinking on a prompt that would have fit. |
| `fix(providers)` | The Responses converter ended with `json.dumps(content)` for any non-string tool result, so a multimodal result reached the model as its own base64 rendered as prose. Nothing raised. |
| `fix(*)` (MCP) | Same shape, different call site: the MCP wrapper ended with `str(block)`, and `str()` on a pydantic model is its repr -- so an MCP tool returning `ImageContent` put `type='image' data='iVBORw0KGgo...'` into the prompt. Live today for any MCP server that returns a screenshot. |
| `feat(agent)` (persistence) | `_save_turn` guarded truncation with `if role == "tool" and isinstance(content, str)`, and the branch below it was `elif role == "user"`. A list-shaped tool result matched neither, so it was persisted verbatim -- the 16,000-char cap never applied. |

All four are the same failure mode: **a type check standing in for a capability check**. `isinstance(x, str)`, `json.dumps`, `str()` each answer "can I stringify this" when the question was "what is this" -- and they answer wrong silently as soon as the type widens. The persistence one is the only defect here that cannot be reverted: code can be changed back, bytes already written to session JSONL cannot.

### How images travel

- **Detection** is by magic bytes, not extension, so a PNG named `.txt` is handled as an image and a `.png` full of text is not.
- **Preprocessing** clears three independent limits, because exceeding any one gets the image rejected or silently altered: 2000px per side, 4.5MB of base64 (under the 5MB per-image cap Bedrock and Vertex AI enforce -- one payload per turn may go to any target, so it must satisfy the lowest bound), and 1568 patch tokens. Both size limits are checked rather than just the long edge: the docs are explicit that "Images larger than either limit are downscaled", so passing the pixel check alone still leaves an image the server resizes, which makes any local token estimate wrong.
- **Delivery** depends on the transport, not the model. `supports_image_tool_result` returns True only for measured targets (`anthropic`, `vertex_ai`, `bedrock`) and, behind a gateway, only for measured model families (`anthropic/claude-`, `google/gemini-`); everything else takes a placeholder-plus-attachment path, the shape OpenClaw uses against Chat Completions endpoints. The list is deliberately a whitelist of what was probed rather than a blacklist of known failures, because the dangerous failure is silent: `openrouter/openai/gpt-4.1-mini` accepts the request, discards the image and confabulates a colour, while the same image in a user message gets a correct answer from it. The gateway key is a family prefix rather than the creator segment for the same reason -- `openrouter/google/gemma-3-27b-it` is served only by third-party OpenAI-compatible hosts, two of which reject the payload with a 422 while two accept it, and OpenRouter picks the host per request. Absences mean unprobed, not known-bad: `gemini` almost certainly works (LiteLLM routes it through the same transform as the listed `vertex_ai`) but there was no direct key to prove it. `ProviderSpec.image_tool_result_override` lets a deployment opt in.
- **Lifetime** is one turn. Images degrade to a placeholder on the way to the session, so they are never replayed on resume or re-sent to the model in a later turn. The text kept alongside names the file path, so the model can re-read an image rather than only learning that one existed. Inbound IM images now do the same, mirroring what the non-image branch of that loop has always done for PDFs.

### Design notes for review

- **`ToolOutput` stays a `str` subclass.** Its docstring already explains why -- the sentinel executor, subagent manager, context curator and tracing all put the registry's return value straight into a message or preview. `blocks` rides along as an attribute, so those four callers are untouched and need no knowledge that multimodal results exist. This makes `model_text` a contract rather than a courtesy: it is what text-only transports send and what lands in the transcript, so a tool setting `blocks` puts the metadata and the path in it, never "see the image above".
- **Graceful degradation is built into the type.** A provider that ignores `blocks` still gets a self-sufficient text result naming the file, rather than nothing.
- **The content-part shape is written in one place.** It was hand-written in five and re-derived in nine, with nothing verifying any of them -- a mistyped key (`imageURL`) would drop the picture with a clean log. Construction now goes through `image_block()` / `text_block()`, and a test greps the tree to keep the literal from coming back.
- **TypedDicts type what Raven produces, not what it reads.** Inbound content legitimately carries parts the union does not model (an Anthropic part with `cache_control`, an MCP audio block, a future extension), and pass-through code would otherwise become a type error for behaving correctly. Read-side helpers keep taking `Any` and checking shape at runtime; a test pins that permissiveness. Worth stating plainly: CI runs no type checker, so these annotations gate nothing on merge -- they document the contract and give an editor language server something to check.
- **Security posture is unchanged in kind.** `wrap_untrusted_blocks` fences text parts and leaves image bytes untouched, because a text boundary cannot contain pixels and rewriting the bytes would corrupt the picture. Instructions rendered *into* an image stay unfenced; the defense for those is the one that already covers text -- privileged actions are gated by policy regardless of what the model just looked at. This matches OpenClaw and Claude Code, neither of which tracks taint from an image to a later action.
- **Pillow is pinned directly.** It already arrives via `everos[multimodal]` -> `everalgo-parser`, but that is a three-hop transitive path, so it is pinned for the same reason `tiktoken` already is.

### Deliberately not included

Terminal rendering of images (needs the Kitty/iTerm2 graphics protocol and would hit the 1 MiB JSON-RPC frame cap; the goal here is images reaching the model, not the eye), a structured `details` channel for program-readable tool metadata (nothing needs it yet -- `CallToolResult.structuredContent` is the shape to copy when something does), and native PDF, audio or video input.

### Fixed during review

Three defects found in review, all on the fallback path, plus one recovery mechanism added on top:

| Item | Defect |
|---|---|
| R1 | `read_file` on a `.svg` regressed from returning the text to returning an error: the extension guess sent XML into the image branch, Pillow could not decode it, and nothing fell back to the text path. |
| R2 | The image carrier message was appended inside the per-tool-call loop, so a batch whose non-last call returned an image put a `user` message between two tool results. Measured: gpt-4o returns 400, Anthropic accepts it, so it fires only on the transport that actually takes this path. |
| R3 | The capability verdict was cached from `self.model` while the turn ran on `effective_model`, which the router selects per turn. A wrong `True` is the silent image loss the whitelist exists to prevent. |
| R5 | The carrier message persisted as a user turn holding `[image]`, which reads on resume and in export as something the user sent. Filtered where the turn's message list is finalized, upstream of persistence, `after_turn` extraction and `backend.store` alike. |
| N1 | The gateway whitelist keyed on the creator segment, admitting routes served by unmeasured third-party hosts. Narrowed to measured family prefixes. |

`ErrorClassification` gained a fourth flag so a refused image is recoverable rather than fatal: classify, move the picture to a user message, retry the iteration once, cache the verdict per model. Worth stating the limit plainly -- this heals a wrong guess only when the endpoint says so. A silent drop stays undetectable, which is why the whitelist is the primary defence and the retry is the backstop.

## Type

- [x] Fix
- [x] Feature
- [x] Refactor

## Verification

Commands run:

```
uv run --extra dev ruff check raven/ tests/     # All checks passed
uv run --extra dev ruff format --check raven/ tests/
uv run --extra dev pytest tests/ -q             # 23 failed, 5036 passed, 30 skipped
make check-commits                              # clean
make check-large-files                          # clean
```

The 23 failures are pre-existing on `main` and unrelated (`test_sandbox_debug_server.py`, `test_cli_theme.py`). Verified by running the suite in a clean worktree at `origin/main` and diffing failing test IDs: **no test fails on this branch that passes on `main`**. The baseline worktree showed three extra failures in `test_channels_errors.py`, which are an artifact of `slack_sdk` being absent from that fresh environment, not a fix from this branch.

65 tests in `tests/test_read_file_image.py`, no skips, no xfail, no weakened assertions. One pre-existing test changed on purpose: `test_image_inlined_as_block` pinned the old user-content text block exactly, and now pins the new shape plus asserts the path and re-read hint are present.

Every commit was checked to import cleanly on its own, since the series is meant to be read commit by commit.

Two surfaces, both updated: `read_file`'s tool description, and `TOOLS.md` (the workspace file for non-obvious tool constraints, loaded into the system prompt) which now documents content-based detection, downscaling, the one-turn lifetime, and the follow-up-message delivery path. Checked that no other doc enumerates tools -- neither README nor README.zh-CN mentions `read_file`.

## Risk

Backward compatibility: a tool returning `str` behaves exactly as before, and `read_file` on a text file is unchanged. `blocks` is additive and only `read_file` and the MCP wrapper set it. The registry keeps its 4-argument `add_tool_result` call for text results so no existing caller or test double sees a signature change.

Rollback: reverting the branch restores prior behavior. The one asymmetry worth naming is the persistence fix -- if the block-carrying commit had landed without it, session JSONL written in between would contain base64 permanently. They are in the same commit for that reason.

## Related Issues

N/A


0xKT:
Reviewed at `5307e84` in a detached worktree, with `origin/main` (`16bdca7`) checked out
side by side as the behaviour baseline. CI is green and all 43 new tests pass; none of the
items below are covered by them. IDs are stable, please reference them in replies.

Verdict: three small fixes before merge, two follow-ups.

Verification boundary, stated up front: R1 is measured end to end (same file, same call,
both branches). R2 and R3 are measured up to the message list that gets built; the
consequence (a 400 from Chat Completions, a silently dropped image) follows from the API
contract and was **not** confirmed against a live endpoint. Fixes for R1 and R2 were
applied to a scratch copy and are verified below.

---

## R1 (before merge) -- `read_file` on a `.svg` regresses from returning the text to returning an error

**File:** `raven/agent/tools/filesystem.py:115-131`, comment at `raven/agent/tools/media.py:29`

**What:** when the magic bytes do not match, the extension guess is trusted
(`mimetypes.guess_type` maps `.svg` to `image/svg+xml`, which passes
`startswith("image/")`), so the image branch runs. Pillow has no SVG decoder, the
`except Exception` catches `UnidentifiedImageError`, and the tool returns an error
string instead of the file. Nothing falls back to the text path.

The `media.py:29` comment states "Anything else (BMP, TIFF, HEIC, SVG) is converted to
JPEG rather than rejected". That holds for BMP/TIFF/ICO but not for SVG (XML, no Pillow
decoder) or HEIC (needs `pillow-heif`).

**Repro:** same file, same call, both branches:

```
                 main (16bdca7)                          PR (5307e84)
diagram.svg  ->  1| <?xml version="1.0"?> ...            Error decoding image /tmp/.../diagram.svg
empty.svg    ->  (Empty file: /tmp/.../empty.svg)        Error decoding image /tmp/.../empty.svg
pic.png      ->  Error reading file: 'utf-8' codec ...   ToolResult(model_text='[image: ...')
pic.bmp      ->  Error reading file: 'utf-8' codec ...   ToolResult(model_text='[image: ...')
notes.txt    ->  1| line one ...                         1| line one ...
```

PNG/BMP/TIFF/ICO are the intended new capability. SVG is a text format an agent reads and
edits, and it stops being readable.

**Fix:** only report a decode failure when the bytes actually sniffed as an image; when it
was an extension guess, fall through to the text path.

```diff
-            mime = detect_image_mime(head)
+            sniffed = detect_image_mime(head)
+            mime = sniffed
             if mime is None:
                 guessed = mimetypes.guess_type(fp.name)[0]
                 # Extension says image but the magic bytes did not match a format
-                # we inline (BMP/TIFF/HEIC/SVG): still an image, prepare_image
-                # converts it.
+                # we inline (BMP/TIFF/ICO): still an image, prepare_image converts
+                # it. The guess can also be wrong -- SVG is XML, and a .png stub
+                # may hold text -- so a decode failure here falls back to reading
+                # the file as text rather than refusing to read it at all.
                 if guessed and guessed.startswith("image/"):
                     mime = guessed
             if mime is not None:
@@
                 except ImageTooLargeError as e:
                     return f"Error: {e}"
                 except Exception as e:
-                    return f"Error decoding image {path}: {e}"
+                    if sniffed is not None:
+                        return f"Error decoding image {path}: {e}"
```

**Verify:** with the patch, `diagram.svg` and `empty.svg` return the same text as `main`
while PNG/BMP/TIFF/ICO still return a `ToolResult`. Suggested test in
`tests/test_read_file_image.py`:

```python
def test_read_file_falls_back_to_text_when_the_extension_lied(tmp_path: Path) -> None:
    (tmp_path / "diagram.svg").write_text('<svg xmlns="http://www.w3.org/2000/svg" width="1" height="1"/>')
    out = _read(tmp_path, "diagram.svg")

    assert isinstance(out, str)
    assert "<svg" in out
```

---

## R2 (before merge) -- the placeholder fallback inserts a `user` message between two tool results

**File:** `raven/agent/loop/main.py:1760-1761`

**What:** `messages.append({"role": "user", "content": attach_blocks})` runs inside
`for tool_call in response.tool_calls`. When an assistant message carries several tool
calls and a non-last one returns an image, the image message lands between two tool
results. OpenAI Chat Completions rejects that: an assistant message with `tool_calls` must
be followed by the tool messages answering each `tool_call_id`. This is the path every
non-`anthropic`/`vertex_ai`/`bedrock` target takes, i.e. exactly where the fallback is
needed. `HistoryTrimmer.structural_errors` does not catch it because it validates pairing,
not adjacency.

**Repro:** fake provider, one assistant message with `read_file(a.png)` + `read_file(b.txt)`,
`supports_image_tool_result` False. Messages sent on the next iteration:

```
role=system
role=user
role=assistant  tool_calls=['c1', 'c2']
role=tool       tool_call_id=c1
role=user       content=list        <-- c2 still unanswered here
role=tool       tool_call_id=c2

HistoryTrimmer.structural_errors -> []
```

Trigger in practice: "look at these two screenshots" on any OpenAI-compatible endpoint.

**Fix:** collect across the batch, append once after the loop. It has to go after the
loop-break nudge, which requires `messages[-1]` to still be the tool result.

```diff
+                # Images the transport cannot carry in a tool result. Collected
+                # across the whole batch and attached *after* the last tool
+                # result: a user message between two tool results would leave an
+                # assistant tool_call unanswered at the point the API checks it.
+                pending_images: list[dict[str, Any]] = []
                 tool_call_dicts = [tc.to_openai_tool_call() for tc in response.tool_calls]
@@
                     if attach_blocks:
-                        messages.append({"role": "user", "content": attach_blocks})
+                        pending_images.extend(attach_blocks)
@@
                     loop_fail_streak = 0  # fire once per fresh streak
+                # After the nudge, which needs the last message to still be the
+                # tool result it appends to.
+                if pending_images:
+                    messages.append({"role": "user", "content": pending_images})
                 prev_had_tool_calls = True
```

**Verify:** with the patch the order becomes `assistant -> tool(c1) -> tool(c2) -> user`,
and 103 tests across `test_read_file_image.py`, `test_plugin_tools.py`,
`test_agent_loop_tool_loop_break.py`, `test_agent_loop_context_overflow.py`,
`test_openai_codex_provider.py`, `test_token_estimation.py`, `test_security_trust.py`
still pass. Worth asserting the order in a loop-level test (see R4).

While in there: `main.py:1782` does `str(messages[-1].get("content", "")) + ...` on what is
now possibly a block list. Checked and it is unreachable today (an image result is not a
hard failure, so the streak resets before the nudge can fire on it), so this is not a bug
report -- just noting that an `isinstance(..., str)` guard is one line and this is the same
shape the branch is fixing elsewhere.

---

## R3 (before merge) -- the capability is cached from `self.model`, but the turn may run on another model

**File:** `raven/agent/loop/main.py:755-770`, override read at `main.py:1545`

**What:** `_supports_image_tool_result()` resolves `self.model` and caches one bool for the
loop's lifetime. The model the turn actually runs on is `effective_model = model or self.model`
(`main.py:1545`), and the caller passes `model=routed_model` from
`self.router.select_model_chain(content)` (`main.py:2233`) -- the router picks a model per
turn and can hand back a chain of fallbacks. So this is not a hypothetical path: with a
router configured, `self.model` and the model in flight routinely differ, and the log line
right there (`Router: {} -> {}`) says so.

Measured:

```
self.model = anthropic/claude-opus-4-5  -> cached True
truth for a turn running gpt-4o         -> False   (cached True is what gets used)
```

A wrong `True` puts the image blocks inside the tool result for a Chat Completions target,
where the schema has nowhere to put them. That is the silent image loss the whitelist was
built to prevent -- `openrouter` is excluded even when it fronts Claude, on the grounds
that a wrong `True` loses the image quietly, and this cache reintroduces exactly that on a
different axis. Raising it to before-merge for that reason: it undercuts the branch's own
stated safety argument.

**Fix:** take the model as an argument (`self._supports_image_tool_result(effective_model)`)
and key the cache on it (`dict[str, bool]`). The `fallback_models` retry path deserves a
thought too -- by the time LiteLLM switches target, the blocks are already in the message
list -- but a per-model cache is the part that is cheap and clearly right.

---

## R4 (follow-up) -- no loop-level test for the fallback path

The new tests cover `image_placeholder_text` in isolation; nothing drives the loop with a
provider that returns False. That is the gap R2 came through. One test with two tool calls
asserting `[m["role"] for m in sent]` closes it.

---

## R5 (follow-up) -- the synthetic image message persists as a user turn

**File:** `raven/agent/loop/main.py:2357-2378` (the `elif role == "user"` branch)

`_save_turn` strips the base64 correctly, but the attachment message survives as a user
turn with no text:

```
{'role': 'tool', 'content': '[image: /tmp/a.png] | 40x40px'}
{'role': 'user', 'content': [{'type': 'text', 'text': '[image]'}]}
```

On resume and in `session/export.py` this reads as the user having sent `[image]`, which
they did not. The tool result above already names the path, so the message carries nothing
that is not already persisted. Suggest marking it non-persistent the way
`_recovery_synthetic` scaffolding already is (a separate key, not that one, to keep the
meanings apart).

---

Not blocking, worth saying: the token-accounting fix is the most valuable part of this
branch (2,813,577 estimated vs ~1,300 real on a 2.9MB PNG is a budget bug that would bite
long before anyone asked for images), and routing every content part through
`image_block()` with a grep test to keep it there is the right call.

gloryfromca:
Thanks -- all three before-merge items are fixed, R4 is closed, and R5 is fixed too. Rebased onto `origin/main` (`af83cbf`), which required merging the approval flow's `retryable`/`abort_action` into the same `ToolResult`/`ToolOutput`/`registry.execute` surface this branch touches; notes on that at the end.

| ID | Status | Commit |
|---|---|---|
| R1 | fixed | `ad4f134` |
| R2 | fixed, and now measured live | `5f6e907` |
| R3 | fixed | `88ac2db` |
| R4 | closed | `5f6e907`, `533e4ea` |
| R5 | fixed | `5f6e907` |

---

## R2 -- your inference was right, and here is the wire evidence

You flagged that R2's consequence was not confirmed against a live endpoint. It is now. Same two-tool batch, image from the first call, only the position of the carrier message differs:

```
openrouter/openai/gpt-4o
  tool(c1), user, tool(c2)  -> 400 "An assistant message with 'tool_calls' must be
                                    followed by tool messages responding to each
                                    'tool_call_id'. The following tool_call_ids did..."
  tool(c1), tool(c2), user  -> 200  "The image is primarily blue in color."

openrouter/anthropic/claude-sonnet-4.5
  tool(c1), user, tool(c2)  -> 200
  tool(c1), tool(c2), user  -> 200
```

Two things follow. Anthropic does not enforce the adjacency rule, so this only fires on Chat Completions -- which is the only transport that takes this path at all, meaning it fires whenever it can. And the batched form does not merely avoid the 400: the model named the test image's colour, so the whole placeholder-plus-attachment mechanism is confirmed end to end on a real OpenAI endpoint.

Applied your fix as written -- collect across the batch, append after the loop-break nudge. One addition from the rebase: the append also has to sit after the approval flow's `abort_action` branch, which ends the turn in runtime code. There is no further model call to show a picture to, so an aborted action drops it rather than leaving a dangling user message. That is commented at the call site.

On the `main.py:1782` note: agreed, and unreachable for the reason you gave -- an image result is not a hard failure, so the streak resets before the nudge can fire on it. Worth flagging that the batching fix thinned the surface slightly: the inline append used to leave a `user` message as the last entry right after an image result, so `messages[-1].get("role") == "tool"` was false there for a second, independent reason. That barrier is gone now and only the streak reset remains. Still unreachable, but on one argument instead of two -- say the word and I will add the `isinstance` guard.

## R1

Fixed as you suggested -- track whether the MIME was sniffed or guessed, and let a guess that will not decode fall through to the text path. Reproduced your table before fixing:

```
diagram.svg  ->  Error decoding image ...     (main: 1| <?xml version="1.0"?> ...)
empty.svg    ->  Error decoding image ...     (main: (Empty file: ...))
```

I considered the narrower fix of whitelisting the extensions worth trusting (`bmp`/`tiff`/`ico`) and rejected it: `.ico` resolves to two different MIME spellings depending on the platform's `mimetypes` table, and `.tif`/`.tiff` differ, so the whitelist becomes its own source of drift. The sniffed-vs-guessed flag covers SVG, HEIC and anything else Pillow cannot open, for one variable.

Also corrected the `media.py:29` comment, which claimed SVG and HEIC convert to JPEG. No vision API accepts SVG at all (Anthropic, OpenAI and Gemini list raster formats only), so the comment now says Pillow-undecodable formats are not images as far as that module is concerned.

Your suggested test is in, parameterized over `.svg` and a `.png` stub holding text, plus the inverse: a file whose PNG magic bytes are intact but whose body is corrupt still returns `Error decoding image`, so the fallback cannot swallow a genuine decode failure.

## R3

Fixed exactly as you described: take the model as an argument, key the cache on it. Your framing -- "it undercuts the branch's own stated safety argument" -- is the right one, and it is what made me look at the rest of that argument, which turned out to be partly wrong. See below.

On `fallback_models`: still open, and I agree it deserves a thought rather than a fix here. By the time LiteLLM switches target the blocks are already in the list, so the correct place is probably the same 400 recovery described below rather than the pre-flight probe.

## R4

Closed. Four loop-level tests now drive the fallback with a provider that returns `False`:

- two tool calls, one returning an image, asserting `[assistant, tool, tool, user]`
- the carrier message is absent from `session.messages`
- the carrier message is absent from what `context_engine.after_turn` receives
- a refused image is demoted and the turn recovers (see below)

The ordering test was mutation-checked: reverting the collect-across-batch change turns it red.

## R5

Fixed, with one change from your suggestion. Marking it non-persistent in `_save_turn` is not sufficient -- the same list feeds `context_engine.after_turn` and `_dispatch_backend_store`, so guarding only the JSONL write still hands the base64 to extraction and indexing. The filter now runs where the turn's message list is finalized, upstream of all three. The `_save_turn` guard stays as the last gate before the one write that cannot be undone, and is unit-tested directly.

Used a separate key as you asked, for the reason you gave.

## New in this round: recovering from a refused image

R3 made me re-measure the capability argument, and one row of it was wrong. The old comment claimed a wrong `True` on a gateway fails loudly. It does not always:

```
openrouter/anthropic/claude-sonnet-4.5  -> 200, answered "blue"
openrouter/google/gemini-2.5-flash      -> 200, answered "Blue"
openrouter/openai/gpt-4o                -> 400, refused
openrouter/openai/gpt-4.1-mini          -> 200, answered "red"
```

Two models on the same backend, two different failure modes: one refuses, the other drops the image and confabulates. The same image in a user message gets a correct answer from both, so the model is not blind -- the picture is discarded in transit. That comment is corrected, and it is now the stated reason the target list stays a whitelist of measured backends.

For the loud half, added a fourth `ErrorClassification` flag mirroring `should_compress`: classify, apply one bounded correction, retry the same iteration, cache the verdict per model. The correction rebuilds onto the placeholder path rather than inventing a third message shape, which a test pins by comparing against `image_placeholder_text` directly.

The rejection wordings are half measured here and half borrowed from Hermes (MIT, already vendored in `LICENSES/`), which had accumulated them across Xiaomi MiMo, Alibaba DashScope and OpenAI-compatible servers failing schema validation. Neither list covered the other. Its `remember_model=False` distinction for 413 is a real gap on our side -- a payload-too-large is not a capability verdict -- but 413 recovery is a separate capability and I left it out rather than widen this branch.

Worth being explicit about the limit: this makes a wrong guess self-healing only when the endpoint says so. A silent drop stays undetectable by any mechanism, which is why the whitelist is still the primary defence and the retry is the backstop.

## Rebase notes

`origin/main` moved to `af83cbf` (approval flow, #239), which changed `ToolResult`, `ToolOutput` and `ToolRegistry.execute` -- the same three files this branch changes. Merged rather than re-applied:

- `ToolResult` and `ToolOutput` carry `retryable`, `abort_action` and `blocks`. `blocks` moved behind the `*` to match the keyword-only convention #239 introduced.
- `registry.execute` threads all three. The error path keeps #239's `suffix = _hint if retryable else ""` and drops `blocks`, since an error replaces the result it came with.
- The tool loop runs the block routing first (it decides `model_text`), then the `abort_action` branch.

## Verification

```
uv run --extra dev ruff check raven/ tests/            # All checks passed
uv run --extra dev ruff format --check raven/ tests/
uv run --extra dev pytest tests/ -q                    # 23 failed, 5029 passed, 30 skipped
make check-commits                                     # clean
make check-large-files                                 # clean
```

The 23 are the `origin/main` baseline (`test_sandbox_debug_server.py` x22, `test_cli_theme.py` x1); the failing ID set is unchanged, not just the count. 58 tests in `test_read_file_image.py` now (18 added this round), no skips, no xfail. Each of the four new commits was checked to lint and import on its own.

R1, R2 and R5 were each mutation-checked -- short-circuit the fix, watch the test go red -- because the first version of the R5 test passed with the guard removed: two independent gates both prevented the write, so the test could not tell which one did it. That is what prompted the `after_turn` test.

0xKT:
Re-reviewed at `533e4ea` (detached worktree, `af83cbf` checked out beside it as baseline).

R1, R2, R3, R5 re-verified by running them, not by reading the diff: `.svg` and an empty
`.svg` return text again while PNG/BMP/TIFF/ICO still return a `ToolResult`; a two-call
batch now sends `assistant -> tool(c1) -> tool(c2) -> user`; the cache is keyed by model
and the call site passes `call_model or effective_model`; a full turn leaves the carrier
message and its base64 out of `session.messages` entirely. R4's four loop tests are there
and the mutation check is the right way to have proved the ordering one. 228 tests pass
across the image, approval and registry files the rebase touched, and the merge of #239's
`retryable`/`abort_action` into `ToolResult`/`ToolOutput`/`registry.execute` reads clean.

Your R5 fix is better than what I suggested -- I stopped at `_save_turn` and missed that
the same list feeds `after_turn` and `backend.store`. And thanks for running the live
probe on R2; that was my inference and it should not have shipped on my say-so alone.

One item before merge, and it is about the new gateway whitelist rather than anything I
raised last round. Same ID scheme continued: N-prefix for this round.

---

## N1 (before merge) -- `GATEWAY_IMAGE_TOOL_RESULT_BACKENDS` admits routes nobody measured

**File:** `raven/providers/capabilities.py:57-58`, resolved at `:100-101`, helper at `:105-112`

`_gateway_backend` returns the creator segment, so the `"google"` entry admits every
`openrouter/google/*` model, not the one that was probed. That segment names a **model
creator, not a serving stack**, and the two come apart immediately:

```
openrouter/google/gemini-2.5-flash  -> hosts: Google, Google AI Studio        (measured, works)
openrouter/google/gemma-3-27b-it    -> hosts: DeepInfra, Parasail, Nebius,
                                              Novita, Phala                   (never measured)
```

(OpenRouter's public endpoints API, no key needed.) Zero Google hosts on the second row --
five third-party OpenAI-compatible inference providers, which is precisely the environment
this module's own docstring describes as having nowhere to put an image in a tool message.
The measured route and the admitted route share no serving infrastructure.

That matters more than a normal untested-path complaint because of which failure mode is
reachable there. `should_drop_tool_images` fires only on a body matching
`_TOOL_IMAGE_REJECTION_PATTERNS` (`providers/base.py:23-39`, classification at `:452`).
A permissive server that accepts the list content, discards the image part and returns 200
never reaches `_demote_tool_images` (`agent/loop/main.py:1449`, gated at `:1755-1780`), and the per-model
cache (`main.py:777-794`) only ever learns `False` from a loud refusal -- so a silent drop
locks in a wrong `True` for the life of the process. That is the one failure the whitelist
exists to prevent, and the recovery you added deliberately does not cover it.

To be fair to your own evidence, and correcting how I put this to a colleague earlier: the
silent drop you measured was on `openrouter/openai/gpt-4.1-mini`, and `openai` is not
whitelisted. So "he measured a silent drop on this gateway" is true of the gateway but not
of a whitelisted backend, and I am not claiming otherwise. The point is that the same
mechanism becomes reachable *inside* the `google` entry via gemma, on hosts of exactly the
kind where you observed it.

**Fix:** keep the gateway support, narrow the key from creator segment to the measured
model-family prefix. `str.startswith` takes a tuple, so this stays one comparison:

```diff
-GATEWAY_IMAGE_TOOL_RESULT_BACKENDS = frozenset({"anthropic", "google"})
+# Measured model-family prefixes, not the bare creator segment: "google" also
+# names gemma-*, which OpenRouter serves only from third-party OpenAI-compatible
+# hosts (DeepInfra, Parasail, Nebius, Novita, Phala) -- none of them Google, none
+# of them probed.
+GATEWAY_IMAGE_TOOL_RESULT_PREFIXES = ("anthropic/claude-", "google/gemini-")
@@
     if target in GATEWAY_TARGETS:
-        return _gateway_backend(model) in GATEWAY_IMAGE_TOOL_RESULT_BACKENDS
+        return _gateway_route(model).startswith(GATEWAY_IMAGE_TOOL_RESULT_PREFIXES)
@@
-def _gateway_backend(model: str) -> str:
-    parts = model.split("/")
-    return parts[1].lower() if len(parts) >= 3 else ""
+def _gateway_route(model: str) -> str:
+    """``openrouter/google/gemini-2.5-flash`` -> ``google/gemini-2.5-flash``.
+
+    A bare ``openrouter/some-model`` yields the model name, which matches no
+    prefix and takes the placeholder path.
+    """
+    _, _, route = model.partition("/")
+    return route.lower()
```

**Verify:** `openrouter/google/gemini-2.5-flash` and `openrouter/anthropic/claude-sonnet-4.5`
stay `True`; `openrouter/google/gemma-3-27b-it`, `openrouter/openai/gpt-4o` and
`openrouter/mystery-model` are `False`. Worth a parameterized test over those five, since
the current whitelist has no test pinning what it excludes.

---

## N2 (before merge, or a deliberate decline) -- the direct `gemini` target is stricter than the same route through the gateway

**File:** `raven/providers/capabilities.py:32`

Measured against this branch:

```
gemini/gemini-2.5-flash             -> False
openrouter/google/gemini-2.5-flash  -> True
```

The path with an extra opaque translation layer is treated more permissively than the
direct one. There is code-level reason to think the direct one works: litellm routes
`gemini` and the whitelisted `vertex_ai` through the same transform --
`llms/vertex_ai/gemini/transformation.py:609` calls `convert_to_gemini_tool_call_result`,
whose signature at `:677` is `Literal["vertex_ai", "vertex_ai_beta", "gemini"]`, and
`factory.py:1539` extracts image parts into `inline_data` for tool results.

I am deliberately **not** asking you to just add `gemini` to `IMAGE_TOOL_RESULT_TARGETS` on
that basis, because "only what was actually probed" is the rule N1 is asking you to hold to
and I should not ask you to break it in the same comment. Either run your existing probe
against `gemini/gemini-2.5-flash` and add it if it names the colour, or leave it out and
note in the comment that the direct target is unprobed rather than known-bad -- so the
asymmetry reads as a gap in coverage instead of a claim about the transport.

---

## N3 (before merge, docs) -- the PR description now contradicts the code, and it becomes the squash body

The description still says openrouter "returns False even when fronting Claude, because
the proxy's tool-result translation is unverified". As of `533e4ea` that is no longer true,
and since this repo squashes with `squash_merge_commit_message=PR_BODY`, that sentence
lands verbatim on `main` as the permanent commit body. Worth updating before merge along
with the Verification block's test counts.

---

## Nits, not blocking

- The claude measurement in the comment at `capabilities.py:40` records a 200 but not
  which of OpenRouter's hosts for `anthropic/claude-sonnet-4.5` answered it (Anthropic,
  Bedrock, Google, Azure). Given the "reassigns providers per request" note two lines
  below, worth saying the host was not pinned rather than leaving it implied.
- Trust-fence asymmetry between the two paths that produce the same placeholder text: on
  the normal path `image_placeholder_text` runs on raw blocks and the whole string is then
  fenced by `add_tool_result`, so the system's own "[1 image attached...]" line sits inside
  the fence; on the demotion path (`_demote_tool_images`) it runs on already-fenced content,
  so that line lands outside it. The demotion path is arguably the correct one. No security
  consequence either way -- the text is code-generated in both -- but they should agree.

Everything else in this round looks right to me, including the parts I would have gotten
wrong: keeping the demotion output identical to `image_placeholder_text` so the retry lands
on a tested shape, gating the retry on `demoted > 0` so an unrelated 400 cannot cause a
blind retry, and putting the batched append after the `abort_action` branch.

0xKT (dismissed):

gloryfromca:
N1, N2 and N3 done at `a24649d`. Nit 1 is answered with data; nit 2 I agree with and am deferring, reasoning below.

## N1 -- confirmed by measurement, and it is worse than untested

You were right, and pinning the hosts turned "nobody measured this" into a demonstrated failure. OpenRouter takes `provider.only` with `allow_fallbacks=false`, so every host can be probed individually:

```
anthropic/claude-sonnet-4.5   Anthropic  blue | Bedrock  blue
                              Azure      blue | Google   blue   -> 4/4 ok
google/gemini-2.5-flash       Google     blue | AI Studio blue   -> 2/2 ok
google/gemma-3-27b-it         Parasail   blue | Nebius    blue
                              DeepInfra  422  | Novita    422    -> 2 of 4 fail
                              Phala      429 (rate-limited, no result)
```

gemma is not merely unmeasured: it works on two of its four answering hosts and hard-fails on the other two, and an unpinned call landed on Parasail and worked. So a creator-level entry buys *intermittent* failure, which is worse than the consistent kind because it looks like a flake. That is now the stated reason in the comment for the key being a family prefix.

Applied your fix as written -- `_gateway_route` plus a tuple to `startswith`. Parameterized test over eight routes including the three you named, `gpt-4.1-mini`, and a mid-string case (`openrouter/someone/not-anthropic/claude-clone`) so the prefix cannot match at a segment boundary it should not. Mutation-checked: reverting to the creator segment turns the gemma row red.

One thing to carry forward from those numbers. DeepInfra rejects with `Input should be a valid string` on `messages.2.function...`, and it is a **422, not a 400**. That wording was not in `_TOOL_IMAGE_REJECTION_PATTERNS`, so even the retry would not have caught it -- it classifies only because the body happens to contain `invalid_request_error`, which the first clause matches. Added the pattern. It is unreachable for gemma now that the prefix excludes it, but it is real wording from a real OpenAI-compatible host and the next such host may be whitelisted.

## N2 -- declined as you framed the option, because I cannot probe it

Took your second option rather than the first. There is no direct Gemini key on this machine -- only OpenRouter -- so running the probe against `gemini/gemini-2.5-flash` is not something I can do, and adding it on the strength of the LiteLLM transform alone is exactly the rule N1 asks me to hold to. The comment on `IMAGE_TOOL_RESULT_TARGETS` now says absences mean unprobed rather than known-bad, names the transform you found, and states that the `gemini/*` vs `openrouter/google/gemini-*` asymmetry is a gap in coverage rather than a claim about the transport.

If you have a Gemini key, that probe is about thirty seconds and I would rather have the row than the caveat.

## N3 -- description updated

You were right that this one matters more than a docs nit, since the body lands verbatim on `main`. Rewrote the Delivery bullet (it now describes family prefixes and leads with the silent-drop failure rather than the unverified-proxy claim, which was the part that had gone stale), corrected the test counts, and added a short table of what the two review rounds changed -- the squash body is the permanent record of this work, and it was describing a design that no longer exists.

## Nit 1 -- answered

Not implied any more: all four hosts for `anthropic/claude-sonnet-4.5` were pinned and each answered correctly, Azure included. That was the row I was least sure about, so thanks for asking -- I had been treating the unpinned 200 as evidence about the route when it was only evidence about one host.

## Nit 2 -- agreed, deferring

Verified it: on the normal path `image_placeholder_text` runs on raw blocks and `add_tool_result` then fences the whole string, so the system's own `[1 image attached...]` line ends up inside the fence; on the demotion path it runs on already-fenced parts, so the line lands outside. You are right that the demotion path is the correct one -- a fence means "do not follow instructions in here", and putting our own note inside it tells the model to distrust us.

Deferring rather than folding it in: fixing the normal path means splitting `image_placeholder_text` so the note can be appended after fencing rather than before, which touches the security module's call shape. That is the one structural change left in this set and it has no behavioural consequence today (both strings are code-generated), so it is better as its own change than as a sixth item in a branch that is otherwise ready. Will open it with the 413 `remember_model=False` gap, which is the other thing I left out.

## Branch state

`main` moved twice while this round was in flight and the branch picked up a merge commit from another machine (`2e89b05`, #252). Rebased my commit on top of that rather than force-pushing over it, so the push was a fast-forward and nothing of yours was rewritten.

```
uv run --extra dev ruff check raven/ tests/            # All checks passed
uv run --extra dev ruff format --check raven/ tests/
uv run --extra dev pytest tests/ -q                    # 23 failed, 5089 passed, 30 skipped
make check-commits                                     # clean
```

Same 23 baseline failures (`test_sandbox_debug_server.py` x22, `test_cli_theme.py` x1), ID set unchanged. 65 tests in `test_read_file_image.py`.

0xKT (approved):

Issue #197, opened 2026-07-23
feat: highlight MiniMax as an open-source partner in onboarding


Yangtze-Seventh:
## Problem

Raven now supports MiniMax, and EverMind and MiniMax have started an open-source collaboration. However, MiniMax is already present in the provider registry but is not surfaced in the curated provider picker shown during Step 1 of onboarding.

This makes the collaboration—and MiniMax as a supported direct provider—easy for new users to miss. It also misses an opportunity to show that Raven welcomes model providers that collaborate with the project in open source.

## Proposal

Update the Step 1 provider picker so that:

- **MiniMax appears near the top of the list**, ideally immediately after OpenRouter and before the other direct providers.
- MiniMax receives a short, clear collaboration marker, for example: `MiniMax (open-source partner)` / `MiniMax（开源合作伙伴）`.
- The ordering and copy remain neutral and useful to users: OpenRouter may stay the recommended multi-provider gateway, while MiniMax becomes the first highlighted direct provider.
- Add a short note in onboarding or provider documentation inviting other model providers to collaborate with Raven through open-source integrations.

Suggested order:

1. OpenRouter (recommended gateway)
2. MiniMax (open-source partner)
3. OpenAI
4. Anthropic
5. Gemini
6. DeepSeek
7. GitHub Copilot
8. Codex
9. Other

### Acceptance criteria

- MiniMax is visible in the initial provider picker without using the custom endpoint flow.
- English and Chinese labels are included.
- Provider selection, API-key setup, model selection, and connectivity tests work through the existing MiniMax registry entry.
- Tests cover the updated curated provider order and labels.

## Alternatives considered

Keep the existing provider order and mention the collaboration only in the README. This is less effective because onboarding is where new users actually choose a model provider.

## Area

Providers

## Additional context

Relevant implementation points:

- `raven/cli/onboard_commands.py`: `_CURATED_PROVIDERS` controls the Step 1 picker.
- `raven/providers/registry.py`: MiniMax is already registered as a direct provider.

Current onboarding order:

```text
OpenRouter (recommended)
OpenAI
Anthropic
Gemini
DeepSeek
GitHub Copilot
Codex
Other (OpenAI-compatible endpoint)
```

PR #252, opened 2026-07-31
feat(*): offer every supported provider in the onboarding picker

### LINKED ISSUE #203: feat: show all supported providers in the onboarding picker
## Problem

During onboarding, the **Choose your LLM provider** step currently shows only a small curated subset of Raven's supported providers, followed by:

```text
Other (OpenAI-compatible endpoint)
```

Raven's provider registry supports more providers than the onboarding picker displays. When a provider is missing from the visible list, less-experienced users may assume that Raven does not support it. They are then forced into the generic OpenAI-compatible flow, where they need to know the correct base URL, model identifier, and other provider-specific details.

This creates unnecessary setup cost and makes the provider support already implemented in Raven difficult to discover.

## Proposal

Make the onboarding provider picker display **all providers supported for Raven's main chat model**, instead of maintaining a small hard-coded subset.

Suggested behavior:

- Populate the picker from the provider registry, or add explicit registry metadata such as `show_in_onboarding` and ordering/group information.
- Keep intentional ordering rules, such as OpenRouter at the top as the recommended gateway and highlighted open-source partners near the top.
- Let users navigate the complete list with the **Up/Down arrow keys**.
- When the list is taller than the terminal, scroll the visible viewport while keeping the current selection clear.
- Optionally support type-to-filter for long provider lists.
- Keep **Other (OpenAI-compatible endpoint)** as the final item for genuinely unsupported or custom endpoints.
- After the provider is selected and credentials are entered, continue using the existing provider-specific model discovery/selection flow.

Example structure:

```text
Choose your LLM provider:

  OpenRouter (recommended — one key, many models)
  MiniMax (open-source partner)
  OpenAI
  Anthropic
  Gemini
  DeepSeek
  ...all other supported providers...
  GitHub Copilot (OAuth)
  Codex (OAuth)
  ─────────────────────────
  Other (OpenAI-compatible endpoint)

  ↑/↓ Navigate · Enter Select
```

### Acceptance criteria

- Every user-configurable provider supported for the main Raven agent can be discovered in the onboarding picker.
- Users can browse the complete list with Up/Down arrow keys, including in a small terminal window.
- The selected provider uses its existing registry metadata and provider-specific configuration flow.
- `Other (OpenAI-compatible endpoint)` remains available and is always the final option.
- English and Chinese labels are supported.
- Tests ensure that eligible registry providers appear in onboarding and that `Other` remains last.

## Alternatives considered

Keep the current curated list and add documentation for the missing providers. This still requires users to leave onboarding and already know which provider names or endpoints to search for.

## Area

Providers / CLI / TUI

## Additional context

The current picker is defined separately from the full registry:

- `raven/cli/onboard_commands.py`: `_CURATED_PROVIDERS`
- `raven/providers/registry.py`: `PROVIDERS`

Using the registry as the source of truth would reduce the risk that provider support and onboarding visibility drift apart.

Related: #197 proposes highlighting MiniMax as an open-source partner near the top of the same picker.

### LINKED ISSUE #221: fix(tui): show Moonshot models after API key setup
### Summary

In Raven Agent v0.1.6, the TUI accepts and saves a Moonshot API key, then advances to the model selection step with an empty list:

```text
Select model (step 2/2)
Moonshot
no models listed for this provider
```

This makes setup look as if the API key failed, even though it was stored successfully.

In v0.1.6, `_provider_models()` builds the picker list from `providers.<slug>.models` plus `common_models_for(slug)`. Moonshot has no entry in `COMMON_MODELS`, and saving the key does not add or discover models. A newly configured Moonshot provider therefore has credentials but zero selectable models.

Relevant files:
- `raven/tui_rpc/methods/model.py`
- `raven/providers/common_models.py`
- `ui-tui/src/components/modelPicker.tsx`

### Steps to reproduce

1. Start Raven v0.1.6 and open the TUI.
2. Run `/model`.
3. Select Moonshot.
4. Enter a non-empty Moonshot API key and save it.
5. Observe the model selection step.

### Expected behavior

After Moonshot credentials are saved, the user can select a supported Moonshot model without leaving the picker. The curated list should contain supported models, or the empty state should clearly prompt the user to add a model inline and explain that the key was saved successfully.

### Actual behavior

The provider is marked configured, but the list is empty and the picker displays `no models listed for this provider`. The saved configuration has a Moonshot API key while `providers.moonshot.models` remains `[]`, and the previous default model remains selected.

Workaround: press `a`, add the model ID manually, and then select it, or run `/model <model-id> --provider moonshot`.

### Environment

OS: macOS
Shell: zsh
Python: unknown
Node: unknown
Installation method: unknown
Raven: v0.1.6
Surface: TUI `/model` picker
Provider: Moonshot

### Logs or screenshots

```text
Select model (step 2/2)
Moonshot
no models listed for this provider


The API key and session identifier are intentionally omitted.
```

### LINKED ISSUE #248: bug(tui): model candidates come from different sources in the picker and wizard
## Summary

Model candidates come from different places depending on where you are, so the
same provider offers different models in the TUI picker and in the wizard.

## Detail

- The TUI picker lists the models written into the provider's config section,
  then a curated shortlist compiled into the package.
- The onboarding wizard fetches `GET /v1/models` from the provider and offers
  what it returns, falling back to the registry's default model.

Neither is wrong on its own. Together they mean a model offered in one place may
be absent from the other, and a stale shortlist is invisible next to a live
fetch.

## Proposal

Decide one order of precedence and use it in both -- for example the user's
configured models, then a live fetch when credentials allow one, then the
shortlist -- so the two views cannot disagree. The shortlist's role should be
explicit: an offline starting point, not a competing catalogue.

0xKT:
## Summary

The onboarding picker offered 13 hardcoded rows while the provider layer could
already reach every vendor LiteLLM routes to, so a supported provider was only
configurable by hand-editing the config file. This opens the picker up and makes
the model step work for everything it now offers.

The picker lists all 21 registered providers in four groups (key, OAuth, local,
fallback), plus a searchable second step for the vendors Raven carries no spec
for. A spec is metadata, not permission: the name gate now passes those through
and normalizes the spelling instead of rejecting it.

Local deployments (Ollama, vLLM) are configured by address rather than by key.
The old path demanded an API key with a minimum length, which nothing running on
localhost has, so the wizard could not finish. The address field is seeded with
whatever is already configured, falling back to the registry default.

Model candidates come from one chain everywhere: what the user configured, then
the curated shortlist, then LiteLLM's own catalogue. Eleven providers offered
zero models before -- picking Moonshot after saving a key showed "no models
listed for this provider", which reads as a rejected key. Moonshot now offers 22,
Ollama 21, Copilot 27. The catalogue is read from the installed package and
loaded on first use, not at import.

Every model id the wizard returns carries its provider's route prefix. A bare id
is routed by keyword and fallback rather than to the provider just configured, so
a vendor with no spec would have been configured correctly and then called with
another provider's key.

"How is this provider reached" is now answered in one function rather than
derived independently at thirteen call sites: a token file, an address, an
endpoint plus a key, or a key alone. Each of those sites disagreeing with the
others is what produced a rollback that crashed the wizard on OAuth providers, a
menu that offered a key prompt to one, and a failure menu that offered to re-enter
a key a local server does not have. A guard test fails if the auth flags are read
anywhere else.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
uv run --all-extras pytest tests/ -q -p no:randomly
  4939 passed, 30 skipped, 13 deselected

uv run ruff check raven/ tests/    All checks passed
uv run ruff format --check          unchanged
```

Every regression test added here was checked by breaking the implementation it
covers and confirming it fails, in both directions where the invariant has two
sides. That found one guard whose branches were unreachable in effect, which is
deleted rather than pinned.

## Risk

Model candidates track LiteLLM's price table, which ships with the package, so
the list moves when that dependency is upgraded rather than staying fixed. This
is deliberate: pinning a local copy would have put candidates out of step with
the pricing and context-window numbers read from the same table, and an earlier
attempt to pin it also pinned the whole process to a table 278 entries behind.
Candidates are a starting list, not a whitelist -- the field accepts free text.

Providers already in a config file keep working: names are matched
spelling-insensitively, and nothing rewrites existing entries.

Rollback is the revert of this branch. It restores the 13-row picker; providers
configured through the new rows stay in the config file and keep working, since
the execution layer already supported them before this change.

## Related Issues

Fixes #203
Fixes #221
Fixes #248

Partially addresses #197 -- MiniMax is marked as an open-source partner in the
picker, placed above the other direct providers rather than immediately after
OpenRouter. The documentation note inviting other providers to collaborate is
left open for someone else to pick up.

### DIFF
--- a/raven/cli/onboard_commands.py
+++ b/raven/cli/onboard_commands.py
@@ -794,7 +794,7 @@ def _run_oauth_login(provider: str) -> bool:
     from raven.providers.registry import find_by_name
 
     spec = find_by_name(provider)
-    if not spec or not spec.is_oauth:
+    if _credential_kind(provider, spec) != CRED_OAUTH:
         console.print(
             _t(
                 f"  [red]✗ {provider} is not an OAuth provider.[/red]",
@@ -855,7 +855,7 @@ def _verify_provider(provider: str, *, skip_test: bool = False) -> tuple[bool, s
     # A local deployment has no key to verify -- what is being checked is that
     # the address answers, and saying "API key" there describes a field the user
     # was never asked for.
-    if spec and spec.is_local:
+    if _credential_kind(provider, spec) == CRED_LOCAL:
         console.print(_t("  [dim]⏳ Reaching the server…[/dim]", "  [dim]⏳ 正在连接服务…[/dim]"))
     else:
         console.print(_t("  [dim]⏳ Verifying your API key…[/dim]", "  [dim]⏳ 正在验证 API Key…[/dim]"))
@@ -928,6 +928,35 @@ def _model_routes_to_provider(model: str, spec: Any) -> bool:
     Defers to the spec so this guard cannot disagree with the routing it guards.
     """
     return bool(model and spec and spec.claims(model))
+
+
+# How a provider proves who it is. Every decision the wizard makes about a
+# provider -- which field to prompt for, what a failure offers to change, what
+# "remove" clears, whether a rollback applies -- follows from this one question,
+# and it was being answered independently at thirteen sites off two spec flags.
+# Each of the last two review rounds found a site that disagreed with the others:
+# a rollback that wrote credentials to an OAuth provider and killed the wizard, a
+# menu that offered a key prompt to one, a prompt that half-guarded a spec it had
+# already dereferenced. Answer it once.
+CRED_OAUTH = "oauth"  # a token file, written by `raven provider login`
+CRED_LOCAL = "local"  # reached by address; there is no key
+CRED_ENDPOINT = "endpoint"  # a key plus a base URL the user supplies
+CRED_KEY = "key"  # a key alone, including vendors Raven carries no spec for
+
+
+def _credential_kind(provider: str, spec: Any) -> str:
+    """Which credential shape this provider uses.
+
+    Derived rather than stored: `spec` is optional metadata, and a vendor with no
+    spec is reached with a key like most others.
+    """
+    if spec is not None and spec.is_oauth:
+        return CRED_OAUTH
+    if spec is not None and spec.is_local:
+        return CRED_LOCAL
+    if provider == "custom":
+        return CRED_ENDPOINT
+    return CRED_KEY
 
 
 def _litellm_spelling(provider: str) -> str:
@@ -1110,7 +1139,7 @@ def _roll_back_provider_fields(provider: str, spec: Any, *, old_key: Optional[st
     layer refuses to write credential fields for them, and doing it anyway turned
     a failed verification into a dead wizard.
     """
-    if spec is not None and spec.is_oauth:
+    if _credential_kind(provider, spec) == CRED_OAUTH:
         return
     _write_provider_fields(provider, {"api_key": old_key or "", "api_base": old_base})
 
@@ -1287,8 +1316,9 @@ def _configure_one_provider(
                 continue
 
         spec = find_by_name(provider)
-        is_oauth = bool(spec and spec.is_oauth)
-        is_custom = provider == "custom"
+        kind = _credential_kind(provider, spec)
+        is_oauth = kind == CRED_OAUTH
+        is_custom = kind == CRED_ENDPOINT
         # The interactive picker already echoes the chosen provider; only print
         # an explicit confirmation when it came from --provider (no echo then).
         if flag_provider:
@@ -1316,7 +1346,7 @@ def _configure_one_provider(
             provider,
             is_oauth=is_oauth,
             is_custom=is_custom,
-            is_local=bool(spec and spec.is_local),
+            is_local=kind == CRED_LOCAL,
             api_key=api_key,
             base_url=base_url,
             model=model,
@@ -1507,7 +1537,11 @@ def _resolve_model_with_test(
                     # A local deployment that cannot be reached is usually a
                     # wrong address, and this is the branch it lands in -- so
                     # retry alone left the one thing worth changing unreachable.
-                    *([(_t("Re-enter server URL", "重新填服务地址"), "rebase")] if spec and spec.is_local else []),
+                    *(
+                        [(_t("Re-enter server URL", "重新填服务地址"), "rebase")]
+                        if _credential_kind(provider, spec) == CRED_LOCAL
+                        else []
+                    ),
                     (_t("Continue anyway", "仍然继续"), "continue"),
                 ]
                 if status == "network_error"
@@ -1517,9 +1551,9 @@ def _resolve_model_with_test(
                     # left a mistyped address with no way back to the field.
                     (
                         (_t("Sign in again", "重新登录"), "reauth")
-                        if spec and spec.is_oauth
+                        if _credential_kind(provider, spec) == CRED_OAUTH
                         else (_t("Re-enter server URL", "重新填服务地址"), "rebase")
-                        if spec and spec.is_local
+                        if _credential_kind(provider, spec) == CRED_LOCAL
                         else (_t("Re-enter key", "重新填 Key"), "rekey")
                     ),
                     (_t("Switch provider", "更换服务商"), "switch"),
@@ -1591,7 +1625,7 @@ def _resolve_model_with_test(
             provider,
             non_interactive=non_interactive,
             warnings=warnings,
-            is_oauth=bool(spec and spec.is_oauth),
+            is_oauth=_credential_kind(provider, spec) == CRED_OAUTH,
         )
         if result == "switch":
             return None
@@ -1650,7 +1684,7 @@ def _configure_existing_provider_model(*, non_interactive: bool) -> bool:
         provider,
         non_interactive=False,
         warnings=[],
-        is_oauth=bool(spec and spec.is_oauth),
+        is_oauth=_credential_kind(provider, spec) == CRED_OAUTH,
     )
     if result == "reauth":
         return _run_oauth_login(provider)
@@ -1703,7 +1737,7 @@ def _manage_existing_providers(*, non_interactive: bool) -> None:
             continue
         if action == "update":
             target_spec = find_by_name(target)
-            if target_spec and target_spec.is_oauth:
+            if _credential_kind(target, target_spec) == CRED_OAUTH:
                 # Nothing here to update: the credential is a token file, and the
                 # ops layer refuses credential writes for these -- so offering the
                 # key prompt ended the wizard instead of editing anything.
@@ -1716,7 +1750,7 @@ def _manage_existing_providers(*, non_interactive: bool) -> None:
                     )
                 )
                 continue
-            if target_spec and target_spec.is_local:
+            if _credential_kind(target, target_spec) == CRED_LOCAL:
                 # A local deployment holds no key; what there is to update is
                 # where it lives. Offering the key prompt wrote a credential into
                 # a provider that never reads one, and left the address alone.
@@ -1771,7 +1805,7 @@ def _manage_existing_providers(*, non_interactive: bool) -> None:
             # to clear and refuses the write, so it is told where its credential
             # actually lives instead of ending the run.
             target_spec = find_by_name(target)
-            if target_spec and target_spec.is_oauth:
+            if _credential_kind(target, target_spec) == CRED_OAUTH:
                 console.print(
                     _t(
                         f"  [dim]{_provider_label(target)}'s credential is an OAuth token, not a config field, "
--- a/raven/cli/onboard_commands.py
+++ b/raven/cli/onboard_commands.py
@@ -1594,10 +1594,7 @@ def _configure_existing_provider_model(*, non_interactive: bool) -> bool:
     from raven.cli._styles import RAVEN_STYLE
     from raven.providers.registry import find_by_name
 
-    choices = [
-        questionary.Choice(_provider_label(name), value=name)
-        for name in _configured_providers()
-    ]
+    choices = [questionary.Choice(_provider_label(name), value=name) for name in _configured_providers()]
     if not choices:
         return False
     provider = questionary.select(
--- a/raven/cli/onboard_commands.py
+++ b/raven/cli/onboard_commands.py
@@ -930,6 +930,23 @@ def _model_routes_to_provider(model: str, spec: Any) -> bool:
     return bool(model and spec and spec.claims(model))
 
 
+def _litellm_spelling(provider: str) -> str:
+    """How LiteLLM spells this vendor, for use as a route prefix.
+
+    Config section names are matched spelling-insensitively, so the name reaching
+    here may be underscored where LiteLLM hyphenates. Only LiteLLM's own form
+    works as a prefix.
+    """
+    from raven.providers.litellm_provider_names import LITELLM_PROVIDER_NAMES
+    from raven.providers.registry import normalize_provider_name
+
+    wanted = normalize_provider_name(provider)
+    for name in LITELLM_PROVIDER_NAMES:
+        if normalize_provider_name(name) == wanted:
+            return name
+    return wanted
+
+
 def _format_model_for_provider(provider: str, spec: Any, model_id: str) -> str:
     """Apply the provider's route prefix to a raw ``/v1/models`` id when needed.
 
@@ -943,12 +960,16 @@ def _format_model_for_provider(provider: str, spec: Any, model_id: str) -> str:
     Its section name is LiteLLM's own name for the vendor, which is exactly the
     prefix LiteLLM routes on, so that is what goes in front.
     """
-    from raven.providers.registry import normalize_provider_name
 
     if not model_id:
         return model_id
     if spec is None:
-        prefix = normalize_provider_name(provider)
+        # LiteLLM's own spelling, not the normalized one. A config section may be
+        # written either way and both resolve, but the wire prefix has to be the
+        # name LiteLLM routes on -- it hyphenates three vendors, and handing it
+        # "nano_gpt/..." is rejected with "LLM Provider NOT provided", so the
+        # provider configured fine and then could not be called.
+        prefix = _litellm_spelling(provider)
         return model_id if model_id.startswith(f"{prefix}/") else f"{prefix}/{model_id}"
     if spec.name in {"minimax_global", "minimax_cn"}:
         public_prefix = spec.name.replace("_", "-")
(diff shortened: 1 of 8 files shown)


arelchan (approved):

gloryfromca:
Reviewed by checking out the branch and driving the code, not just reading the diff. Overall this is a well-built change: the credential-kind consolidation is the right shape, the rollback design (read before write, restore exactly) is correct, and the tests are unusually strong (call-site pinning, subprocess import-cost probes, mutation-checked). I reproduced the Verification claims: the 4 touched test files pass (208 tests), ruff check/format are clean, and the candidate counts match exactly (moonshot 22, ollama_chat 21, github_copilot 27 against litellm 1.85.0).

One must-fix bug (confirmed with a repro), two should-fixes, and a few minor notes.

### Must fix: stale credential flags survive "Switch provider" and kill the wizard

Confirmed by driving `_configure_one_provider` with stubs:

```
raven onboard --provider deepseek --api-key sk-bad
  -> verification fails -> "Switch provider"
  -> pick Ollama (newly offered by this PR)
  -> BadParameter: "ollama_chat is a local deployment and takes no --api-key; pass --base-url instead"
  -> the whole wizard exits on a usage error; steps 2-4 are lost
```

Both rewind branches (back-out at https://github.com/EverMind-AI/Raven/blob/2dd09699d47ddd6f8e3992614fbb8a29efe3373a/raven/cli/onboard_commands.py#L1353-L1357, switch at https://github.com/EverMind-AI/Raven/blob/2dd09699d47ddd6f8e3992614fbb8a29efe3373a/raven/cli/onboard_commands.py#L1369-L1389) reset only `flag_provider`; `api_key` / `base_url` / `model` carry into the next iteration. The comment says "drop the flag so the second pass prompts rather than reusing the failed flag value" -- honored for the provider flag only.

Two more consequences of the same root cause (both pre-existing on main, but far more reachable now that local providers are in the picker and `--base-url` is their documented flag):

- a stale `--api-key` is written to the newly picked key provider without prompting, so the user watches a second unexplained failure;
- a stale `--base-url` leaks into the switched-to provider's `api_base` via the `elif base_url:` branch, silently pointing e.g. deepseek at an Ollama box.

Suggested fix: null out `api_key = base_url = model = None` alongside `flag_provider = None` in both rewind branches. The local-deployment `BadParameter` then fires only when the flags target a local provider on the first pass, which is the legitimate case it was written for.

### Should fix: three RPC handlers read the catalogue on the event loop

`model_save_key` / `model_add_model` / `model_remove_model` build their response entry synchronously (https://github.com/EverMind-AI/Raven/blob/2dd09699d47ddd6f8e3992614fbb8a29efe3373a/raven/tui_rpc/methods/model.py#L183-L186 and the two below), which calls `litellm_models_for`. Only `model.options` warms the cache off the loop. With a cold cache (call order is not guaranteed, and the failure guard deliberately clears it), the ~2s LiteLLM import runs on the event loop -- the exact stall the warm-up comment in `model_options` says must not happen. Moving the entry build into the existing `to_thread`, or warming at the top of each handler, closes it.

### Should fix: false "Couldn't reach the provider" on the four biggest providers' happy path

anthropic / openai / deepseek / gemini ship no api_base, so `_verify_provider` returns `(True, "skipped", None)` on every run, and `_pick_model` (https://github.com/EverMind-AI/Raven/blob/2dd09699d47ddd6f8e3992614fbb8a29efe3373a/raven/cli/onboard_commands.py#L1055-L1066) then prints "Couldn't reach the provider for its model list" immediately after "Skipping the model-list pre-check". Nothing failed. Threading the skipped-vs-failed distinction through fixes the copy.

### Minor

- The catalogue index is keyed by LiteLLM's raw spelling while the spec-less lookup normalizes (`index.get(normalize_provider_name(slug))` in `litellm_models_for`). No impact today (nano-gpt has no chat rows; the `vertex_ai-*_models` sub-keys would not match either way), but a future hyphen-named vendor with rows silently gets zero candidates. Looking up by `_litellm_spelling_for(slug)` hardens it. Related data quirks check out: `amazon_nova/amazon-nova/nova-micro-v1` does route via `get_llm_provider` -- the normalization actually rescues rows whose own catalogue spelling does not route.
- `_litellm_spelling` (onboard_commands.py L960) and `_litellm_spelling_for` (common_models.py L193) are the same function twice. Given this PR's own "answer it once" theme, worth exporting one and importing it.
- Title scope: the change spans cli / providers / tui_rpc / config; AGENTS.md 3.1 says multi-scope -> omit the scope or use `(*)`. CI's regex will not catch it, so maintainer's call.
- Two one-liners: the guard test's `\bspec\.is_(oauth|local)` regex does not catch reads off e.g. `target_spec` (underscore is a word char, so no boundary) -- fine as a heuristic, just not a hard guarantee. And the manage-menu "update" for `custom` (CRED_ENDPOINT) still re-prompts only the key, never the URL -- pre-existing, just more visible under the new taxonomy.


---

## Fix brief (agent-ready)

A machine-actionable version of the findings above: exact locations, a deterministic repro, required fixes, and acceptance criteria. Paste this section to whatever agent works the branch.

Context: repo `EverMind-AI/Raven`, branch `feat/onboard_picker_all_providers`,
HEAD `2dd09699d47ddd6f8e3992614fbb8a29efe3373a`. All file:line references are at
that SHA; if the branch has moved, locate code by the quoted symbols instead.

House rules that apply to these fixes (from AGENTS.md):
- Run tests with `uv run pytest ...`, never bare pytest.
- New regression tests go into the EXISTING files named below (5.4) - do not
  create new test files.
- Comments and commit messages in English; Conventional Commits; do NOT amend
  existing commits - add new ones. Do not commit/push without the maintainer's
  explicit instruction.
- After edits: `uv run ruff check raven/ tests/` and `uv run ruff format --check`.

---

### Finding 1 - MUST FIX: stale credential flags survive rewinds in `_configure_one_provider`

**File:** `raven/cli/onboard_commands.py`
**Symbol:** `_configure_one_provider` (loop at L1295-1391)

**Defect.** The two rewind branches reset only `flag_provider`:
- back-out rewind: `if custom_model is _BACK:` (L1353-1357)
- switch rewind: `if chosen_model is None:` (L1369-1389)

`api_key`, `base_url`, `model` keep their `--flag` values into the next loop
iteration. Consequences:
1. CONFIRMED CRASH: stale `--api-key` reaches a local provider's new guard in
   `_collect_credentials` (L1440-1445) and the `typer.BadParameter` escapes the
   wizard (exit code 2, steps 2-4 lost).
2. Stale `--api-key` is silently written to a switched-to key provider without
   prompting (the `pure_interactive` gate at L1470 sees `api_key` set).
3. Stale `--base-url` is written into the switched-to provider's `api_base`
   via the `elif base_url:` branch (L1503-1504).
4. Stale `--model` is passed as `user_model_flag` to the next provider and gets
   prefixed for it.

**Repro (deterministic).** Save as `repro_finding1.py` in the repo root on the
PR branch, run `uv run python repro_finding1.py`:

```python
import tempfile
from pathlib import Path
from unittest.mock import patch

import typer

import raven.cli.onboard_commands as oc

tmp = Path(tempfile.mkdtemp())
(tmp / ".raven").mkdir()

picks = iter(["ollama_chat"])  # what the user picks after "Switch provider"
prompted_for_base = []

with patch.object(Path, "home", lambda: tmp), \
     patch.object(oc, "_select_provider", lambda: next(picks)), \
     patch.object(oc, "_write_provider_fields", lambda p, f: None), \
     patch.object(oc, "_verify_provider", lambda p, skip_test=False: (False, "invalid_key", None)), \
     patch.object(oc, "_failure_choice", lambda options, non_interactive: "switch"), \
     patch.object(oc, "_roll_back_provider_fields", lambda *a, **k: None), \
     patch.object(oc, "_prompt_local_api_base",
                  lambda spec, current="", allow_back=False:
                      prompted_for_base.append(1) or "http://127.0.0.1:11434"):
    try:
        oc._configure_one_provider(
            provider="deepseek", api_key="sk-bad", base_url=None, model=None,
            non_interactive=False, warnings=[],
        )
        print("UNEXPECTED: wizard returned success")
    except typer.BadParameter as e:
        print(f"BUG REPRODUCED - BadParameter escaped the wizard: {e}")
    except StopIteration:
        assert prompted_for_base, "local provider was never asked for its address"
        print("FIXED - stale --api-key dropped; local provider prompted for address")
```

Expected before the fix: `BUG REPRODUCED - BadParameter escaped the wizard: ...`
Expected after the fix: `FIXED - stale --api-key dropped; ...`
Delete the script afterwards; it must not be committed.

**Required fix.** In BOTH rewind branches, clear the one-shot flag credentials
alongside `flag_provider = None`:

```python
api_key = None
base_url = None
model = None
flag_provider = None
```

Invariant to preserve: `--provider X --api-key Y` targeting a local provider on
the FIRST pass must still raise `BadParameter` (that guard is correct and has a
test). Flags are one-shot: they apply to the provider they arrived with, never
to a provider picked later.

**Regression tests** (add to the existing `tests/test_cli_onboard_commands.py`;
verify each by reverting the fix and confirming it fails):
1. Flag key + failed verify + switch to `ollama_chat`: no exception escapes;
   `_prompt_local_api_base` is called; the fields written for `ollama_chat`
   contain `api_base` only (spy `_write_provider_fields`).
2. Flag key + failed verify + switch to a key provider (e.g. `anthropic`):
   `_prompt_api_key` IS called (stale key not silently reused).

---

### Finding 2 - SHOULD FIX: three RPC handlers read the LiteLLM catalogue on the event loop

**File:** `raven/tui_rpc/methods/model.py`
**Symbols:** `model_save_key` (entry build at L184-186), `model_add_model`
(L204-207), `model_remove_model` (L216-219)

**Defect.** These handlers call `_build_provider_entry` -> `_provider_models`
-> `litellm_models_for` synchronously inside the async handler. Only
`model_options` warms the catalogue via `await asyncio.to_thread(...)` (L142).
With a cold cache - RPC call order is not guaranteed, and
`_litellm_chat_models_by_provider` deliberately calls `cache_clear()` on
failure/empty - the ~2s LiteLLM import runs ON the event loop, stalling all
concurrent RPC traffic. This contradicts the module's own policy (docstring:
sync work is wrapped in `to_thread`).

**Required fix.** In each of the three handlers, move the entry build off the
loop, e.g.:

```python
entry = await asyncio.to_thread(_build_provider_entry, parsed.slug, current_provider=current_provider)
```

(also covers the disk reads inside `_build_provider_entry`), or await the same
warm-up line `model_options` uses at the top of each handler.

**Regression test** (add to the existing `tests/test_tui_rpc_model.py`):
monkeypatch `raven.providers.common_models._cached_chat_models_by_provider` (or
`litellm_models_for`) with a stub that records
`threading.current_thread() is threading.main_thread()`, run one of the three
handlers under `asyncio.run(...)` on the main thread with a tmp config, and
assert the stub ran OFF the main thread.

---

### Finding 3 - SHOULD FIX: false "Couldn't reach the provider" on the skipped-precheck path

**Files:** `raven/cli/onboard_commands.py`
**Symbols:** `_verify_provider` (returns `(True, "skipped", None)` for providers
with no public /models endpoint - anthropic / openai / deepseek / gemini hit
this on EVERY run), `_resolve_model_with_test` (drops the status),
`_pick_model` fallback block (`if not model_ids:` at L1055-1067, message at
L1062: "Couldn't reach the provider for its model list - offering the ones we
know.")

**Defect.** `_pick_model` cannot distinguish "fetch failed" from "fetch was
skipped by design", so the four most common onboarding paths print a failure
message right after "Skipping the model-list pre-check...". Nothing failed.

**Required fix.** Thread the distinction into `_pick_model` (e.g. a
`fetch_status: Optional[str]` keyword passed from both call sites -
`_resolve_model_with_test` and `_configure_existing_provider_model`):
- `status == "skipped"` -> neutral copy, e.g. "This provider publishes no model
  list here - choose from the models we know." (update the zh string in the
  same `_t(...)` call);
- actual failure -> keep the current "Couldn't reach..." copy.
Both `_t` branches (en/zh) must be updated together.

**Regression test** (existing `tests/test_cli_onboard_commands.py`): capture
`console.print` output; with a stubbed `_verify_provider` returning
`(True, "skipped", None)` assert the neutral copy is printed and "Couldn't
reach" is NOT; with `(False, "network_error", None)` -> "continue" -> assert
the failure copy still appears where applicable.

---

### Finding 4 - MINOR (latent): spec-less catalogue lookup uses normalized key against raw-keyed index

**File:** `raven/providers/common_models.py`
**Symbol:** `litellm_models_for`, spec-less branch:
`index.get(normalize_provider_name(slug), ())`

**Defect.** `_cached_chat_models_by_provider` keys the index by LiteLLM's raw
`litellm_provider` spelling; the spec-less lookup normalizes the slug
(hyphens -> underscores). Any hyphen-spelled vendor (e.g. `nano-gpt`) that
gains catalogue rows will silently produce zero candidates. Zero impact today
(nano-gpt has no chat rows; `vertex_ai-*_models` sub-keys would not match under
either spelling), so this is hardening.

**Required fix.** Look up by the raw spelling the function already computes for
the prefix: `index.get(_litellm_spelling_for(slug), ())` (falls back to the
normalized form when the name is not in `LITELLM_PROVIDER_NAMES`, so behavior
for `mistral`-like names is unchanged).

**Regression test** (wherever `litellm_models_for` is already tested):
monkeypatch `_litellm_chat_models_by_provider` to return
`{"nano-gpt": ("nano-gpt/gpt-4o",)}` and assert
`litellm_models_for("nano_gpt") == ["nano-gpt/gpt-4o"]`.

---

### Finding 5 - MINOR: duplicated helper

`_litellm_spelling` (`raven/cli/onboard_commands.py` L960) and
`_litellm_spelling_for` (`raven/providers/common_models.py` L193) are the same
function. Keep the one in `common_models.py`, delete the CLI copy, import it in
`onboard_commands.py` (the codebase already imports private helpers across
these modules, see `_litellm_chat_models_by_provider` in
`raven/tui_rpc/methods/model.py`). No behavior change; existing tests cover
both call paths.

---

### Verification checklist (run all after the fixes)

```
uv run pytest tests/test_cli_onboard_commands.py tests/test_tui_rpc_model.py \
              tests/test_config_provider_matching.py tests/test_per_model_provider.py -q
uv run ruff check raven/ tests/
uv run ruff format --check raven/ tests/
uv run python repro_finding1.py   # must print "FIXED - ..."; then delete the file
```

Baseline before fixes: the four test files pass 208 tests on HEAD `2dd0969`;
the repro prints "BUG REPRODUCED". Every new regression test must be
mutation-checked: revert the fix it covers, confirm the test fails, restore.

PR #265, opened 2026-08-03
refactor(*): cover untested wizard paths and drop two inert lines

0xKT:
## Summary

Five gaps a three-round review listed as follow-up. None was a defect: they were
paths that work with nothing holding them down. Four are now driven end to end;
the fifth turned out to be two lines that never did anything.

**The retype-address branch.** A local deployment that cannot be reached is
almost always a typo, so the failure menu offers the address back -- and what
that choice runs had no test: it re-reads the stored address, seeds the prompt
with it, writes what comes back and re-verifies. Ctrl+C there quits, as in the
sibling re-enter-key branch; returning None would read as "switch provider" and
roll the setup back instead of retrying.

**Both flag-path guards.** A local deployment refuses `--api-key` rather than
dropping it silently, and a `--base-url` with no scheme is refused rather than
stored and failing at first use.

**Removing a provider that serves the default model** warns first, including for
a vendor Raven holds no spec for, which is decided by prefix alone. Treating "no
spec" as "not the source" skipped the warning and left the default pointing at a
provider whose key had just been removed.

**The catalogue index.** An empty result -- a LiteLLM whose table has not loaded
-- must not be cached, or every provider without a curated shortlist shows no
models for the life of the process with no way to retry.

Two deletions, both found by mutation rather than by reading:

* The `cache_clear()` in the failure branch never had anything to clear.
  `lru_cache` stores only successful returns, so a raised exception re-runs on the
  next call by itself. The mutation that removed the line changed nothing, which
  is what exposed it.
* The `None` check after the endpoint URL prompt was unreachable: that prompt,
  `_prompt_base_url`, raises on Ctrl+C rather than returning None. Its sibling
  `_prompt_local_api_base` does return None, which is why the two call sites that
  translate that None -- into an exit in the retype-address branch, into a
  `continue` in the manage menu -- are still live and still needed. Converging the
  two contracts would delete both translations; left for a separate change.

Left undone deliberately: `_is_priced_template` and the borrowed-driver guard in
`common_models`. Both affect candidate quality only, and the guard has no
observable effect against today's catalogue.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [x] Refactor
- [ ] Other

## Verification

```
bash .claude/scripts/preflight_ci.sh   exit 0
```

Seven mutations were applied. Six turn a new test red; the seventh changed
nothing, which is how the dead `cache_clear()` was found. Two of the tests were
rewritten after their first version failed to bite -- one stubbed out the very
`cache_clear` it was testing, and one never exercised the Ctrl+C path it claimed
to cover.

## Risk

Production code shrinks by five lines, all of them unreachable or inert; the rest
is tests. No behaviour changes.

Rollback is the revert of this branch.

## Related Issues

N/A


0xKT (commented):
Reviewed independently in a throwaway worktree at 63cb0af (main tree untouched,
`git status` empty afterwards). Verdict: **mergeable**. What follows is one
before-merge item that is a wording fix, one judgement call, and two follow-up
nits. No defect found in the code.

## What I verified

`uv run pytest tests/test_cli_onboard_commands.py tests/test_provider_resolution_invariants.py`
-> 290 passed, 30 skipped.

Then five mutations, to check the new tests bite rather than merely pass:

| Mutation | Result |
|---|---|
| `_resolve_model_with_test`: `current=stored` -> `current=""` | killed |
| `_resolve_model_with_test`: `raise typer.Exit(1)` -> `return None` in the `rebase` branch | killed |
| `_manage_existing_providers`: spec-less branch -> `was_default_source = False` | killed |
| `_litellm_chat_models_by_provider`: drop the empty-index `cache_clear()` | killed |
| **Re-add** the deleted `cache_clear()` in the `except` branch | all green -- the line was indeed inert |

Both deletions check out independently: `_prompt_base_url` raises
`typer.Exit(1)` on None (`raven/cli/onboard_commands.py:749`), so the removed
`if retyped_url is None: continue` was unreachable; and `lru_cache` does not
store exceptions, so the removed `cache_clear()` had nothing to clear.
`find_by_name("mistral")` returns None, so U12 does exercise the spec-less
branch rather than the spec branch.

## R1 -- before merge (description only, no code change)

- **Severity**: describe
- **File**: PR description, paragraph "Two deletions"; same sentence in the commit body
- **Claim**: "every prompt in that module raises on Ctrl+C rather than returning None"
- **Why it is wrong**: `_prompt_local_api_base` returns None on Ctrl+C
  (`raven/cli/onboard_commands.py:714-715`). That is precisely why the `rebase`
  branch needs its `raise typer.Exit(1)` at line 1570 -- the branch this PR adds
  a test for -- and why the live `if retyped is None: continue` at line 1759
  still has to exist. A reader applying the sentence as written would conclude
  line 1759 is dead too, and it is not.
- **Fix**: narrow it to the prompt the deletion is actually about, e.g.
  "`_prompt_base_url` raises on Ctrl+C rather than returning None".
- **Why before merge**: the repo squash-merges with
  `squash_merge_commit_message=PR_BODY`, so this sentence lands on `main` as the
  commit body.
- **Verify**: read `_prompt_local_api_base` (line 714) next to `_prompt_base_url`
  (line 749).

## R2 -- judgement call for the author

- **Severity**: describe
- **File**: PR title / commit type
- **Claim**: type is `test`, but the diff removes five lines of production code
  in `onboard_commands.py` and `common_models.py`.
- **Why**: AGENTS.md 3.1 defines `test` as "Tests only". The PR title -- which
  becomes the squash subject -- does not mention the deletions at all (the
  commit body does), so `main` will record a production deletion under a
  tests-only heading.
- **Fix**: either retype as `refactor(*)`, or extend the title to name the
  deletions. Not a blocker; the author's call.

## R3 -- follow-up, test placement

- **Severity**: nit / follow-up
- **File**: `tests/test_provider_resolution_invariants.py:505` and `:539`
- **Claim**: the two catalogue-cache tests sit in a file whose docstring scopes
  it to provider-resolution invariants and source-shape tripwires. Caching
  behaviour of `common_models` is neither.
- **Fix**: `tests/test_provider_catalog.py` is where `common_models` is already
  imported and tested. Move on a later pass; not worth a re-push here.

## R4 -- follow-up, the inconsistency behind R1

- **Severity**: nit / follow-up
- **File**: `raven/cli/onboard_commands.py:714`
- **Claim**: `_prompt_local_api_base` returns None on Ctrl+C while every sibling
  prompt in the module (`_prompt_api_key:673`, `_prompt_base_url:749`) raises
  `typer.Exit(1)`. Each call site then has to remember which contract it is
  holding: line 1570 converts None to an exit, line 1759 converts it to a
  `continue`.
- **Why it matters**: this divergence is what made the R1 sentence read as true.
  Converging it would delete both call-site translations.
- **Fix**: separate PR.

arelchan (approved):

Issue #278, opened 2026-08-06
bug(tui): the mouse-tracking opt-out is turned back on at startup


0xKT:
## Summary

`RAVEN_TUI_DISABLE_MOUSE` is honoured when the TUI starts and undone a moment
later, so mouse tracking is on for a user who asked for it off and selecting text
in the terminal stops working.

The startup config sync normalizes every display field it knows about and patches
the result into the UI store. Mouse tracking is the one field whose default comes
from the environment rather than from the store, so normalizing a config that says
nothing about it yields `true` and overwrites the opt-out.

## Steps to reproduce

1. Run `RAVEN_TUI_DISABLE_MOUSE=1 raven tui`.
2. Wait for the first config sync to land.
3. Try to select text with the mouse.

## Expected behavior

A config that says nothing about mouse tracking leaves it alone, and the
environment opt-out holds for the session.

## Actual behavior

The application captures the selection; the terminal cannot select text.
## Environment

```
OS: macOS 26.5 (arm64)
Shell: zsh
Python: 3.12.12
Node: v24.14.0
Raven: 3087174 (origin/main)
Install: source checkout, uv sync
litellm: 1.85.0
```

0xKT:
Fixed by #279 (squashed to main as 53aeb0c). The startup config sync only patches mouseTracking when the config actually carries mouse_tracking or tui_mouse (ui-tui/src/app/useConfigSync.ts:149), so a config that says nothing about it leaves the environment opt-out alone.

Issue #277, opened 2026-08-06
bug(token_wise): a call on a subscription plan is costed at $0.00


0xKT:
## Summary

Two wrongs in the same place.

A call on a subscription is costed at zero. ChatGPT (Codex), a Copilot seat, and a
MiniMax coding plan are not billed per token, so no per-token figure describes what
a call cost. LiteLLM files those models at a price of zero, the estimate reads that
as a real number, and the session reports `$0.00` -- indistinguishable from a free
model, and accumulating into a session total that means nothing.

The subscription prices Raven displays were written into the source without a
source.

## Steps to reproduce

1. Sign in to `openai-codex` and select one of its models.
2. Send any message.
3. Look at the cost reported for the turn and for the session.

## Expected behavior

"No per-token price" is a distinct answer from zero and is not accumulated into a
session total. How a provider bills is declared rather than inferred from whether it
uses OAuth -- Vertex is OAuth and metered.

## Actual behavior

The turn and the session both report `$0.00` on a paid plan.
## Environment

```
OS: macOS 26.5 (arm64)
Shell: zsh
Python: 3.12.12
Node: v24.14.0
Raven: 3087174 (origin/main)
Install: source checkout, uv sync
litellm: 1.85.0
```

0xKT:
Fixed by #279 (squashed to main as 53aeb0c). How a provider bills is now declared (ProviderSpec.billing == 'plan') rather than inferred from whether it uses OAuth, estimate_cost_usd returns None for a plan-billed model, and UsageTracker._add_into skips None instead of summing it as zero -- so a subscription call reports no per-token price and does not land in the session total.

Issue #276, opened 2026-08-06
bug(tui): importing LiteLLM prints over the TUI's rendering


0xKT:
## Summary

The first action that reaches LiteLLM prints to the terminal underneath the TUI's
own rendering, and the screen stays unreadable until something forces a full redraw.

`_strip_tty_stream_handlers` removes the handlers that write to the terminal at
startup, but LiteLLM is imported lazily, after that -- and its import installs its
own stderr handler. The root logger's level is 0, so DEBUG records pass as well.

## Steps to reproduce

1. Run `raven tui`.
2. Type `/model`.
3. Open a provider whose model list is not cached yet, so LiteLLM is imported.

## Expected behavior

Nothing LiteLLM logs reaches the terminal while the TUI owns it. The log file under
`~/.raven` already exists and is the destination.

## Actual behavior

Log lines land mid-render and the layout is broken until the next full redraw.

## Logs or screenshots

```
LiteLLM:WARNING: common_utils.py:979 - litellm: could not pre-load bedrock-runtime
response stream shape - Bedrock event-stream decoding will be unavailable.
Error: No module named 'botocore'
```
## Environment

```
OS: macOS 26.5 (arm64)
Shell: zsh
Python: 3.12.12
Node: v24.14.0
Raven: 3087174 (origin/main)
Install: source checkout, uv sync
litellm: 1.85.0
```

0xKT:
Fixed by #279 (squashed to main as 53aeb0c). raven/providers/litellm_setup.py detaches the stderr StreamHandler litellm installs on its own LiteLLM* loggers at import time. _strip_tty_stream_handlers could not reach it because the import is lazy and happens after startup.

Issue #275, opened 2026-08-06
bug(config): a completed OAuth sign-in can report itself unauthenticated


0xKT:
## Summary

Each OAuth family decides for itself where its credential lives, and the code that
writes one is not always the code that reads it. A sign-in can complete, store a
valid token, and leave every status surface reporting the provider as
unauthenticated -- permanently, because nothing will look where the token was put.

Three roots are in play: `openai_codex` writes through `oauth_cli_kit`'s data dir
while LiteLLM's chatgpt driver reads `CHATGPT_TOKEN_DIR`; `github_copilot`'s
authenticator defaults to `~/.config/litellm/github_copilot`; the two MiniMax
families use `oauth_cli_kit`'s auth dir under platformdirs.

The quieter half of the same problem: "the file exists" is treated as "the
credential is usable". A truncated write, a hand-edited file, or an empty one all
pass a path check and then fail on the first request, having already reported the
provider as ready.

## Steps to reproduce

1. Run `raven provider login openai-codex` and complete the flow.
2. Run `raven provider list`.
3. Run `raven tui`, type `/model`, and look at the OpenAI Codex row.

## Expected behavior

A provider reports itself authenticated after a sign-in that stored a token, and
"configured" means the credential reads as a credential rather than that a file sits
at the path.

## Actual behavior

`provider list` still says the provider needs `raven provider login openai-codex`,
and the `/model` row still shows the same warning. Signing in again does not change
it.

## Logs or screenshots

```
$ raven provider list
openai_codex        (empty)     run `raven provider login openai-codex` to authenticate
```
## Environment

```
OS: macOS 26.5 (arm64)
Shell: zsh
Python: 3.12.12
Node: v24.14.0
Raven: 3087174 (origin/main)
Install: source checkout, uv sync
litellm: 1.85.0
```

0xKT:
Fixed by #279 (squashed to main as 53aeb0c). One function now answers where a credential lives -- `get_oauth_dir()` in raven/config/paths.py, everything under ~/.raven/oauth -- and the chatgpt, copilot and both minimax paths all ask it instead of each deciding for itself. "Configured" now means the file parses as a credential rather than that a path exists.

PR #279, opened 2026-08-06
fix(*): sign in to an oauth provider from /model

### LINKED ISSUE #263: /model provider picker does not respond when selecting MiniMax OAuth provider
## Summary

In the current version, after starting Raven and entering `/model`, selecting `MiniMax Global (OAuth)` or `MiniMax CN (OAuth)` and pressing Enter does not trigger any visible action.

The UI shows the provider entry as selectable, and also shows the warning:

```text
warning: run `raven provider login minimax-global` to authenticate
```

But pressing Enter on the MiniMax provider does not start login, open the OAuth flow, switch provider, or show any further prompt.

## Steps to reproduce

1. Start Raven:

```bash
raven
```

2. Enter:

```text
/model
```

3. Move selection to:

```text
MiniMax Global (OAuth) · (sign in)
```

or:

```text
MiniMax CN (OAuth) · (sign in)
```

4. Press Enter.

## Expected behavior

Pressing Enter should do one of the following:

- start the MiniMax OAuth login flow;
- or show a clear instruction / confirmation screen;
- or tell the user to run:

```bash
raven provider login minimax-global
```

and then return safely.

## Actual behavior

Nothing appears to happen after pressing Enter.

The selection stays on the MiniMax provider entry, and the user does not know whether Raven accepted the input, failed silently, or is waiting for another action.

## Additional context

Screenshot available from the report.

This may be related to the new MiniMax Global/CN OAuth provider entries. The provider appears in the `/model` picker, but the Enter handler for OAuth sign-in providers may not be wired correctly in the TUI model selection flow.

## Environment

- OS: macOS
- Terminal: iTerm2
- Provider: MiniMax Global / MiniMax CN OAuth
- Raven version: current version

0xKT:
## Summary

`/model` listed the four OAuth providers and did nothing when Enter was pressed on
one. Enter now hands the terminal to `raven provider login <family>` and lands on
that provider's model list when the sign-in succeeds. Everything the fix walked
through on the way is here too, because a working key was not enough to reach a
model:

**Where a credential lives.** Each OAuth family resolved its own path -- Codex
through `oauth_cli_kit`, Copilot under `~/.config/litellm`, MiniMax under
platformdirs -- and the code that wrote one was not always the code that read it,
so a completed sign-in could report itself unauthenticated forever. One function
now answers where a credential lives, under `~/.raven/oauth`, and "configured"
means the file parses as a credential rather than that a path exists. LiteLLM's
driver owns the Codex flow end to end, which retires the `oauth-cli-kit`
dependency.

**Signing in.** `provider login` is now the driver's own flow, which is what makes
the credential it writes the one the request path reads. Two hazards come with
handing the flow over, and both are closed here rather than left for a user to
find: asking LiteLLM for a token falls through to starting a device flow when a
refresh fails, so a `model.options` RPC or a `provider test` could print a device
code and poll for fifteen minutes with nobody waiting to type it -- both entry
points are refused on the instance that only reads. And the driver treats its own
unfinished-attempt timestamp as a flow in progress, so an interrupted sign-in would
make the next one wait five silent minutes; the login path discards it. Signing in
while already signed in says so instead of announcing a flow it skips, a revoked
credential falls through to a real sign-in, and the failure text names both reasons
a refresh can fail -- LiteLLM wraps a revoked token and an unreachable network in
one error, so claiming either one would send half the users to fix the wrong thing.

**Which Codex models exist.** Two ids were hard-coded, in the registry and in the
provider constructor, and a live account refused both with "not supported when
using Codex with a ChatGPT account". Which slugs an account may use is only
knowable by asking it, so the account's own catalogue answers -- cached per
credential, with failures cached too so being offline costs one timeout rather than
one per keystroke, and never asked while signed out.

**Probes.** `provider test` fetched a ChatGPT token for every non-MiniMax OAuth
family, so Copilot was probed with somebody else's credential. Each family is
dispatched explicitly now -- and one with no check written yet says so, rather than
borrowing the nearest one. Codex and Copilot each get a probe that asks the way
their backend accepts being asked: Codex's account catalogue, and for Copilot the
endpoint out of its own API key file with the editor headers its driver sends,
because the generic `{api_base}/v1/models` request could reach neither -- for
Copilot it answered "api_base is empty" without asking, and asking would have had
a valid seat refused. A refused credential reads as a credential problem and an
unreachable host as a network fault.

Not verified: that GitHub answers 200 to the Copilot request. That needs a live
seat; what is pinned by test is the request that goes out.

**Cost.** A call on a subscription was costed at $0.00, indistinguishable from a
free model. Plan billing is declared in the registry, "no per-token price" is a
distinct answer from zero, the boot banner no longer opens a plan-billed session at
$0.00, and the invented subscription prices are gone.

**The wizard and the terminal.** The wizard's model step defaults to the newest
model the account offers, and says which one it used when the answer was empty --
Codex has no static default to fall back on any more, and a substitution nobody
mentions is how a test message ends up sent with a model the user did not pick.
Three more: importing LiteLLM printed over the TUI's rendering,
`RAVEN_TUI_DISABLE_MOUSE` was turned back on a moment after start, and the TUI
spawned whatever `raven` came first on `PATH` rather than the one that started the
session -- which is how a sign-in writes a credential the running process will
never read.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
make lint-python   ruff check: all checks passed; ruff format --check: clean
make lint-tui      eslint + lint:rpc + tsc --noEmit: exit 0 (pre-existing warnings only)
make test-tui      Test Files 85 passed (85), Tests 969 passed (969)
make test-python   5525 passed, 48 skipped, 13 deselected, 2 warnings in 187.32s
```

The suite was reading the developer's real OAuth credentials before this branch:
`import_litellm` publishes the token-directory variables process-wide and they
outlive the test that triggered the import, so a later test that faked `$HOME`
still resolved a live credential -- which made providers report themselves
configured inside a sandboxed home and sent the Codex catalogue lookup to the
network. On a signed-in machine that was a red suite; it is green now, and the
variables point at a temp directory for every test.

Review on this branch kept turning up tests that passed either way, so the guards
were re-checked by mutation: break the implementation in an isolated worktree, run
the test, confirm it goes red.

- [ ] User-facing docs or screenshots are updated when needed -- nothing under
  `docs/` or `README.md` names a credential location or the wizard's startup
  condition, so there was nothing to update

## Risk

Five changes are visible to someone who upgrades:

1. **Existing OAuth sign-ins are not migrated.** Credentials are read from
   `~/.raven/oauth` now, so a signed-in user signs in once more. The old files are
   left where they are, untouched.
2. **`~/.raven/oauth` is created 0700, and an existing one is tightened to it.** A
   user who widened that directory on purpose will find it narrowed; the files in it
   are credentials, and the drivers that write them do not set a mode themselves.
3. **`provider login` over a credential that already works reports and exits**
   instead of re-running the browser flow. `raven provider reset <name>` is the way
   to sign in as somebody else, and the command says so.
4. **Startup now notices a default model whose provider has no credentials**, which
   it did not before: it asks the config which provider serves the model rather than
   accepting any configured key as proof. The answer is a notice naming the model
   and pointing at `/model` -- not the wizard, which would restart at the language
   screen to fix one line. The wizard still runs for a config with nothing usable.
5. **Plan-billed providers stop reporting `$0.00`** and are left out of the session
   total rather than adding zero to it.

Removed: a five-second `config.get {key: 'mtime'}` poll in the TUI. `config.get`
answers with its four whitelisted keys whatever it is handed, so `mtime` was never
in the response, the poll always took its early return, and the MCP reload it
guarded never ran -- while costing a round trip on the same stdio pipe that carries
`prompt.submit`. `/reload-mcp` is unaffected.

Rollback is a revert of the merge: the credential move is a read-path change, so
reverting reads the old locations again and a user signed in before the upgrade is
signed in after the revert.

## Related Issues

Fixes #263, #212, #218, #269, #275, #276, #277, #278


gloryfromca:
## Adversarial review (verified against the branch, not just the diff)

Method: fetched `refs/pull/279/head` (11521ed), read the pinned litellm 1.85 authenticator sources, executed the copilot probe scenario against branch code, and ran the touched Python test files locally (648 passed, 48 skipped). Findings are ordered by severity: items 1-2 are worth fixing before merge, 3-6 are cheap to fix here or fine as follow-ups. A "verified non-issues" list is at the end so nothing gets re-litigated.

### 1. OAuth credential files lose the 0600 permission the removed stack applied (security regression)

- The removed `oauth_cli_kit` chmods token files to 0600 (`oauth_cli_kit/storage.py:70`).
- Its replacements do not: litellm's ChatGPT authenticator writes `auth.json` with a plain `open(..., "w")` (`litellm/llms/chatgpt/authenticator.py`, `_write_auth_file`), and the Copilot authenticator has no chmod anywhere. Under the default umask 022 that leaves `~/.raven/oauth/chatgpt/auth.json`, `github_copilot/access-token`, and `github_copilot/api-key.json` world-readable.
- raven's own MiniMax writer does 0600 + fsync (`raven/providers/minimax_oauth.py:142`), so the same directory now holds credentials at two permission levels.
- `clear_abandoned_device_code()` (`raven/providers/chatgpt_token.py:83-91`) rewrites `auth.json` via `tmp.write_text` + `os.replace`; `os.replace` keeps the tmp file's mode, so this resets a user-tightened 0600 back to 0644.

Suggested fix: create `get_oauth_dir()` with `mode=0o700` in `_point_oauth_tokens_at_raven()` (it runs before litellm's `makedirs`, so the parent mode shields the files regardless of their own mode), and chmod the credential files to 0600 after a successful login and after the tmp-file rewrite.

### 2. `provider test github-copilot` still cannot succeed for a validly signed-in seat

Executed against this branch: with a valid `access-token` and a fresh (unexpired) `api-key.json` on disk,

```
test_provider("github_copilot", ...) ->
{'ok': False, 'status': 'not_configured',
 'error': 'api_base is empty and provider has no default'}
```

The new copilot branch fetches a real API key via `Authenticator().get_api_key()` (a network refresh when the key is expired), then falls into the generic probe, which dies on the `if not api_base` check because the registry declares `default_api_base=""` for github_copilot (`raven/providers/registry.py:358`). The key just fetched is discarded. litellm's `api-key.json` even carries the endpoint (`Authenticator().get_api_base()` reads `endpoints.api`).

The failure mode predates this PR, but the PR summary says each family is dispatched explicitly now -- for copilot only the bad-credential half is fixed; the happy path is unreachable, and no test pins it because it cannot pass. Suggested fix: a copilot-specific probe like `_probe_codex_catalog` (hit `{get_api_base()}/models` with the fetched key), or at minimum fill `api_base` from `get_api_base()` before the generic probe.

For the record (verified in the litellm 1.85 sources, so nobody needs to re-check): the device-flow hazard IS closed on this path -- `get_access_token` returns the stored token without validating it, and a revoked token surfaces as `RefreshAPIKeyError` from `_refresh_api_key`, not as a `_login()` call.

### 3. Duplicate test definition

`test_resetting_an_unrelated_provider_stays_quiet` is defined twice in `tests/test_cli_provider_commands.py` (lines 540 and 601, identical bodies). The second shadows the first; ruff is silent because `tests/**` ignores F811. Delete one copy.

### 4. Ctrl-C during the TUI sign-in reports "exited with code null"

The oauthLogin screen promises "Ctrl+C while it runs cancels the sign-in and comes back here", but a child killed by a signal resolves `{code: null}` (`ui-tui/src/lib/externalCli.ts`, the `exit` handler), and `signIn`'s `result.code !== 0` branch renders that as an error: `` `raven provider login minimax-global` exited with code null `` (`ui-tui/src/components/modelPicker.tsx`). Special-case `code === null` with no `error` as a cancel (return to idle, no error text), or at least word it as cancelled.

### 5. Boot banner still injects `cost_usd: 0.0` for plan-billed sessions

`raven/tui_rpc/methods/session.py:121` hard-codes `"cost_usd": 0.0` in the session-create usage baseline, so a plan-billed session shows `$0.0000` in the status bar until the first `message.complete` overwrites it with null -- the exact "subscription reads as free" this PR removes elsewhere. Make the baseline None (or omit the key) when the default model's provider is plan-billed.

### 6. Stale docstring in `raven/tui_rpc/methods/reload.py`

The module docstring justifies the no-op handler with "useConfigSync.ts:202 polls reload.mcp every 5 seconds, hard-coded... the hermes call frequency itself cannot be touched (it's pulled from the fork)". This PR deleted that poller. The handler is still reachable from `/reload-mcp`, but the stated rationale is now false; rewrite it.

### Verified non-issues (checked; no changes needed)

- `estimated_cost_usd: None` threads safely end to end: `raven/tui_rpc/models.py` already declares `cost_usd: float | None`, `appChrome.tsx:355` gates on `typeof === 'number'`, `session.ts:648` on `!= null`, and `_add_into` guards the sum.
- The mtime-poll removal claim is accurate: the tui_rpc `config.get` whitelist never served `mtime`.
- The codex catalogue cache invalidates across processes via the auth-file fingerprint (mtime_ns + size), so `provider login` not calling `reset_cache()` is correct.
- The per-instance stubbing of `_login_device_code` / `_wait_for_access_token` is brittle by design but pinned by `test_the_refusal_is_wired_to_the_methods_the_driver_actually_falls_back_to`.
- Branch commits: all Conventional-Commits compliant, all English, zero non-ASCII characters.

Minor observations, no action expected: `codex_catalog`'s module-level cache is unsynchronized across `asyncio.to_thread` callers (benign duplicate fetch at worst); signals arriving during a terminal handoff are dropped rather than re-delivered after restore (a SIGHUP mid-device-flow can leave a headless TUI); `own_entry_point()` matches any argv[0] basename starting with "raven".

0xKT:
Thank you -- all six verified independently, all six real, nothing waved away. Fixed
in 9182f3c, 8ccad09, 13bb188 (plus c54b644 for what the ruff hooks took with them).
Point by point, including where I read the severity differently.

### 1. 0600 -- confirmed, and fixed as a class rather than at the three named sites

Every citation checked out. I pulled the 0.1.6 wheel to read the removed
dependency rather than take it on trust: `oauth_cli_kit/storage.py:70` is
`os.chmod(path, 0o600)`, exactly as you said.

One correction to the attribution: this is a regression for `openai_codex` only.
Copilot's files were already 0644 on `main` under `~/.config/litellm` -- this PR
relocated them rather than widening them. The remedy is the same for both, so the
finding stands as written; the account differs.

Fixed in three parts, because only some of these writers are ours:

- `get_oauth_dir()` creates 0700 and tightens an existing directory that is not.
  That is the part that holds for a file raven never sees -- including one a family
  added later writes.
- A sign-in restricts what it left behind, at one call site in `provider_login`
  rather than in each handler, over the list that already answered "every file a
  sign-in can leave behind" for disconnect. A driver rewriting its own credential
  opens the path for truncation and keeps the mode, so once per sign-in holds
  across every refresh after it.
- `clear_abandoned_device_code` sets the mode on the replacement, since
  `os.replace` leaves the old inode's mode with the old inode.

Your suggestion to do it in `_point_oauth_tokens_at_raven` would also have worked;
I put it in `get_oauth_dir()` because every caller derives the path through it, and
`paths.py` already creates the directories it names.

While in there: that shared list derived MiniMax's path a second time instead of
asking the module that writes it, so it pointed at the wrong file whenever
`MINIMAX_OAUTH_TOKEN_DIR` was set -- disconnect would have missed the credential it
means to delete. It asks now, like the other three families.

### 2. Copilot probe -- confirmed, and it needed more than filling `api_base`

Reproduced your execution. But filling `api_base` from `get_api_base()` and
falling into the generic probe would have swapped one wrong answer for another:
`get_copilot_default_headers` (`litellm/llms/github_copilot/common_utils.py:60`)
sends `copilot-integration-id`, `editor-version` and five more on every call, and
the backend refuses a request carrying only `Authorization`. A valid seat would
have come back 401 and been reported as a bad credential. The generic URL is also
wrong for it -- `"/v1" not in api_base` turns `api.githubcopilot.com` into
`/v1/models`.

So Copilot gets its own probe alongside Codex's, taking both the endpoint and the
headers from the driver. The request and how its answer is reported are shared with
the generic path, so there is one vocabulary and two callers deciding what to ask.

Two tests that could not exist before now do: the happy path asserts the URL, the
bearer and the editor headers on the outgoing request, and a credential with no
endpoint reports a credential problem rather than "provider has no default". What
is still unverified is that GitHub answers 200 to that shape -- it needs a live
seat, and the PR says so rather than implying otherwise.

You were right that the summary overclaimed. It now says which families get their
own probe and why, and carries the unverified part.

### 3. Duplicate test -- confirmed, deleted

Mine, from restoring a file I had clobbered by appending to it. F811 is ignored
under `tests/**`, so nothing was going to say so.

### 4. Ctrl+C -- confirmed, and worse than cosmetic

The screen promises Ctrl+C cancels, so `exited with code null` told the user their
own key press was a fault. `code === null` with no error is now a cancel, pinned by
a test that mounts the picker with `{code: null}`.

### 5. Boot banner -- confirmed

`session.py` opened a plan-billed session at $0.00, which is the same "a
subscription reads as free" this PR removes from the estimate: an incomplete sweep,
not a separate bug. The predicate that decides is now public and asked wherever a
dollar figure is reported.

### 6. `reload.py` docstring -- confirmed, and there were two copies

The module docstring and `tests/test_tui_rpc_reload.py`'s both justified the
handler by the poll this PR deleted. Both rewritten; the handler stays, since
`/reload-mcp` reaches it.

### Verification

Each fix was mutation-checked in a throwaway worktree -- directory not tightened,
sign-in not restricting, `tmp.chmod` removed, banner back to 0.0, Copilot dispatch
removed, Copilot headers reduced to `Authorization` -- and the matching test goes
red for each. Full suites after: 5525 passed / 48 skipped (was 5516), 969 TS tests
(was 968). All eight CI checks green.

On your minor observations: agreed on all three, and none touched here. The
unsynchronized catalogue cache costs a duplicate fetch; a signal during a terminal
handoff is dropped rather than re-delivered; `own_entry_point()` matches any argv[0]
basename starting with "raven". Left alone deliberately rather than overlooked.

Your "verified non-issues" section saved a round of re-litigation -- I spot-checked
the `cost_usd: float | None` thread and the `_add_into` guard and found them as you
described. Worth the space it took.

gloryfromca (approved):

PR #285, opened 2026-08-10
feat(*): send user attachments to the model, gated on vision support

arelchan:
## Summary

Before this change a picture a user attached never reached the model. `TurnRequest.media` was wired end to end and channels already used it, but `turn.send` had no `media` field, so a front end could only paste the path into the message text. This adds that field and the delivery rules behind it.

**Getting the file in.** `turn.send` takes an optional `media` array of workspace paths, resolved with the filesystem tools' own policy (so `uploads/shot.png` means the same thing it means to every other tool). A path that does not resolve, resolves outside the allowed directory, or cannot be read costs its own note in the message rather than the whole turn. The list is bounded at 64 in the schema, and one message inlines at most 16 images or 16 MB of base64.

**Deciding how it reaches the model.** A vision-capable model gets the image inlined as a base64 `image_url` block in the user message; anything else gets a text note naming the file, and the note names a description tool only when one is actually registered. The image is preprocessed by the same `prepare_image` that `read_file` already used: a 4032x3024 phone photo goes out as 1180x885, which is 30x less base64 than inlining it raw.

**Where the verdict comes from.** The gateway catalog Raven already fetches and caches for pricing, which publishes `input_modalities` for every model it lists. That completeness is why it is the source rather than LiteLLM's price table, which states `supports_vision` on under a third of its rows -- reading that silence as a denial would take a picture that reaches Grok, Llama 4 and the Qwen-VL family today and replace it with prose.

The catalog is read from cache only, never fetched inside a turn, and warmed on a background thread with a cooldown. The warm is not optional: the pricing path asks LiteLLM first and only reaches this catalog when that table misses, which excludes every model Raven ships a default for, so without it the probe would answer optimistically forever on a fresh install.

**Which way it fails.** Absence, staleness, a lookup that raises, and a caller-chosen deployment name all read as "no answer", which resolves to optimism and is never cached. Being wrong that way fails loudly at the endpoint. Being wrong the other way is silent: the picture never arrives and the model answers from the surrounding text as though it had seen one, and unlike the tool-result path there is no automatic recovery. Deployment names are excluded for that reason -- `gpt-4` is the name Azure's own quickstarts use, and a team keeps the deployment name while repointing it at a newer model, so joining it against a vendor catalog answers about somebody else's model.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests -q` -> 1 failed, 5334 passed, 58 skipped
- `LITELLM_LOCAL_MODEL_COST_MAP=true uv run pytest tests -q` -> 3 failed, 5332 passed, 58 skipped
- `uv run ruff check raven tests` -> clean; `uv run ruff format --check raven tests` -> 791 files already formatted
- `npm run lint:rpc` -> generated.ts in sync; `npm run type-check` -> clean; `npm test` -> 969 passed
- All 4 failures above reproduce on a pristine `main` worktree and are unrelated to this change: `test_cli_theme.py::test_bold_accent_renders_styled_not_bare` is a `rich` version issue, the two `test_token_wise_pricing.py` MiniMax-M3 assertions need a cost map the bundled LiteLLM does not carry, and `test_default_context_engine.py::TestTwoTrackConcurrency::test_skill_and_memory_run_concurrently` is a load-sensitive timing assertion.
- Verified against a live catalog that the probe denies the five text-only models Raven ships defaults for and grants every vision model tested, while every deployment and local-runtime name stays optimistic.
- Verified on a cold cache that the first turn answers optimistically without blocking, the background warm lands, and the next turn on the same long-lived loop gets the real verdict.

## Risk

`media` is optional, so every existing caller is unaffected, and the channels producer of `TurnRequest.media` is untouched. Attachment paths are resolved with the same policy as the file tools, so `restrict_to_workspace` still bounds them and a path outside it is dropped.

Pricing is deliberately unchanged: the catalog entry gained an `input_modalities` field and the on-disk cache version was bumped, but the pricing lookup itself has no deleted or modified lines, and cost and context-window resolution were compared against `main` across 40 real model ids in both cost-map modes with identical results.

Rollback is `ProviderSpec.vision_override` per provider for a wrong verdict, and reverting the commit otherwise.

**Known scope boundary.** Channels populate `TurnRequest.media` today and benefit immediately. The TUI does not reach this lane yet, and not for want of a client: it already implements both attachment gestures (a dropped path via `image.attach`, a clipboard image via `clipboard.paste`), but the server side of those is a `not supported in v0.1` stub or absent entirely, and the TUI catches the failure silently. Connecting them is a follow-up PR, and it is server-side work -- `ui-tui/` needs no change.

Two follow-ups deliberately left out of scope: a `media` list that fails validation surfaces as `-32603 internal_error` rather than `-32602`, which is how every `TurnSendParams` validation failure already behaves and needs a dispatcher change to fix; and the vision verdict is computed for the routed primary while a LiteLLM fallback further down the chain is sent the same message list, which the sibling transport probe shares and which needs the chain assembled per candidate.

## Related Issues

N/A

### DIFF
--- a/raven/providers/capabilities.py
+++ b/raven/providers/capabilities.py
@@ -13,6 +13,11 @@ Two questions that look alike but are not:
 The second question decides whether ``read_file`` hands the model a picture or a
 text placeholder plus a follow-up attachment, so it is answered per target, not
 per model.
+
+The first is :func:`supports_vision`, answered per model from the gateway
+catalog. Both have to be asked: a model that cannot see is not helped by a
+transport that could have carried the picture, and a model that can see still
+loses it over a transport that cannot.
 """
 
 from __future__ import annotations
@@ -66,6 +71,11 @@ IMAGE_TOOL_RESULT_TARGETS = frozenset({"anthropic", "vertex_ai", "bedrock"})
 # model is not blind -- the picture is discarded in transit. A refusal is
 # recoverable (ErrorClassification's should_drop_tool_images retries on the
 # placeholder path); a silent drop is undetectable by any mechanism.
+# Route prefixes whose second segment is a name the user chose rather than a
+# model id. LiteLLM's spelling for an Azure deployment; the registry has no
+# route by this name, so nothing else recognizes it.
+_DEPLOYMENT_NAME_PREFIXES = frozenset({"azure", "azure_ai", "azure_text"})
+
 GATEWAY_TARGETS = frozenset({"openrouter"})
 GATEWAY_IMAGE_TOOL_RESULT_PREFIXES = ("anthropic/claude-", "google/gemini-")
 
@@ -114,6 +124,116 @@ def supports_image_tool_result(provider: Any, model: str, spec: "ProviderSpec | 
     return target in IMAGE_TOOL_RESULT_TARGETS
 
 
+def _model_id_is_caller_chosen(model: str, provider: Any, spec: "ProviderSpec | None") -> bool:
+    """Does this route's model string name a deployment rather than a model?
+
+    Azure takes the name of a deployment the user created, and a local runtime
+    takes whatever tag the user pulled or served under. Either can be spelled
+    exactly like a vendor id it does not serve -- ``gpt-4`` is the deployment name
+    Azure's own quickstarts use, and teams keep the name while repointing the
+    deployment at a newer model -- so joining it against a vendor catalogue
+    answers about somebody else's model. Only a *denial* does damage (a grant is
+    what absence already gives), and a denial here is the silent failure this
+    module exists to avoid, so the catalogue is not consulted for these at all.
+
+    Asked three ways because Azure arrives three ways. Configured as a Raven
+    provider it is served by ``AzureOpenAIProvider`` and the model string is a
+    bare deployment name -- no prefix resolves ``find_by_model`` to the Azure
+    spec, and ``gpt-4`` alone is indistinguishable from OpenAI's own id, so only
+    the live provider instance knows. Routed through LiteLLM instead it carries
+    LiteLLM's ``azure/`` prefix, which the registry does not answer to and which
+    always introduces a deployment name. And a local runtime is named by a spec
+    that says so.
+    """
+    from raven.providers.azure_openai_provider import AzureOpenAIProvider
+    from raven.providers.registry import split_model_id
+
+    if isinstance(provider, AzureOpenAIProvider):
+        return True
+    if split_model_id(model)[0] in _DEPLOYMENT_NAME_PREFIXES:
+        return True
+    if spec is None:
+        return False
+    return bool(spec.is_local) or spec.client == "azure"
+
+
+def vision_verdict(
+    model: str,
+    spec: "ProviderSpec | None" = None,
+    provider: Any = None,
+) -> bool | None:
+    """What is *known* about ``model`` seeing images: True, False, or unknown.
+
+    ``None`` is the load-bearing case and the reason this sits under
+    :func:`supports_vision` rather than inside it. It means no answer exists yet
+    -- an unlisted model, a deployment name, or a catalog not warm -- which a
+    caller must not memoize: caching the optimistic default that ``None`` becomes
+    would freeze a cold-start guess for the life of the process and the warm
+    behind it would fill a table nobody re-reads.
+    """
+    if spec is not None and spec.vision_override is not None:
+        return spec.vision_override
+    if _model_id_is_caller_chosen(model, provider, spec):
+        return None
+
+    # Imported inside the call: pricing reaches back into this package, so a
+    # module-level import here would close the loop.
+    from raven.token_wise.pricing import openrouter_input_modalities, warm_catalog_in_background
+
+    try:
+        mods = openrouter_input_modalities(model)
+    except Exception as e:
+        # The catalog degrades rather than raising, but a capability probe must
+        # never be the thing that fails a turn.
+        logger.debug("vision_verdict: catalog lookup failed for {}: {}", model, e)
+        return None
+    if mods is None:
+        warm_catalog_in_background()
+        return None
+    return "image" in mods
+
+
+def supports_vision(
+    model: str,
+    spec: "ProviderSpec | None" = None,
+    provider: Any = None,
+) -> bool:
+    """Whether ``model`` can see an image at all.
+
+    The other half of this module's opening question, and the one that decides
+    whether a picture is inlined into the user message or replaced by a note
+    telling the model to read it another way.
+
+    Answered from the gateway catalog Raven already fetches and caches for
+    pricing (:func:`raven.token_wise.pricing.openrouter_input_modalities`), which
+    publishes ``input_modalities`` for every model it lists. That completeness is
+    the reason it is the source rather than LiteLLM's price table: the table
+    states ``supports_vision`` on under a third of its rows, and reading the
+    silence on the other two thirds as a denial would take a picture that reaches
+    Grok, Llama 4 and the Qwen-VL family today and replace it with prose.
+
+    No entry at all means yes, which leaves the model exactly where it was before
+    this function existed. Being wrong that way is loud -- the endpoint refuses
+    the request and the turn fails. Being wrong the other way is silent: the
+    picture never arrives and the model answers from the surrounding text as
+    though it had seen one. There is no automatic recovery in either direction on
+    the attachment path (``should_drop_tool_images`` rescues an image out of a
+    *tool result*, not out of a user message), so the choice is between a visible
+    failure and an invisible one. :attr:`ProviderSpec.vision_override` settles a
+    model the catalog gets wrong or never lists.
+
+    The catalog is read from cache only, never fetched here, so on a cold install
+    the first answers are the optimistic default while a background warm fills
+    it. The pricing path cannot be left to do that warming -- it reaches this
+    catalog only for models LiteLLM's static table misses, which excludes every
+    model Raven ships a default for.
+
+    A caller that caches this answer wants :func:`vision_verdict` instead, which
+    says whether there was an answer to cache.
+    """
+    return vision_verdict(model, spec, provider) is not False
+
+
 def _gateway_route(model: str) -> str:
     """The gateway's own model id, with the gateway prefix stripped.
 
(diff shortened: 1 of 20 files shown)


0xKT (approved):
Reviewed the full diff and verified the referenced symbols against main
(split_model_id, spec.client/is_local, _resolve_path, prepare_image,
Media fields, the attach_blocks -> pending_images hand-off). The
fail-direction design (unknown -> optimistic -> never cached) is
consistent across vision_verdict / _supports_vision and locked by tests.
CI is green and the 4 local test failures listed in Verification
reproduce on a pristine main. LGTM.

Three non-blocking notes, none worth holding the merge for:

R1 (nit) raven/context_engine/segments/render.py, build_user_content:
the can_see_images check runs after the file is read whole, so a blind
model attaching a large image reads up to 64MB just to emit a note.
Fix: hoist the can_see_images branch above `raw = head + handle.read()`.
Verify: existing test_the_attachment_note_names_no_tool_when_none_is_registered
still passes; optionally assert reads == [_SNIFF_BYTES] for the blind path.

R2 (follow-up) raven/agent/loop/main.py, _supports_vision: self.provider
is passed for any routed id, so when the primary provider is
AzureOpenAIProvider every routed model classifies as caller-chosen and
stays optimistic. Only bites mixed Azure-primary + router setups and
fails in the loud direction; worth a line in the known-boundaries list
or a per-candidate provider when the fallback-chain follow-up lands.

R3 (nit) tests/test_read_file_image.py grew ~770 lines of vision/warm
coverage; the file's subject is read_file image handling. Consider
splitting a test_providers_capabilities.py in a follow-up.

raven/providers/capabilities.py
```diff
@@ -114,6 +124,116 @@ def supports_image_tool_result(provider: Any, model: str, spec: "ProviderSpec |
...
+    spec, and ``gpt-4`` alone is indistinguishable from OpenAI's own id, so only
+    the live provider instance knows. Routed through LiteLLM instead it carries
+    LiteLLM's ``azure/`` prefix, which the registry does not answer to and which
+    always introduces a deployment name. And a local runtime is named by a spec
+    that says so.
+    """
+    from raven.providers.azure_openai_provider import AzureOpenAIProvider
+    from raven.providers.registry import split_model_id
+
+    if isinstance(provider, AzureOpenAIProvider):
```
gloryfromca, line 151: **`LazyProvider` defeats this guard on the only path that reaches `turn.send media`.**

The docstring above is right that "only the live provider instance knows" for a
bare Azure deployment name -- but `raven tui` builds the loop with
`make_lazy_provider()`, which is a plain proxy, not an `AzureOpenAIProvider`. So
for `model="gpt-4"` on an Azure deployment that has been repointed at gpt-4o:

- arm 1: `isinstance(provider, AzureOpenAIProvider)` -> False (it is the lazy proxy)
- arm 2: `split_model_id("gpt-4")[0]` is not in `_DEPLOYMENT_NAME_PREFIXES` -> False
- arm 3: `find_by_model("gpt-4")` returns *OpenAI's* spec, so `spec.is_local` is
  False and `spec.client == "azure"` is False

`_model_id_is_caller_chosen` returns False, `vision_verdict` consults the catalog,
the catalog answers about OpenAI's `gpt-4`, and vision is denied. The picture is
silently replaced by prose -- the failure mode the PR description singles out as
the one with no automatic recovery.

Verified at this head: `vision_verdict("gpt-4", find_by_model("gpt-4"), lazy)`
returns `False`, versus `None` when passed a real `AzureOpenAIProvider`.

Worth noting arm 3's `spec.client == "azure"` is unreachable for the same reason:
by the time a bare deployment name reaches here, `find_by_model` has already
resolved it to a vendor spec. Unwrapping the lazy proxy (or asking the provider a
question instead of type-testing it) would fix both arms.

raven/providers/capabilities.py
```diff
@@ -114,6 +124,116 @@ def supports_image_tool_result(provider: Any, model: str, spec: "ProviderSpec |
...
+        return True
+    if split_model_id(model)[0] in _DEPLOYMENT_NAME_PREFIXES:
+        return True
+    if spec is None:
+        return False
+    return bool(spec.is_local) or spec.client == "azure"
```
gloryfromca, line 157: **`custom` belongs in this predicate, and the documented rollback is dead for it.**

Any OpenAI-compatible endpoint configured as `custom` carries a *served* model
name chosen by whoever runs the endpoint -- the same situation as Azure, ollama and
vLLM, which are all excluded here. `spec.is_local` does not cover it (a `custom`
endpoint is frequently remote), so a served name like `gpt-4` or `llama-3-vision`
gets joined against the vendor catalog and answered about somebody else's model.

The second half is worse: the PR describes `ProviderSpec.vision_override` as the
rollback for a wrong verdict, but for `custom` that escape hatch cannot be
reached. `find_by_model()` resolves a bare or `openai/`-prefixed id to the
*vendor's* spec, so the spec that arrives here is never the `custom` one --
`route_names == {"custom"}` while `model_prefix == "openai"`. An override written
on the `custom` spec is never read, which means this case has no rollback at all,
not just a wrong default.

raven/token_wise/pricing.py
```diff
@@ -261,6 +275,115 @@ def _fetch_openrouter_models() -> dict[str, dict]:
...
+
+    Retried on a cooldown rather than attempted once. An attempt that fails
+    proves nothing about the next one -- the first turn of a session routinely
+    runs before a VPN is up or a proxy has authenticated -- and a single latched
+    attempt would leave the reader answering from an empty catalog for the whole
+    process. A success needs no cooldown: the filled cache is itself the guard.
+    """
+    global _WARM_AT
+
+    if _OPENROUTER_CACHE:
```
gloryfromca, line 299: **A stale on-disk catalog suppresses the background warm permanently.**

`_cached_catalog_only` (below) deliberately accepts a table of any age and, on the
way out, assigns it to `_OPENROUTER_CACHE`. This early return then sees a
non-empty cache and bails on every subsequent call.

So on a long-lived `serve` or TUI session that starts with a days-old cache file:
the first modality lookup populates `_OPENROUTER_CACHE` from disk, and the warm
never runs again for the life of the process. The cooldown reasoning in the
docstring covers the *fetch-failed* case but not this one -- there was never an
attempt to cool down from.

The two functions each make sense alone; the coupling is that "good enough for a
modality question" (any age) and "fresh enough that no warm is needed" are the
same flag. Gating this return on `_OPENROUTER_CACHE_TIME` freshness rather than on
`_OPENROUTER_CACHE` being non-empty would separate them.

raven/token_wise/pricing.py
```diff
@@ -261,6 +275,115 @@ def _fetch_openrouter_models() -> dict[str, dict]:
...
+        return _OPENROUTER_CACHE
+    disk = model_catalog_cache.load()
+    if disk is None:
+        return {}
+    # Re-checked after the read, not just before it: ``load()`` touches the
+    # filesystem and releases the GIL, so a background warm can land in that
+    # window with both a fresher table and a fresh ``_OPENROUTER_CACHE_TIME``.
+    # Overwriting it with this stale copy would leave that timestamp vouching for
+    # the wrong table, and the fetch's TTL check would then skip the refetch.
+    if _OPENROUTER_CACHE:
```
gloryfromca, line 342: **This re-check is still an unlocked check-then-act, so it does not close the window the comment describes.** (Lower confidence than the other notes -- I did not reproduce it.)

The comment above is precise about the hazard: a warm landing during `load()`
leaves a fresh `_OPENROUTER_CACHE_TIME` that would end up vouching for the stale
table. But the guard is a read at line 342 and a write at line 348 with no lock
between them, so the same interleaving simply moves:

1. line 342 reads `_OPENROUTER_CACHE` -- still empty, falls through
2. the warm thread completes: fresh table into `_OPENROUTER_CACHE`, fresh
   `_OPENROUTER_CACHE_TIME`
3. line 348 overwrites with `disk[0]`

which is the exact state the comment says must not happen -- and now pricing, not
just modality lookups, serves the stale table until the TTL expires (up to an
hour), because the fresh timestamp survives.

The window is much narrower than the one before `load()`, so this may be an
acceptable trade rather than a bug. If it is, saying so in the comment would be
worth more than the current wording, which reads as though the re-check
eliminates the race.

raven/tui_rpc/methods/turn.py
```diff
@@ -38,6 +40,57 @@
...
+        return ()
+
+    out: list[Media] = []
+    for raw in paths:
+        if not isinstance(raw, str) or not raw.strip():
+            continue
+        try:
+            resolved = _resolve_path(raw.strip(), workspace, allowed)
+            if not resolved.is_file():
+                logger.warning("turn.send: attachment {} does not resolve to a file", raw)
```
gloryfromca, line 80: **A dropped attachment is invisible to both the model and the front end.**

This `continue` (and the one at line 88) drops the path before
`build_user_content` ever sees it, so unlike an unreadable-but-resolved file --
which the description says "costs its own note in the message" -- a mis-spelled or
out-of-workspace path produces no `[Attachment: ...]` note at all. The RPC still
returns success, and the log line goes to the server log the user is not reading.

Net effect for a typo'd path: the model answers as though nothing was attached,
and the user believes it saw the image. That is the same silent shape the PR
argues against for the vision verdict, and it is reachable by ordinary user error
rather than by a catalog edge case.

Emitting a note for the dropped path (it does not need the file to exist -- the
raw string is enough) would make it visible in the transcript, which is where the
user would notice.

gloryfromca (commented):
Adversarial review pass. I read every hunk, opened the surrounding code, and
verified the first two findings by executing the code in a throwaway worktree at
this head (also ran `tests/test_read_file_image.py test_plugin_tools.py
test_tui_rpc_turn_send.py test_token_wise_pricing.py` -> 201 passed).

The design is careful and the "absence means yes" bias is the right one -- the
asymmetry argued in the description holds up. The problems are all in the
*exclusion* list: the code that decides when a model string must not be looked up
in the vendor catalog. A name that is wrongly allowed through lands in exactly the
silent failure mode the description says must be avoided.

Five inline notes below, most severe first. Nothing here is a design objection.

Checked and found sound: the routing/assembly reorder (the router needs only
`content`); `media_paths` does not filter on `Media.kind`/`mime`, so `kind="file"`
is fine downstream; `prepare_image` meta keys and the 64-byte sniff are adequate
for `detect_image_mime`; the count and byte ceilings, and the header-only read for
blind and non-image attachments; `_save_turn`'s image stripping keeps the path
notes; `AssemblyContext`/`TurnContext` are only ever constructed with keywords;
the Codex Responses provider does translate user-side `image_url` parts; pricing's
TTL logic is genuinely unchanged by the decision to leave `_OPENROUTER_CACHE_TIME`
alone.

PR #296, opened 2026-08-11
fix(tui_rpc): surface the cause of an internal_error, not just the code

arelchan:
## Summary

A -32603 reached the user as exactly this, and nothing else:

```
error: [rpc -32603] internal_error
```

No cause, no file, no log path. The information existed the whole time and was discarded
three separate times:

1. **The dispatcher dropped `detail`.** `raven/tui_rpc/dispatcher.py` used
   `if exc.data is not None: ... elif exc.detail: ...`, so a raiser that set both lost the
   detail. `_build_tui_agent_loop` always sets both.
2. **`turn.send` emitted only `{code, message}`** for a latched init crash, although the
   `ErrorEvent` schema already has a `detail` field and `chatStream.onError` already renders
   it.
3. **The client formatted `[rpc <code>] <message>`**, where `message` is a fixed code name
   (`internal_error`), and every call site prints `err.message`.

So an AgentLoop that cannot start produced a dead end. A real instance: a config file
containing a key the running branch does not know about fails validation, `_build_tui_agent_loop`
raises `InternalError` with the full pydantic message and `log_path`, and the TUI showed the
line above.

The fix keeps `message` protocol-faithful and puts the cause where callers already look:

- `error_data(exc)` in `raven/tui_rpc/errors.py` folds `detail` into `data`, and both the
  dispatcher and the turn error events use it, so the two paths carry the same context.
- `turn.send` passes the init-crash detail plus its `log_path` into the emitted event.
- `formatRpcError` in `ui-tui/src/rpc/errors.ts` builds the message from `detail`,
  `exception_message` or `reason`, and appends the log path. A multi-line cause (a config
  error listing each offending field) keeps its line breaks below the summary line.

Same failure after the change, rendered by feeding a real dispatcher frame through the client
path:

```
error: [rpc -32603] internal_error:
Config at /Users/admin/.raven/config.json fails schema validation:
1 validation error for Config
subagents
  Extra inputs are not permitted
(details in ~/.raven/logs/tui.log)
```

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

`uv run pytest tests/ -k "tui_rpc or dispatcher or tui_bootstrap"` -> 523 passed. New cases:
the dispatcher keeps `detail` next to `reason` and `log_path` in one frame; `turn.send` emits
the build-error cause with its log path; and it omits `detail` entirely when the error has no
cause to report.

`npx vitest run` in `ui-tui/` -> 86 files, 988 tests passed, including six new
`rpcErrorFromFrame` cases: a bare frame keeps the old one-line form; `data` context is
surfaced with the log path; `exception_message` and `reason` are read when `detail` is absent;
a multi-line cause keeps its shape; non-object or blank `data` cannot corrupt the message; and
typed subclass selection plus raw `data` access are unchanged.

`npm run type-check`, `npx eslint`, `npx prettier --check`, `ruff check`, `ruff format --check`
all clean.

End to end, the exact production shape: an `InternalError` raised the way
`_build_tui_agent_loop` raises it was dispatched to a JSON-RPC frame, and that frame was fed
through `rpcErrorFromFrame` with `tsx`, producing the output shown above.

## Risk

Notes: `error.code` and `error.message` are unchanged on the wire, so nothing that branches on
them is affected; the added `data.detail` and `payload.detail` are both already in the schema.
`err.message` gains a suffix, which the one message-matching call site
(`SESSION_BUSY_RE` in `useSubmission.ts`) is unaffected by, since it matches text that still
appears. Detail text originates from server-side exceptions and is already written to
`~/.raven/logs/tui.log`, so this exposes nothing new to the client; the client accepts it only
when it is a string and ignores any other shape. Rollback is a revert of this commit.

## Related Issues

N/A

### DIFF
--- a/raven/tui_rpc/methods/turn.py
+++ b/raven/tui_rpc/methods/turn.py
@@ -140,6 +140,13 @@ def _build_error_detail(exc: RpcError) -> str | None:
     ``message`` is only the code name (``internal_error``), so without this the
     TUI shows a turn failing for no stated reason. ``log_path`` rides along
     because an init crash is usually only fully diagnosable from the log.
+
+    It rides in *front* of the cause, not behind it: the TUI renders the first
+    line of this string and nothing else (``chatStream.ts``), and the crash this
+    exists for -- a config the running build cannot parse -- raises a
+    ``ValidationError`` whose ``str()`` is always multi-line (7 lines for two bad
+    fields). Appended, the pointer landed on the last line and never reached
+    anyone; the one case that needs the log was the one case that lost it.
     """
     data = error_data(exc) or {}
     detail = data.get("detail") or data.get("exception_message")
@@ -147,7 +154,7 @@ def _build_error_detail(exc: RpcError) -> str | None:
         return None
     log_path = data.get("log_path")
     if isinstance(log_path, str) and log_path.strip():
-        return f"{detail.strip()} (details in {log_path.strip()})"
+        return f"(details in {log_path.strip()}) {detail.strip()}"
     return detail.strip()
 
 
--- a/raven/tui_rpc/methods/turn.py
+++ b/raven/tui_rpc/methods/turn.py
@@ -26,7 +26,7 @@ from pydantic import ValidationError
 
 from raven.spine import ChatType, Media, Origin, Source, TurnHandle, TurnRequest
 from raven.spine.scheduler import Scheduler, SchedulerDrainingError
-from raven.tui_rpc.errors import RpcError, TurnInProgressError
+from raven.tui_rpc.errors import RpcError, TurnInProgressError, error_data
 from raven.tui_rpc.models import (
     TurnCancelParams,
     TurnSendParams,
@@ -134,16 +134,38 @@ def _resolve_model(parsed: TurnSendParams) -> str:
 # ---------------------------------------------------------------------------
 
 
+def _build_error_detail(exc: RpcError) -> str | None:
+    """Human-readable cause of a latched init crash.
+
+    ``message`` is only the code name (``internal_error``), so without this the
+    TUI shows a turn failing for no stated reason. ``log_path`` rides along
+    because an init crash is usually only fully diagnosable from the log.
+    """
+    data = error_data(exc) or {}
+    detail = data.get("detail") or data.get("exception_message")
+    if not isinstance(detail, str) or not detail.strip():
+        return None
+    log_path = data.get("log_path")
+    if isinstance(log_path, str) and log_path.strip():
+        return f"{detail.strip()} (details in {log_path.strip()})"
+    return detail.strip()
+
+
 async def _emit_start_then_error(
-    emitter: SubscriptionEmitter, session_key: str, turn_id: str, code: int, message: str
+    emitter: SubscriptionEmitter,
+    session_key: str,
+    turn_id: str,
+    code: int,
+    message: str,
+    detail: str | None = None,
 ) -> None:
     # message.start first so the front-end has a turn to clear, then the error
     # clears it (its onError resets turnId) — same shape the old per-turn task used.
     await emitter.emit(session_key, {"type": "message.start", "payload": {"turn_id": turn_id}})
-    await emitter.emit(
-        session_key,
-        {"type": "error", "payload": {"code": code, "message": message, "reason": "internal"}},
-    )
+    payload: dict[str, Any] = {"code": code, "message": message, "reason": "internal"}
+    if detail:
+        payload["detail"] = detail
+    await emitter.emit(session_key, {"type": "error", "payload": payload})
 
 
 async def turn_send(
@@ -182,7 +204,12 @@ async def turn_send(
         if emitter is not None:
             if build_error is not None:
                 await _emit_start_then_error(
-                    emitter, parsed.session_key, turn_id, build_error.code, build_error.message
+                    emitter,
+                    parsed.session_key,
+                    turn_id,
+                    build_error.code,
+                    build_error.message,
+                    _build_error_detail(build_error),
                 )
             else:
                 await _emit_start_then_error(emitter, parsed.session_key, turn_id, -32008, "model_not_available")
--- a/raven/tui_rpc/dispatcher.py
+++ b/raven/tui_rpc/dispatcher.py
@@ -26,6 +26,7 @@ from raven.tui_rpc.errors import (
     METHOD_NOT_FOUND,
     PARSE_ERROR,
     RpcError,
+    error_data,
 )
 
 Handler = Callable[[dict[str, Any]], Awaitable[dict[str, Any]]]
@@ -106,10 +107,9 @@ class Dispatcher:
                 "code": exc.code,
                 "message": exc.message,
             }
-            if exc.data is not None:
-                err_payload["data"] = exc.data
-            elif exc.detail:
-                err_payload["data"] = {"detail": exc.detail}
+            payload_data = error_data(exc)
+            if payload_data is not None:
+                err_payload["data"] = payload_data
             return {"jsonrpc": "2.0", "id": frame_id, "error": err_payload}
         except SystemExit as exc:
             # Click/Typer can leak SystemExit even with standalone_mode=False;
--- a/raven/tui_rpc/errors.py
+++ b/raven/tui_rpc/errors.py
@@ -61,6 +61,20 @@ class RpcError(Exception):
         return self.MESSAGE
 
 
+def error_data(exc: RpcError) -> dict[str, Any] | None:
+    """Wire ``error.data`` for an exception: ``data`` with ``detail`` folded in.
+
+    ``message`` is a fixed code name, so ``detail`` is the only place the cause
+    is spelled out; callers that set both must not lose it. Shared by the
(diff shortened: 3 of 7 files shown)


gloryfromca (commented):
No blockers; this can merge as far as I am concerned. One non-blocking note inline about the `log_path` suffix not reaching the user in the multi-line case.

### What I checked

- The diff itself, plus the callers on both sides: `_build_tui_agent_loop` (`raven/cli/tui_commands.py:483-525`) is the only producer of these `InternalError`s, and it always passes `data`, which is exactly the case the old dispatcher dropped `detail` for. The consumer of the new event field is `onError` in `ui-tui/src/app/chatStream.ts:210-224`, which is live (wired through `createChatStream` in `useMainApp.ts`), so the server half is not dead code.
- Schema: `ErrorEventPayload.detail` already exists in `raven/tui_rpc/models.py:212-216` and in the checked-in `ui-tui/src/rpc/generated.ts`, so nothing needed regenerating and the strict payload model still accepts the emitted dict.
- Backward compatibility: both new fields are additive. An older client ignores the extra `detail` key; a newer client against an older server just formats the bare code name, which is what the first test in `rpc.test.ts` pins.
- `error_data` semantics: `setdefault` means a raiser that sets both `data["detail"]` and `exc.detail` keeps the `data` one, and `data or None` now omits `error.data` when a raiser passes an empty dict. Neither has a caller in the tree, and no consumer requires `data` to be present, so I am not raising them.
- Project rules: commit header follows Conventional Commits with a real scope, the whole message is ASCII (`git log -1 --format=%B | grep -nP "[^\x00-\x7F]"` prints nothing), no new test files were created where an existing one exists (AGENTS.md 5.4), no assets.
- Tests were not weakened. Both new Python tests assert real behaviour, and the negative one (`..._omits_detail_when_the_build_error_has_no_cause`) is a genuine assertion, not a skip. The TS `ignores non-object and blank data` case loops over six hostile shapes rather than asserting one.

### What I ran

```
uv run pytest tests -k tui -q
611 passed, 7 skipped, 5335 deselected in 40.42s
```

The 7 skips are the pre-existing optional-channel import guards (`dingtalk_stream`, `lark_oapi`, `nio`, `botpy`, `slack_sdk`, `telegram`, `wecom_aibot_sdk`) and are unrelated to this change.

```
cd ui-tui && npx vitest run
Test Files  86 passed (86)
     Tests  988 passed (988)

npx tsc --noEmit      # clean
uv run ruff check raven/tui_rpc tests/test_tui_rpc_system.py tests/test_tui_rpc_turn_send.py
All checks passed!
```

(The suite needs `packages/hermes-ink` built first -- `npx esbuild src/entry-exports.ts --bundle --platform=node --format=esm --packages=external --outdir=dist` -- otherwise 41 files fail on a missing `./dist/entry-exports.js`. That is an environment prerequisite, not this branch.)

### Nit, take it or leave it

The doc comment at `ui-tui/src/rpc/errors.ts:21` writes the example as `(see ~/.raven/logs/tui.log)` but the code emits `(details in ...)`.

raven/tui_rpc/methods/turn.py
```diff
@@ -134,16 +134,38 @@ def _resolve_model(parsed: TurnSendParams) -> str:
...
+    TUI shows a turn failing for no stated reason. ``log_path`` rides along
+    because an init crash is usually only fully diagnosable from the log.
+    """
+    data = error_data(exc) or {}
+    detail = data.get("detail") or data.get("exception_message")
+    if not isinstance(detail, str) or not detail.strip():
+        return None
+    log_path = data.get("log_path")
+    if isinstance(log_path, str) and log_path.strip():
+        return f"{detail.strip()} (details in {log_path.strip()})"
```
gloryfromca: Non-blocking: the `log_path` suffix does not survive to the user in the case that motivated it.

The renderer on the other end keeps only the first line of `detail`, capped at 200 chars (`ui-tui/src/app/chatStream.ts:221`, pre-existing):

```ts
const extra = detail ? `: ${detail.split('\n')[0].slice(0, 200)}` : ''
```

`ValidationError` is one of the caught init-crash types and its `str()` is always multi-line -- I checked rather than assumed:

```
$ uv run python -c "...M(a='x', b='y')..."
"2 validation errors for M\na\n  Input should be a valid integer ..."
LINES 7
```

So for the exact scenario the commit message cites -- a config the running branch cannot parse -- the appended `(details in ~/.raven/logs/tui.log)` lands on line 7 and the user sees only `error: internal_error (code=-32603): 2 validation errors for RavenConfig`. No log path, which is the one thing this docstring says the path is riding along to provide.

Single-line causes (`FileNotFoundError`, `OSError`, `MissingCredentialsError`) are fine, so this is a partial gap rather than a regression -- main showed nothing at all. Cheapest fix is to put the pointer where the truncation cannot reach it, e.g. `f"{log_path} <- {first_line_of_detail}"`, or prepend it:

```python
return f"(details in {log_path.strip()}) {detail.strip()}"
```

gloryfromca (dismissed):

raven/tui_rpc/methods/turn.py
arelchan: Fixed in the pushed commit -- verified the failure first rather than taking the shape of it on faith.

```
ValidationError lines: 7
what the user saw:     "2 validation errors for M"
log path visible:      False
```

So the pointer was lost for precisely the scenario the commit message cites, exactly as you described. Prepended it, which is the cheaper of the two options you offered and keeps single-line causes reading the same.

Also replaced the assertion, since that is the part that let this through: the old test compared the whole `detail` string, which a trailing pointer satisfies just as well as a leading one. The new test builds a multi-line cause and asserts against `detail.split("\n")[0][:200]` -- the renderer's own slice -- so the next person to move this pointer gets told.

Rebased on current main (it was 25 behind) and re-ran: 52 passed across the turn-send and session-init files, ruff clean. Needs another approval since the push landed after yours.

gloryfromca: Confirmed at `e8417b22`. Prepending is the right resolution, and the added assertion is the part that matters -- it slices `detail` the way the renderer does, so a future trailing pointer fails the test instead of passing it.

```
uv run pytest tests -k tui -q
617 passed, 7 skipped, 5827 deselected

cd ui-tui && npx vitest run
Test Files  86 passed (86)
     Tests  990 passed (990)
```

gloryfromca (commented):

Re-reviewing only the delta since `b6b8ab48`, which is where I left off. The first commit is byte-identical after the rebase (`diff <(git show b6b8ab4851b7 --format="") <(git show 9c7e8797 --format="")` is empty), so the substance I already reviewed has not moved and I am not repeating it. The new commit `e8417b22` is the one thing I asked for and nothing else: the log pointer moved to the front of `detail`, and the existing whole-string assertion was updated rather than relaxed, with a second test added that slices `detail` the way `chatStream.ts` does -- so the shape that failed before now fails the suite instead of passing it. That is the right place to put the assertion; the previous one passed a trailing pointer just as happily, which is exactly how the gap got in.

On the delta I checked: that the renderer this is calibrated against is still the same after the rebase (`ui-tui/src/app/chatStream.ts:250`, moved from :221 by main's changes, logic unchanged); that no other test or source pins the old appended order (`grep -rn "details in"` finds only the new Python string, its test, and the TS frame path, which puts the pointer on its own line and is unaffected); that the 200-char slice still leaves ~165 chars for the cause after the ~35-char pointer; and that the commit message is ASCII with a real scope.

```
uv run pytest tests -k tui -q
617 passed, 7 skipped, 5827 deselected in 25.57s

cd ui-tui && npx vitest run
Test Files  86 passed (86)
     Tests  990 passed (990)

uv run ruff check raven/tui_rpc tests/test_tui_rpc_turn_send.py
All checks passed!
```

Count went 611 -> 617: one is the new test here, the rest came in with the rebase onto main. The 7 skips are the same pre-existing optional-channel import guards.

gloryfromca (approved):

gloryfromca (commented):

The new commit fixes the prior truncation gap: the streamed error now puts the log pointer before a multiline ValidationError, and the regression test asserts the same first-line, 200-character slice used by chatStream. I found no code issue worth raising.

I covered the full diff, the new delta, the startup-error source, dispatcher and streamed-error callers, client reconstruction and transcript rendering, history, backward compatibility of typed RPC errors and raw data, test strength, and the AGENTS.md/CONTEXT architecture and naming constraints.

Verification:
- `uv run pytest tests -k tui -q`: 617 passed, 7 skipped, 5827 deselected.
- `cd ui-tui && npm test`: 86 files passed; 977 passed, 13 skipped (990 total).
- `cd ui-tui && npm run type-check && npm run lint:rpc`: passed; generated RPC types are in sync.

Integration note: GitHub currently reports this head as CONFLICTING because current main renamed `raven/tui_rpc` to `raven/rpc` and the matching tests. The patch therefore needs a mechanical rebase/port before GitHub can merge it, followed by review of that new revision. I am not treating target-branch drift as a late-round product blocker because it is not introduced behavior and it has a direct author-side resolution.

I also resolved the earlier log-path thread after confirming its fix at this head.

PR #303, opened 2026-08-12
fix(cli): surface memory backend start failures on stderr

### LINKED ISSUE #271: bug(plugin): an invalid memory identity turns memory off without saying so
## Summary

`EverosBackend._validate_identity` rejects a `memory.userId` / `memory.agentId` that
EverOS cannot accept -- the value becomes a directory segment on its write path -- by
raising `ValueError` from `start()`. Every call site of `start()` swallows it:

- `raven/cli/agent_commands.py:402-407` and `:499-505`
- `raven/cli/tui_commands.py:655-663`
- `raven/cli/gateway_commands.py:368-373`

all wrap the call in `except Exception: logger.exception("memory backend start failed;
continuing with legacy memory path")`. Under the TUI, loguru is redirected to a file
(`tui_commands.py:1046`), so an invalid identity turns long-term memory off with no
signal on screen at all. The agent REPL and the gateway do print the traceback, so the
silent case is the TUI specifically.

The same backend takes the opposite decision one function away: when the server is up but
embedding is unavailable, it prints to stderr through rich
(`raven/plugin/memory/everos/backend.py:386,435`) precisely because a file-only log is
not visible enough for a degraded-memory condition. An identity that disables memory
outright deserves at least as much.

Found by code review of #264 (see that PR's discussion) and deliberately not fixed there:
the four call sites are in files that branch does not touch, and the fix is either a
config-load-time validation or four changed `except` blocks, which is its own change.

## Steps to reproduce

Derived from reading the code path, not reproduced at runtime.

1. Set an identity EverOS rejects in `~/.raven/config.json`, e.g.
   `"memory": {"userId": "team/alice"}` (the guard requires
   `^[a-zA-Z0-9_.@+-]+$` and forbids `.` / `..`).
2. Run `raven tui`.
3. Hold a conversation, then ask about something said earlier in a previous session.

## Expected behavior

The wizard, `raven doctor`, or the first start says the identity is invalid and names the
`userId` key, the same way an unavailable embedding provider is reported on stderr.

## Actual behavior

Startup continues, memory is off for the whole session, and the only record is a traceback
in `~/.raven/logs/`. Recall returns nothing and store never runs, which reads as an agent
that simply does not remember.

## Environment

Not environment-specific -- the visibility depends on which surface starts the backend
(TUI silent, agent REPL and gateway print). Observed while reviewing
`feat/hermes_cold_start_import` at 723ac26 on macOS / Python 3.12.

## Logs or screenshots

Expected shape of the only surviving record:

    ERROR ... memory backend start failed; continuing with legacy memory path
    ValueError: memory.userId='team/alice' is not accepted by EverOS: it becomes a
    directory segment on the write path, so it must match ^[a-zA-Z0-9_.@+-]+$ and must
    not be '.' or '..'.

icn5381:
## Summary

The four memory-backend start() call sites (agent REPL x2, gateway, TUI) caught failures with logger.exception only. Under the TUI loguru is redirected to a file, so an invalid memory identity (rejected by EverosBackend._validate_identity in start()) silently turned long-term memory off with nothing on screen. The agent REPL and gateway already surface the traceback via stderr; only the TUI was silent.

Add a warn_memory_start_failed helper that prints to stderr (mirroring the embedding-unavailable warning in everos/backend.py) and call it from every start() call site, so the degraded-memory condition is visible in every entry point.

## Type

- [x] Fix

## Verification

- [x] Relevant tests pass locally: `uv run pytest tests/test_memory_warn.py -v` -> 1 passed
- [x] Relevant lint / type checks pass locally: `uv run ruff format` (5 files unchanged), `uv run ruff check` (all passed)
- [ ] User-facing docs or screenshots updated: N/A (internal warning, no doc surface)

## Risk

- [x] Security impact considered: adds a stderr notice only; no new code path, no shell, reads only the exception type name
- [x] Backward compatibility considered: existing logger.exception calls are unchanged; the warning is additive
- [x] Rollback path is clear: revert restores the four bare except blocks

## Related Issues

Fixes #271


gloryfromca:

```bash
git fetch upstream main
git push --force-with-lease
```

PR #305, opened 2026-08-12
feat(tools): add MiniMax voice cloning

octo-patch:
Reason: Add executable voice cloning for the configured MiniMax backend.

## Summary

- Add a `voice_clone` tool that uploads MP3, M4A, or WAV source audio and creates a reusable voice.
- Support global and China endpoints, optional prompt audio, transcript validation, normalization, noise reduction, preview text, and watermark controls.
- Inherit the configured MiniMax provider key and register the tool only when its media configuration is enabled.
- Parse upload and clone API status responses and return the created voice identifier.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run --frozen pytest -q -p no:everos_plugin tests/test_media_gen_voice_clone_minimax.py` (4 passed)
- `uv run --frozen pytest -q -p no:everos_plugin tests/test_config_raven_sections.py` (20 passed)
- `uv run --frozen ruff format --check raven/agent/tools/media_gen.py raven/agent/loop/main.py raven/config/schema.py tests/test_media_gen_voice_clone_minimax.py`
- `uv run --frozen ruff check --ignore I001 raven/agent/tools/media_gen.py raven/agent/loop/main.py raven/config/schema.py tests/test_media_gen_voice_clone_minimax.py`
- `uv run --frozen python -m py_compile raven/agent/tools/media_gen.py raven/agent/loop/main.py raven/config/schema.py tests/test_media_gen_voice_clone_minimax.py`
- `git diff origin/main...HEAD --check`

- [x] Relevant tests pass locally
- [x] Relevant lint / type checks pass locally
- [ ] User-facing docs or screenshots are updated when needed

## Risk

- [x] Security impact considered
- [x] Backward compatibility considered
- [x] Rollback path is clear for risky changes

The tool is opt-in and preserves existing media behavior. Rollback consists of removing the tool registration and its configuration field.

## Related Issues

N/A

PR #312, opened 2026-08-13
feat(agent): withhold web_search without a key, and make tool capabilities visible

Handsome-wzw:
## Summary

A WeChat user asked for a recommendation and the bot answered "the web search
tool is not configured here". That sentence was not the model's own phrasing --
it was the tool's error text, relayed outward. `web_search` needs a Serper key
and that deployment has never had one, but the tool was registered
unconditionally, so on every search-shaped question the model saw it, reached
for it, and passed the setup error on to whoever was in the chat.

Withholding the tool fixes that and creates a second problem: an unregistered
tool is invisible. Nothing in a running Raven then says the capability exists at
all -- the model is never offered it, no document lists it, and `raven doctor`
reports on providers and memory but has never mentioned tools. The only way to
learn that web search is one account and one edit away was to read the source.

This does both halves, because either alone leaves a real gap: the deployer is
told what this install can and cannot do, and the model is still not offered a
tool it cannot run.

### Withholding the unusable tool

`web_search` is gated on a resolved key, in the main loop and in the sub-agent
surface (`subagent/manager.py`) -- a sub-agent that reaches for a search it
cannot run reports the failure to its caller, and that text lands in the parent
turn, the same leak one level down.

The gate asks the tool, not the config. `WebSearchTool.api_key` resolves at call
time from the constructor value *or* `SERPER_API_KEY`, so reading
`tools.web.search.apiKey` alone would withdraw a working tool from any deploy
that exports the variable and configures nothing.

The tool's error message also hard-coded `~/.raven/config.json` while the
gateway runs with `--config` elsewhere, so following it meant editing a file the
process never reads. It names the path actually in force now. That text is
reachable only if the key disappears after registration, which is exactly why it
should be right: it is the message for the case the gate cannot cover.

### Three rules, and nowhere to read them

Registration is decided per family, each a different shape:

| family | rule |
|---|---|
| `web_search` | a resolved key, asked of the built tool |
| `web_fetch` | nothing -- always registered, a key only improves extraction |
| media x3 | an `api_key` *or* a `model`, either counting as configured |

For media those are two questions, not one. A section naming only a model is
registered, because a model alone counts as asking for the tool, and then every
call returns a missing-key error. Whether a capability is *offered* and whether
it *works* come apart there, and a report treating them as one fact ticks a
capability that cannot run.

Each rule is defensible where it sits. What is missing is anywhere to read them.
Providers had the same sprawl once and answered it with `providers.auth`: a
declarative table plus `credential_status` as the single authority, with an AST
invariant enforcing that authority, because six surfaces had answered the same
question six ways and each looked reasonable alone. Tools never got the
equivalent.

### What this adds

**Each tool answers for itself,** twice where the questions differ.
`WebSearchTool.is_configured` reads the config value *or* `SERPER_API_KEY`,
because those are two sources and only the tool consults both. The media base
answers `is_configured` on a model *or* a key -- which is what stops an
OpenRouter credential set for chat from silently switching on three tools that
bill per call -- and answers `has_key` separately on the chain it actually
resolves at call time: its own section, the borrowed provider key, then
`OPENROUTER_API_KEY`. Both rules live with the credential they read.

**`capabilities.py` describes the five** for a human deciding what to set up:
what each does in one line, how much work it is (nothing / reuse a credential
you already have / obtain an account), where the key goes, where to get one, and
what it costs. It rules on nothing -- `is_configured` and `has_credential` both
ask the tools -- so it cannot become a second opinion.

**`raven doctor` grows a section**, listing every capability configured or not,
ordered by how much the deployer has to do:

```
(markers render as a green check and a yellow bang; spelled [ok] and [!] here to
keep this description ASCII, since it becomes the squash commit body)

Tool capabilities
  web_fetch:       [ok] Read a web page the agent already has the URL for
  image_generate:  [ok] Generate an image  (borrowed: providers.openrouter.apiKey)
  text_to_speech:  [!] Generate speech from text
                   no key resolves; calls will fail
                   set: tools.media.speech.apiKey
                   or env: OPENROUTER_API_KEY
  video_generate:  -  Generate a video
                   switch on: tools.media.video.model
                   key: reusing providers.openrouter.apiKey
                   Billed per call; needs prepaid OpenRouter credit.
  web_search:      -  Search the web
                   set: tools.web.search.apiKey
                   or env: SERPER_API_KEY
                   key from: https://serper.dev
  2 capability(s) available but not set up; the agent is not offered them.
```

Three deliberate details.

Naming the credential is load-bearing in both directions. A row that cannot
distinguish a reused credential from a missing one sends someone to create an
account they already have -- and a row that claims a reuse with nothing to reuse
is worse, because acting on it means setting a model, getting a registered tool,
and watching every call fail on a credential they were told they had. So the
reuse line prints only when a key is genuinely there to pick up, and a
capability already in that broken state says so outright instead of showing a
satisfied tick.

The credential is named at the path that holds it. For the media family
`config_path` names the *model*, so reusing it as the key source pointed the
deployer at a line with no credential in it.

And each fact is on its own line rather than in a sentence, because the terminal
wraps a long line mid-path and a config key broken across two rows cannot be
copied, which is the only thing that row is for.

Nothing here moves the exit code, including the warned row. An install without
image generation is a choice, not a fault, and a doctor that fails on it teaches
people to ignore doctor. The half-finished one is arguable -- the memory section
does exit non-zero for a role the user configured that the server could not
build -- but that failure is silent where it happens, recall just returning
nothing, whereas this one returns an error string to the model on every call.
Say the word and it becomes an exit code instead.

### Not in this change

Registration still lives in `AgentLoop`. Having it read the table is the point
of this shape and removes the last duplicate reader, but it edits
`agent/loop/main.py`, which is under active change, and it is worth doing on its
own once that settles. Until then the table is a description, and the tests
below are what keep it honest.

`deep_research` is deliberately absent: it is moving to the sub-agent surface
and its tool is going away.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

Mixed on purpose: the first commit is a fix and the rest are the feature it
made necessary. Splitting them would land a change that hides a capability
without landing the one that makes it discoverable.

## Verification

```
make lint                                            exit 0
make build                                           exit 0
npm run test --prefix ui-tui                         969 passed, 13 skipped
uv run --all-extras pytest -q                        6186 passed, 35 skipped, 1 failed
uv run pytest tests/test_tool_capabilities.py \
              tests/test_cli_doctor_commands.py \
              tests/test_agent_loop_web_tools.py     64 passed
uv run --extra dev ruff check raven tests scripts    All checks passed
uv run --extra dev ruff format --check               828 files already formatted
npx commitlint --from origin/main --to HEAD          exit 0
scripts/check_commit_messages.py origin/main..HEAD   exit 0
```

The one failure is `test_read_file_image.py::test_an_attachment_that_cannot_be_
read_costs_a_note_not_the_turn`, and it is not from this branch: it fails the
same way on `main` at `1cb604a` with these commits absent. The case makes a file
unreadable with `chmod 000`, which does not block a root user, so it fails for
anyone running the suite as root and passes in CI. This branch does not touch
that file or the code under it.

The capability tests drive a real `AgentLoop` and compare what it registered
against what the table predicts, rather than asserting the table against itself.
The gated set is derived -- tools present once credentials are supplied, absent
without -- so a sixth gated tool whose author forgets the table fails here rather
than going unnoticed.

That last claim was false when first written, and the last commit is what makes
it true. The fixture listing the media tools by hand never switched a fourth one
on, so a new one never joined the gated set and the assertion held over an
already-incomplete table. #305, which adds a MiniMax voice-clone tool, is that
case: merged against this branch the assertion passed. Read from
`MediaGenConfig` instead, it fails and names the tool -- `gated but undeclared:
['voice_clone']`.

Which means whichever of the two lands second turns this red, deliberately. The
fix is one table entry, and it needs the rule that PR settles: `voice_clone`
counts as configured on `api_key or api_base or model` in
`effective_media_config` while registration still gates on `api_key or model`,
so an `apiBase`-only install with no MiniMax key is configured by one rule and
withheld by the other. That is the divergence this table exists to make visible,
and it is worth resolving there rather than papering over here.

Fifteen mutations, each caught:

```
the web_search gate removed from the main loop    3 failed
the same gate removed from the sub-agent surface  1 failed
media rule reads only the key                     7 failed
web_search rule reads only the config             1 failed
a tool removed from the table                     6 failed
the doctor section not rendered                   6 failed
doctor lists only configured tools                4 failed
unconfigured treated as a failure                17 failed
the reuse line is printed unconditionally         1 failed
the key source falls back to the model path       3 failed
a keyless registered capability is not flagged    1 failed
has_key always answers yes                        1 failed
has_credential collapses into is_configured       1 failed
the credential chain drops OPENROUTER_API_KEY     3 failed
the reuse check ignores the environment           3 failed
```

Four of those tests exist only because a mutation pass found the earlier
versions insufficient, and the last rounds are why the shape changed. A media
case that sets both a model and an OpenRouter key proves nothing about the "or
model" half: the borrow fills the key in, so a rule reading only the key still
answers correctly. The case that pins it has nothing to borrow.

The sub-agent case is the same argument applied to the second call site: with
the gate present only in the main loop, every test above still passes, because
nothing was watching what the sub-agent surface registers.

More usefully, mutating `has_key` to always answer yes changed no test at all in
the first version, which said the ruling was not load-bearing: the doctor was
inferring "no credential" from an *empty source string* rather than from the
tool's answer, so a fourth credential source would have been reported as a
missing one. `has_credential` is now its own fact, asked of the tools, and that
mutation fails.

Separately, autouse fixtures clear `SERPER_API_KEY` and `OPENROUTER_API_KEY`,
without which several of these pass for the wrong reason on any machine where a
developer exported one.

Rendering was checked against a real config in six states -- nothing set, a
provider key present, a model with nothing to borrow, a tool with its own key, a
model plus a borrowable key, and an exported variable as the only source --
rather than only asserted on.

## Risk

One behaviour change: `web_search` is no longer offered to the model when no key
resolves. That is the fix. A deployment that has a key, in config or in the
environment, is unaffected; one that has none was getting an error string in
place of an answer.

The rest is additive. No other registration logic changed, so which tools an
agent is offered is otherwise exactly what it was; `is_configured` moved the
existing predicates onto the tools without altering them, and the mutation
results above are what pins that. `has_key` and `_resolve_key` are new names for
the chain `api_key` already resolved -- the property calls the extracted one, so
callers see the same answers.

`raven doctor` gains a section and no new exit code. Its zero-network guarantee
holds: the table reads config and environment only.

Two allowlist entries were added to
`test_only_the_auth_module_decides_configuredness_from_a_key`, argued in place:
`subagent/manager.py`, which asks the built tool whether a key resolved so an
unusable search is withheld, and `capabilities.py`, which reads keys to *report*
-- which source supplied one, and whether one is there to reuse. Both ruling
halves are delegated to the tools, whose files were already listed.

Rollback is a revert. Nothing is written and no configuration is read
differently.

## Related Issues

N/A

### DIFF
--- a/raven/agent/tools/capabilities.py
+++ b/raven/agent/tools/capabilities.py
@@ -13,6 +13,13 @@ the model, one per family, each a different shape:
     web_fetch    nothing -- always registered, a key only improves extraction
     media x3     an api_key *or* a model, either one counting as configured
 
+For the media family, being offered to the model and being usable are two
+different questions: a section naming only a model is registered, because a
+model alone counts as asking for the tool, and then every call fails on a
+missing key. :func:`is_configured` answers the first and the tools' ``has_key``
+answers the second -- collapsing them is how a report ends up ticking a
+capability that cannot run.
+
 Each rule is defensible where it sits. What is missing is anywhere to *read*
 them. A deployer cannot ask what this install lacks, and since an unconfigured
 tool stopped being registered there is no surface at all saying the capability
@@ -39,6 +46,10 @@ from typing import TYPE_CHECKING, Any
 if TYPE_CHECKING:
     from raven.config.schema import Config
 
+#: The credential the media family falls back to, named once so what a doctor
+#: row tells the deployer to set cannot drift from what is read here.
+_OPENROUTER_KEY = "providers.openrouter.apiKey"
+
 
 class Need(Enum):
     """What stands between a capability and working."""
@@ -61,9 +72,11 @@ class Capability:
     #: One line in the deployer's terms rather than the code's.
     summary: str
     need: Need
-    #: Dotted config path the deployer would edit to switch this on.
+    #: Dotted config path the deployer would edit to switch this on. For the
+    #: media family this is the *model* field, not a credential -- see
+    #: :attr:`key_path`.
     config_path: str = ""
-    #: Environment variable accepted instead of the config field, if any.
+    #: Environment variable accepted instead of this capability's credential.
     env_var: str = ""
     #: Where to go when ``need`` is NEW_ACCOUNT.
     obtain_from: str = ""
@@ -72,6 +85,16 @@ class Capability:
     #: Stated before the deployer switches it on rather than after: these cost
     #: money per call, and one cannot run at all without prepaid credit.
     cost_note: str = ""
+
+    @property
+    def key_path(self) -> str:
+        """Dotted path to this capability's own credential field.
+
+        Distinct from :attr:`config_path`, which for the media family is the
+        model field: naming a model path as where a key came from sends the
+        deployer to edit a line that holds no credential.
+        """
+        return f"tools.media.{self.media_attr}.apiKey" if self.media_attr else self.config_path
 
 
 #: Ordered by how much the deployer has to do, least first, because that is the
@@ -89,6 +112,7 @@ CAPABILITIES: tuple[Capability, ...] = (
         need=Need.OWN_CREDENTIAL,
         config_path="tools.media.image.model",
         media_attr="image",
+        env_var="OPENROUTER_API_KEY",
         cost_note="Billed per image.",
     ),
     Capability(
@@ -97,6 +121,7 @@ CAPABILITIES: tuple[Capability, ...] = (
         need=Need.OWN_CREDENTIAL,
         config_path="tools.media.speech.model",
         media_attr="speech",
+        env_var="OPENROUTER_API_KEY",
         cost_note="Billed per call.",
     ),
     Capability(
@@ -105,6 +130,7 @@ CAPABILITIES: tuple[Capability, ...] = (
         need=Need.OWN_CREDENTIAL,
         config_path="tools.media.video.model",
         media_attr="video",
+        env_var="OPENROUTER_API_KEY",
         cost_note="Billed per call; needs prepaid OpenRouter credit to run at all.",
     ),
     Capability(
@@ -144,6 +170,28 @@ def is_configured(cap: Capability, config: "Config") -> bool:
     return WebSearchTool.is_configured(config.tools.web.search.api_key)
 
 
+def has_credential(cap: Capability, config: "Config") -> bool:
+    """Whether a credential actually resolves for this capability.
+
+    A different question from :func:`is_configured`, which answers whether the
+    deployment asked for the tool. They come apart for the media family: a
+    section naming only a model is registered and offered to the model, and
+    every call then fails on a missing key. A report that collapses the two
+    ticks a capability that cannot run, which is the one thing it must not do.
+
+    The same answer twice for the other families -- web_search is configured by
+    a resolved key and nothing else, and web_fetch needs none -- so this only
+    ever diverges where the rule itself does.
+    """
+    if cap.need is Need.NOTHING:
+        return True
+    if cap.media_attr:
+        from raven.agent.tools.media_gen import _OpenRouterMediaTool
+
+        return _OpenRouterMediaTool.has_key(_resolved_media(cap, config))
+    return is_configured(cap, config)
+
+
 def configured_from(cap: Capability, config: "Config") -> str:
     """Where a satisfied capability got its credential, for a doctor row.
 
@@ -153,21 +201,54 @@ def configured_from(cap: Capability, config: "Config") -> str:
     have" and "needs a key" are different instructions, and a row that cannot
     tell them apart sends someone to create an account they already have.
 
-    Empty when the capability is unconfigured, and for the ones needing nothing
-    -- there is no credential to report on either.
+    Empty when the capability is unconfigured, when it needs nothing, and --
+    for the media family only -- when it is registered with no credential at
+    all: a section naming just a model is offered to the model and fails on
+    every call, so there is no source to name.
+
+    Names a source; it does not rule on whether one exists. Callers wanting that
+    answer ask :func:`has_credential`, which asks the tools -- inferring it from
+    an empty string here would make the two facts one, and they are not.
     """
     if cap.need is Need.NOTHING or not is_configured(cap, config):
         return ""
     if cap.media_attr:
+        if getattr(config.tools.media, cap.media_attr).api_key:
+            return cap.key_path
         # effective_media_config resolves the borrow, so a key present after it
         # but absent in the raw section came from the provider entry.
-        raw = getattr(config.tools.media, cap.media_attr)
-        if not raw.api_key and _resolved_media(cap, config).api_key:
-            return "providers.openrouter.apiKey (borrowed)"
-        return cap.config_path
-    if config.tools.web.search.api_key:
-        return cap.config_path
+        if _resolved_media(cap, config).api_key:
+            return f"borrowed: {_OPENROUTER_KEY}"
+    elif config.tools.web.search.api_key:
+        return cap.key_path
     return cap.env_var if os.environ.get(cap.env_var) else ""
 
 
-__all__ = ["CAPABILITIES", "Capability", "Need", "configured_from", "is_configured"]
+def borrowable_credential(cap: Capability, config: "Config") -> str:
+    """Where an unconfigured media capability would get its key once switched on.
+
+    "Reuse the OpenRouter key you already have" and "get a key as well" are
+    different instructions, and only this tells them apart. Stating the first
+    unconditionally is worse than saying nothing: the deployer sets a model,
+    the tool is registered because a model alone counts, and every call then
+    fails on a credential they were told they already had.
+
+    Empty for the other families, whose own rows already name what to set.
+    """
+    if not cap.media_attr:
+        return ""
+    openrouter = config.providers.get("openrouter")
+    if openrouter and openrouter.api_key:
+        return _OPENROUTER_KEY
+    return cap.env_var if os.environ.get(cap.env_var) else ""
+
+
+__all__ = [
+    "CAPABILITIES",
+    "Capability",
+    "Need",
+    "borrowable_credential",
+    "configured_from",
+    "has_credential",
+    "is_configured",
+]
--- a/raven/agent/tools/capabilities.py
+++ b/raven/agent/tools/capabilities.py
@@ -1,0 +1,173 @@
+"""What each credential-bearing tool needs, in one place.
+
+Providers have had a declarative credential model for a while --
+``providers.auth`` names what each connection method requires and
+``credential_status`` is the single authority on whether a provider is usable.
+An AST invariant enforces that authority, because six surfaces once answered
+the same question six ways and each looked reasonable alone.
+
+Tools never got the equivalent. Three rules decide whether a tool is offered to
+the model, one per family, each a different shape:
+
+    web_search   a resolved key, asked of the built tool
+    web_fetch    nothing -- always registered, a key only improves extraction
+    media x3     an api_key *or* a model, either one counting as configured
+
+Each rule is defensible where it sits. What is missing is anywhere to *read*
+them. A deployer cannot ask what this install lacks, and since an unconfigured
+tool stopped being registered there is no surface at all saying the capability
+exists: the model is not offered it, no document lists it, and ``raven doctor``
+reports on providers and memory but has never mentioned tools.
+
+This module is that surface. It describes the rules rather than replacing them
+-- ``is_configured`` mirrors the loop's judgement instead of inventing a second
+one -- and ``tests/test_tool_capabilities.py`` pins the description against what
+the loop actually registers, so the two cannot drift apart quietly. Drifting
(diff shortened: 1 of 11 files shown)


0xKT:
@claude

gloryfromca (commented):

Reviewed the full diff against `main` (5 commits, 11 files), the three `AgentLoop` construction sites, `effective_media_config`, and the doctor rendering; ran the tests rather than reading the table in the description.

**What I actually checked**

*The gate and the report read the same values.* All three call sites pass `brave_api_key=config.tools.web.search.api_key or None` and `media_config=config.effective_media_config()`, and `SubagentManager` inherits both from the loop, so `capabilities.is_configured` is consulting exactly what registration consults. For the media family `effective_media_config` borrows a key only into sections that are already configured, so `bool(api_key or model)` returns the same answer on the raw section and on the resolved one -- the table cannot tick a tool the loop withholds, which is the failure mode a second description of an existing rule invites.

*The tests bite.* 111 passed across the five touched test files, plus 38 in the neighbouring loop / subagent / media suites; `ruff check` and `ruff format --check` clean. I reverted the sub-agent gate to an unconditional `register(...)` and `test_the_subagent_surface_applies_the_same_rule` failed -- that is the call site a main-loop-only test would have missed. Nothing is skipped, xfailed or loosened. The one pre-existing test edited (`test_agent_loop_tool_search.py`) now supplies a key so its subject stays present, which is the honest fix rather than a weakened assertion.

*The derived gated set.* I checked #305 instead of taking the claim: it adds `voice_clone: MediaToolConfig` to `MediaGenConfig`, so `_media_attrs()` does pick it up and the table will fail loudly for whichever of the two lands second, naming the tool. Worth knowing the derivation keys on the annotation being `MediaToolConfig` -- a future media tool with its own config class would still slip past it.

*Rendering.* Drove `_render_tool_capabilities` against real configs in the borrow / nothing-to-borrow / keyless-model states. The rows say what they claim to, and rich wraps at spaces, so the dotted paths stay copyable in one piece.

*Project rules.* Commits are Conventional and ASCII-clean; test naming follows the tests section (`test_cli_doctor_commands.py` extended rather than a new CLI file); no assets; comments are all why-comments in English. The oldest commit is the only one without the `Co-authored-by` trailer -- squash collects it from the other four, so nothing is lost.

*Backward compatibility.* `doctor --json` grows a `tools` key with no in-tree consumer, and `web_search` disappears only where no key resolves from either source. `has_key` / `_resolve_key` preserve the chain `api_key` already walked, so existing callers get the same answers.

**One follow-up, not a blocker**

`DeepResearchTool.description` still ends with "for a single quick lookup use `web_search`" (`raven/agent/tools/deep_research.py:81`). In a deployment with a MiroThinker key and no Serper key, the model now reads that while `web_search` is absent from its tool list -- the mild version of the leak this PR fixes. It costs at most one rejected tool call, and that tool is on its way out per the description, so leaving it is reasonable.

gloryfromca (dismissed):

raven/agent/tools/capabilities.py
```diff
@@ -0,0 +1,254 @@
...
+    if cap.need is Need.NOTHING:
+        return True
+    if cap.media_attr:
+        from raven.agent.tools.media_gen import _OpenRouterMediaTool
+
+        return _OpenRouterMediaTool.is_configured(_resolved_media(cap, config))
+
+    from raven.agent.tools.web import WebSearchTool
+
+    return WebSearchTool.is_configured(config.tools.web.search.api_key)
```
gloryfromca, line 174: This answers only the credential gate, but production applies `tools.disabledTools` afterward in `AgentLoop._apply_disabled_tools`. With a Serper key plus `disabledTools: ["web_search"]`, I reproduced `is_configured=True` and a green doctor row naming `tools.web.search.apiKey`, while `loop.tools.has("web_search")` is false. That makes the new inventory claim a capability is offered when Raven has explicitly removed it. Please represent the disabled state separately (simply treating it as unconfigured would wrongly tell the deployer to set a key), and pin the doctor/table contract against the final registry with a credentialed disabled tool.

gloryfromca (commented):
Blocking: `raven doctor` can report a tool as configured even though Raven has removed it from the agent; see the inline note.

This is a fresh verdict for `669db655a085`. I compared it with the reviewed `71936e82f7a2`: four feature commits have identical stable patch IDs, while the doctor commit was relocated around the intervening `main` changes. The current branch merges cleanly and `git diff --check` is clean.

I re-read the full 11-file diff and covered the production `AgentLoop` construction paths through final tool registration, `SubagentManager`, `effective_media_config`, both doctor renderings, history/range-diff, backward compatibility, the repository rules, and whether the edited tests weakened their subjects. The existing test adjustment supplies the newly required key rather than weakening an assertion; I found no new skips or xfails.

Verification on this head: the five touched test files pass (`140 passed`), adjacent subagent/web/media tests pass (`23 passed`), focused Ruff check and format check pass, and every current GitHub check is green, including the full unit job. A direct credential-plus-disable probe reproduces the inline mismatch: doctor says configured and names `tools.web.search.apiKey`, while the final registry says `web_search` is absent.

raven/cli/doctor_commands.py
```diff
@@ -488,6 +577,74 @@ def _render_memory_capabilities(memory: MemoryInfo) -> None:
...
+    # is the only thing these rows are for.
+    indent = f"{'':<19}"
+    for cap in tools.capabilities:
+        label = f"  {cap.tool + ':':<17}"
+        if cap.configured:
+            where = f"  [dim]({cap.source})[/dim]" if cap.source else ""
+            # The marker carries the answer too: a green tick above a line
+            # saying every call fails is the same misreport in miniature.
+            mark = "[green]✓[/green]" if cap.has_credential else "[yellow]![/yellow]"
+            if cap.disabled:
```
gloryfromca, line 607: The renderer hides `disabled` whenever the same capability is unconfigured, so its setup advice omits the off switch that will still remove the tool. With no Serper key and `disabledTools: ["web_search"]`, I get `configured=False`, `disabled=True`, and `offered=False`, but the human output only says to set `tools.web.search.apiKey`; it never mentions `tools.disabledTools`. Following that advice and restarting still leaves search absent because `AgentLoop._apply_disabled_tools` removes it. The JSON row does carry `disabled=true`, so this is specifically the nested human-rendering branch. Please surface the disabled decision independently of credential setup and cover the configured/disabled state matrix rather than configuring every capability before testing the off switch.

gloryfromca (commented):
Blocking: the new disabled-state rendering still hides the off switch when the same capability is also unconfigured; see the inline note.

The new commit does close the previously reported credentialed-disabled case: Doctor now renders an `x`, names `tools.disabledTools`, preserves `configured=true` in JSON, and the new final-registry test passes. The prior five commits are unchanged according to `git range-diff`, so I focused the fresh review on this sixth commit and its integration with the already-reviewed capability table, Doctor renderings, and `AgentLoop._apply_disabled_tools`.

I also covered the full current diff scope, caller/history context for the disable rule, backward compatibility of the additive JSON field, repository rules, and test integrity. The new tests add assertions without weakening or skipping existing behavior, but they configure every capability before disabling it, so they do not exercise the independent false/true `configured`/`disabled` combination in the inline note. The commit is Conventional and ASCII-clean; comments are English; the existing correctly named test files were extended; no dependencies or assets changed.

Verification on `93b2d7b045ce`: the five touched test files pass (`148 passed`), adjacent subagent/web/media tests pass (`23 passed`), focused Ruff check and format check pass, commit-message validation passes, `git diff --check` is clean, and the merge tree is clean. All reported GitHub checks are green except the full unit job, which is still in progress. A direct renderer probe reproduces the inline output.

gloryfromca (commented):

Fresh verdict for `3fd13c9a680c`. The first five feature commits are unchanged; the amended sixth commit is the only feature delta from the prior review. It closes the remaining two-state rendering gap: a keyless capability named in `tools.disabledTools` now shows both the off switch and the credential setup path, while keyed/on and keyed/off continue to render correctly. I drove all four web-search configured/disabled combinations through the real gatherer and human renderer rather than relying only on the new assertions.

I covered the current full diff, the amended commit, the final-registry caller and disable-rule history, human and JSON compatibility, repository rules, and test integrity. The new tests expand the state matrix without weakening, skipping, or replacing an existing assertion. Test files remain in the required locations; comments and commit text are English and ASCII-clean; no PR dependency or asset changes were introduced. The target advanced only by the `0.1.13` version bump in `pyproject.toml` and `uv.lock`, with no path overlap, and the merge tree is clean.

Verification on this head: the five touched test files pass (`154 passed`), adjacent subagent/web/media tests pass (`23 passed`), focused Ruff check and format check pass, commit-message validation passes, and `git diff --check` is clean. Every reported GitHub check is green except the full unit job, which is still in progress.

gloryfromca (approved):

PR #313, opened 2026-08-13
test(context_engine): skip the chmod attachment case when the runner is root

Handsome-wzw:
## Summary

`test_an_attachment_that_cannot_be_read_costs_a_note_not_the_turn` makes a file
unreadable with `chmod 000` and asserts the renderer notes the failed attachment
instead of failing the whole turn. `chmod 000` does not block a root user, so
the file stays readable, the error path is never entered, and the case fails for
anyone running the suite as root. It passes in CI, which does not run as root,
so the failure is invisible there and shows up only on a developer's machine.

Skipped for root rather than reworked: the behaviour under test is what happens
when a read raises, and making that happen without permissions means faking the
error, which would test the mock. The case is worth keeping for the runs that
can exercise it honestly.

The commit is `0xKT`'s, cherry-picked from the GitLab trunk where it already
landed. Authorship is preserved.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other

Test-only.

## Verification

```
uv run pytest tests/test_read_file_image.py    105 passed, 1 skipped
```

Before this, the same command as root gave `1 failed, 105 passed`, with the
failure on this case. The skip reason names the cause, so a reader is not left
guessing why it is absent.

## Risk

None to the product: no source file is touched. Non-root runs, CI included,
execute the case exactly as before.

## Related Issues

N/A

### DIFF
--- a/tests/test_read_file_image.py
+++ b/tests/test_read_file_image.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import asyncio
 import base64
 import json
+import os
 import threading
 import time
 from pathlib import Path
@@ -1615,14 +1616,13 @@ def test_an_attachment_that_cannot_be_prepared_is_named_not_dropped(tmp_path: Pa
     assert "could not be prepared" in out
 
 
+@pytest.mark.skipif(os.geteuid() == 0, reason="chmod 000 does not block root")
 def test_an_attachment_that_cannot_be_read_costs_a_note_not_the_turn(tmp_path: Path) -> None:
     """Resolution only proved the path pointed at a file. Permissions can change
     between then and the read, and the file can be gone -- and this renderer runs
     deep inside turn assembly, where an ``OSError`` reaches the caller as a failed
     turn rather than as a message about one attachment.
     """
-    import os
-
     from raven.context_engine.segments import render
 
     locked = _write_image(tmp_path / "locked.png", (60, 40))


0xKT:
@claude  review

arelchan (approved):

gloryfromca (commented):

## What I checked

The diff itself, the surrounding test, the repo's own conventions (AGENTS.md sections 1/2/3/5), whether the skip weakens coverage, CI configuration, and prior art in the tree.

**The skip is honest, not a green-wash.** The concern with `skipif` is always that a case stops asserting anything. Here the guard is keyed to `os.geteuid() == 0` only, so every non-root run still executes the full body. Verified locally at euid 501:

```
uv run pytest tests/test_read_file_image.py -q -rs
106 passed in 16.44s        # 0 skipped, the case ran its assertions
uv run ruff check tests/test_read_file_image.py
All checks passed!
```

That matches the description's root-side figure (105 passed, 1 skipped) as the complement. The unit matrix in `.github/workflows/ci.yml` is `ubuntu-latest` and does not run as root, so CI keeps exercising the case exactly as before; the skip only fires on a root developer box.

**The alternative would have been worse.** The behaviour under test is what `build_user_content` does when the read raises. As root, DAC is bypassed, so the only ways to reach that path are patching the read or forcing a delete between resolution and read -- both of which assert against a mock rather than against the renderer. Guarding the case is the right trade, and the reason string names the cause so nobody has to rediscover it.

**It matches prior art.** `tests/test_importer_hermes_skills.py` already carries the identical decorator and reason string at two sites, so this is the established spelling in this repo rather than a new local invention. Moving `import os` to module scope is required, since the decorator is evaluated at collection time.

**Conventions.** Branch `test/skip_chmod_case_as_root` fits the `<type>/<snake_desc>` form; the commit header is 74 chars, lowercase, no trailing period, scope is a real subpackage; commit message and PR description are both ASCII-clean (grepped with `[^\x00-\x7F]`, zero matches); the `Co-authored-by` trailer is present in the commit and absent from the PR description, which is what squash-merge wants. No source file is touched, so there is no backward-compatibility surface.

**One pre-existing note, not a finding against this diff:** `os.geteuid` does not exist on Windows, so a module-scope call would be an `AttributeError` at collection there. This is already the situation in `test_importer_hermes_skills.py`, the unit matrix never runs on Windows, and the one `windows-latest` job runs a single unrelated integration file -- so nothing this PR does makes any run fail. Worth a `sys.platform` guard someday if the suite ever needs to collect on Windows; out of scope here.

PR #316, opened 2026-08-14
fix(tracing): reject a dashboard viewer whose ui files are gone

arelchan:
## Summary

The dashboard opened as unstyled HTML that never connected: `/` and `/api/health` answered 200 while `/app.css` and `/app.js` both 404'd.

The port-reuse guard added in #141 asks the viewer whether it is alive, not whether it can still serve a page. Both the health payload and the HTML shell are produced from memory, while `STATIC_DIR` is read per request -- so a viewer whose install directory was **removed** rather than overwritten keeps reporting itself healthy for as long as the process stays up, and the launcher happily reuses it. Observed after the local install changed form (a switch to an editable install rebuilt the environment on a different Python version, deleting the directory the running viewer had been started from); a plain in-place upgrade overwrites the same path and does not trigger it.

Two changes, because either probe alone accepts a viewer that cannot render:

- `_viewer_health` now also fetches an asset the page cannot work without, and requires 200. This catches any stale viewer regardless of the code it is running -- including one started before this change.
- `/api/health` now reports the UI as part of health, answering `503 {"ok": false, "ui": "missing"}` when its static directory no longer holds `app.js`, so a fresh process is honest about the same condition to any other consumer.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
uv run pytest tests/test_cli_tracing_commands.py tests/test_tracing_api.py -q
  -> 27 passed

uv run ruff check raven tests            -> All checks passed!
uv run ruff format --check raven tests   -> 814 files already formatted
node --check raven/tracing/viewer/server.js -> ok
```

The new test is not vacuous -- against the same fake viewer (health 200, assets 404) the previous implementation of `_viewer_health` returns `True` and the new one returns `False`.

End-to-end against a real viewer started with `app.js` absent from its static directory:

```
GET /api/health -> 503 {"ok":false,"port":4399,"stateDir":"...","ui":"missing"}
_viewer_health(4399) -> False
```

And against a healthy viewer, `/`, `/app.css`, `/app.js` and `/api/health` all return 200 with `"ui":"ok"`.

## Risk

The health endpoint gains a field and can now return 503. The only in-tree consumer is `_viewer_health`, which is updated here; an external consumer that treated any 200 as healthy would newly see a 503, which is the intended signal. No change to what the dashboard renders when it is working.

`_VIEWER_UI_PROBE` adds one localhost request per launch, on the path already taken only when a port is found occupied.

Rollback is reverting the commit: the guard returns to health-only and the endpoint to a bare `ok: true`.

## Related Issues

N/A -- follow-up to #141, which introduced the guard this hardens.

### DIFF
--- a/raven/cli/tracing_commands.py
+++ b/raven/cli/tracing_commands.py
@@ -37,6 +37,12 @@ def _viewer_dir() -> Path:
     return Path(__file__).resolve().parent.parent / "tracing" / "viewer"
 
 
+# Asset the viewer must still be able to read off disk for the page to work.
+# Served by ``serveStatic``, unlike the health payload and HTML shell, which the
+# process answers from memory and therefore survive their files being deleted.
+_VIEWER_UI_PROBE = "/app.js"
+
+
 def _port_live(port: int) -> bool:
     """True if something is already listening on 127.0.0.1:port."""
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
@@ -45,12 +51,20 @@ def _port_live(port: int) -> bool:
 
 
 def _viewer_health(port: int) -> bool:
-    """True only if *our* tracing viewer is serving on ``port``.
+    """True only if *our* tracing viewer is serving on ``port`` and can still
+    serve its UI.
 
-    A live port is not enough: a stale viewer from an older layout, or an
-    unrelated server (e.g. another observability tool), can hold it and 404
-    every request. The viewer answers ``/api/health`` with ``{"ok": true}``;
-    a foreign server does not, so this distinguishes reuse from a clash.
+    Two probes, because either one alone accepts a viewer that cannot render:
+
+    * ``/api/health`` separates our viewer from an unrelated server (e.g.
+      another observability tool) holding the port -- a foreign server does not
+      answer ``{"ok": true}``.
+    * ``_VIEWER_UI_PROBE`` separates a working viewer from one whose files were
+      deleted from under it. The health payload and the HTML shell are both
+      answered from memory, so a viewer whose install directory was replaced
+      (a version upgrade, a switch to an editable install) keeps passing the
+      first probe indefinitely while every asset read 404s -- the page then
+      renders unstyled and never connects.
     """
     try:
         with urllib.request.urlopen(f"http://127.0.0.1:{port}/api/health", timeout=0.5) as resp:
@@ -59,7 +73,13 @@ def _viewer_health(port: int) -> bool:
             data = json.loads(resp.read().decode("utf-8"))
     except Exception:  # noqa: BLE001 — any failure means "not our viewer"
         return False
-    return isinstance(data, dict) and data.get("ok") is True
+    if not (isinstance(data, dict) and data.get("ok") is True):
+        return False
+    try:
+        with urllib.request.urlopen(f"http://127.0.0.1:{port}{_VIEWER_UI_PROBE}", timeout=0.5) as resp:
+            return resp.status == 200
+    except Exception:  # noqa: BLE001 — a 404 raises here; either way the UI is gone
+        return False
 
 
 def _find_free_port(start: int, span: int = 20) -> int | None:
--- a/raven/tracing/viewer/server.js
+++ b/raven/tracing/viewer/server.js
@@ -981,7 +981,17 @@ const server = http.createServer((req, res) => {
   const url = new URL(req.url, `http://${req.headers.host || '127.0.0.1'}`);
 
   if (url.pathname === '/api/health') {
-    sendJson(res, 200, { ok: true, port: PORT, stateDir: STATE_DIR });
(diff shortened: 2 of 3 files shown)


raven/cli/tracing_commands.py
```diff
@@ -59,7 +73,13 @@ def _viewer_health(port: int) -> bool:
...
     except Exception:  # noqa: BLE001 — any failure means "not our viewer"
         return False
-    return isinstance(data, dict) and data.get("ok") is True
+    if not (isinstance(data, dict) and data.get("ok") is True):
+        return False
+    try:
+        with urllib.request.urlopen(f"http://127.0.0.1:{port}{_VIEWER_UI_PROBE}", timeout=0.5) as resp:
+            return resp.status == 200
+    except Exception:  # noqa: BLE001 — a 404 raises here; either way the UI is gone
+        return False
```
gloryfromca, line 82: Nonblocking, and about the call site rather than this line.

When this returns False for the newly covered reason - our own viewer, assets gone - `_open_dashboard` falls into the else branch and prints, at `raven/cli/tracing_commands.py:141`:

> Port {port} is held by another process (not the tracing viewer); starting on {free} instead.

In that case it *is* the tracing viewer, and the message points the user away from the only action that clears it. The stale process keeps holding the default port, so every later `raven tracing dashboard` walks to a fresh port and the user never learns why.

The two cases are already distinguishable here (health ok but UI probe failed, vs no health at all), so the branch could say something like "a stale tracing viewer is holding {port} with no UI assets; kill it or pass --port". Not a merge blocker - the fallback still hands the user a working dashboard.

gloryfromca (commented):

### What I actually verified

**The fix changes the outcome, checked against the real viewer (not only the tests).** Copied the viewer out of the tree and ran it under node: with `ui/app.js` present, `/api/health` answers `200 {"ok":true,...,"ui":"ok"}`; after deleting `ui/app.js` the same server answers `503 {"ok":false,...,"ui":"missing"}` and `/app.js` returns 404. So both halves bite, and they cover different populations - the server-side `uiOk` covers viewers started after this lands, and the client-side `_VIEWER_UI_PROBE` covers the case that motivated the change, an already-running viewer from the old build that keeps answering `ok:true` from memory forever.

**The probe target is the right asset.** `SHELL_HTML` is `require('./ui/shell.js')` at startup, so the shell is in memory and survives its file being deleted, while `serveStatic` resolves `/app.js` under `STATIC_DIR` per request. `ui/shell.js` loads exactly `/app.js`, so the probe asserts the asset the page needs. There is no SPA fallback that would make it vacuously pass: `serveStatic` special-cases only `/` and `/index.html`, everything else 404s when missing.

**The test is non-vacuous and nothing was weakened.** `uv run pytest tests/test_cli_tracing_commands.py -q` -> 5 passed, 0 skipped. With `raven/cli/tracing_commands.py` reverted to main and the new test kept, `test_viewer_health_false_when_ui_assets_are_gone` fails - so it would catch the regression. `test_viewer_health_true_for_our_viewer` still asserts True via the default `ui_ok=True`, so the happy path was not loosened to accommodate the new case.

**Backward compatibility.** An old CLI against a new server gets an `HTTPError` from the 503 and returns False, which is the direction you want (do not reuse a UI-less viewer). Nothing in `ui/` reads `/api/health`, so the new 503 has no in-page consumer to break.

**Repo rules and packaging.** Comments are English and explain a hidden constraint (in-memory shell vs per-request asset read) rather than restating the code, per AGENTS.md 1.1/1.2; branch and commit grammar match 2.1/3.1; the test lands in `tests/test_cli_tracing_commands.py`, which is the file 5.1 requires for `raven/cli/tracing_commands.py`. `raven/tracing/viewer/**/*.js` is in the hatch wheel include list, so `app.js` ships and a normal install will not trip the new probe.

Covered: the diff, the callers (`_open_dashboard` reuse and startup-wait paths), AGENTS.md, backward compatibility in both directions, and whether the tests were weakened. One nonblocking note inline, about the message the newly-covered False path lands on.

gloryfromca (approved):

raven/cli/tracing_commands.py
arelchan: Merged as-is, and this one is worth doing -- tracking it as a follow-up rather than dropping it.

You are right that the call site now lies for the case this PR added: the health probe says "not the tracing viewer" precisely when it is our own viewer with its assets gone, and it points away from the only action that clears the port. Since the stale process keeps the default port, every later run walks to a fresh one and the reason stays invisible -- so the fallback working is what hides it.

Taking it in the same follow-up as a couple of leftovers from #341, which touch the neighbouring CLI files. The branch will distinguish the two cases the probe already separates (health ok but UI probe failed, vs no health at all) and name the stale viewer in the message.

PR #317, opened 2026-08-14
feat: add python line and branch coverage governance to ci

forrestjgq:
## Summary

Adds Python coverage governance to CI so existing coverage cannot silently
regress and new code stays well tested, without pretending the project is
already near 100%.

- `scripts/coverage_gate.py` plus `make coverage*` targets compute line and
  branch coverage for the `raven` package and drive four gates:
  - **summary** of totals and the lowest-covered files;
  - **ratchet** of total line and branch coverage against the audited
    `.github/coverage-baseline.json` (0.05pp tolerance for rounding noise);
  - **diff coverage** requiring 90% on the executable lines a PR changes
    (comments, deletions, and non-executable lines are excluded);
  - **baseline-check** rejecting any proposed baseline that lowers line or
    branch coverage relative to the target branch.
- The CI unit job runs the suite once with coverage and reuses the report for
  every gate, uploads xml/json/html artifacts on every run (`if: always()`),
  and keeps diff coverage as a PR-only gate.
- `pyproject.toml` configures `coverage.py` (branch mode, `raven` source, three
  documented omissions); generated reports are gitignored.
- Key decision: the ratchet uses a manually bumped, audited baseline rather
  than auto-raising, so the diff-coverage gate is what guards new code while the
  ratchet only blocks regression.

The production-file pathspec uses git glob magic (`:(glob)raven/**/*.py`) so
top-level modules such as `raven/__init__.py` are not silently dropped from diff
coverage; a regression test covers this.

## Type

- [x] Feature

## Verification

- `uv run --frozen --python 3.12 pytest tests/test_coverage_gate.py -q` -> 13 passed
- `uv run --frozen --python 3.12 --extra dev ruff check scripts/coverage_gate.py tests/test_coverage_gate.py` -> All checks passed
- `uv run --frozen --python 3.12 --extra dev ruff format --check scripts/coverage_gate.py tests/test_coverage_gate.py` -> already formatted

## Risk

No runtime `raven` code changes; this is CI and tooling only. The new gates can
fail PRs that lower coverage, which is the intended behavior. Rollback is a
plain revert of this commit (or disabling the coverage steps in the unit job).

## Related Issues

N/A

### DIFF
--- a/.github/coverage-baseline.json
+++ b/.github/coverage-baseline.json
@@ -81,11 +81,11 @@
       "statements": 90
     },
     "raven/agent/loop/main.py": {
-      "branch_percent": 82.207207,
+      "branch_percent": 81.981982,
       "branches": 444,
-      "covered_branches": 365,
-      "covered_lines": 1037,
-      "line_percent": 87.069689,
+      "covered_branches": 364,
+      "covered_lines": 1036,
+      "line_percent": 86.985726,
       "statements": 1191
     },
     "raven/agent/loop/recovery.py": {
@@ -148,8 +148,8 @@
       "branch_percent": 94.444444,
       "branches": 18,
       "covered_branches": 17,
-      "covered_lines": 53,
-      "line_percent": 98.148148,
+      "covered_lines": 52,
+      "line_percent": 96.296296,
       "statements": 54
     },
     "raven/agent/tools/base.py": {
@@ -169,11 +169,11 @@
       "statements": 264
     },
     "raven/agent/tools/file_search.py": {
-      "branch_percent": 78.571429,
+      "branch_percent": 64.285714,
       "branches": 70,
-      "covered_branches": 55,
-      "covered_lines": 170,
-      "line_percent": 82.926829,
+      "covered_branches": 45,
+      "covered_lines": 148,
+      "line_percent": 72.195122,
       "statements": 205
     },
     "raven/agent/tools/filesystem.py": {
@@ -908,8 +908,8 @@
       "branch_percent": 80.681818,
       "branches": 88,
       "covered_branches": 71,
-      "covered_lines": 239,
-      "line_percent": 95.219124,
+      "covered_lines": 238,
+      "line_percent": 94.820717,
       "statements": 251
     },
     "raven/cli/gateway_commands.py": {
@@ -945,9 +945,9 @@
       "statements": 1012
     },
     "raven/cli/onboard_everos.py": {
-      "branch_percent": 43.442623,
+      "branch_percent": 43.032787,
       "branches": 244,
-      "covered_branches": 106,
+      "covered_branches": 105,
       "covered_lines": 300,
       "line_percent": 54.644809,
       "statements": 549
@@ -961,11 +961,11 @@
       "statements": 83
     },
     "raven/cli/provider_commands.py": {
-      "branch_percent": 84.375,
+      "branch_percent": 83.333333,
       "branches": 96,
-      "covered_branches": 81,
-      "covered_lines": 369,
-      "line_percent": 89.346247,
+      "covered_branches": 80,
+      "covered_lines": 368,
+      "line_percent": 89.104116,
       "statements": 413
     },
     "raven/cli/sandbox_commands.py": {
@@ -1017,11 +1017,11 @@
       "statements": 91
     },
     "raven/cli/tui_commands.py": {
-      "branch_percent": 74.637681,
+      "branch_percent": 75.362319,
       "branches": 138,
-      "covered_branches": 103,
-      "covered_lines": 365,
-      "line_percent": 81.111111,
+      "covered_branches": 104,
+      "covered_lines": 363,
+      "line_percent": 80.666667,
       "statements": 450
     },
     "raven/cli/update_notice.py": {
@@ -1097,11 +1097,11 @@
       "statements": 236
     },
     "raven/config/update_everos.py": {
-      "branch_percent": 85.714286,
+      "branch_percent": 92.857143,
       "branches": 14,
-      "covered_branches": 12,
-      "covered_lines": 61,
-      "line_percent": 91.044776,
+      "covered_branches": 13,
+      "covered_lines": 63,
+      "line_percent": 94.029851,
       "statements": 67
     },
     "raven/config/update_providers.py": {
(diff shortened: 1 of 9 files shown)


gloryfromca (commented):

My own pass over `github/main...e6e4b31f` (independent of the bot review already on the thread). What I covered: AGENTS.md rules, the whole diff (workflow, Makefile, pyproject, gate script, baseline, tests, tests/README.md), the gate's callers and the CI job log for the current head, backward compatibility of the changed Make targets, and whether the tests actually assert.

**Verification I ran**
- `uv run --frozen --python 3.12 --all-extras pytest tests/test_coverage_gate.py -q` -> **14 passed, 0 skipped**. No graceful-skip, no env-gated assertion; `test_git_diff_inputs_includes_top_level_raven_files` builds a real temp repo and exercises the `:(glob)raven/**/*.py` pathspec, which is exactly the trap the comment above it describes.
- Read the unit job log for this head: every gate really executed. `Line 75.56% / Branch 64.04%`, `Ratchet line delta=+0.00pp`, `Ratchet branch delta=+0.00pp`. The committed baseline reproduces bit-for-bit on CI, so the ratchet is not sitting on rounding luck.
- Note for the record: `Diff coverage: 100.00% (0/0 executable changed lines)` -- this PR touches no `raven/` file, so the diff gate is green vacuously here. Its logic is covered by unit tests, not by this run.

**Both earlier findings are genuinely fixed in `e6e4b31f`**, not just closed: `uv sync --locked --all-extras --dev --python 3.12` restores the lockfile-freshness check while keeping the matrix pin (`--frozen` on the later `uv run` calls only skips the lock *check*, the sync already enforced it), and `parse_changed_lines` now resets `path = None` on a non-`b/` target with a test for the `+++ /dev/null` case.

**Things I looked at and decided are not findings**
- `OMITTED_PATHS` duplicated between `pyproject.toml` and the script: already guarded by `test_documented_omissions_match_run_and_report_configuration`, so drift fails loudly.
- `exclude_also` rather than `exclude_lines`, so coverage.py's `pragma: no cover` defaults survive. Correct choice.
- `--python $(PYTHON_VERSION)` on every `uv run` will rebuild a contributor's 3.13 `.venv` at 3.12. That is the point of a reproducible baseline, and `tests/README.md` documents the override.
- Baseline JSON is 96 KiB and not an asset type AGENTS.md section 7 blocks; `repository files` is green.

Three inline notes, all non-blocking. The first one is the only one I would actually act on, and it can be a follow-up.

scripts/coverage_gate.py
```diff
@@ -0,0 +1,355 @@
...
+def check_baseline_update(current: dict[str, Any], previous: dict[str, Any]) -> bool:
+    if current.get("schema_version") != SCHEMA_VERSION or previous.get("schema_version") != SCHEMA_VERSION:
+        raise ValueError("current and target-branch baselines must use the supported schema")
+    passed = True
+    for metric in ("line_percent", "branch_percent"):
+        value = float(current["totals"][metric])
+        target = float(previous["totals"][metric])
+        delta = value - target
+        print(f"Baseline {metric}: proposed={value:.6f}% target={target:.6f}% delta={delta:+.6f}pp")
+        if value < target:
```
gloryfromca, line 136: **[non-blocking, but the one worth acting on]** The two gates together make deleting well-covered production code unmergeable, with no documented escape.

Overall coverage is `C/S`. Remove a chunk whose own coverage `c/s` is above the project average and the total *drops*: deleting a 100%-covered 200-statement module takes line coverage from `32494/43003 = 75.5622%` to `32294/42803 = 75.4497%`, i.e. `-0.11pp`, past the `0.05pp` tolerance -> `coverage-ratchet` fails. The obvious remedy is to lower the baseline in the same PR, but `check_baseline_update` rejects any decrease (`if value < target`, strict, no tolerance), so `coverage-baseline-check` fails too. `tests/README.md` closes the last door explicitly: "Never edit the percentages by hand or lower the baseline to make a change pass."

So the author's only options are editing `COVERAGE_RATCHET_TOLERANCE` in the Makefile (a permanent loosening for a one-off deletion) or padding unrelated tests until the number comes back. This is not an exotic case -- removing a dead but well-tested module is routine, and 200 statements is enough to trigger it.

Worth defining the sanctioned path before people hit it. Options, cheapest first: (a) document that a baseline *decrease* is allowed when the PR is a net deletion, and have `check_baseline_update` accept it behind an explicit marker rather than a blanket ban; (b) compare against coverage recomputed on the target branch instead of a static file, so deletions cancel out; or (c) ratchet on `covered/total` counts of surviving files rather than the whole-project ratio. Same reasoning applies to `branch_percent`.

Secondary point on this function: the ratchet tolerates `0.05pp` of noise but `check_baseline_update` tolerates none, so a candidate refreshed from a `main` run that dipped by `0.000001pp` is rejected. Worth sharing one tolerance.

.github/coverage-baseline.json
```diff
@@ -0,0 +1,3576 @@
...
+      "branch_percent": 80.0,
+      "branches": 10,
+      "covered_branches": 8,
+      "covered_lines": 33,
+      "line_percent": 94.285714,
+      "statements": 35
+    }
+  },
+  "python": "3.12",
+  "reference_commit": "840d05101b393354e24bba0191b6d9ae40790df6",
```
gloryfromca, line 3565: **[nit]** This sha is not resolvable in a clone. `git cat-file -t 840d0510` -> `fatal: could not get object info`; via the API it is `Merge af70757b into cd686453`, i.e. the ephemeral `refs/pull/317/merge` commit GitHub synthesises for the PR, reachable from no branch and gone after the squash-merge.

It comes from `coverage-baseline-candidate` running `git rev-parse HEAD` while the unit job sits on the merge ref (the `pre-commit` job pins `ref: github.event.pull_request.head.sha` for exactly this reason; the unit job does not). `tests/README.md` tells the reviewer to "Review the reference commit and totals in the candidate before replacement", which this value cannot support.

Post-bootstrap candidates from `main` push runs will record a real commit, so the damage is limited to the tracked bootstrap baseline plus any candidate a contributor pulls from a PR run to sanity-check a refresh. Cheap fix: thread `--commit ${{ github.event.pull_request.head.sha || github.sha }}` through the Make target.

AGENTS.md
```diff
@@ -22,6 +22,9 @@ Hard constraints only (violations get reverted / rejected). Soft suggestions and
 
 ### §1.1 Top rule: don't add comments unless necessary
 
+- Every new code file must document its purpose in English with an appropriate
```
gloryfromca, line 25: **[nit]** This is a new repo-wide hard constraint riding along in a CI-coverage PR. AGENTS.md is self-describing as "hard constraints only (violations get reverted / rejected)", so from merge onward every new file in the repo is subject to it, and reviewers looking at a coverage PR are not the audience that would weigh in on it. The bullet itself is fine and the follow-on sentence resolves the tension with the "don't comment unless necessary" top rule -- it is the bundling I would flag.

Also: the summary table at the top still gives section 1 as "Don't comment unless necessary; comments in English", which no longer describes a section that now *mandates* a module docstring in every new file. If the bullet stays, the gist row should mention it.

gloryfromca (approved):

PR #324, opened 2026-08-16
feat(session): auto-name sessions and persist bare create

0xKT:
## Summary

Session UX improvements found in an internal usability review of v0.1.11.
`sessions create` now persists minimal metadata immediately (with or
without --title), so create/list/delete round-trips work instead of
create handing out an id that list cannot see and delete reports as not
found. Sessions are auto-titled from the first user message (first line,
whitespace-collapsed, 40-char cap) at save time; explicit titles are
never overwritten, and forks only inherit human-given titles. One
pre-existing test was renamed and inverted (test_create_lazy_no_file to
test_create_bare_persists_file) because it pinned the exact lazy behavior
this change removes.

## Type

- [x] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_cli_session_commands.py tests/test_session_manager.py tests/test_session_export.py -x`: 117 passed.
- Full suite `uv run pytest -q`: 1 failed / 6149 passed; the single failure is the pre-existing test_read_file_image failure already present on main.
- Sandbox-HOME real runs: create -> id visible in list (Messages=0) -> delete exits 0; deleting an unknown id still reports not found with exit 1; a real turn produces a 40-char truncated EN title and a correctly rendered CJK title.
- All changes were written red-first against frozen acceptance assertions.

## Risk

Behavior change by design: bare `sessions create` now writes a metadata
file immediately instead of being lazy. Existing sessions are untouched;
old sessions without titles keep showing "-". Rollback = revert the
branch; the metadata format is unchanged.

## Related Issues

N/A - found in an internal usability review of v0.1.11; no tracked GitHub issues.

### DIFF
--- a/raven/cli/session_commands.py
+++ b/raven/cli/session_commands.py
@@ -4,14 +4,15 @@ User-facing "session id" is the bare chat_id (strip "cli:" prefix for
 display; re-prepend internally). Full session key = "cli:<chat_id>".
 
 Persistence semantics:
-- ``session create`` (bare): mints a new id and prints it. Nothing is
-  written to disk — the id materialises on first use (lazy). Note: a
-  lazily-minted id that was never used cannot be found by ``resume``
-  because no file exists yet; the output includes a reminder.
-- ``session create --title TEXT``: mints + immediately persists metadata
-  so the title survives process exit. This diverges from the TUI's lazy
-  semantics because the CLI process dies immediately after the command
-  returns, leaving no opportunity for a later lazy flush.
+- ``session create`` mints a new id and immediately persists minimal
+  metadata (zero messages, title optional), so the id shows up in
+  ``list`` and can be ``resume``d or ``delete``d right away. The CLI
+  process dies as soon as the command returns, so a lazy (in-memory)
+  session would be lost — hence eager persistence, diverging from the
+  TUI's lazy semantics.
+- Untitled sessions are auto-named from their first user message (see
+  ``raven.session.manager``); ``--title`` names the session up front
+  and is never overwritten.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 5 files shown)


0xKT:
Review comments addressed in 3 follow-up commits:

1. [blocker] empty-session recency hijack -- fixed in SessionManager.find_most_recent_chat_id. Recency now runs two tracks: the most-recently-updated session WITH messages always wins; zero-message sessions (the immediately-persisted bare `sessions create` files) are only the fallback when the channel has no messaged session at all. This protects both `agent --continue` and cron/sentinel proactive delivery. Red test was committed first and turned green by the fix: an empty session with a later updated_at no longer beats the newest messaged session; a second test pins the all-empty fallback.

2. [nit] fork title-inheritance false negative on exact-text coincidence -- fixed. save() now stamps `title_auto: true` in metadata when it auto-names; fork() checks that marker instead of comparing the title text against the first-user-message derivation (the comparison is gone). A title without the marker counts as human, so sessions persisted before the marker existed keep inheriting as before (back-compat). `Session.set_title()` is the new rename choke point that clears the marker; the TUI rpc title setter and `sessions create --title` route through it, so renaming an auto-named session correctly flips it to human. New tests: a human title verbatim-equal to the auto-derived text still yields `<title> (fork)`; an auto-named source keeps the marker across a disk round-trip and its child stays untitled.

Test run: `uv run pytest tests/test_session_manager.py -x` -> 68 passed; adjacent files (test_cli_session_commands, test_tui_rpc_session, test_cron_delivery, test_cli_cron_handler, test_sentinel_persistence, test_session_export, test_agent_loop_session_stamps) -> 169 passed. ruff format/check clean.

0xKT:
Commit hashes for the addressed items above, for the record: 7e6038a (red test for the recency hijack), 61f5e8d (blocker fix: prefer messaged sessions), 57540e9 (nit fix: title_auto marker replaces the text compare).

arelchan (approved):

PR #327, opened 2026-08-16
fix(*): enforce skill hub install safety policy at a SkillPolicy gate

0xKT:
## Summary

Fixes a critical skill hub supply-chain gap: a low-quality or
malicious hub skill could be silently installed and take over a core
capability with zero user awareness, because no install-time policy
existed anywhere between hub content and host execution.

What was broken:

- The `min_safety` guard was a no-op on the standard path: the catalog
  payload omits `score_safety` (it only exists in the detail response),
  and nothing checked the detail-level score. The config docstring
  claimed `use_skill` refuses low-scored skills; the implementation had
  no such check.
- The post-gate hydrate installed hub bundles silently on the context
  assembly hot path: no notification, no audit trail, no uninstall.
- No operator kill switch existed for a bad skill, and a skill body
  pointing at another product's home dotdir (the tag-memory
  `~/.openclaw` incident) installed without objection.

What this PR adds, in two commits:

1. `fix(*)`: enforce the safety bar on both install paths. The segment
   builder drops hub hits whose detail `score_safety` is below
   `min_safety` right after the pre-gate hydrate (the first place the
   score is actually available), and `use_skill` refuses before
   download, reusing the vetted detail metadata as `prefetched_meta` so
   no extra round-trip is added. Adds `skillForge.blocklist` (dropped
   from the pool for every source, case-insensitive) plus an install
   audit JSONL and a one-line install notice on both paths.
2. `feat(*)`: consolidate all checks into a single `SkillPolicy`
   chokepoint (`raven/skill_hub/policy.py`) that both install paths
   consult: blocklist, safety bar, and a new external home-dotdir lint
   that refuses skills referencing foreign product directories rather
   than rewriting them. Adds `raven skill block/unblock/remove` for the
   operator lifecycle.

Release note for existing installs: users who already have tag-memory
on disk should run `raven skill block tag-memory && raven skill remove
tag-memory --yes` after upgrading; a server-side takedown cannot clean
already-installed bundles.

Follow-ups tracked outside this PR: server-side takedown and publish CI,
`score_safety` in the catalog payload, and the `autoInstall`
prompt/auto/off consent modes (stacked PR `feat/skill_consent_visibility`
builds on this branch).

## Type

- [x] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_skill_hub_policy.py tests/test_skill_segment_builder.py tests/test_skill_hub_tools.py tests/test_cli_skill_commands.py -q` - 79 passed (41 new tests, written failing-first to reproduce the gap)
- `uv run pytest -q` on this branch rebased onto upstream main (cd68645) - 1 failed / 6186 passed; the single failure is the documented pre-existing red (`test_read_file_image.py::test_image_block_is_the_only_place_the_shape_is_written`, see `.design/94b-upstream-baseline-failures.txt`), zero regression from this change
- `uv run ruff check` on all touched files - clean
- Isolated real-machine acceptance (fresh HOME, real LLM, local hub stub mirroring the real catalog shape): low-score skill dropped with zero disk writes and no foreign dotdir created; a 0.9-scored skill with a foreign-path body refused by the lint alone; blocklist enforced for hub hits and for an already-installed copy surfacing via the local registry; `use_skill` refusal verified with the model actively calling it; install audit record written; `skill remove` deletes the bundle

## Risk

- [ ] Rollback path is clear for risky changes

Security: this PR is itself a security hardening; the lint refuses
rather than rewrites, so no third-party content is modified.
Backward compatibility: deployments whose detail payload omits
`score_safety` keep working (missing score passes); `blocklist`
defaults to empty; the audit file is append-only and best-effort.
Rollback: revert the two commits; no data migration involved.

## Related Issues

N/A - found in an internal security review of the skill hub; no tracked GitHub issues.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -418,6 +418,18 @@ candidates into the weighted RRF (weight 0.85, below Local 1.0 and Everos 0.9), 
 `read_skill` / `use_skill` tools do on-demand body fetch / script materialization. Replaces
 the retired "Mass" source.
 
+**SkillPolicy** (`skill_hub/policy.py`):
+The install-time safety decision both Hub install paths consult before any
+`SkillHubClient.install()` — the segment builder's post-gate hydrate and the `use_skill`
+tool. `refusal_for_detail()` checks, in order: the operator blocklist
+(`skillForge.blocklist`, matched case-insensitively against name / slug / native id), the
+`min_safety` bar against the *detail*-level `score_safety` (the catalog payload omits the
+score; a missing or malformed score passes), and an external home-dotdir lint over the
+skill body (`~/.raven` is allowed; any other dotdir reference refuses the install). A hub
+candidate whose detail fetch fails is unvetted and dropped — it never reaches `install()`.
+Every install that passes is appended to a JSONL audit trail
+(`<workspace>/skills/hub/installs.jsonl`, `skill_hub/audit.py`).
+
 **Episode**:
 A distilled event note the Consolidation step writes to `episodes.md`.
 
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -481,6 +481,12 @@ class AgentLoop:
             workspace,
             skill_forge_router_config,
         )
+        # Install-policy knobs for the use_skill tool (registered later in
+        # ``_register_builtin_tools``, which no longer sees these configs).
+        self._skill_min_safety = float(
+            getattr(getattr(skill_forge_router_config, "hub", None), "min_safety", 0.7),
+        )
+        self._skill_blocklist = list(getattr(skill_forge_config, "blocklist", None) or [])
 
         self.context_engine: "ContextEngine" = build_context_engine(
             workspace=workspace,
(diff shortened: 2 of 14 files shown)


0xKT:
Confirmed real and fixed in e99e010.

The report was accurate: a transient `get()` failure during the pre-gate
hydrate left the candidate in the pool unvetted (`content=""`, absent from
`prefetched_meta`), and the post-gate step then called
`install(prefetched_meta=None)`, whose internal re-fetch downloaded and
installed the bundle without ever consulting `SkillPolicy` -- a network flap
turned into a full policy bypass.

Fix (fail closed, two layers):

1. `_hydrate_bodies` now drops any hub candidate whose detail fetch failed --
   an unvetted candidate never stays in the pool.
2. `_hydrate_refs` additionally refuses to install a hub hit that has no
   vetted detail metadata, so `install()` is never handed a hub candidate
   with `prefetched_meta=None` (this also closes the prefilled-content shape
   that skips the hydrate step entirely).

Tests, written red-first against the pre-fix code:

- `test_hub_hit_dropped_when_detail_fetch_fails`: `get()` raising a
  transient error -> candidate dropped, `install()` unreachable.
- `test_hub_hit_without_vetted_meta_never_installs`: a hydrate-skipping hit
  -> install refused.

Full suite after the fix: 1 failed / 6188 passed; the single failure is the
documented pre-existing red (`test_read_file_image.py`, see
`.design/94b-upstream-baseline-failures.txt`).

0xKT:
Nit addressed in b383151: added a SkillPolicy entry to CONTEXT.md's Skill Hub section (decision order, blocklist matching, detail-level score check, external-dotdir lint, fail-closed drop on fetch failure, and the install audit trail) per AGENTS.md section 6. CI green on the previous head; docs-only change.

0xKT:
Nit addressed in 65edb39: set_skill_blocked added to the update module's __all__. The [blocker]-tagged inline on skills.py is the first-pass anchor, already fixed in e99e010 and verified in the follow-up review ("Prior blocker verified fixed ... No blockers; this can merge").

arelchan (approved):

PR #329, opened 2026-08-17
refactor(cli): retire the interactive repl and the cli cron channel

0xKT:
## Summary

Consolidate the interactive surface on the TUI. Bare `raven agent` no
longer hosts a REPL: it prints a pointer to `raven tui` and exits
non-zero, while `raven agent -m "..."` keeps working for scripted
one-shot turns. The REPL-only plumbing goes with it (the interactive
loop, slash-command handling, the exit-time pending-reminder notice);
CliOutlet and build_repl stay because the one-shot path renders through
them.

Cron follows the same consolidation (this builds on the fire-at-origin
ownership rule): the one-shot path no longer registers the cron tool -
with no REPL there is no runner left for a cli-bound reminder, so
creating one from a one-shot turn would mint a job that never fires.
The "cli" channel value retires: `cron add` validates `--channel`
against `tui` plus the enabled IM channels, and jobs stored with
`channel="cli"` by older builds migrate to `"tui"` at load time, one
log line per migrated job (old jobs.json files keep loading unchanged
otherwise).

## Type

- [x] Refactor

## Verification

  5991 passed / 2 failed, both failures pre-existing on main and on the
  base branch run the same way (test_read_file_image, plus the
  order-dependent test_cli_theme flake that fails whenever its file
  runs as a whole) - zero failures introduced.
- Affected-area suites (onboard, agent CLI, cron CLI, repl spine,
  migration): 308 passed.
- Real-machine checks (real CLI subprocesses under an isolated HOME):
  bare `agent` exits non-zero with the tui pointer; `cron add
  --channel cli` is rejected listing the valid channels; a legacy
  jobs.json with a cli-bound job shows the job migrated to tui on read.

## Risk

Behavior change by design: the interactive REPL entry point is removed
(pointer + non-zero exit); scripts using `raven agent -m` are
unaffected. Stored cli-bound cron jobs migrate to tui automatically;
no data is dropped. Rollback = revert the squash commit.

## Related Issues

N/A


raven/cli/agent_commands.py
```diff
@@ -210,31 +112,28 @@ def agent(
...
-        # Startup gate: a config that cannot reach a model is settled first. Only
-        # on an interactive TTY -- scripted one-shots (`-m`) and non-TTY pipes must
-        # fail loudly later rather than block on prompts.
-        if message is None and _stdout_isatty():
-            from raven.cli.onboard_commands import ensure_ready_to_start
-
-            ensure_ready_to_start()
+        if message is None:
+            console.print(
+                "[yellow]The interactive REPL was removed. Use [bold]raven tui[/bold] "
```
gloryfromca, line 121: `docs/dev.md:57` still says:

| `raven agent` | Start interactive chat session |

That row is now wrong -- following the getting-started doc gets you this pointer and exit code 2. The PR updated `CONTEXT.md` but not the user-facing command table, so the one doc a new contributor reads first now advertises a command that fails. Suggest replacing the row with `raven tui` and keeping the `raven agent -m "Hello"` row below it.

gloryfromca (commented):

### What I covered

Re-based the read on the real PR base (`github/main` = d594acbb, i.e. #326 already merged), so the scope here is just: delete the REPL (`_repl_slash.py`, `run_repl_loop`, prompt-toolkit session, `warn_about_pending_cli_reminders`), turn bare `raven agent` into a pointer + exit 2, and retire the `cli` channel value (load-time `cli` -> `tui` migration, `default_channel="tui"`, `_INTERACTIVE_CHANNELS = ("tui",)`).

- **Diff correctness.** The one-shot `-m` body is byte-identical to main's apart from unwrapping the `if message:` and `interactive=message is None` -> `interactive=False` (same value, since `-m` is now required). Dropping `cron_service` from the `-m` `AgentLoop` is right, not a loss: post-#326 a job the one-shot created would be bound to a runner that exits immediately, so it could never fire.
- **The `cli` -> `tui` migration.** In-memory at `_load_store`, persisted by the next save; `start()` always saves, so it lands on first runner start rather than re-logging forever. Gateway leaves the migrated job alone (`tui` is outside its partition) but still persists the rewrite, and the TUI picks it up. Non-`cli` channels and `channel=None` pass through unchanged, which matches the three cases the new test asserts.
- **Onboarding gate.** Removing the `raven agent` gate is safe: `raven tui` has its own `_stdout_isatty()`-guarded `ensure_ready_to_start()` (`raven/cli/tui_commands.py:1024`), and the deleted `test_agent_gate_*` cases are replaced by `test_agent_bare_exits_with_pointer_without_wizard` plus the tui-only parametrisation of `test_a_stale_default_model_does_not_restart_the_wizard`.
- **Callers of the retired surface.** No remaining references to `_repl_slash`, `run_repl_loop`, `EXIT_COMMANDS`, `--legacy-repl`, or `channel="cli"` outside the migration itself. `prompt_toolkit` is still a live dependency via onboard/questionary.
- **Tests.** Not weakened. The removed tests covered code that no longer exists, and the new `tests/test_cron_service_migration.py` asserts real behaviour (channel rewrite, exactly one INFO line, persistence on next save, pass-through for other channels).

### Test run

`uv run pytest tests -q -p no:randomly` -> **1 failed, 5992 passed, 43 skipped**. The one failure is `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`, which is **pre-existing**: I ran the same selection in a clean `github/main` worktree and it fails there too. It passes standalone and only fails in a full-suite run, so it is cross-test pollution of rich's global state, unrelated to this diff. Not this PR's to fix, but worth someone's ticket.

### Non-blocking nits

1. Inline note below on `docs/dev.md`.
2. `CONTEXT.md:180` still carries `_Avoid_: using "CLI" for the interactive REPL (retiring)` under the **CLI** term. The PR updates the Fire-at-origin entry right above it but leaves this pointing at a REPL that no longer exists; per AGENTS.md section 6 the definition should be verifiable against the code.
3. `get_cli_history_path` (`raven/config/paths.py:63`, re-exported in `raven/config/__init__.py`) had exactly one caller, the deleted `_init_prompt_session`. It is now dead, and `~/.raven`'s history file is orphaned.
4. Scope nit on the header: the diff spans `raven/cli/**` and `raven/proactive_engine/schedulers/cron/service.py`, so AGENTS.md section 3.1 wants the scope omitted or `(*)` rather than `(cli)`. The PR title becomes the squash header, so it is the title that matters here. Minor.

gloryfromca (dismissed):

0xKT:
Good catch - fixed in c44fa12: the command table now points interactive chat at `raven tui` (the `-m` one-shot row stays). Also swept the whole docs/ tree for other stale interactive `raven agent` mentions; that table row was the only one.

gloryfromca (approved):

PR #330, opened 2026-08-17
fix: provider auth error rendering with exit codes and guidance

0xKT:
## Summary

Stacked on the REPL retirement (#329); retargets automatically as the
chain merges. Provider authentication failures used to reach the user
as a raw exception dump - the exception class name printed twice, the
provider's JSON error body rendered as if it were the agent's reply,
the process exited 0, and the memory query-rewrite layer echoed the
same failure once more. A zero-credential first run pointed at hand
editing config.json and never mentioned the guided path.

Changes:

- Provider errors carry the existing error classification through to
  the CLI instead of being stringified into content. The one-shot path
  renders an auth failure as two lines (Error: API key invalid
  (openrouter 401). / Fix: raven provider test openrouter  or  raven
  onboard) and exits 1 via an exit-code seam that survives the
  delivery-hub worker boundary, where a plain typer.Exit would be
  swallowed.
- The raw JSON error body and the duplicated exception prefix are
  stripped from the canonical message; the retry ladder and failover
  semantics are untouched.
- The query-rewrite failure echo drops to debug (it already falls back
  to plain retrieval).
- Both zero-credential raise sites in check_provider_credentials append
  "Or run raven onboard for guided setup." to the existing remedy line.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_cli_helpers.py tests/test_cli_agent_commands.py -x`: 48 passed.
- Full suite (no -x): 6232 passed / 2 failed, both pre-existing on main (the image-block test and the order-dependent theme test).
- Live sandbox with a fake OpenRouter key (real 401): before - bare exception double-write, JSON body, exit 0, rewrite warning; after - the two-line message, exit 1, zero AuthenticationError occurrences. Zero-config sandbox now prints the onboard line and exits 1.
- Each change landed only after a test reproducing the old behavior failed against the unmodified base.

## Risk

Scripts that relied on a zero exit code from a failed one-shot turn
will now see exit 1; that is the point of the fix. Revert the branch to
roll back.

## Related Issues

N/A - no tracked GitHub issues (internal v0.1.11 usability review).

### DIFF
--- a/raven/cli/_helpers.py
+++ b/raven/cli/_helpers.py
@@ -70,10 +70,7 @@ def check_provider_credentials(config: Config) -> None:
         raise MissingCredentialsError(
             status.summary,
             provider=provider_name,
-            remedy=(
-                "Run: raven provider set <name> --api-key <key>\n"
-                "Or run `raven onboard` for guided setup."
-            ),
+            remedy=("Run: raven provider set <name> --api-key <key>\nOr run `raven onboard` for guided setup."),
         )
 
 
--- a/raven/cli/_helpers.py
+++ b/raven/cli/_helpers.py
@@ -70,7 +70,7 @@ def check_provider_credentials(config: Config) -> None:
         raise MissingCredentialsError(
             status.summary,
             provider=provider_name,
-            remedy=("Run: raven provider set <name> --api-key <key>\nOr run `raven onboard` for guided setup."),
+            remedy="Run `raven onboard` for guided setup.",
         )
 
 
--- a/raven/cli/_helpers.py
+++ b/raven/cli/_helpers.py
@@ -59,12 +59,19 @@ def check_provider_credentials(config: Config) -> None:
             "no provider configured",
             # A command, not a config path: the old text pointed at
             # ~/.raven/config.json, the layout the CLI exists to hide.
-            remedy="Run: raven provider set <name> --api-key <key>, then raven provider use <name>/<model>",
+            remedy=(
+                "Run: raven provider set <name> --api-key <key>, then raven provider use <name>/<model>\n"
+                "Or run `raven onboard` for guided setup."
+            ),
         )
 
     status = credential_status(provider_name, config.providers.get(provider_name), include_external=True)
     if not status.ok:
-        raise MissingCredentialsError(status.summary, provider=provider_name)
+        raise MissingCredentialsError(
+            status.summary,
+            provider=provider_name,
+            remedy="Or run `raven onboard` for guided setup.",
+        )
 
 
 def make_provider(config: Config):
--- a/raven/cli/_helpers.py
+++ b/raven/cli/_helpers.py
@@ -70,7 +70,10 @@ def check_provider_credentials(config: Config) -> None:
         raise MissingCredentialsError(
             status.summary,
             provider=provider_name,
-            remedy="Or run `raven onboard` for guided setup.",
+            remedy=(
+                "Run: raven provider set <name> --api-key <key>\n"
+                "Or run `raven onboard` for guided setup."
+            ),
         )
 
 
--- a/raven/cli/agent_commands.py
+++ b/raven/cli/agent_commands.py
@@ -53,6 +53,20 @@ def _print_agent_response(response: str, render_markdown: bool) -> None:
     console.print()
 
 
+# Category-apt hints for non-auth provider errors. Credential guidance would
+# mislead here: a rate limit or a network drop is not fixed by re-checking the
+# key. Categories absent from this map (invalid_request, unknown, ...) render
+# the error line alone.
+_NON_AUTH_HINTS = {
+    "rate_limit": "Hint: the provider is rate limiting; retry in a moment.",
+    "network": "Hint: network problem; check connectivity and retry.",
+    "context_overflow": "Hint: the input exceeds the model's context window; shorten it.",
+    "server": "Hint: provider-side error; retry later or switch models.",
+    "model_unavailable": "Hint: model not served; pick another with raven provider use <name>/<model>.",
+    "billing": "Hint: billing or quota issue; check your provider account.",
+}
+
+
 def _print_llm_error(content: str) -> bool:
     """Render a provider error as a diagnosis + fix hint instead of a fake
     agent reply. Returns True when handled; marks the one-shot path to exit
@@ -69,10 +83,13 @@ def _print_llm_error(content: str) -> bool:
     if category == "auth":
         where = f"{provider} 401" if provider else "401"
         console.print(f"[red]Error: API key invalid ({escape(where)}).[/red]")
+        target = provider or "<name>"
+        console.print(f"Fix: raven provider test {escape(target)}  or  raven onboard")
     else:
         console.print(f"[red]Error: LLM call failed ({escape(category)}): {escape(detail[:200])}[/red]")
-    target = provider or "<name>"
-    console.print(f"Fix: raven provider test {escape(target)}  or  raven onboard")
+        hint = _NON_AUTH_HINTS.get(category)
+        if hint:
+            console.print(hint)
     console.print()
     _ONE_SHOT_EXIT["code"] = 1
     return True
--- a/raven/cli/agent_commands.py
+++ b/raven/cli/agent_commands.py
@@ -81,8 +81,11 @@ def _print_llm_error(content: str) -> bool:
     category, provider, detail = parsed
     console.print()
     if category == "auth":
-        where = f"{provider} 401" if provider else "401"
-        console.print(f"[red]Error: API key invalid ({escape(where)}).[/red]")
+        # No status code here: the auth bucket also fires on 403, on
+        # PermissionDeniedError and on substring matches, so naming one would
+        # be a guess. The detail carries the provider's own reason instead.
+        where = f" ({escape(provider)})" if provider else ""
+        console.print(f"[red]Error: provider rejected the credentials{where}: {escape(detail[:200])}[/red]")
         target = provider or "<name>"
         console.print(f"Fix: raven provider test {escape(target)}  or  raven onboard")
     else:
--- a/raven/cli/agent_commands.py
+++ b/raven/cli/agent_commands.py
@@ -35,14 +35,47 @@ from raven.utils.helpers import sync_workspace_templates
 console = Console()
 
 
+# One-shot ``-m`` exit code: set by the error renderer, checked after the
+# turn. Module-level because the render callback runs inside the delivery
+# hub's worker task, where raising typer.Exit would be swallowed.
+_ONE_SHOT_EXIT = {"code": 0}
+
+
 def _print_agent_response(response: str, render_markdown: bool) -> None:
     """Render assistant response with consistent terminal styling."""
     content = response or ""
+    if _print_llm_error(content):
+        return
     body = Markdown(content) if render_markdown else Text(content)
     console.print()
     console.print(f"[cyan]{__logo__} Raven[/cyan]")
     console.print(body)
     console.print()
+
+
+def _print_llm_error(content: str) -> bool:
+    """Render a provider error as a diagnosis + fix hint instead of a fake
+    agent reply. Returns True when handled; marks the one-shot path to exit
+    non-zero."""
+    from rich.markup import escape
+
+    from raven.providers.base import parse_llm_error
+
+    parsed = parse_llm_error(content)
+    if parsed is None:
+        return False
+    category, provider, detail = parsed
+    console.print()
+    if category == "auth":
+        where = f"{provider} 401" if provider else "401"
+        console.print(f"[red]Error: API key invalid ({escape(where)}).[/red]")
+    else:
+        console.print(f"[red]Error: LLM call failed ({escape(category)}): {escape(detail[:200])}[/red]")
+    target = provider or "<name>"
+    console.print(f"Fix: raven provider test {escape(target)}  or  raven onboard")
+    console.print()
+    _ONE_SHOT_EXIT["code"] = 1
+    return True
 
 
 # ---------------------------------------------------------------------------
@@ -344,11 +377,14 @@ def register(app: typer.Typer) -> None:
                             "memory backend stop failed; continuing shutdown",
                         )
 
+        _ONE_SHOT_EXIT["code"] = 0
         asyncio.run(run_once())
         # Native runtimes loaded by the agent loop (lancedb's Rust/tokio
         # thread, torch) segfault during interpreter finalization. The exit
         # chokepoint in raven.cli.commands.run hard-exits past finalization
         # when that hazard is live, so this path just returns normally.
+        if _ONE_SHOT_EXIT["code"]:
+            raise typer.Exit(_ONE_SHOT_EXIT["code"])
 
 
 __all__ = ["register"]
--- a/raven/providers/base.py
+++ b/raven/providers/base.py
@@ -78,18 +78,25 @@ _LLM_ERROR_CONTENT_RE = re.compile(
 
 
 def _strip_json_error_body(text: str) -> str:
-    """Replace a raw JSON error body with its human-readable message."""
+    """Replace a raw JSON error body with its human-readable message.
+
+    Rewrites only when a message is actually extracted, and only within the
+    parsed object's own boundary -- trailing text after the JSON survives, and
+    a body yielding no message leaves the text unchanged rather than truncated.
+    """
     idx = text.find("{")
     if idx == -1:
         return text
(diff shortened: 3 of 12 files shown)


raven/cli/agent_commands.py
```diff
@@ -35,16 +35,49 @@
...
+        return False
+    category, provider, detail = parsed
+    console.print()
+    if category == "auth":
+        where = f"{provider} 401" if provider else "401"
+        console.print(f"[red]Error: API key invalid ({escape(where)}).[/red]")
+    else:
+        console.print(f"[red]Error: LLM call failed ({escape(category)}): {escape(detail[:200])}[/red]")
+    target = provider or "<name>"
+    console.print(f"Fix: raven provider test {escape(target)}  or  raven onboard")
```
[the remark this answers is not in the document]
0xKT: Fixed in 80c1004. The credential fix line now renders only for the `auth` category; the other categories each get a category-apt one-line hint instead (`rate_limit` -> retry in a moment, `network` -> check connectivity, `context_overflow` -> shorten the input, `server` -> retry later or switch models, `model_unavailable` -> pick another model via `raven provider use`, `billing` -> check the provider account), and categories with no apt hint (`invalid_request`, `unknown`, ...) render the error line alone. Covered by `test_print_llm_error_non_auth_categories_get_apt_hint_not_key_guidance` (red before the change, green after).

raven/providers/base.py
```diff
@@ -569,15 +644,17 @@ async def _chat_attempt_with_retry(
...
             if response.finish_reason != "error":
                 return response
 
             # Prefer a provider-attached classification (it had the live
             # exception); else classify the exception we caught, else the string.
-            classification = response.error_classification or self.classify_error(exc, response.content)
+            classification = response.error_classification or self.classify_error(exc, response.content or None)
             response.error_classification = classification
+            if exc is not None and not response.content:
+                response.content = format_llm_error(exc, classification, provider=getattr(self, "provider_name", None))
```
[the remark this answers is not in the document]
0xKT: Fixed in eae0de2. Went with the first option: all four swallow sites now build their content with `format_llm_error` at the point where the live exception exists, matching the litellm path -- azure's non-200 branch, azure's raised-exception branch, azure's response-parse failure (previously `Error parsing Azure OpenAI response: ...` with no classification attached at all), and codex's raised-exception branch. Provider heads are `azure_openai` / `openai_codex`, so the CLI renders the same diagnosis + hint and exits 1. Classification and retry semantics are untouched; the ladder's `exc is not None` branch stays as the safety net for providers that raise. Covered red-first by three new assertions in `tests/test_azure_openai_provider.py` (non-200, raised exception, parse failure) and one in `tests/test_openai_codex_provider.py`, each asserting the content round-trips through `parse_llm_error` with the right category and provider.

raven/providers/base.py
```diff
@@ -68,6 +69,80 @@ class ErrorClassification:
...
+    candidate = text[idx:].strip()
+    if not candidate.startswith(('{"', "{'")):
+        return text
+    message = ""
+    try:
+        obj = json.loads(candidate)
+    except ValueError:
+        obj = None
+    if isinstance(obj, dict):
+        err = obj.get("error")
```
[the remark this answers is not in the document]
0xKT: Fixed in ab4131c. `_strip_json_error_body` now finds the object's own boundary with `json.JSONDecoder().raw_decode`: a rewrite preserves any trailing text after the JSON, the regex rescue only searches within the parsed boundary (whole remainder when the body is malformed, as the swallowed litellm shape is), and when no message can be extracted the text is returned unchanged instead of truncated at the first `{` -- your `'Config invalid: {"foo": "bar"} ... retry with a valid key'` example now comes back intact. Also added the direct unit coverage you suggested to `tests/test_error_classification.py`: a real exception fed through `format_llm_error` (prefix collapse + JSON body replacement + `parse_llm_error` round-trip), a parse rejection case, and the two boundary cases above.

raven/cli/agent_commands.py
```diff
@@ -35,16 +35,66 @@
...
+
+    from raven.providers.base import parse_llm_error
+
+    parsed = parse_llm_error(content)
+    if parsed is None:
+        return False
+    category, provider, detail = parsed
+    console.print()
+    if category == "auth":
+        where = f"{provider} 401" if provider else "401"
```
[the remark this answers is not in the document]
gloryfromca: Fixed in a77804c. You are right that the status was fabricated: the `auth` bucket also fires on 403, on `PermissionDeniedError`, and on the `"permission denied"` / `"unauthorized"` substrings, so nothing downstream knows the failure was a 401.

Rather than trying to recover the real code (`parse_llm_error` only carries category / provider / detail, and regex-guessing a status back out of rendered text is exactly what `classify_error`'s own comment argues against), the line now names no status at all and carries the provider's reason instead:

```
Error: provider rejected the credentials (openrouter): AuthenticationError: User not found.
Fix: raven provider test openrouter  or  raven onboard
```

That also closes a second hole in the same branch: `detail` was being dropped entirely for `auth`, so a permission failure printed no reason at all -- and the reason is precisely what separates a rejected key from a key without access to that model.

The fix hint stays as-is for both. For the 403 permission case, `raven provider test` doing a free `GET /v1/models` and passing is itself the useful answer -- it tells the user the credential is fine and the problem is the model/plan.

Coverage: `test_print_llm_error_renders_diagnosis_and_marks_exit` updated for the new line, plus a new `test_print_llm_error_auth_does_not_fabricate_a_status_code` that runs a `status_code = 403` exception through `classify_error` -> `format_llm_error` -> `_print_llm_error`, asserts the classification really is `auth`, and asserts `"401" not in out`. `uv run pytest tests/test_cli_agent_commands.py tests/test_error_classification.py -q` -> 54 passed, 0 skipped.

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. One cosmetic nit inline, not worth holding the merge for.

This revision is a clean rebase, not new work: `git range-diff` reports all six commits identical to the pre-rebase revision (`a54f0e8..a77804c` -> `266e1c3..014e2ab`, every pair `=`), and the diff against the new base is still +560/-17 across the same 12 files, so nothing leaked in or out when the base moved from `refactor/retire_repl_and_cli_channel` to `main`.

The rebase-specific risk was real and I checked it: `main` landed #329, which rewrote `raven/cli/agent_commands.py` heavily (-443 lines net on the main side of the same file this PR edits) and cut 33 lines from `raven/cli/_helpers.py`. The merge was textually clean, so the question is whether it is still semantically live. It is -- `_print_agent_response` is still the render callback handed to `build_repl` (`agent_commands.py:351`), `_print_llm_error` still gates it (`:47`), and the one-shot exit seam still resets before the run and raises `typer.Exit` after it (`:400`, `:406-407`). Both `MissingCredentialsError` raise sites in `_helpers.py` survived main's deletions intact.

Back-compat: nothing in production code matches the old `Error calling LLM: ` literal prefix. The only remaining matches are `test_provider_prompt_cache.py` and `test_sentinel_planner.py`, which feed synthetic strings straight into the *string* classifier -- that path never read the prefix, so the new `Error calling LLM (<category>@<provider>): ` shape does not break them.

Tests, run rather than claimed:
- `uv run pytest tests -q` on this head: **6012 passed, 43 skipped, 1 failed** in 149s.
- The failure is `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare` (rich emits 256-color `38;5;221` instead of truecolor `38;2;251;226;63`). I checked it out on `main` in a separate worktree and ran the full suite there: **5998 passed, 43 skipped, 1 failed -- the same test**. Pre-existing, untouched by this PR, and it passes in isolation on both sides.
- Skip count is identical on both sides (43, all optional-import or not-applicable guards), so this PR adds no skips, and the +14 tests are real assertions: the new 403 test drives a live `status_code = 403` exception through `classify_error` -> `format_llm_error` -> `_print_llm_error` and asserts the classification really is `auth` before asserting `"401" not in out`.

Conventions checked: all six commit subjects are Conventional Commits with valid scopes and no non-ASCII anywhere in the messages; test changes go into the existing `test_cli_agent_commands.py` / `test_cli_helpers.py` / `test_error_classification.py` / provider-specific files rather than new phase-suffixed ones (AGENTS.md 5.1/5.4); no assets or large files.

raven/cli/_helpers.py
```diff
@@ -59,12 +59,19 @@ def check_provider_credentials(config: Config) -> None:
...
+            ),
         )
 
     status = credential_status(provider_name, config.providers.get(provider_name), include_external=True)
     if not status.ok:
-        raise MissingCredentialsError(status.summary, provider=provider_name)
+        raise MissingCredentialsError(
+            status.summary,
+            provider=provider_name,
+            remedy="Or run `raven onboard` for guided setup.",
```
gloryfromca: [nit] This remedy starts with `Or`, but nothing precedes it here. The sole outlet prints summary then remedy verbatim (`raven/cli/commands.py:164-165`), and `CredentialStatus.summary` is just `"<provider> needs <field>"`, so a configured provider missing its key renders:

```
Error: openrouter needs OPENROUTER_API_KEY.
Or run `raven onboard` for guided setup.
```

A dangling conjunction with no alternative before it. The zero-provider site above reads fine because its remedy names the `raven provider set` command first, and this one inherited the second half of that pair without the first. The TUI path has the same shape -- it ships the string as a structured `remedy` field (`tui_commands.py:493`).

Dropping the `Or ` is enough, or pair it the way the other site does: `Run: raven provider set <name> --api-key <key>\nOr run \`raven onboard\` for guided setup.` The new `test_no_api_key_error_mentions_onboard` / `test_azure_missing_key_same_onboard_guidance` only substring-match `raven onboard`, so either wording keeps them green.

0xKT: Fixed in a5c3dcf: the credential-status site now pairs the remedy the same way the zero-provider site does (provider set line first, onboard as the alternative), so the conjunction no longer dangles.

gloryfromca (dismissed):

gloryfromca (commented):
No blockers; this can merge. But the one-line change in `a5c3dcf` should be reverted first, and the fault is mine -- my nit was wrong on its facts.

**My earlier nit does not hold.** I claimed the `Or run ...` remedy dangled because `CredentialStatus.summary` is just `"<provider> needs <field>"`. It is not. `summary` uses `m.hint or m.label`, and every `Requirement` in `raven/providers/auth.py` (lines 215, 218, 219, 224, 234) ships a hint with `{public}` substituted to the real provider name (line 398), so the label-only branch I quoted is unreachable. The pre-`a5c3dcf` output actually read:

```
Error: openrouter needs an API key -- run `raven provider set openrouter --api-key <key>`.
Or run `raven onboard` for guided setup.
```

The summary already ends in a concrete command, so the `Or` had its antecedent and the message was fine as it stood. Sorry for the churn.

**The fix made it worse**, which is the inline note below: the new `Run:` line is unconditional, while what is missing varies. Rendered through `check_provider_credentials` on this head:

```
Error: github_copilot needs a sign-in -- run `raven provider login github-copilot`.
Run: raven provider set <name> --api-key <key>
Or run `raven onboard` for guided setup.
```

An OAuth provider told to set an API key -- the same wrong-remedy shape as the very first review thread on this PR, and in the ordinary key case it is a `<name>`-placeholder duplicate of the specialized command the line above already gave. Reverting `a5c3dcf` (restoring `remedy="Or run \`raven onboard\` for guided setup."`) is the whole fix; `868ce42` is formatting-only on top of it and goes with it.

**Rest of the revision.** The first six commits are the same SHAs as the head I reviewed at `014e2ab` (`range-diff` shows six `=` pairs, no rebase), so the delta is exactly this one string plus its reformat, and nothing else needs re-reviewing. `868ce42` is value-preserving -- implicit concatenation of two literals collapsed into one, same `\n`.

**Verified, not claimed:** `uv run ruff format --check` and `ruff check` on `raven/cli/_helpers.py` both clean; `uv run pytest tests/test_cli_helpers.py tests/test_cli_agent_commands.py tests/test_error_classification.py tests/test_azure_openai_provider.py tests/test_openai_codex_provider.py -q` -> 108 passed, 0 skipped. The two helper tests stay green either way because they substring-match `raven onboard`, which is why neither the fix nor the revert is caught by them -- worth knowing, not worth adding a test for a line that should come out.

raven/cli/_helpers.py
```diff
@@ -59,12 +59,19 @@ def check_provider_credentials(config: Config) -> None:
...
+            ),
         )
 
     status = credential_status(provider_name, config.providers.get(provider_name), include_external=True)
     if not status.ok:
-        raise MissingCredentialsError(status.summary, provider=provider_name)
+        raise MissingCredentialsError(
+            status.summary,
+            provider=provider_name,
+            remedy=("Run: raven provider set <name> --api-key <key>\nOr run `raven onboard` for guided setup."),
```
gloryfromca: [nit] This line is unconditional, but what the provider is missing is not. `check_provider_credentials` reaches here for any unmet requirement, and only one of them is an API key. Rendered through the real gate on this head:

```
Error: github_copilot needs a sign-in -- run `raven provider login github-copilot`.
Run: raven provider set <name> --api-key <key>
Or run `raven onboard` for guided setup.
```

```
Error: ollama_chat needs an address -- run `raven provider set ollama-chat --api-base <url>`.
Run: raven provider set <name> --api-key <key>
Or run `raven onboard` for guided setup.
```

An OAuth provider is told to set an API key; an address-missing provider is pointed at the wrong flag. Even in the ordinary key case the line is a `<name>`-placeholder copy of the command the summary already gave with the provider filled in.

This is my fault -- the nit that prompted it was wrong (see the review summary: `summary` always carries a substituted hint, so the `Or` never dangled). Suggest reverting to `remedy="Or run \`raven onboard\` for guided setup."` and letting `summary` keep naming the requirement-specific command.

0xKT: Fixed in 868ce42: the status-site remedy is now requirement-neutral (just the guided-setup line, no conjunction) since the summary already names the exact unmet requirement and its command. The api-key pairing stays only at the zero-provider site where it is always accurate.

0xKT: Correction: the previous reply cited the wrong sha (a tooling slip on my side - the edit had not actually landed). The real fix is 154825a, now pushed: the status-site remedy is requirement-neutral (guided-setup line only), and the api-key pairing remains only at the zero-provider site.

gloryfromca (commented):

`154825a` resolves the finding I raised on `868ce42`, and it went one better than the revert I suggested: dropping the `Or ` as well makes the line stand on its own instead of leaning on an antecedent that varies. Verified by driving `check_provider_credentials` for each requirement shape on this head:

```
Error: github_copilot needs a sign-in -- run `raven provider login github-copilot`.
Run `raven onboard` for guided setup.

Error: ollama_chat needs an address -- run `raven provider set ollama-chat --api-base <url>`.
Run `raven onboard` for guided setup.

Error: openrouter needs an API key -- run `raven provider set openrouter --api-key <key>`.
Run `raven onboard` for guided setup.
```

The specific command comes from `summary` (correct per requirement), the remedy adds the guided path (correct for all of them). No duplication, no wrong flag, no dangling conjunction.

I also checked the zero-provider site is not dead code, since its paired `provider set` + `Or run onboard` remedy is only right if nothing is configured -- it is reachable with an empty `agents.defaults.model`, and there the pairing reads correctly:

```
Error: no provider configured.
Run: raven provider set <name> --api-key <key>, then raven provider use <name>/<model>
Or run `raven onboard` for guided setup.
```

Delta scope: the eight commits below `154825a` are the same SHAs I reviewed at `868ce42`, so this one string is the entire change.

Verified, not claimed: `ruff format --check` and `ruff check` on `raven/cli/_helpers.py` clean; `uv run pytest tests/test_cli_helpers.py tests/test_cli_agent_commands.py tests/test_error_classification.py tests/test_azure_openai_provider.py tests/test_openai_codex_provider.py -q` -> 108 passed, 0 skipped. (The full-suite result and the one pre-existing `test_cli_theme.py` failure I verified against `main` two revisions ago still stand; nothing since then touches those paths.)

One optional follow-up, explicitly not blocking and not worth another round on its own: `test_no_api_key_error_mentions_onboard` writes `{}` and its docstring calls that a zero-provider config, but the loader defaults the model to an anthropic id, so it actually exercises the status site -- the same one the other test covers. The zero-provider branch (empty model) has no test, and its remedy string was edited by this PR. Both tests substring-match `raven onboard`, so they pass either way. Worth a line in some later cleanup, not here.

gloryfromca (approved):

PR #332, opened 2026-08-17
feat(tui): surface reminders missed while the tui was closed

0xKT:
## Summary

Complete the missed-reminder story for tui-bound one-shot reminders.
Under fire-at-origin ownership, a one-shot reminder bound to the tui
lapses at the next tui startup when its session was closed at fire
time - previously with only a warning log the user never sees. Now:

- The startup cleanup records each dropped reminder as structured data
  (name, message, originally scheduled time) instead of only logging.
- The tui backend fans them out as a single `cron.missed` event right
  after its cron service starts.
- The frontend renders one compact summary block: "missed N reminders",
  one line per item with the originally scheduled local time.

Startup events race the client's subscription (the event fired before
`turn.subscribe` was previously lost), so `SubscriptionEmitter` gains a
one-shot startup buffer flushed to the first subscriber. The new event
is registered in the RPC schema and `generated.ts` is regenerated.

This matches how established reminder systems treat a closed surface:
the notification ping is best-effort, and the store's overdue state is
shown at the next open.

## Type

- [x] Feature

## Verification

  6000 passed / 2 failed, both failures pre-existing on main run the
  same way (test_read_file_image, plus the order-dependent
  test_cli_theme flake that fails whenever its file runs as a whole) -
  zero failures introduced.
- Affected-area suites (tui events, subscription emitter, cron startup,
  tui bootstrap/commands, rpc models): 178 passed.
- `make lint-tui`: eslint zero errors, RPC drift check reports
  generated.ts in sync; `ui-tui` vitest: 984 passed, covering the
  summary rendering across its review rounds (English chrome, dated
  form for non-today timestamps vs bare HH:MM today, and the reminder
  message text).
- New tests: startup-drop collection, cron.missed model and fanout,
  emitter startup buffer, frontend rendering.

## Risk

Additive feature: no schema or storage format changes beyond a new
event type; clients that never subscribe simply never receive it.
Rollback = revert the squash commit.

## Related Issues

N/A

### DIFF
--- a/ui-tui/src/__tests__/chatStream.test.ts
+++ b/ui-tui/src/__tests__/chatStream.test.ts
@@ -410,3 +410,43 @@ describe('createChatStream — cancel preserves streamed content', () => {
     expect(sysCalls).toContain('interrupted')
   })
 })
+
+describe('createChatStream — cron.missed startup notice', () => {
+  beforeEach(() => {
+    resetTurnState()
+    resetUiState()
+    turnController.fullReset()
+  })
+
+  it('renders one compact summary block via sys with per-item name and scheduled HH:MM', async () => {
+    const fake = makeFakeRpc()
+    const sysCalls: string[] = []
+    const stream = createChatStream({
+      rpcClient: fake,
+      sessionKey: 'tui:default',
+      sys: m => sysCalls.push(m)
+    })
+    await stream.attach()
+
+    fake.__pushEvent({
+      type: 'cron.missed',
+      payload: {
+        count: 2,
+        items: [
+          { message: '记得喝水', name: 'hydrate', scheduled_at: '2025-06-04T10:03:00+00:00' },
+          { message: '起来活动一下', name: 'stretch', scheduled_at: '2025-06-04T10:40:00+00:00' }
+        ]
+      }
+    })
+
+    expect(sysCalls).toHaveLength(1)
+    const block = sysCalls[0]
+    expect(block).toContain('错过 2 条提醒')
+    expect(block).toContain('hydrate — 原定')
+    expect(block).toContain('stretch — 原定')
+    // scheduled_at renders as local HH:MM, so assert the shape, not the value.
+    expect(block).toMatch(/原定 \d{2}:\d{2}/)
+    // A missed notice is a transcript block, not a turn — the UI must stay idle.
+    expect(stream.isTurnActive()).toBe(false)
+  })
+})
--- a/ui-tui/src/__tests__/chatStream.test.ts
+++ b/ui-tui/src/__tests__/chatStream.test.ts
@@ -441,11 +441,11 @@ describe('createChatStream — cron.missed startup notice', () => {
 
     expect(sysCalls).toHaveLength(1)
     const block = sysCalls[0]
-    expect(block).toContain('错过 2 条提醒')
-    expect(block).toContain('hydrate — 原定')
-    expect(block).toContain('stretch — 原定')
+    expect(block).toContain('missed 2 reminders')
+    expect(block).toContain('hydrate — scheduled')
+    expect(block).toContain('stretch — scheduled')
     // scheduled_at renders as local HH:MM, so assert the shape, not the value.
-    expect(block).toMatch(/原定 \d{2}:\d{2}/)
+    expect(block).toMatch(/scheduled \d{2}:\d{2}/)
     // A missed notice is a transcript block, not a turn — the UI must stay idle.
     expect(stream.isTurnActive()).toBe(false)
   })
--- a/ui-tui/src/__tests__/chatStream.test.ts
+++ b/ui-tui/src/__tests__/chatStream.test.ts
@@ -442,11 +442,14 @@ describe('createChatStream — cron.missed startup notice', () => {
     expect(sysCalls).toHaveLength(1)
     const block = sysCalls[0]
     expect(block).toContain('missed 2 reminders')
-    // Fixed 2025 timestamps are never "today": the dated form must render.
-    expect(block).toMatch(/hydrate — scheduled 06-04 \d{2}:\d{2}: 记得喝水/)
-    expect(block).toMatch(/stretch — scheduled 06-04 \d{2}:\d{2}: 起来活动一下/)
-    // Non-today timestamps render the dated local form: MM-DD HH:MM.
-    expect(block).toMatch(/scheduled 06-04 \d{2}:\d{2}/)
+    // Fixed 2025 timestamps are never "today" in any timezone: the dated
+    // form must render — but the calendar day is LOCAL while the fixture is
+    // UTC, so assert the shape, never a literal date.
+    expect(block).toMatch(/hydrate — scheduled \d{2}-\d{2} \d{2}:\d{2}: 记得喝水/)
+    expect(block).toMatch(/stretch — scheduled \d{2}-\d{2} \d{2}:\d{2}: 起来活动一下/)
+    // Non-today timestamps render the dated local form (shape only: the
+    // local calendar day shifts with the machine's timezone).
+    expect(block).toMatch(/scheduled \d{2}-\d{2} \d{2}:\d{2}/)
     // A missed notice is a transcript block, not a turn — the UI must stay idle.
     expect(stream.isTurnActive()).toBe(false)
   })
--- a/ui-tui/src/__tests__/chatStream.test.ts
+++ b/ui-tui/src/__tests__/chatStream.test.ts
@@ -442,11 +442,37 @@ describe('createChatStream — cron.missed startup notice', () => {
     expect(sysCalls).toHaveLength(1)
     const block = sysCalls[0]
     expect(block).toContain('missed 2 reminders')
-    expect(block).toContain('hydrate — scheduled')
-    expect(block).toContain('stretch — scheduled')
-    // scheduled_at renders as local HH:MM, so assert the shape, not the value.
-    expect(block).toMatch(/scheduled \d{2}:\d{2}/)
+    // Fixed 2025 timestamps are never "today": the dated form must render.
+    expect(block).toMatch(/hydrate — scheduled 06-04 \d{2}:\d{2}: 记得喝水/)
+    expect(block).toMatch(/stretch — scheduled 06-04 \d{2}:\d{2}: 起来活动一下/)
+    // Non-today timestamps render the dated local form: MM-DD HH:MM.
+    expect(block).toMatch(/scheduled 06-04 \d{2}:\d{2}/)
     // A missed notice is a transcript block, not a turn — the UI must stay idle.
     expect(stream.isTurnActive()).toBe(false)
+  })
+
+  it('renders a bare HH:MM when the missed reminder was scheduled today', async () => {
+    const fake = makeFakeRpc()
+    const sysCalls: string[] = []
+    const stream = createChatStream({
+      rpcClient: fake,
+      sessionKey: 'tui:default',
+      sys: m => sysCalls.push(m)
+    })
+    await stream.attach()
+
+    const today = new Date()
+    today.setHours(9, 30, 0, 0)
+    fake.__pushEvent({
+      type: 'cron.missed',
+      payload: {
+        count: 1,
+        items: [{ message: 'drink water', name: 'hydrate', scheduled_at: today.toISOString() }]
+      }
+    })
+
+    expect(sysCalls).toHaveLength(1)
+    expect(sysCalls[0]).toContain('hydrate — scheduled 09:30: drink water')
+    expect(sysCalls[0]).not.toMatch(/scheduled \d{2}-\d{2} /)
   })
 })
--- a/ui-tui/src/app/chatStream.ts
+++ b/ui-tui/src/app/chatStream.ts
@@ -159,8 +159,9 @@ const dispatch = (
     case 'cron.missed': {
       if (sys) {
         const { count, items } = event.payload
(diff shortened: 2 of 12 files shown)


0xKT:
Fixed in 859184b: the cron.missed block now renders in English ("missed N reminders" / "scheduled HH:MM"), matching cron.delivered and the rest of the sys chrome; the vitest rendering case is updated with it. eslint and the full ui-tui suite (983) pass.

ui-tui/src/app/chatStream.ts
```diff
@@ -146,6 +156,15 @@ const dispatch = (
       }
       return
     }
+    case 'cron.missed': {
+      if (sys) {
+        const { count, items } = event.payload
+        const lines = items.map(item => `${item.name} — scheduled ${formatScheduledAt(item.scheduled_at)}`)
```
gloryfromca: Non-blocking, but worth a thought before merge: `formatScheduledAt` renders only `HH:MM`, and this is the one event type where the day is the whole point.

Concrete case: TUI closed Friday evening, reopened Monday morning. A one-shot reminder that was due Saturday 09:00 renders as `weekend-review — scheduled 09:00`, which reads as *this morning*. The user has no way to tell a three-day-old miss from a two-hour-old one, and the gap is exactly what a "missed while the TUI was closed" notice exists to communicate. `cron.delivered` gets away with a bare timestamp because it always fires in the here-and-now; this one does not.

A date-when-not-today rule keeps the block compact in the common case and honest in the uncommon one — e.g. `09:00` for today, `Sat 09:00` (or `06-14 09:00`) otherwise.

Separately, `item.message` is carried all the way through the payload, the Pydantic model and the OpenRPC schema, then never rendered. If the compact block is deliberate that is fine, but a job `name` can be terse or autogenerated, and the message text is what makes the notice actionable — either show it or consider dropping the field.

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. One non-blocking display note inline.

**What I checked**

- The prior Chinese-literal blocker: fixed. The `cron.missed` block now uses the same chrome shape as `cron.delivered` right above it (`─── ⏰ … ───` + a 40-char rule), all English.
- The startup-buffer design. This is the part that could plausibly break, and it holds up: `emitter.register()` enqueues the buffered event before returning the `sub_id`, so the `turn.subscribe` response frame is written first and the coalesce loop only ships the event ≥16 ms later. The client registers its handler in `RpcClient.subscribe` right after that response resolves, so the event cannot land on an unregistered `subscription_id`. The buffer is one-shot and only ever holds this one event.
- Ownership. `_build_gateway_channels` deliberately excludes `tui`, so the gateway never drops a past-due TUI reminder out from under the TUI, and `_recompute_next_runs` removes the dropped jobs and persists, so the notice cannot repeat on the next launch. `raven cron *` builds its service with `allowed_channels=None` but never calls `start()`, so it does not steal the drops either.
- Back-compat. The TUI ships with the runtime, so there is no old-client/new-server pairing; even so an unknown variant falls through `dispatch`'s `default` harmlessly.
- Conventions. Comments and commits are English; `CONTEXT.md` already covers the drop-at-next-startup rule under Fire-at-Origin, so no new domain term is being coined; no new test files (both Python additions extend existing suites); no assets.

**Tests — actually run, not claimed**

- `uv run pytest tests/test_rpc_schema_match.py tests/test_subscription_emitter.py tests/test_cron_service_startup.py tests/test_tui_cron_delivered_event.py tests/test_tui_cron_tool_wired.py -q` → 134 passed, 0 skipped.
- `npm run lint:rpc` → `generated.ts` in sync with `openrpc.json` (so the hand-written generated file is not drifting).
- `npm run type-check` → clean; `npm test` → 86 files, 983 passed.
- `uv run ruff check raven/ tests/` → clean.

Nothing is skipped, xfail'd, or weakened. The new Python tests assert real structure (`CronStartupDrop` equality, the foreign-channel negative case, the one-shot buffer not re-delivering on a second `register`), and the vitest case asserts the turn stays idle so a notice cannot be mistaken for a turn.

gloryfromca (approved):

0xKT:
Both addressed in 6f85afa: (1) non-today timestamps now render the dated local form - `scheduled 06-14 09:00` - while today keeps bare HH:MM, so a reminder missed over a weekend no longer reads like this morning; (2) the reminder message now renders after the schedule ("name - scheduled when: message") rather than riding the whole pipeline unused - kept the field, rendering is what makes the notice actionable. vitest covers both branches (dated + today) and the message rendering; 984 pass.

ui-tui/src/__tests__/chatStream.test.ts
```diff
@@ -410,3 +410,69 @@ describe('createChatStream — cancel preserves streamed content', () => {
...
+          { message: '起来活动一下', name: 'stretch', scheduled_at: '2025-06-04T10:40:00+00:00' }
+        ]
+      }
+    })
+
+    expect(sysCalls).toHaveLength(1)
+    const block = sysCalls[0]
+    expect(block).toContain('missed 2 reminders')
+    // Fixed 2025 timestamps are never "today": the dated form must render.
+    expect(block).toMatch(/hydrate — scheduled 06-04 \d{2}:\d{2}: 记得喝水/)
```
gloryfromca: Non-blocking, but this tightening made the assertion host-dependent. The fixture timestamps are fixed UTC (`2025-06-04T10:03:00+00:00`), while `formatScheduledAt` renders the **local** calendar date - so `06-04` is only the right answer for hosts within roughly UTC-10..UTC+13.

Demonstrated, not hypothetical:

```
$ TZ=Pacific/Kiritimati npx vitest run src/__tests__/chatStream.test.ts
 Tests  1 failed | 14 passed (15)
```

(UTC+14 pushes 10:03Z to `06-05`; UTC-11, e.g. `Pacific/Pago_Pago`, pulls it back to `06-03`.) CI runs UTC so it stays green, which is exactly what makes it the annoying kind of failure - it only ever bites one contributor on one machine.

The behaviour under test is "not today renders the dated form, today renders the bare form", and the day-of-month value carries none of that. `\d{2}-\d{2}` on lines 446/447/449 asserts the same thing and cannot drift:

```js
expect(block).toMatch(/hydrate — scheduled \d{2}-\d{2} \d{2}:\d{2}: 记得喝水/)
```

Worth noting the pre-fix version of this assertion was timezone-robust; the new today-case test right below is too. This is the only one that is not.

gloryfromca (commented):
Still no blockers - the dating fix is right and my earlier note is fully addressed; one non-blocking test-determinism nit inline.

I only re-reviewed `859184b..6f85afa6` (the rest of the PR I already covered and it has not moved).

**The fix itself is correct.** `formatScheduledAt` compares `getFullYear`/`getMonth`/`getDate` against `new Date()`, i.e. a local-calendar same-day test, so there is no UTC-vs-local off-by-one in the today/not-today decision itself, and the weekend case I raised now renders `06-14 09:00` instead of a bare `09:00`. Dropping the year means a miss older than twelve months would read as this year, but that needs a one-shot `at` job to sit pending across a year of TUI launches - not worth the extra width. `item.message` is now rendered, and `--message` is a required option on `cron add`, so the `name - scheduled ...: <empty>` shape is not reachable outside a hand-corrupted store.

**Tests - run, not claimed**

- `npm test` -> 86 files, 984 passed (was 983; the new today-case test is the +1). No skips.
- `npm run type-check` -> clean. `npm run lint` -> 0 errors (22 pre-existing warnings, none in the touched files). `npm run lint:rpc` was unaffected - this commit does not touch the schema.
- The new today-case test is timezone-safe by construction: it builds `new Date()` + `setHours(9,30)` locally and round-trips through `toISOString()`, so the local wall-clock value it asserts is the one it created.

The inline note is about the *other* test in the same block, which went the opposite way.

0xKT:
Fixed in aa827bd: the three assertions now match the dated form's shape (`\d{2}-\d{2} \d{2}:\d{2}`) instead of a literal local date, with a comment explaining the local-day-vs-UTC-fixture trap. The Kiritimati repro was a nice touch - thanks. vitest: 984 pass.

ui-tui/src/__tests__/chatStream.test.ts
gloryfromca: Confirmed fixed in `aa827bd`, and the loosened regex did not cost the test anything.

Re-ran the original repro plus the opposite extreme:

```
Pacific/Kiritimati (UTC+14)   Tests  15 passed (15)
Pacific/Pago_Pago  (UTC-11)   Tests  15 passed (15)
UTC                           Tests  15 passed (15)
Asia/Shanghai      (UTC+8)    Tests  15 passed (15)
```

The obvious worry with swapping a literal for `\d{2}-\d{2}` is that the assertion goes slack, so I checked rather than assumed: forcing `formatScheduledAt` to always return the bare `hhmm` (i.e. regressing the exact behaviour this test guards) still fails it — `Tests 1 failed | 14 passed`. The bare form has no dash-separated pair, so the dated-vs-bare distinction is still genuinely asserted, and the today-case test below keeps the other direction pinned with its `not.toMatch`. Reverted my local mutation; worktree clean.

Nothing outstanding from my side.

gloryfromca (commented):
No blockers and nothing left from me; this can merge as far as I am concerned.

`6f85afa6..aa827bde` is a six-line test-assertion change and it is exactly the fix. I re-reviewed only that range - everything below it I covered on the two prior revisions and it has not moved.

The one thing worth verifying rather than assuming: replacing a literal date with `\d{2}-\d{2}` is the kind of edit that quietly turns a real assertion into a tautology. It did not. Details and the mutation check are on the resolved thread; short version is that the test still fails when the dated form is regressed away, both branches of the today/not-today split stay pinned, and the suite passes at UTC+14, UTC-11, UTC and UTC+8 rather than only in CI's timezone.

`npm test` -> 86 files, 984 passed, no skips.

All three review threads on this PR (the Chinese-literal blocker, the missing date, the host-dependent assertion) are now closed with code changes rather than argument, and each fix came with test coverage rather than just a green run.

gloryfromca (dismissed):

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. Re-confirming on the rebased head - the sign-off from `aa827bd` does not carry across a branch move, so here it is against the code that would actually land.

This revision is a rebase, not new work: the target moved from `refactor/retire_repl_and_cli_channel` to `main` now that #329 has landed as `e21f855`, and every sha shifted. A clean textual rebase can still break things semantically, so I checked rather than assumed.

- **Content is unchanged.** Diffing this head against the previously-reviewed `aa827bd` over `raven/`, `tests/` and `ui-tui/` shows drift only in `config/raven.py`, `context_engine/factory.py` and `memory_engine/skill_forge/*` - that is #290 arriving from main. Not one of the PR's own twelve files moved.
- **No semantic overlap.** Nothing main gained between the old base and now touches `tui_rpc`, `cron`, `tui_commands` or `ui-tui`, so there is no silent interaction for the rebase to have introduced.
- **Four commits, linear, no merge commits**, sitting directly on main's tip.

**Gates re-run on the rebased tree, not carried over:**

- `uv run pytest` over the cron/tui_rpc/schema suites -> 159 passed, 0 skipped (more than last round because `test_tui_rpc_spine.py` is in scope on this base).
- `npm test` -> 86 files, 984 passed. `npm run type-check` clean. `npm run lint:rpc` -> `generated.ts` still in sync with `openrpc.json`.
- `uv run ruff check raven/ tests/` clean.
- `npx commitlint --from e21f8550 --to HEAD` passes against the new base, and all four commit messages are pure ASCII per the repo's all-English rule.

Everything I raised across the three earlier rounds is closed in code with tests. Nothing new from me.

gloryfromca (approved):

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. Second rebase, so re-confirming against the code that would actually land - a branch move drops the previous sign-off.

Another rebase-only revision: main picked up #330 and #331, the branch was replayed onto `5e5f546`, and the PR's own eleven files are byte-identical to the `fba4673` head I reviewed last round. Four commits, linear, sitting on main's tip.

The thing actually worth checking here was #331 - "fold status output and add per-turn usage summary" sounds adjacent to the turn/event surface this PR extends, and a schema or `TurnEvent`-union collision is exactly the kind of thing a clean textual rebase hides. It is not adjacent: #330/#331 touch `cli/_helpers.py`, `_repl_spine.py`, `agent_commands.py`, `status_commands.py`, `config/schema.py`, the providers and `utils/helpers.py`. Zero file intersection with this PR, and nothing in `tui_rpc`, `openrpc.json`, `ui-tui` or the cron scheduler.

**Gates re-run on the rebased tree:**

- Full `uv run pytest tests/` this time rather than a subset: **6031 passed, 43 skipped**.
- `npm test` -> 86 files, 984 passed. `type-check` clean, `lint:rpc` still in sync, `ruff` clean.
- `commitlint --from 5e5f5466 --to HEAD` passes against the new base; all four messages pure ASCII.

**One thing to know before someone reads a red run and blames this PR:** `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare` fails in a full-suite run. It is not this branch. I ran the same full suite in a clean worktree at pristine `5e5f546` and it fails there identically - a pre-existing test-isolation bug on main (it passes in isolation, so something earlier in the run leaves theme state behind). Out of scope for this PR, but worth its own issue. The count delta between the two runs is exactly +9, which is precisely this PR's nine new Python tests, all passing.

Everything I raised across the earlier rounds is closed in code with tests. Nothing new from me.

PR #333, opened 2026-08-17
feat(proactive_engine): add cron anti-runaway protection and missed-reminder notices

0xKT:
## Summary

Two follow-ups that complete the fire-at-origin cron story, both built
end-to-end.

Anti-runaway protection: a recurring reminder the user never engages
with now disables itself. Fires are counted at the delivery point
(after a successful turn, recurring schedules only), the counter resets
whenever a genuine user message arrives on the job's channel/chat
(chained after the existing user-inbound hook and guarded to USER
origin, so a cron fire can never reset its own counter), and crossing
the per-job limit (default 12) disables the job with a warning log.
The store-side disable is sticky against the post-run writeback, and
cron list/get surface the live counter.

Missed-reminder notices: the gateway observes - read-only, never
mutating jobs it does not own - one-shot reminders bound to interactive
surfaces that are more than a grace window past due (their session was
closed at fire time), and enqueues one deduplicated system event per
job; the heartbeat surfaces it to the user. Gated by
cron.notify_missed (default on).

## Type

- [x] Feature

## Verification

  6053 passed / 2 failed, both failures pre-existing on main run the
  same way (test_read_file_image, plus the order-dependent
  test_cli_theme flake that fails whenever its file runs as a whole) -
  zero failures introduced.
- Affected-area suites (silent-fire, missed notices, cron handler/CLI,
  claim/wake, event wake, gateway/tui wiring): 209 passed.
- Review rounds added: user-activity reset for legacy jobs without
  channel attribution, a counter reset on re-enable (auto-disable is no
  longer a one-way door), a user-visible auto-disable notice riding the
  heartbeat path plus a default-view list hint - each with tests.
- Real-machine harness (real CLI subprocesses, real CronService driver
  processes, a real `raven gateway`, isolated HOME): full 11 scenarios /
  46 checks green, including the anti-runaway scenario (recurring job
  auto-disables at the limit and never overruns) and the
  missed-notification scenario (a real gateway enqueues the notice
  exactly once across wake passes, and the config toggle silences it).

## Risk

Old jobs.json files load unchanged (the counter fields reuse the JSON
names an earlier build wrote, and readers tolerate their absence).
Both features are conservative by default: the runaway limit only
disables after 12 unengaged fires and logs loudly; the missed-notice
observer is read-only over foreign jobs and can be turned off with
cron.notify_missed=false. Rollback = revert the squash commit.

## Related Issues

N/A

### DIFF
--- a/raven/cli/cron_commands.py
+++ b/raven/cli/cron_commands.py
@@ -152,6 +152,15 @@ def _format_next_run(j: CronJob) -> str:
     return ts.strftime("%Y-%m-%d %H:%M")
 
 
+def _format_silent(j: CronJob) -> str:
+    n = j.state.silent_fire_count
+    if n == 0:
+        return "0"
+    if n >= 5:
+        return f"[yellow]{n} ⚠[/yellow]"
+    return str(n)
+
+
 def _resolve_id(service: CronService, prefix: str, *, include_disabled: bool = True) -> CronJob:
     """Find a job by full id or unique prefix. Exits with friendly
     error on no-match or ambiguous-prefix."""
@@ -238,6 +247,7 @@ def cron_list(
 
     Helps answer:
     - "What reminders does the user currently have?"
+    - "Which jobs are firing too often (silent-fire ≥ 5)?"
     - "When's the next wake-up?"
 
     Default hides disabled jobs; pass ``--all`` to include them.
@@ -281,6 +291,7 @@ def cron_list(
     table.add_column("Schedule")
     table.add_column("Next Run")
     table.add_column("Last", style="dim")
+    table.add_column("Silent", justify="right")
     table.add_column("Channel")
     table.add_column("Name")
     for j in jobs:
@@ -298,6 +309,7 @@ def cron_list(
             _format_schedule(j.schedule),
             _format_next_run(j),
             last_styled,
+            _format_silent(j),
             j.payload.channel or "-",
             (j.name or "-")[:36],
         )
@@ -312,7 +324,7 @@ def cron_get(
     id_prefix: str = typer.Argument(..., metavar="ID", help="Job id or unique prefix"),
 ):
     """Show full detail of one cron job (schedule, payload, state,
-    claim status).
+    silent-fire counter, claim status).
 
     Helps answer:
     - "Why is this job firing weird?"
@@ -331,6 +343,7 @@ def cron_get(
     table.add_row("Enabled", "[green]True[/green]" if job.enabled else "[red]False[/red]")
     table.add_row("Schedule", _format_schedule(job.schedule))
     table.add_row("delete_after_run", str(job.delete_after_run))
+    table.add_row("silent_fire_limit", str(job.silent_fire_limit) if job.silent_fire_limit else "-")
 
     # Payload
     table.add_row("─ Payload ─", "")
@@ -354,6 +367,7 @@ def cron_get(
     table.add_row("last_status", job.state.last_status or "-")
     if job.state.last_error:
         table.add_row("last_error", f"[red]{job.state.last_error[:200]}[/red]")
+    table.add_row("silent_fire_count", _format_silent(job))
     if job.state.claimed_by_pid:
         table.add_row("claimed_by_pid", str(job.state.claimed_by_pid))
         if job.state.claimed_at_ms:
--- a/raven/cli/cron_commands.py
+++ b/raven/cli/cron_commands.py
@@ -282,6 +282,22 @@ def cron_list(
     )
     _print_runner_status(service)
 
+    if not all_:
+        # An auto-disabled reminder (counter at its limit) would otherwise
+        # vanish from the default view without a trace.
+        auto_disabled = [
+            j
+            for j in service.list_jobs(include_disabled=True)
+            if not j.enabled and j.silent_fire_limit and j.state.silent_fire_count >= j.silent_fire_limit
+        ]
+        if auto_disabled:
+            names = ", ".join(f"'{j.name}' ({j.id})" for j in auto_disabled[:3])
+            more = "..." if len(auto_disabled) > 3 else ""
+            console.print(
+                f"[yellow]⚠[/yellow] {len(auto_disabled)} reminder(s) auto-disabled after repeated "
+                f"fires with no reply: {names}{more} — re-enable with `raven cron enable <id>`"
+            )
+
     if not jobs:
         console.print("[dim]  (no jobs to list — pass --all to include disabled)[/dim]")
         return
--- a/raven/proactive_engine/schedulers/cron/service.py
+++ b/raven/proactive_engine/schedulers/cron/service.py
@@ -918,6 +918,10 @@ class CronService:
                     job.updated_at_ms = self._now_ms()
                     if enabled:
                         job.state.next_run_at_ms = _compute_next_run(job.schedule, self._now_ms())
+                        # Re-enabling is deliberate user engagement with this
+                        # job: without a counter reset, one auto-disabled at
+                        # the limit would re-disable on its very next fire.
+                        job.state.silent_fire_count = 0
                     else:
                         job.state.next_run_at_ms = None
                     self._save_store()
--- a/raven/proactive_engine/schedulers/cron/service.py
+++ b/raven/proactive_engine/schedulers/cron/service.py
@@ -40,6 +40,10 @@ _MAX_WAKE_INTERVAL_S = 30.0
 # Backoff after a failed tick so a persistent error cannot spin the loop.
 _ERROR_BACKOFF_S = 5.0
 
+# Grace before a past-due foreign one-shot counts as "missed" — its owning
+# session may just be slow (a long agent turn) rather than closed.
+_MISSED_GRACE_MS = 5 * 60 * 1000
+
 
 def _now_ms() -> int:
     return int(time.time() * 1000)
@@ -127,6 +131,10 @@ class CronService:
         self.lock_path = store_path.with_suffix(store_path.suffix + ".lock")
         self.on_job = on_job
         self.allowed_channels = allowed_channels
+        # Missed-reminder observer (gateway wires this): called with the
+        # past-due foreign one-shots on start and once per wake-loop pass.
+        # Read-only towards the store — the callback must not mutate jobs.
+        self.on_missed_foreign: Callable[[list[CronJob]], None] | None = None
         self._store: CronStore | None = None
         # Nanosecond precision: float st_mtime collapses writes ~238ns apart
         # into one value, serving a stale cache after an external rewrite.
@@ -216,10 +224,12 @@ class CronService:
                                 last_error=j.get("state", {}).get("lastError"),
                                 claimed_by_pid=j.get("state", {}).get("claimedByPid"),
                                 claimed_at_ms=j.get("state", {}).get("claimedAtMs"),
+                                silent_fire_count=j.get("state", {}).get("silentFireCount", 0),
                             ),
                             created_at_ms=j.get("createdAtMs", 0),
                             updated_at_ms=j.get("updatedAtMs", 0),
                             delete_after_run=j.get("deleteAfterRun", False),
+                            silent_fire_limit=j.get("silentFireLimit", 12),
                         )
                     )
                 self._store = CronStore(jobs=jobs)
@@ -265,10 +275,12 @@ class CronService:
                         "lastError": j.state.last_error,
                         "claimedByPid": j.state.claimed_by_pid,
                         "claimedAtMs": j.state.claimed_at_ms,
+                        "silentFireCount": j.state.silent_fire_count,
                     },
                     "createdAtMs": j.created_at_ms,
                     "updatedAtMs": j.updated_at_ms,
                     "deleteAfterRun": j.delete_after_run,
+                    "silentFireLimit": j.silent_fire_limit,
                 }
                 for j in self._store.jobs
             ],
@@ -294,6 +306,7 @@ class CronService:
             self._load_store()
             self._recompute_next_runs()
             self._save_store()
+        self._check_missed_foreign()
         self._loop_task = asyncio.create_task(self._run_loop())
         logger.info("Cron service started with {} jobs", len(self._store.jobs if self._store else []))
 
@@ -394,6 +407,7 @@ class CronService:
             except Exception:
                 logger.exception("Cron: wake tick failed; continuing in {}s", _ERROR_BACKOFF_S)
                 await asyncio.sleep(_ERROR_BACKOFF_S)
+            self._check_missed_foreign()
             delay = self._compute_wake_delay()
             try:
                 await asyncio.wait_for(self._wake_event.wait(), timeout=delay)
@@ -507,8 +521,11 @@ class CronService:
                     j.state.last_run_at_ms = job.state.last_run_at_ms
                     j.state.last_status = job.state.last_status
                     j.state.last_error = job.state.last_error
-                    j.state.next_run_at_ms = job.state.next_run_at_ms
-                    j.enabled = job.enabled
+                    # A store-side disable is sticky: record_fire's auto-disable
+                    # (or a user disabling mid-run) lands between claim and this
+                    # writeback, and the in-memory copy must not resurrect it.
+                    j.enabled = j.enabled and job.enabled
+                    j.state.next_run_at_ms = job.state.next_run_at_ms if j.enabled else None
                     j.updated_at_ms = job.updated_at_ms
                     break
             # Handle "at"-kind delete_after_run (_execute_job removed from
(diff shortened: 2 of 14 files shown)


0xKT:
Fixed in 983b81d: a falsy channel/to on the job side now acts as a wildcard in notify_user_active, mirroring the claim-side rule (pre-attribution jobs are claimable by any runner, so symmetrically any genuine user activity resets them). Added a regression test for the legacy-job reset path; affected suites and lint pass.

gloryfromca (commented):
Blocking: recovery from an auto-disable is broken -- `cron enable` does not clear `silent_fire_count`, so the job fires once and auto-disables again. See the inline note; the fix is one line.

**Scope:** single commit `0c0e435` on `origin/feat/tui_missed_reminders` (859184b), reviewed as `git diff github/feat/tui_missed_reminders...HEAD`.

**What I covered**
- The diff itself, plus the callers and the surrounding cron lifecycle (`_process_due` / `_execute_job` / `_writeback_locked` / `enable_job` / `run_job`), the two wiring sites (gateway, TUI), and the AgentLoop hook gate.
- Repo rules (AGENTS.md): commit header/body are ASCII-clean, test files follow the naming rules (`test_cron_service_*`, `test_cli_*_commands.py`), no new assets.
- Backward compatibility of the store format.
- Whether tests were weakened to look green.

**Things I checked and am satisfied with**
- Sticky-disable writeback. `j.enabled = j.enabled and job.enabled` composes correctly with `record_fire`'s persisted disable and with the `at`-kind path (`_execute_job` sets `job.enabled = False`, so the `and` still yields False). `record_fire` reloads the store mid-`on_job`, which drops `self._store` while `_process_due` still holds the old job objects -- harmless, because `_writeback_locked` re-reads and patches by id, and the claim fields survive the reload so `claimed_by_pid == my_pid` still matches.
- Origin gate. `_SKIP_USER_INBOUND_ORIGINS` is `{SENTINEL, SUBAGENT}`, so CRON/HEARTBEAT turns do reach the hook -- the `Origin.USER` guard in `chain_cron_activity_reset` is load-bearing and correct.
- TUI binding. TUI turns are built with `channel="tui"`, `chat_id="default"` (`raven/tui_rpc/methods/turn.py:197`), matching `CronTool.set_context("tui", "default")`, so the reset actually reaches TUI-created jobs.
- Store round-trip. `silentFireLimit` / `silentFireCount` default on read and are written back; an explicit `null` limit round-trips as "no limit".

**Tests**
- `uv run pytest tests/ -q --ignore=tests/integration` -> `6045 passed, 43 skipped, 13 deselected`. The skips are all missing optional channel SDKs and provider-catalog guards, none from this diff.
- The 7 touched/new test files run 166 assertions with no skip, xfail or env-gated bypass; the new tests genuinely exercise the behaviour (including the writeback-resurrection case).
- Note: `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare` fails without `COLORTERM=truecolor` in the shell. Pre-existing and unrelated to this PR, but worth knowing.

I did not repeat the `payload.channel is None` point already raised in the existing review comment.

raven/proactive_engine/schedulers/cron/service.py
```diff
@@ -556,12 +573,113 @@ async def _execute_job(self, job: CronJob) -> None:
...
+            for j in store.jobs:
+                if j.id != job_id:
+                    continue
+                if j.schedule.kind not in ("every", "cron"):
+                    return False
+                j.state.silent_fire_count += 1
+                limit = j.silent_fire_limit
+                disabled = False
+                if limit is not None and limit > 0 and j.state.silent_fire_count >= limit:
+                    j.enabled = False
```
gloryfromca, line 600: **Blocking: there is no way back from an auto-disable.**

`enable_job` (line 906) flips `enabled` and recomputes `next_run_at_ms`, but never touches `state.silent_fire_count`. After an auto-disable the count is already at the limit, so the first fire after `cron enable` re-trips this branch immediately.

Repro against this branch:

```python
s = CronService(d / "jobs.json")
j = s.add_job("hourly", CronSchedule(kind="every", every_ms=3600_000), "ping",
              channel="feishu", to="chat1")
for _ in range(12):
    s.record_fire(j.id)          # -> auto-disabled, count=12

s.enable_job(j.id, enabled=True) # enabled=True, count still 12
s.record_fire(j.id)              # -> True: disabled again, count=13
```

Output:

```
after cron enable -> enabled: True count: 12 next_run: True
after ONE more fire -> auto-disabled again: True enabled: False count: 13
```

So the user-facing recovery path (`raven cron enable <id>`, documented as "I disabled a routine reminder during vacation, how do I resume?") yields exactly one more fire and then dies again, indefinitely. And there is no other reset surface: `silent_fire_limit` is not settable by `add_job`, by `CronTool`, by any CLI flag, or by config -- it is hardcoded to 12 for every job, existing ones included (`j.get("silentFireLimit", 12)` on load). `notify_user_active` is the only reset, and it skips disabled jobs (`if not j.enabled ... continue`), so "user chats a lot, then re-enables" does not help either -- only "enable, then chat, then wait" does. A user who cannot or does not reply on that binding has no supported escape except deleting and re-adding the job.

One line in `enable_job` fixes it:

```python
if enabled:
    job.state.silent_fire_count = 0
    job.state.next_run_at_ms = _compute_next_run(job.schedule, self._now_ms())
```

Re-enabling is an explicit statement of engagement, so clearing the counter there is the semantically right place.

raven/proactive_engine/schedulers/cron/service.py
```diff
@@ -556,12 +573,113 @@ async def _execute_job(self, job: CronJob) -> None:
...
+                    j.state.next_run_at_ms = None
+                    disabled = True
+                    logger.warning(
```
gloryfromca, line 603: [non-blocking] The auto-disable is invisible to the user: a `logger.warning` is the only signal, and `cron list` hides disabled jobs unless `--all` is passed, so the job simply vanishes from the default view.

Concrete case: a daily reminder on a channel where the user reads but never replies (a med reminder, a standup ping) reaches 12 silent fires after 12 *days* and stops, with nothing telling the user it stopped. The counter itself is only discoverable by running `cron list` before the disable happens -- after it, the row is hidden.

The `~1 day of hourly fires` rationale in `types.py:73` is right for `every_seconds=3000`-style runaways, which is the case this feature is for, but the same fixed count reads very differently for a daily `cron` expression.

This PR already builds the machinery to say something: `make_on_missed_foreign` enqueues a `SystemEvent` + `request_wake_now` so the heartbeat surfaces a notice. Emitting the same kind of event on auto-disable ("I paused '<name>' after N unanswered fires -- reply to resume") would close the loop and would also give the user a natural way to trigger the `notify_user_active` reset. Fine as a follow-up.

0xKT:
Thanks for both - the blocking one was a real one-way door. Fixed in 27cf52f:

**Re-enable resets the counter**: `enable_job` now zeroes `silent_fire_count` when enabling - re-enabling is the clearest user-engagement signal there is, so counting restarts from zero instead of insta-re-disabling on the next fire. Regression test covers the disable -> enable -> fire cycle.

**Auto-disable is now visible twice over**: (1) at disable time the handler enqueues a system event on the same heartbeat path as the missed-reminder notices (your hunch about the mechanism fitting was right), telling the user which reminder was disabled and the exact re-enable command; (2) the default `cron list` view prints a hint for auto-disabled jobs it is hiding ("N reminder(s) auto-disabled ... re-enable with ...") instead of letting a daily reminder silently vanish from view.

Affected suites (143) and lint pass.

raven/proactive_engine/schedulers/cron/service.py
gloryfromca: Confirmed fixed in `27cf52f`. Re-ran my repro against the PR head (`27cf52f`), with the same script as above:

```
after cron enable -> enabled: True count: 0 next_run: True
after ONE more fire -> auto-disabled again: False enabled: True count: 1
```

`enable_job` now zeroes `silent_fire_count` on the enable branch only, so `cron disable` still preserves the count and the one-way door is gone. `test_reenable_resets_the_runaway_counter` covers the disable -> enable -> fire cycle and asserts the third fire returns `False`, which is exactly the regression that would reappear.

Dropping this as blocking.

raven/cli/cron_commands.py
```diff
@@ -272,6 +282,22 @@ def cron_list(
     )
     _print_runner_status(service)
 
+    if not all_:
+        # An auto-disabled reminder (counter at its limit) would otherwise
+        # vanish from the default view without a trace.
+        auto_disabled = [
+            j
+            for j in service.list_jobs(include_disabled=True)
+            if not j.enabled and j.silent_fire_limit and j.state.silent_fire_count >= j.silent_fire_limit
```
gloryfromca, line 291: [non-blocking] AGENTS.md 5.4: "when changing/adding a CLI command, update the matching `test_cli_<module>_commands.py`". This commit changes `cron list` output and `tests/test_cli_cron_commands.py` is untouched -- the base commit did add tests there, so this is a slip in the follow-up rather than a pattern.

Worth closing because the condition on this line is doing real discrimination that a regression would break silently: it must fire for an auto-disabled job and *not* for one the user disabled by hand (count below limit). I verified both hold today by running this against `27cf52f`, so it is a paste-in, not new work:

```python
def test_list_flags_auto_disabled_jobs(fake_cron_dir, runner):
    svc = CronService(fake_cron_dir / "jobs.json")
    j = svc.add_job("hourly", CronSchedule(kind="every", every_ms=3600_000),
                    "ping", channel="tui", to="default")
    for _ in range(12):
        svc.record_fire(j.id)
    result = runner.invoke(cron_app, ["list"])
    assert result.exit_code == 0
    assert "auto-disabled" in result.output


def test_list_does_not_flag_a_manually_disabled_job(fake_cron_dir, runner):
    svc = CronService(fake_cron_dir / "jobs.json")
    j = svc.add_job("hourly", CronSchedule(kind="every", every_ms=3600_000),
                    "ping", channel="tui", to="default")
    svc.enable_job(j.id, enabled=False)
    assert "auto-disabled" not in runner.invoke(cron_app, ["list"]).output
```

The auto-disable `SystemEvent` added in `_cron_handler.py` is likewise uncovered; `tests/test_cli_cron_handler.py` already has the fixtures for it if you want the same treatment there.

gloryfromca (commented):
No blockers left; this can merge as far as I am concerned. My earlier blocking finding is genuinely fixed -- I re-ran the repro against `27cf52f` and it now reads `after cron enable -> count: 0` / `after ONE more fire -> auto-disabled again: False`. One non-blocking note inline about test coverage for the new CLI banner.

**Scope of this pass:** `git diff 0c0e4357..27cf52f` (the two follow-up commits `983b81d` + `27cf52f`), plus the code they touch. I did not re-review the base commit -- that was covered in the previous round.

**The three prior findings**
- **Blocking, re-enable was a one-way door** -- fixed. `enable_job` zeroes `silent_fire_count` on the enable branch only, so `cron disable` still preserves the count. Verified by repro and by `test_reenable_resets_the_runaway_counter`, which asserts the post-enable fire returns `False`.
- **Non-blocking, invisible auto-disable** -- addressed twice over: a `SystemEvent` + `request_wake_now` at disable time (same heartbeat path as the missed notices), and a `cron list` banner. I exercised the banner through `CliRunner` on a temp cron dir: it fires for an auto-disabled job and correctly stays silent for a manually-disabled one (count below limit), including in the `no jobs to list` case, since it is printed before that early return.
- **claude bot's `payload.channel is None` nit** -- addressed. The falsy-side wildcard in `notify_user_active` mirrors `_owns_channel`'s claim-side rule, and the guard still filters correctly in the mixed cases I walked through (job with a real channel but `to=None` is still channel-scoped; job with a real channel is not reset by activity on a different one).

**What I covered:** the incremental diff, the callers (`cron enable` / `cron disable` are the only `enable_job` callers, and `cron_enable` early-returns when already enabled so the reset cannot be triggered spuriously), backward compatibility of the store, repo rules (AGENTS.md: new commit messages are ASCII-clean, no new assets), and whether tests were weakened.

**Tests:** `COLORTERM=truecolor uv run pytest tests/ -q --ignore=tests/integration` -> `6047 passed, 43 skipped, 13 deselected` (was 6045 before these two commits; the two new tests account for the delta). The 14 tests in `test_cron_service_silent_fire.py` all run under `asyncio_mode=auto` -- none skipped, none weakened; the new ones assert real state transitions rather than just non-failure.

**One thing worth knowing, not a finding:** the in-conversation auto-disable notice is gated on `system_events`/`wake`, which only the gateway wires -- `tui_commands.py` has no `SystemEventQueue` or `WakeScheduler` at all. So a `tui`-bound recurring job that auto-disables while the TUI is open tells the user nothing in-session; they only see it via `raven cron list`. That is a gap in the plumbing, not in this PR.

0xKT:
Test gap closed in 8a673dd: two cases pin the new list behavior - the default view prints the auto-disabled hint naming the hidden job, and --all shows the job row itself without the hint. (And thanks for re-running your repro script on the re-enable fix.)

raven/cli/cron_commands.py
gloryfromca: Closed by `8a673dd` -- resolving this thread.

The two new cases pin the positive side (hint present in the default view, naming the hidden job) and the flag side (`--all` suppresses the hint). Neither is the negative I actually cared about -- a *manually* disabled job must not be labelled "auto-disabled after repeated fires" -- so I checked whether that property is pinned anywhere by mutating the condition to a bare `if not j.enabled` and running the file:

```
FAILED tests/test_cli_cron_commands.py::test_list_hides_disabled_by_default
1 failed, 73 passed
```

The pre-existing `test_list_hides_disabled_by_default` catches it, because the hint would name a hand-disabled job's id in the default view and that test asserts the id is absent. So the discrimination is covered after all, just not by the new tests. Nothing further needed.

Full suite at `8a673dd`: `6049 passed, 43 skipped, 13 deselected`.

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. Re-stating for the rebased head so the sign-off attaches to the code that will actually land.

`cc219bf` is a pure rebase of `8a673dd` onto the new base tip `fba4673d`. I verified that rather than assuming it: the PR's own three-dot diff is byte-identical to the one I reviewed last round.

```
diff <(git diff 859184b...8a673dd) <(git diff fba4673d...cc219bf)   # only index lines differ
```

The delta a two-dot diff shows (skill-forge RRF `rrf_k` plumbing, the `local` weight drop to 0.96, the `chatStream.ts` dated-timestamp rendering, the `docs/dev.md` `raven agent` -> `raven tui` fix) all comes from the base branch moving, not from this PR. I checked the one place the two could have collided -- the base's `cron.missed` TUI block and this PR's `make_on_missed_foreign` are separate paths (TUI startup-drops vs the gateway's system-event/heartbeat notice), so the rebase is semantically clean, not just textually.

**Tests at this head:** `COLORTERM=truecolor uv run pytest tests/ -q --ignore=tests/integration` -> `6055 passed, 43 skipped, 13 deselected`. The delta from 6049 last round is the base branch's own new tests; nothing from this PR changed count or state, and no skip comes from the cron work.

All four threads I opened or was involved in are resolved: the blocking re-enable one-way door (`27cf52f`, repro re-run), the invisible auto-disable (`27cf52f`, banner exercised through CliRunner), the CLI test gap (`8a673dd`, plus the mutation check showing the manual-disable negative is pinned by the pre-existing `test_list_hides_disabled_by_default`), and claude bot's falsy-channel nit (`983b81d` / `818c1fd9`). Nothing new to raise.

gloryfromca (approved):

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. Sign-off re-attached to the new head after the second rebase.

`e648d87` is again a pure rebase -- this time because the base branch itself was rebased onto a newer main (picking up #330 and #331). The PR's own three-dot diff against the new base tip `92f47fee` is byte-identical to the one I reviewed at `cc219bf`:

```
diff <(git diff fba4673d...cc219bf) <(git diff 92f47fee...e648d87)   # only index lines differ
```

The one place the two sides could have collided is `raven/config/schema.py`, which both the base delta and this PR modify. They do not overlap semantically: the base adds a new `CliConfig` section plus its `Config.cli` field, while this PR adds `notify_missed` to `CronConfig`. Different classes, no shared keys, no precedence interaction.

**Tests at this head:** `COLORTERM=truecolor uv run pytest tests/ -q --ignore=tests/integration` -> `6079 passed, 43 skipped, 13 deselected`. The jump from 6055 is the base's own new tests (`test_cli_status_commands`, `test_error_classification`, `test_openai_codex_provider`, `test_cli_repl_spine`); nothing from the cron work changed state, and none of the skips come from it.

Nothing new to raise -- all four earlier threads stay resolved.

gloryfromca (dismissed):

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. Sign-off re-attached after the retarget to `main`.

What changed since my last pass is the target, not the code: #332 was squash-merged to `main` as `13ea802`, and this PR was rebased onto it. Two checks rather than an assumption:

```
git diff --stat 92f47fee github/main          # empty -- squash produced the same tree as the old base tip
diff <(git diff 92f47fee...e648d87) <(git diff github/main...HEAD)   # only index lines differ
```

So the diff under review is the same 14 files / +1348/-7 I reviewed at `e648d87`, now expressed against `main`. New commit shas (`2e9d396` / `b986a80` / `64f4e9d` / `e1abbc5`) are the rebase renumbering the same four commits; their messages are still ASCII-clean per AGENTS.md 3.1.1.

**Tests at this head:** `COLORTERM=truecolor uv run pytest tests/ -q --ignore=tests/integration` -> `6079 passed, 43 skipped, 13 deselected` -- same counts as at `e648d87`, consistent with the identical trees.

Nothing new to raise; the four earlier threads stay resolved.

gloryfromca (approved):

PR #334, opened 2026-08-17
refactor(cli): drop the inert skill-extraction flags

0xKT:
## Summary

`raven agent -m` exposed two flags whose help text described a
mechanism that no longer exists: --wait-skill-extract and
--flush-skill-buffer only ever reached
AgentLoop.await_pending_extractions, which has been an explicit
documented no-op since the embedded extraction pipeline was replaced by
the memory-backend flush (extraction now fires every turn on the
backend side). The flags never changed any behavior; their help text
promised boundary-detection workflows that silently did nothing.

This removes both flags, their consumption block, and the no-op method
(its only real caller was that block). A guard test pins the help
output so the flags cannot silently return. The proactivity benchmark
driver, which unconditionally appended one of the flags to its CLI
invocations, is updated in the same change - it would have broken at
runtime otherwise (no caller uses its wait_skill_extract kwarg).

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [x] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_cli_agent_commands.py -x`: 29 passed, including the new guard test asserting `raven agent --help` mentions neither flag.
- Full suite (no -x): 6315 passed / 2 failed, both pre-existing on main (the image-block test and the order-dependent theme test).
- `uv run raven agent --help` on a real run shows no skill/extract/flush mention; `uv run ruff format` and `ruff check` clean across raven/, tests/, benchmarks/.

## Risk

Scripts passing the removed flags will now get an unknown-option error
instead of a silent no-op; that is the honest outcome, since the flags
have not done anything for several releases. Revert the branch to roll
back.

## Related Issues

N/A - no tracked GitHub issues.

### DIFF
--- a/benchmarks/proactivity_eval/runners/_common/raven_driver/driver.py
+++ b/benchmarks/proactivity_eval/runners/_common/raven_driver/driver.py
@@ -146,7 +146,6 @@ class RavenDriver:
         *,
         fake_now: str | None = None,
         session_id: str = "cli:direct",
-        wait_skill_extract: bool = False,
     ) -> AgentResponse:
         """Run ``raven agent --message <m> [--fake-now ...]`` once.
 
@@ -160,10 +159,6 @@ class RavenDriver:
         cmd.extend(["--session", session_id])
         cmd.append("--no-markdown")
         cmd.append("--no-logs")
-        if wait_skill_extract:
-            cmd.append("--wait-skill-extract")
-        else:
-            cmd.append("--no-wait-skill-extract")
         if fake_now:
             cmd.extend(["--fake-now", fake_now])
         return self._run(cmd)
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -2421,25 +2421,6 @@ class AgentLoop:
             except Exception:
                 logger.exception("on_turn_complete callback failed")
 
-    async def await_pending_extractions(
-        self,
-        flush_session_id: str | None = None,
-        *,
-        wait: bool = True,
-    ) -> None:
-        """No-op retained for CLI / batch-mode call-site compatibility.
-
-        Previously this flushed the local skill-extraction buffer and
-        blocked on its in-flight tasks. That embedded pipeline was
-        removed — case-to-skill distillation now lives in the
-        :class:`MemoryBackend` plugin (``backend.store`` /
-        ``backend.feedback``), which the after-turn pipeline drives
-        directly. Kept so ``raven agent`` callers don't need changing;
-        the ``flush_session_id`` / ``wait`` knobs are inert.
-        """
-        del flush_session_id, wait
-        return None
-
     async def close_mcp(self) -> None:
         """Close MCP connections and the sandbox executor."""
         if self._mcp_stack:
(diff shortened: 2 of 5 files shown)


gloryfromca (commented):

### What I checked

The diff and its callers, backward compatibility, repo rules (AGENTS.md branch/commit/test-naming, uv-only), and whether the new test is a real guard or a decoration.

**The premise holds.** `AgentLoop.await_pending_extractions` is an explicit no-op at `github/main`'s head (`del flush_session_id, wait; return None`), so both flags were genuinely inert and nothing observable is lost by removing them.

**No dangling callers.** A repo-wide grep (all file types, not just `*.py`) for `skill-extract` / `skill_extract` / `flush_skill` / `pending_extractions` leaves only: the two new `--help` assertions, three dead test-double stubs (see nit), and `min_quality_for_skill_extract` in `raven/config/raven.py`, which belongs to the backend-side extractor and is unrelated. Nothing passes `wait_skill_extract=` to `RavenDriver.send_message` -- the two call sites (`longrun_adapters.py:605`, `backends/raven.py:122`) pass only `content` / `fake_now` / `session_id`, so the signature change is safe.

**The new test is a real guard, not a green-looking no-op.** I was suspicious that a `not in r.stdout` assertion on Rich-rendered help could pass for the wrong reason (column truncation). I built a positive control: a synthetic Typer app with both flags restored at the same 80-col width. Rich ellipsizes the *negated* forms (`--no-wait-skill-ex...`) but renders `--wait-skill-extract` and `--flush-skill-buffer` verbatim, so both assertions would fail if the flags came back. Genuine regression coverage.

**Conventions.** Branch `refactor/drop_inert_skill_extract_flags` matches 2.1; commit message is ASCII-clean end to end (checked with `LC_ALL=C grep '[^ -~]'`) and carries the `Co-authored-by` trailer; the test went into the existing `tests/test_cli_agent_commands.py` rather than a new file, per 5.4.

**Backward compatibility.** Removing the flags does turn a silent no-op into an unknown-option exit for any script still passing them. The PR states this explicitly under Risk, and I agree it is the right trade: keeping a flag whose help text promises boundary-detection work the CLI cannot perform is worse than a loud failure.

### Test results

- `uv run pytest tests/test_cli_agent_commands.py -q` -- 29 passed, including the new guard.
- `uv run pytest tests/ -q -p no:randomly` -- **6079 passed, 1 failed, 43 skipped**. The failure is `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`, which also fails when run alone (`1 failed, 44 passed in 0.23s`) and touches no code in this diff -- pre-existing, not order-dependent as the PR description guesses. The PR describes 2 failures; only 1 remains on this base. The 43 skips are optional-dependency channel imports (`slack_sdk`, `telegram`, ...) plus provider-matrix parametrizations, all pre-existing and unrelated.
- `uv run ruff check raven/ tests/ benchmarks/` -- clean.

### Nit (non-blocking)

Three test doubles still stub the deleted method -- `async def await_pending_extractions(self, **_kw)` at `tests/test_cli_agent_commands.py:140`, `:444`, `:659`. They are harmless (nothing calls them now), but they are the same category of artifact this PR exists to remove: a stand-in for a mechanism that no longer exists. Worth dropping while you are in the file. I could not anchor this inline because those lines are outside the diff.

gloryfromca (approved):

PR #341, opened 2026-08-17
fix(*): stop an old default from capping every model's context window

arelchan:
## Summary

Every config written by 0.1.10 or earlier carries `contextWindowTokens: 65536` verbatim: that was the schema default back then, and the bootstrap dumped every default to disk. 0.1.11 (#287) started sizing the window from the model's own catalogue, but an explicit pin still wins over it -- correctly, that is what a pin means -- so **the fix reached new installs only**. Every upgraded install stayed capped at 64k on a 1M-token model, and `/model` could not shake it loose (`refresh_context_window` returns early for an explicit window, by design).

The blast radius is not cosmetic. The window sizes the per-turn history budget (`_make_token_budget`), the Curator's fast/slow path threshold (`available_history * 0.60`) and the memory consolidator's archive target (`window // 2`). An upgraded install therefore trimmed history early, paid for an extra Curator LLM call per turn, and consolidated at 32k instead of ~500k. The only existing signal was a `logger.warning` in a file nobody reads.

Key decisions:

- **Clear the pin once, under a watermark kept in a sidecar (`config.migrations.json`).** The value carries no provenance -- 65536 we planted and 65536 a user chose are the same three bytes -- so a watermark is the only thing that lets a re-set value stand for good. Without it, a user re-setting 65536 would have it stripped on every load, which is worse than the bug. The watermark deliberately does **not** live in config.json: `Config` is `extra='forbid'` and `load_config` raises rather than falling back, so a key this build knows and an older one does not is a hard boot failure for the older build -- a reverted release, a pinned version, or a second checkout sharing one `~/.raven`, which is already the case in this repo. A sidecar no published build reads is ignored instead of fatal. Because it is our file rather than the user's, it is also stamped unconditionally, which closes the one hole an in-file stamp had to accept: a `contextWindowTokens` the user sets by hand afterwards is never second-guessed, whatever its value.
- **Run it at config load, not in `raven upgrade`.** The upgrade helper is the outgoing version and cannot know a rule that ships with the incoming one; it `execve`s a bare Python bootstrap that never imports raven. And `install.sh` / `uv tool upgrade` never call `raven upgrade` at all. Config load is the one chokepoint every entry point shares, so the user experience is "upgraded, and it was already fixed".
- **A file we did not edit is left byte for byte alone.** config.json is rewritten only where the fossil was actually removed -- `provider use` and friends promise not to touch a config they decided against changing, and that promise is theirs to keep, not ours to spend (`test_use_leaves_the_config_alone_when_it_cannot_tell` caught an earlier draft that stamped unconditionally *in the config*).
- **The rewrite carries the file mode across.** `os.replace` swaps the inode, so a config tightened to 0600 -- it holds `providers.*.apiKey` -- came back 0644 in the first draft. The temp file is chmod-ed to the original's mode before the swap, and named per-PID so two processes migrating at once cannot have one truncate the file the other is about to move into place.
- **The write is surgical.** It re-reads the raw file, because the mapping `load_config` migrated has the extension blocks popped (`pop_extension_keys`) and writing that back would delete the user's `memory` / `plugins` / `skillForge` sections. It never goes through `save_config`, which dumps ~8 KB of defaults and would re-plant exactly this kind of fossil. A read-only home degrades to an in-memory-only migration rather than failing the boot.
- **The user is told where the user is looking, from every entry point.** The notice is drained in the `run()` wrapper, so all thirteen config-loading commands tell it, not the two that print eagerly (`agent`, `gateway`, which keep the early print because they hold the screen for a long time). This matters because the watermark suppresses the next run's notice: an unsaid notice is lost, not deferred, and `raven status` would otherwise migrate the file in silence forever. The TUI and the served page get a new `config_notices` field on the session init bundle, rendered as a system line on both `session.create` and `session.resume` -- resume drains the same one-shot list, so a picker-first user would otherwise see nothing. This also gives the client-side `config_warning` / `credential_warning` handlers their first server-side sibling that is actually populated -- both were dead wiring. The notice names the way back, because a window genuinely smaller than 65536 (a local model, an endpoint served with `--max-model-len`) is a real configuration this clears once.
- **The loop's 65536 warning is gone.** Post-migration, a pin that reaches the constructor is a deliberate choice, and warning about deliberate choices is noise.

Follow-ups, deliberately not in this PR: `raven doctor --fix` so the check can be run on demand rather than only at the next launch, and stopping the bootstrap from materialising defaults at all (`save_config(load_config())` in onboard) -- the root cause, and a change to every new user's config file shape, so it deserves its own review.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/test_config_loader.py                      31 passed (16 new migration cases, incl. the
                                                       sidecar watermark, mode preservation, per-PID temp)
pytest tests/test_tui_rpc_session_init_bundle.py         2 new cases, all green
pytest test_config_loader + test_config_schema + test_config_raven_loader \
     + test_cli_provider_commands + test_tui_rpc_session_init_bundle + test_cli_smoke
                                                       194 passed (re-run after rebasing on main)
pytest tests/ -k "config or agent_loop or provider or helpers or tui_rpc or usage_sink or onboard or gateway"
                                                       failure set identical with and without this branch
                                                       (56 -> 56, zero delta; pre-existing, a local config
                                                       block the schema on main does not accept)
ruff check raven/ tests/                               All checks passed
ruff format --check                                    already formatted
npm run type-check (ui-tui)                            clean
npx vitest run src/__tests__ (ui-tui)                  843 passed (70 files)
npx prettier --check / npx eslint on both changed .ts  0 errors (1 pre-existing warning at line 146)
```

End-to-end on a config shaped like an upgraded install (`contextWindowTokens: 65536`, plus `memory` and `skillForge` blocks):

```
first load    in-memory pin      -> None
              file               -> the line is gone, model/maxTokens/temperature untouched
              watermark          -> config.migrations.json, nothing added to config.json
              extension blocks   -> intact
              notice             -> surfaced once, then drained
user re-sets 65536 afterwards    -> honoured, no further notice
config with no fossil            -> file byte identical, no write
read-only home                   -> migration still correct in memory, no raise
0600 config with an apiKey        -> still 0600 after the migration
migrated config, older build     -> loads fine (the downgrade path the sidecar exists for)
raven status on a fossil config  -> prints the notice (it used to migrate in silence)
```

## Risk

User-visible behaviour changes: on the first launch after this lands, an upgraded install loses the `contextWindowTokens: 65536` line from `~/.raven/config.json` and its context window jumps to whatever the model actually supports. Sessions get a longer history budget, the Curator drops back to its zero-LLM fast path far more often, and the status bar percentage falls accordingly. Users who genuinely need a 64k ceiling (a local model, a self-hosted endpoint with a smaller serving window) must re-add the line, which the notice tells them; those requests would fail loudly at the endpoint rather than silently, so the failure is self-announcing.

Rollback: revert the commits. Nothing of ours is left inside config.json, so an already-migrated config is byte-for-byte a config that simply has no pin -- exactly the state a fresh install has, loadable by every published build. The orphaned `config.migrations.json` sidecar is ignored by any build that does not look for it, and re-applying the change later just re-reads it. An earlier draft of this PR put the watermark in config.json and would have made a revert leave every migrated install unable to boot (`Config` is `extra='forbid'`); that is fixed, and the same reasoning covers pinning an older release or running a second checkout against the same `~/.raven`.

## Related Issues

N/A

### DIFF
--- a/raven/cli/_helpers.py
+++ b/raven/cli/_helpers.py
@@ -352,6 +352,19 @@ def print_deprecated_memory_window_notice(config: Config) -> None:
         )
 
 
+def print_config_migration_notices() -> None:
+    """Tell the user about any config line a migration just changed for them.
+
+    The migrations run inside the loader, which has no terminal; this is the
+    place that does. Call it after the config is loaded and before the command
+    takes over the screen -- once printed, the notices are gone.
+    """
+    from raven.config.loader import drain_migration_notices
+
+    for notice in drain_migration_notices():
+        console.print(f"[yellow]Config updated:[/yellow] {notice}")
+
+
 __all__ = [
     "DEFAULT_PROBE_MESSAGE",
     "make_provider",
@@ -360,4 +373,5 @@ __all__ = [
     "load_runtime_config",
     "parse_fake_now",
     "print_deprecated_memory_window_notice",
+    "print_config_migration_notices",
 ]
--- a/raven/config/loader.py
+++ b/raven/config/loader.py
@@ -2,13 +2,21 @@
 
 import json
 import logging
+import os
 import sys
 from pathlib import Path
 from typing import Any
 
 from pydantic import ValidationError
 
-from raven.config.schema import Config
+from raven.config.schema import CURRENT_CONFIG_VERSION, Config
+
+# The context window every pre-0.1.11 bootstrap wrote to disk verbatim: back
+# then ``AgentDefaults.context_window_tokens`` defaulted to this number and
+# ``save_config`` dumped every default. A pin is honoured over the model's real
+# window by design, so on an upgraded install this exact value silently caps
+# every model at 64k -- see ``_migrate_legacy_context_window``.
+LEGACY_CONTEXT_WINDOW_TOKENS = 65_536
 
 # Single source of truth for Raven extension block keys.
 # Both _migrate_config (pop before base Config validates) and
@@ -37,6 +45,14 @@ _current_config_path: Path | None = None
 # Paths already warned about as malformed in this process; repeated
 # load_config calls (status/doctor load more than once) warn only once.
 _warned_paths: set[str] = set()
+
+# User-facing lines produced by a migration that actually changed something,
+# waiting to be printed by whichever CLI entry point owns the terminal.
+# Migrations run inside the loader, which has no console of its own and whose
+# logs land in a file nobody reads; a config the user can see us edit has to be
+# announced where the user is looking. Drained, not read, so N loads per
+# process yield one telling.
+_migration_notices: list[str] = []
 
 
 def set_config_path(path: Path) -> None:
@@ -90,6 +106,117 @@ def read_raw_or_raise(path: Path) -> dict[str, Any]:
         ) from exc
 
 
+def drain_migration_notices() -> list[str]:
+    """Take the pending migration notices, clearing them.
+
+    Drained rather than read so that a process loading the config several times
+    (status and doctor do; the TUI RPC server reloads every turn) tells the user
+    once. Callers own a console -- see ``cli._helpers``.
+    """
+    notices = list(_migration_notices)
+    _migration_notices.clear()
+    return notices
+
+
+def _config_version(data: dict[str, Any]) -> int:
+    """The migration generation stamped on a raw config mapping.
+
+    Absent / unparseable means generation 0, i.e. a file written before the
+    stamp existed -- every migration still owes it a pass.
+    """
+    raw = data.get("configVersion", data.get("config_version"))
+    try:
+        return int(raw)
+    except (TypeError, ValueError):
+        return 0
+
+
+def _migrate_legacy_context_window(data: dict[str, Any], *, notify: bool = False) -> bool:
+    """Clear the retired 65536 context-window pin. True when ``data`` changed.
+
+    Gated on the stamp, and stamps in turn, because the value carries no
+    provenance: 65536 written by the old bootstrap and 65536 chosen by a user
+    are the same three bytes. Running once means we clear what we planted; from
+    then on the number is the user's, and the ladder in
+    ``providers.rates.effective_context_window`` honours it like any other pin.
+
+    ``notify`` is for the caller that has a user to tell -- the persist pass
+    re-runs this on the raw file and must not double-announce.
+
+    Nothing found means nothing written: the stamp lands only on a file we
+    actually edited, so a config that never carried the fossil is left byte for
+    byte alone. The cost is a narrow hole -- someone who hand-sets 65536 on a
+    still-unstamped config has it cleared once, with the notice explaining why,
+    and their second attempt sticks (that first clear stamps the file). Paid
+    knowingly: touching every config on earth to close it is the worse trade.
+    """
+    if _config_version(data) >= CURRENT_CONFIG_VERSION:
+        return False
+
+    agents = data.get("agents")
+    defaults = agents.get("defaults") if isinstance(agents, dict) else None
+    if not isinstance(defaults, dict):
+        return False
+
+    changed = False
+    for legacy_key in ("contextWindowTokens", "context_window_tokens"):
+        if defaults.get(legacy_key) == LEGACY_CONTEXT_WINDOW_TOKENS:
+            defaults.pop(legacy_key)
+            changed = True
+            logging.getLogger(__name__).info(
+                "Migrated: dropped agents.defaults.%s (the retired 65536 default)",
+                legacy_key,
+            )
+
+    if not changed:
+        return False
+
+    notice = (
+        f"Removed the leftover `contextWindowTokens: {LEGACY_CONTEXT_WINDOW_TOKENS}` from your config (an old "
+        "default, written there by earlier versions). The context window now follows each model's real size. "
+        "Put the line back if you did want that number -- for an endpoint served with a smaller window, for "
+        "instance."
+    )
+    # Deduped, not just drained: the extension loader migrates its own read of
+    # the same file, so one command can walk this path twice before anything is
+    # printed.
+    if notify and notice not in _migration_notices:
+        _migration_notices.append(notice)
+    data["configVersion"] = CURRENT_CONFIG_VERSION
+    return True
+
+
+def _persist_migrated_config(path: Path) -> None:
+    """Apply the stamped migrations to the file itself. Best effort, once.
+
+    Re-reads the raw file instead of reusing the mapping ``load_config`` already
+    migrated: that one has the extension blocks popped (see
+    ``pop_extension_keys``), so writing it back would delete the user's memory /
+    plugins / skillForge sections. ``save_config`` is no use here either -- it
+    dumps every default (~8 KB), re-planting the very kind of fossil this
+    migration exists to pull out. So: surgical edit, atomic replace.
+
+    Silent on failure (read-only home, a lost race with another process): the
+    in-memory migration already made this process correct, and the write only
+    serves to keep the file from disagreeing with it. Two processes racing here
+    write byte-identical content.
+    """
+    try:
+        raw = read_raw_or_raise(path)
+    except ConfigReadError:
+        return
+    if not raw or not _migrate_legacy_context_window(raw):
+        return
+
+    tmp = path.with_name(path.name + ".migrating")
+    try:
+        tmp.write_text(json.dumps(raw, indent=2, ensure_ascii=False), encoding="utf-8")
+        os.replace(tmp, path)
+    except OSError as exc:
+        logging.getLogger(__name__).debug("Could not persist config migration to %s: %s", path, exc)
+        tmp.unlink(missing_ok=True)
+
+
 def load_config(config_path: Path | None = None) -> Config:
     """
     Load configuration from file or create default.
@@ -107,7 +234,13 @@ def load_config(config_path: Path | None = None) -> Config:
         try:
             with open(path, encoding="utf-8") as f:
                 data = json.load(f)
+            # The stamp before and after: a migration only stamps a file it
+            # actually changed, so this pair is also the "did anything happen"
+            # signal -- and the reason an untouched config costs no extra file
+            # read per load (the TUI RPC server reloads every turn).
+            unstamped = _config_version(data) < CURRENT_CONFIG_VERSION
             data = _migrate_config(data)
+            migrated = unstamped and _config_version(data) >= CURRENT_CONFIG_VERSION
         except json.JSONDecodeError as e:
             # Boot on defaults for a malformed file (a transient mid-write race
             # shouldn't brick callers) but warn LOUDLY -- a persistent syntax
@@ -142,6 +275,10 @@ def load_config(config_path: Path | None = None) -> Config:
                 raise ValueError(
                     f"Config at {path} fails schema validation:\n{e}",
                 ) from e
+            # Only once the migrated data is known to validate: a file we
+            # cannot load is a file we have no business rewriting.
+            if migrated:
+                _persist_migrated_config(path)
 
     if config is None:
         config = Config()
@@ -177,6 +314,11 @@ def _migrate_config(data: dict, *, pop_extension_keys: bool = True) -> dict:
     import logging as _logging
 
     _log = _logging.getLogger(__name__)
+
+    # Stamped, run-once migrations first: they are the ones allowed to reach
+    # back out to disk (see _persist_migrated_config), and the stamp they set
+    # must not be confused by anything the shims below do.
+    _migrate_legacy_context_window(data, notify=True)
 
     # Move tools.exec.restrictToWorkspace → tools.restrictToWorkspace
     tools = data.get("tools", {})
--- a/raven/config/loader.py
+++ b/raven/config/loader.py
@@ -9,7 +9,14 @@ from typing import Any
 
 from pydantic import ValidationError
 
-from raven.config.schema import CURRENT_CONFIG_VERSION, Config
+from raven.config.schema import Config
+
+# Generation counter for the run-once config migrations below. Bump it, and add
+# the matching rule, when a migration must run exactly once per config rather
+# than on every load -- the watermark is what lets a user re-set by hand
+# whatever a migration cleared. Kept out of the schema on purpose: see
+# ``_stamp_path``.
+CURRENT_CONFIG_VERSION = 1
 
 # The context window every pre-0.1.11 bootstrap wrote to disk verbatim: back
 # then ``AgentDefaults.context_window_tokens`` defaulted to this number and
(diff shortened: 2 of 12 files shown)


raven/config/schema.py
```diff
@@ -841,6 +847,12 @@ class CliConfig(Base):
 class Config(BaseSettings):
     """Root configuration for raven."""
 
+    # Which generation of one-shot config migrations this file already went
+    # through (see loader._migrate_config). Detection reads the RAW file, where
+    # an absent key means "generation 0", so this default is free to name the
+    # current generation -- and must, so a config written fresh today is never
+    # taken for a pre-migration one.
+    config_version: int = Field(default=CURRENT_CONFIG_VERSION, alias="configVersion")
```
gloryfromca: **Blocking.** `Config` is `extra="forbid"` (line 1015) and `load_config` turns a `ValidationError` into a raised `ValueError` rather than falling back to defaults (`loader.py:270-276`). So a config carrying `configVersion` is not merely unrecognised by an older build -- it is a hard boot failure. Confirmed against `github/main`'s schema:

```
$ # main's Config.model_validate({'configVersion': 1, ...})
OLD SCHEMA REJECTED: ValidationError
configVersion
  Extra inputs are not permitted [type=extra_forbidden, input_value=1, input_type=int]
```

Concrete failures:

1. The rollback in the PR description does not work. `git revert` removes this field, and every install that had already launched once now fails at load with `Config at ~/.raven/config.json fails schema validation`. The description says those configs "stay migrated ... which is the same state a fresh install has" -- they do not, they carry a key the reverted schema rejects. Since the squash body *is* this description, that claim lands on `main` (section 3.7 asks Risk to state how to roll back).
2. It fires inside the repo today, not just on a hypothetical downgrade: worktrees share `~/.raven/config.json`. Run `raven` once on this branch, then `raven status` from a checkout of `main`, and it raises. Same for anyone who pins an older release.

Worth noting the escape hatch: `Base` (line 22) has no `extra` setting, so it defaults to `ignore` -- nested unknown keys are silently dropped by old and new builds alike. Stamping inside a `Base` block, or in a sidecar under `~/.raven/`, would make old builds ignore the stamp instead of dying on it, and keeps a revert clean. If you would rather keep the top-level key, then the minimum is to say so in Risk: the revert must retain `config_version` in the schema, and downgrading to any published release will not boot.

raven/config/loader.py
```diff
@@ -90,6 +106,117 @@ def read_raw_or_raise(path: Path) -> dict[str, Any]:
...
+    try:
+        raw = read_raw_or_raise(path)
+    except ConfigReadError:
+        return
+    if not raw or not _migrate_legacy_context_window(raw):
+        return
+
+    tmp = path.with_name(path.name + ".migrating")
+    try:
+        tmp.write_text(json.dumps(raw, indent=2, ensure_ascii=False), encoding="utf-8")
```
gloryfromca: The write does not carry the file's mode across. `write_text` creates the temp file under the process umask and `os.replace` swaps the inode, so the mode of the original is gone -- which is the exact case `config.paths.restrict_to_owner` warns about ("a rewrite that replaces the file instead has to set the mode on the replacement itself"). `config.json` holds `providers.*.apiKey` (`schema.py:364`, `372`).

Reproduced:

```
$ chmod 600 config.json     # {"agents": {...65536}, "providers": {"anthropic": {"apiKey": "sk-secret"}}}
-rw-------  config.json
$ python -c 'load_config(p)'
mode after migration: 0o644
```

A user who tightened their config to owner-only gets it widened to world-readable, silently, on the first launch after upgrading -- with the key still in it. `os.chmod(tmp, path.stat().st_mode & 0o7777)` before the `os.replace` fixes it (skip when the `stat` fails).

While in here, a smaller one on line 211: the temp name is fixed, so two processes migrating at once use the same path. The second one's `write_text` truncates the file the first is about to `os.replace` into place, which can leave `config.json` momentarily empty for a concurrent reader (that reader boots on defaults with the "not valid JSON" warning). The content does end up whole, so this is not the data loss the fixed name suggests, but `path.name + f".migrating.{os.getpid()}"` removes the window for free.

raven/cli/_helpers.py
```diff
@@ -328,6 +328,19 @@ def print_deprecated_memory_window_notice(config: Config) -> None:
         )
 
 
+def print_config_migration_notices() -> None:
```
gloryfromca, line 355: Only `run` and `gateway` call this, but `load_config()` migrates and rewrites the file from many more commands: `status_commands.py:27`, `doctor_commands.py:172`, `cron_commands.py:80`, `provider_commands.py:375`, `channel_commands.py:439`, `skill_commands.py:41`, `sandbox_commands.py:48`, `import_commands.py:110`.

Because the notice is keyed to the migration and the migration stamps the file, the telling is not deferred -- it is lost. If the first raven command a user happens to run after upgrading is `raven status` (or `provider list`, or `cron list`), the config is edited, `configVersion: 1` is stamped, nothing is printed, and no later run will ever mention it: the next load sees the stamp and produces no notice.

That is the case the notice exists for. The user who genuinely needs a 64k ceiling (local model, endpoint with `--max-model-len`) loses the line and the explanation together. Calling this from `load_runtime_config`, or from the `run()` wrapper after the command returns, would cover every command instead of two. The TUI path is fine -- `tui_commands.py:448` loads in the same process that serves RPC, so `_default_session_info` drains it.

gloryfromca (commented):
**Blocking: the top-level `configVersion` stamp makes every earlier raven build refuse to boot on a config this migration touched -- including the `git revert` rollback the description promises.** The diagnosis and the core fix are right, and the tests are honest; two smaller findings below the blocker.

What holds up, so it is on the record:

- The fix really does fix it. `AgentDefaults.context_window_tokens` is `int | None = None` (`schema.py:267`), so popping the key hands the window back to `providers.rates`; I confirmed end to end on a config shaped like an upgraded install.
- The two design calls that look risky are the right ones. Re-reading the raw file in `_persist_migrated_config` instead of writing back the mapping `load_config` migrated is necessary -- `pop_extension_keys` has already removed `memory` / `plugins` / `skillForge` -- and `test_migration_write_back_keeps_extension_blocks` pins it. Avoiding `save_config` is likewise necessary; `save_config` does `model_dump(by_alias=True)` (`loader.py:300`), which is how the fossil got planted in the first place, and `test_migration_write_back_does_not_materialise_defaults` pins that too.
- Tests are not weakened. 53 passed in `tests/test_config_loader.py tests/test_tui_rpc_session_init_bundle.py`, nothing skipped or xfailed, and each new case asserts observable state (on-disk bytes, drained notices) rather than "did not raise". The read-only-home case does chmod the directory rather than mock the failure.
- Repo rules: branch `fix/config_legacy_context_window_migration` and the header both match sections 2 and 3, the `Co-authored-by` trailer is present, tests went into the existing files instead of new phase-suffixed ones (section 5.4), no assets (section 7).

Covered: the diff, `AGENTS.md` rules, callers of `_migrate_config` (including `config/raven.py:1356`, which is in-memory only and unaffected), the CLI entry points that load config, backward and *forward* compatibility, and whether the tests were weakened. I did not check the ui-tui side beyond reading the two changed files.

arelchan:
All four findings reproduced and fixed in `3156fc3` (rebased on latest main). Nothing was waved off -- the first one was the more serious bug of the two in this PR.

**The stamp in config.json (blocking).** Confirmed exactly as described: a build without the field raises `Config at ... fails schema validation`, and the rollback claim in the description was simply false. The cross-checkout case is not hypothetical here either -- this repo has a dozen worktrees against one `~/.raven`.

Took the sidecar option: `config.migrations.json`, sibling of the config it describes so a `--config` path or a second instance gets its own. Verified the downgrade path directly -- migrate on this branch, then load the same file from a checkout without the change: loads fine.

The suggestion turned out to be strictly better than what it replaced, not just safer. Because the sidecar is ours rather than the user's, it can be stamped unconditionally, which closes the hole the in-file stamp had to accept: previously a user hand-setting 65536 on a never-stamped config had it cleared once. Now config.json is touched only when the fossil is actually removed, *and* a value the user sets afterwards is never second-guessed. The PR description's Risk section is rewritten accordingly, since the squash body is that description.

**File mode (blocking in effect).** Reproduced: 0600 -> 0644 with the apiKey still in the file. Fixed by chmod-ing the temp file to the original's mode before the swap, with the `stat` failure skipped as suggested. `restrict_to_owner`'s docstring was the right pointer -- a replacing writer owns the mode.

Took the per-PID temp name too. Agreed it is not data loss, but the window costs nothing to remove.

**Notice wired to two commands out of thirteen.** The framing is what made this land: the telling is *lost*, not deferred, because the watermark leaves the next load with nothing to report. `raven status` migrating in silence forever is precisely the user this notice exists for.

Drained from the `run()` wrapper, so every command tells it. Kept the eager print in `agent` and `gateway` -- they hold the screen for a long time, so the notice is better placed before they start than after they exit; the drain makes double-printing impossible either way. Verified with `raven status` on a fossil config: it now prints.

**session.resume swallowing the notice.** Correct, and the picker-first flow is the common case, not the exotic one. The resume handler now renders `config_notices` like the create handler, and `SessionResumeResponse.info` types it. Left `config_warning` / `credential_warning` on resume alone -- both are create-only fields today with no producer, so wiring them there would be dead code in a PR that is already wide.

Full gates after the rebase: 194 passed across the config / provider / RPC / smoke files, 31 in `test_config_loader.py` (16 migration cases now, including the sidecar, the mode and the temp path), zero new failures against a stashed baseline on the same selection, ruff clean, ui-tui type-check clean, 843 vitest passing.

gloryfromca (commented):
**No blockers; this can merge as far as I am concerned.** Re-reviewed the new head (`3156fc3c`, rebased on `7bd02804` + the watermark commit). All three findings from my last pass are fixed, and fixed at the root rather than papered over.

Verified rather than taken on trust, on a config shaped like an upgraded install (fossil + `providers.anthropic.apiKey` + a `memory` block, `chmod 600`):

```
window  -> None                       # pin gone, sized from the model
mode    -> 0o600                      # preserved across the os.replace
file    -> {"agents": {"defaults": {"model": "x/y"}}, "providers": {...}, "memory": {...}}
sidecar -> ['config.json', 'config.migrations.json']
notice  -> 1
OLD BUILD SCHEMA: accepts migrated config     # main's Config.model_validate, the case that used to raise
second load pin -> None               # idempotent
```

On each finding:

- **The stamp moved out of the config entirely** (`_stamp_path` sidecar) instead of just documenting the revert hazard. That is the stronger fix: nothing this build knows lands in a file an older build validates with `extra="forbid"`, so a revert, a pinned release and a second checkout sharing `~/.raven` all keep booting -- confirmed above against `main`'s schema. It also closes the hole the old design accepted: the watermark is now written even when nothing needed removing, so a hand-set 65536 is never second-guessed even once. `run_stamped` keeping the extension loader (`config/raven.py:1356`) out of the stamped path is right -- it only reads blocks, and the base config comes from `load_config`.
- **Mode is carried over** before the replace, with the `stat` failure swallowed. Matches what `config.paths.restrict_to_owner` asks of a replacing writer.
- **The notice now covers every command** via the `finally` in `commands.run()`, which wraps `app()`, so `status` / `provider list` / `cron list` all announce a migration they performed. `agent` and `gateway` drain first, so there is no double print. The PID-scoped temp name is in too, and `_persist_migrations` returns before stamping when the config write fails, so a read-only home retries instead of recording a migration that never happened.
- The bot's `session.resume` point is handled as well -- resuming first no longer swallows the notices.

Tests: `tests/test_config_loader.py tests/test_config_raven_sections.py tests/test_cli_smoke.py tests/test_cli_helpers.py tests/test_tui_rpc_session_init_bundle.py` -> **146 passed, 0 skipped**; `ruff check raven/ tests/` clean. The new cases assert observable state (on-disk bytes, sidecar contents, `st_mode`, the temp file's name) and the read-only case chmods a real directory rather than mocking the failure.

One nit, comments only, no behaviour: the test section header still says the pin is cleared "under a `configVersion` stamp", and `test_config_without_the_fossil_is_left_byte_identical`'s docstring still argues "the stamp lands only on a file we actually edited" -- both describe the previous design, and the second is now contradicted by the test right below it. Worth a sweep so the next person bumping `CURRENT_CONFIG_VERSION` reads the mechanism that exists.

I checked the symlinked-config case (`os.replace` detaches a `~/.raven/config.json` symlinked into dotfiles, leaving the tracked file behind with the fossil) and am **not** raising it: `config/update.py::_write_atomic` on `main` already writes that way, so it is the repo's existing convention rather than something this diff introduces -- and this writer is the one that improved on it by keeping the mode.

Covered this round: the delta since `aa56f237`, the new sidecar mechanism and its failure paths (unwritable stamp, malformed stamp, race, `--config` siblings), callers of `_migrate_config`, forward compatibility against `main`'s schema, the CLI wiring for every command that loads a config, and whether the new tests actually exercise it.

gloryfromca (approved):

arelchan:
Both second-round findings are accepted; merging this as-is and taking them in the immediate follow-up rather than re-opening an approved PR for them.

**The `--json` blocker is real and reproduced.** Confirmed on `raven doctor --json` with a fossil-carrying config: the notice lands on stdout after the JSON and `jq` fails to parse the result. Introduced by this PR's own `run()` catch-all, and `doctor`'s help does advertise the output as CI-friendly.

The fix is to print the notices on stderr, which is where the rest of this class of message already goes -- `run()`'s own `ConfigReadError` branch uses `Console(stderr=True)`, and the loader's malformed-config warning prints to stderr too. That makes it correct for `import --json` and any future `--json` command without either having to know about it, which is better than allow-listing commands.

Scope note: it misfires only on the single first run after an upgrade, and only for a config that still carries the pin, so the exposure on main is one run per affected automation rather than an ongoing break. Weighed against holding an approved fix for a 64k cap that is degrading every upgraded install today, landing now and correcting the stderr target next is the better trade -- the follow-up PR (`raven doctor --fix`) touches these same two files.

**Stale test section comment**: correct, the wording predates the move to the sidecar. Same follow-up.

Thanks for both rounds -- the watermark placement in particular was the difference between a fix and a footgun.

gloryfromca:
Withdrawing my sign-off on `3156fc3c`. The `--json` finding is real, and it is in the mechanism I explicitly endorsed in my last summary ("the notice now covers every command via the `finally` in `commands.run()`") -- I checked that every command reaches the drain and did not check what the drain writes to. My miss.

Reproduced on this head, fake HOME with a fossil-carrying config:

```
$ raven doctor --json 2>/dev/null > out.json ; jq -e . out.json
jq: PARSE FAILED
$ tail -4 out.json
Config updated: Removed the leftover `contextWindowTokens: 65536` from your
config (an old default, written there by earlier versions). ...
```

The config was migrated by that same run, so the corruption and the migration are the same event -- once per install, on the command whose help says "CI-friendly".

One thing not yet said, and the reason a green suite told nobody: **nothing in the suite drives `run()`.** Every CLI test imports `app` (or a sub-app) and goes through `CliRunner` -- `test_cli_doctor_commands.py`, `test_cli_import_commands.py`, all of them. The `finally` hook at `commands.py:183` is therefore executed by zero tests, so the 194-passing selection could not have caught this and will not catch the next thing that goes in there. Whatever the fix, it wants a test that actually invokes `run()` with a fossil config and asserts stdout stays parseable; otherwise the follow-up lands as untested as this did.

On merge-now-versus-follow-up, my position is that it belongs here. It is one line (`Console(stderr=True)` for the notice) inside a line this PR itself added, the finding is already agreed, and stderr is the right channel on its own merits -- an advisory about a file we rewrote is not part of a command's output. Deferring it means `main` carries a known stdout regression plus an uncovered hook, and costs a second PR to undo, which is more work than the fix.

Your call and the approver's, not mine to make -- but I would rather not be the sign-off that carried it in.

PR #321, opened 2026-08-16
fix(cli): first-run onboarding hints and host-run confirmation

0xKT:
## Summary

First-run experience fixes found in an internal usability review of v0.1.11. install.sh closing hints
now point to `raven onboard` first, matching the README. The model picker
prints a one-line positioning note for the recommended default. The
memory step's skip/non-interactive exits now print the same
configure-later hint the interactive give-up path already has ("Run
raven onboard again whenever you want to configure it"), re-anchored
into onboard_everos.py after the upstream onboarding split. Choosing the
host run location now shows a prompt-injection risk warning with an
explicit confirm (default No); non-interactive host runs print the
warning without blocking.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_cli_onboard_commands.py -x`: 200 passed.
- Full suite `uv run pytest -q`: 1 failed / 6151 passed; the single failure is the pre-existing test_read_file_image failure already present on main.
- Sandbox-HOME real runs: install.sh closing block greps for raven onboard; non-TTY onboard guard unchanged (exit 2 with the re-run hint); the skip-branch hint prints in a real wizard run; host selection was exercised on a real pseudo-TTY (warning + confirm, refuse returns to the menu without writing config).
- All changes were written red-first against frozen acceptance assertions; two assertions were amended with documented reasons after the upstream onboarding restructure (details in the branch worktree NOTES).

- [x] Relevant tests pass locally
- [x] Relevant lint / type checks pass locally
- [ ] User-facing docs or screenshots are updated when needed

## Risk

- [x] Security impact considered
- [x] Backward compatibility considered
- [x] Rollback path is clear for risky changes

The host-run confirm adds one interactive gate on a security-sensitive
default; non-interactive flows are warned but not blocked, so headless
setups keep working. Rollback = revert the branch.

## Related Issues

N/A - found in an internal usability review of v0.1.11; no tracked GitHub issues.

Issue #309, opened 2026-08-13
test(cli): test_bold_accent_renders_styled_not_bare fails only when its file runs as a whole


gloryfromca:
`test_bold_accent_renders_styled_not_bare` passes alone and fails when its file runs as a whole, so any full-suite run shows one red that has nothing to do with the change under test.

## Reproducing

```
uv run pytest tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare   # passes
uv run pytest tests/test_cli_theme.py                                             # fails
uv run pytest tests/                                                              # fails
```

`-p no:randomly` does not change it, so this is not ordering randomisation.

## What actually differs

```
AssertionError: expected bold + accent(#fbe23f) ANSI, got '\x1b[1;93mX\x1b[0m'
assert '1;38;2;251;226;63' in '\x1b[1;93mX\x1b[0m'
```

Worth noting what `93` is: bright yellow from the 8-colour palette. The test constructs its own `Console(color_system="truecolor", force_terminal=True)`, so this is not the console degrading its colour depth -- it is emitting a *named* colour. The accent resolved to something like `yellow` rather than to `#fbe23f`, and Rich rendered the name.

So the leak is in the theme the console was handed, not in terminal detection.

## Where to look

`tests/test_cli_theme.py:21` has an autouse fixture that resets `_theme._cache` and clears `RAVEN_THEME`, `RAVEN_TERM_BACKGROUND`, `COLORFGBG`, `CI`, `SSH_CONNECTION`, `SSH_TTY` before each test, and nulls the cache again afterwards. Whatever survives between tests is therefore not one of those, and not the cache -- some other module-level state in `raven/cli/_theme.py` reached by an earlier test in the same file and never restored.

I did not narrow it to the specific test that leaves it behind.

## Why it is worth fixing rather than tolerating

A permanently-red suite trains everyone to read `1 failed` as normal, which is exactly when a real regression stops being visible. It also costs every contributor the same investigation -- confirming on a clean base that the red is pre-existing before they can trust their own run.

Found while verifying #308; unrelated to it and present on a clean `main`.

Issue #88, opened 2026-07-03
gateway: enabling Feishu from onboard/channels does not install the optional extra and suggests a source-only uv sync command


xiaozuoyou:
## Summary
When Raven is installed via the official installer / release wheel, enabling the `feishu` channel from `raven onboard` or `raven channels enable feishu ...` only writes config. It does **not** install the `channel-feishu` optional dependency (`lark-oapi`), so `raven gateway` later disables the channel and logs:

```text
feishu channel disabled: missing dependency (...). Run: uv sync --extra channel-feishu
```

That remediation is source-checkout-only. End users installed from the one-line installer do not have a local `pyproject.toml` checkout, so they cannot run `uv sync --extra channel-feishu` successfully.

## Steps to reproduce
1. Install Raven via the official installer / release wheel path.
   - `install.sh` remote mode uses `uv tool install --force "$wheel_url"` at `install.sh:175-188`.
   - `install.ps1` remote mode uses `uv tool install --force $wheelUrl` at `install.ps1:229-232`.
2. Run `raven onboard` and select the `feishu` channel, or run `raven channels enable feishu --app-id ... --app-secret ...`.
3. Start the gateway with `raven gateway`.
4. Observe that Feishu is not started and the warning suggests `uv sync --extra channel-feishu`.

## Expected behavior
Selecting / enabling a channel should either:
- install the required extra automatically before marking the channel usable, or
- block/guide the user with a command that works for release-wheel installs (not a source-tree-only `uv sync` command).

At minimum, the remediation shown at gateway startup should match the actual installation mode.

## Actual behavior
- `raven onboard` / `raven channels enable` only enables config fields and never installs the channel dependency.
- `ChannelManager` catches the deferred `ImportError` and disables the channel at runtime.
- The warning points users at `uv sync --extra channel-feishu`, which requires running inside a Raven source checkout with `pyproject.toml`.

## Root cause analysis
- `pyproject.toml:47-64` declares Feishu support as an optional extra: `channel-feishu = ["lark-oapi>=1.5.0,<2.0.0"]`.
- The official installers do a base `uv tool install` in release mode and do not install any `channel-*` extras:
  - `install.sh:175-188`
  - `install.ps1:229-232`
- Channel onboarding/enabling only writes config and returns success; it never verifies or installs the adapter dependency:
  - `raven/cli/onboard_commands.py:1579-1587`
  - `raven/cli/onboard_commands.py:1776-1789`
[...]

PR #342, opened 2026-08-17
fix(*): first-run flow, channel install hints, and default-model pricing

0xKT:
## Summary

Follow-ups found while running a real-machine regression pass over the fixes in #321-#334. Five independent defects, each verified by reproducing it and then reproducing the fix on this machine rather than only in tests.

**First run is one command again.** The installer's closing block told a new user to run `raven onboard`, then `raven`, then `raven agent -m hello`. Bare `raven` already runs the wizard through the startup gate and continues into the TUI in the same process, so the list presented one continuous flow as three commands -- and the third one cannot work on a fresh install, because `raven agent` never consults the gate. Both installers now branch on an existing `config.json`: a first run is sent to bare `raven`; a re-run over an existing config is told its config is untouched and pointed at `raven upgrade` for the in-place path. `install.ps1` had never carried the onboard line at all, so the two scripts had disagreed since they were written; they now say the same thing. README (both languages) opens with bare `raven`, keeping `raven onboard` in the command tables as the explicit reconfigure entry. On the gate path the wizard also stops printing its "Get started" command list, which the TUI answers a second later by taking the terminal.

**A zero-provider install no longer names a vendor the user never chose.** With nothing configured, routing falls back to the schema default model, and its vendor was reported as the thing to go fix -- sending someone who only holds an OpenRouter key to `provider set anthropic`. The first-run verdict now requires both halves (the model is still the schema default, and no provider anywhere holds credentials), so a user who chose this model, or who has another provider working, keeps the specific message; for the OAuth families that message still names a sign-in rather than a key.

**`uv sync --extra channel-telegram` was uninstalling other channels.** `uv sync` is an exact sync, so the dev hint for one missing channel SDK removed every other channel's SDK on the way in. It now names the umbrella `channels` extra with `--inexact`, matching what the installer puts on an end user's machine. Separately, an enabled channel whose SDK is absent was visible only as a single WARNING line in the gateway log; `channels status` and `doctor` now report it through the same probe the gateway performs.

**The recommended default model could not be priced.** Raven routes Anthropic's `claude-sonnet-4-5`; OpenRouter files the same model as `claude-sonnet-4.5`. Exact-key lookup missed, so the model the wizard itself recommends reported a cost of None on every turn, with an unknown-model warning beside it. The catalog lookup now tries dotted spellings after the exact key misses, one digit boundary at a time so `llama-3-3-70b` resolves as `llama-3.3-70b` rather than `llama-3.3.70b`. It is a fallback only, so a wrong guess degrades to the same None it replaces.

**Two smaller items.** Hub skill ids carry slashes (`openclaw/skills/tag-memory`) and were interpolated into the request URL as extra path segments; they are now escaped into one segment. And the `--every` scheduling contract is written down: it measures from the moment the previous fire completed, so a job that takes 15s repeats every interval + 15s and its clock drifts by design, which is what buys the guarantee that a slow run never overlaps itself or leaves a backlog. That was never stated anywhere, so the drift read as a defect; CONTEXT.md now carries the term and `cron add --every` states it where the schedule is chosen.

Two decisions were taken deliberately and are **not** changed here: `skillForge.autoInstall` stays `auto` (the disclosure line, the two install-path audit records, `skill remove` and `skill block` together already give the user visibility and control), and cron stays fixed-delay (documented rather than converted to fixed-rate, which would need a catch-up policy and would interact with the anti-runaway counter from #333).

Out of scope, reported rather than fixed: the published `tag-memory` bundle on the Hub still ships the upstream author's runtime data at `data/pending_summary.json`, including personal content. That is server-side; the client-side install policy from #327 already refuses the skill over its `~/.openclaw` references, which is what keeps it off users' machines today.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Commands run on macOS against this branch:

- `uv run pytest tests/ -q` -> 6080 passed, 43 skipped, 2 failed locally: `test_cli_theme.py::test_bold_accent_renders_styled_not_bare` (terminal color-depth detection, tracked as #309) and `test_read_file_image.py::test_image_block_is_the_only_place_the_shape_is_written`. Both reproduce on unmodified `main` on this machine, and both pass in CI on this branch, so they are artifacts of the local terminal environment rather than anything this branch does.
- `uv run ruff check raven/ tests/` -> All checks passed
- `uv run ruff format --check raven/ tests/` -> 821 files already formatted
- `sh -n install.sh` -> syntax OK. `install.ps1` was not parse-checked (no PowerShell on this machine); its closing block is covered by a text assertion in `tests/test_cli_onboard_commands.py`, and its braces balance.
- `make check-large-files` -> no findings

Behaviour reproduced by hand, each against an isolated `HOME`:

- Installer closing block, driven with the heavy steps stubbed: no `config.json` prints the bare-`raven` first-run hint; an existing `config.json` prints `Raven updated ... raven upgrade`.
- `raven agent -m hi` with nothing configured -> `Error: no provider is configured yet -- run `raven onboard` for guided setup.`, exit code 1.
- `raven agent -m "say ok"` with a working provider -> `5k in / 4 out tokens - $0.0184 - 4.0s` (the cost figure was absent before this branch).
- `raven channels status` and `raven doctor` with telegram enabled and its SDK absent -> both report the missing SDK plus `Run: uv sync --inexact --extra channels`; with no channels enabled, neither prints anything extra.
- The wizard driven through a pseudo-terminal to step 4, confirming the model picker's default line, the host-run `(y/N)` confirmation, and the memory-embedding prompt.

Regression checks over the areas this touches, on the same machine: a cron job under a real `raven gateway` fired four times at its interval; the Skill Hub install policy refused a blocklisted skill at runtime (verified against a control run where the same prompt installed it); `autoInstall` `off` and `prompt` both installed nothing on a non-interactive run.

## Risk

User-visible changes:

- The installer prints a different closing block, and a re-run over an existing config now says "Raven updated" instead of first-time-setup wording.
- A first run with nothing configured gets a different error sentence from `raven agent`. Any script matching on the old `anthropic needs an API key` text for the zero-config case would need updating; the per-provider message is unchanged for every configured case.
- The wizard prints no "Get started" panel when it was started by the gate. Running `raven onboard` explicitly still prints it.
- Turn-level cost figures appear for OpenRouter models whose ids are hyphenated where the catalog is dotted. This changes reported numbers from absent to present; no billing behaviour changes.

Rollback: each concern is a separate commit, so any one can be reverted alone. Nothing here writes to `~/.raven`, changes a config default, or alters an on-disk format, so a revert needs no migration.

Security: `--inexact` only stops `uv sync` from removing packages, and the umbrella extra is the same set the installer already installs. Escaping the Hub skill id narrows what a hub-supplied id can address in a request URL. The install-policy and blocklist behaviour from #327 and #328 is unchanged; this branch only verifies it.

## Related Issues

Nothing on the tracker covers these five defects; they surfaced while regression-testing #321-#334, so this PR closes no issue.

Referenced for context only:

- #88 is where the missing-dependency hint got its editable-vs-wheel split. This PR refines the editable branch it added; the wheel branch is unchanged.
- #309 tracks `test_bold_accent_renders_styled_not_bare`, one of the two suite failures listed under Verification. The other, `test_read_file_image.py::test_image_block_is_the_only_place_the_shape_is_written`, has no issue open against it; both fail the same way on unmodified `main`.

### DIFF
--- a/raven/cli/_helpers.py
+++ b/raven/cli/_helpers.py
@@ -76,10 +76,15 @@ def check_provider_credentials(config: Config) -> None:
     # so answer the wizard instead. Both halves are required -- a user who picked
     # this model, or who has some other provider working, gets the specific
     # verdict, which for the OAuth families names a sign-in rather than a key.
+    # Names come from the declared fields *and* the extras: an undeclared
+    # provider key is a supported shape, and `ProvidersConfig.get` is the only
+    # place allowed to resolve either kind, so route both through it rather than
+    # reading `__dict__` -- which sees no extras and would call a user whose one
+    # working credential lives there unconfigured.
     chose_a_model = config.agents.defaults.model != type(config.agents.defaults)().model
+    configured = (*config.providers.__dict__, *(config.providers.model_extra or {}))
     if not chose_a_model and not any(
-        credential_status(name, section, include_external=True).ok
-        for name, section in config.providers.__dict__.items()
+        credential_status(name, config.providers.get(name), include_external=True).ok for name in configured
     ):
         raise MissingCredentialsError(
             "no provider is configured yet -- run `raven onboard` for guided setup",
--- a/raven/cli/_helpers.py
+++ b/raven/cli/_helpers.py
@@ -66,12 +66,31 @@ def check_provider_credentials(config: Config) -> None:
         )
 
     status = credential_status(provider_name, config.providers.get(provider_name), include_external=True)
-    if not status.ok:
+    if status.ok:
+        return
+
+    # A first run fails this check while naming a provider the user never chose:
+    # with nothing configured, routing falls back to the schema's default model,
+    # whose vendor then gets reported as the thing to go fix. Sending someone who
+    # only has an OpenRouter key to `provider set anthropic` is the wrong errand,
+    # so answer the wizard instead. Both halves are required -- a user who picked
+    # this model, or who has some other provider working, gets the specific
+    # verdict, which for the OAuth families names a sign-in rather than a key.
+    chose_a_model = config.agents.defaults.model != type(config.agents.defaults)().model
+    if not chose_a_model and not any(
+        credential_status(name, section, include_external=True).ok
+        for name, section in config.providers.__dict__.items()
+    ):
         raise MissingCredentialsError(
-            status.summary,
-            provider=provider_name,
-            remedy="Run `raven onboard` for guided setup.",
+            "no provider is configured yet -- run `raven onboard` for guided setup",
+            remedy="Already have a key? raven provider set <name> --api-key <key>",
         )
+
+    raise MissingCredentialsError(
+        status.summary,
+        provider=provider_name,
+        remedy="Run `raven onboard` for guided setup.",
+    )
 
 
 def make_provider(config: Config):
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -156,6 +156,15 @@ _Avoid_: reintroducing fire-time channel selection (the retired
 channel value is retired with the REPL; stored `cli`-bound jobs migrate to
 `tui` at load time.
 
+**Fixed-delay interval**:
+The scheduling contract for `--every` jobs: the next run is computed from the
+moment the previous fire **completed**, not from the moment it was due. A job
+that takes 15s to run therefore repeats every `interval + 15s`, and its clock
+drifts by design — the property being bought is that a slow run can never
+overlap itself or leave a backlog to catch up on.
+_Avoid_: calling this fixed-rate, or reading `--every 2m` as a promise to fire
+on the two-minute mark; calendar-anchored schedules are what `--cron` is for.
+
 **Predictor**:
 The Sentinel pipeline stage that turns signals into predicted user needs (the
 proactive side of prediction).
--- a/README.md
+++ b/README.md
@@ -64,9 +64,10 @@ irm https://raw.githubusercontent.com/EverMind-AI/Raven/refs/heads/main/install.
 ### Onboard and run
 
 ```bash
-raven onboard
 raven
 ```
+
+That is the whole first run: with nothing configured yet, `raven` walks you through setup and then opens the TUI in the same session. To reconfigure later, run `raven onboard` explicitly.
 
 The bilingual onboarding wizard configures six areas without requiring manual edits to `~/.raven/config.json`:
 
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -67,9 +67,10 @@ irm https://raw.githubusercontent.com/EverMind-AI/Raven/refs/heads/main/install.
 ### 完成引导并运行
 
 ```bash
-raven onboard
 raven
 ```
+
+首次运行只需这一条：尚未配置时，`raven` 会先带你走完引导，然后在同一次会话里直接进入 TUI。之后想重新配置，再显式运行 `raven onboard`。
 
 双语 onboarding 向导会配置六个方面，无需手动编辑 `~/.raven/config.json`：
 
--- a/install.ps1
+++ b/install.ps1
@@ -317,6 +317,10 @@ function Install-Raven([string]$UvPath, [string]$NodePath) {
 }
 
 function Main {
+    # Read before installing so the closing hint can tell a first run from an
+    # upgrade; the install itself never writes config.json (the wizard does).
+    $hadConfig = Test-Path (Join-Path $RavenHome "config.json")
+
     $uv = Ensure-Uv
     $node = Ensure-Node
     Install-Raven $uv $node
@@ -325,11 +329,19 @@ function Main {
     Add-ProcessPath $toolBin
 
     Write-Host ""
-    Write-Ok "All set. Open a new PowerShell window, or continue in this one, then run:"
-    Write-Host ""
-    Write-Host "    raven            # enter the TUI"
-    Write-Host "    raven agent -m `"hello`""
-    Write-Host ""
+    if ($hadConfig) {
+        Write-Ok "Raven updated. Your config in $RavenHome is unchanged."
+        Write-Host ""
+        Write-Host "    raven    # continue where you left off"
+        Write-Host ""
+        Write-Host "  tip: next time you can upgrade in place with 'raven upgrade'"
+        Write-Host ""
+    } else {
+        Write-Ok "All set. Open a new PowerShell window, or continue in this one, then run:"
+        Write-Host ""
+        Write-Host "    raven    # sets you up on first run, then opens the TUI"
+        Write-Host ""
+    }
     if (($env:PATH -split ';') -notcontains $toolBin) {
         Write-Warn "Current PATH does not include $toolBin. Restart PowerShell if 'raven' is not found."
     }
--- a/install.sh
+++ b/install.sh
@@ -259,16 +259,27 @@ install_raven() {
 # --- main ------------------------------------------------------------------
 main() {
   have curl || die "curl is required; please install it first"
+  # Read before installing so the closing hint can tell a first run from an
+  # upgrade; the install itself never writes config.json (the wizard does).
+  if [ -f "$RAVEN_HOME/config.json" ]; then
+    had_config=1
+  else
+    had_config=0
+  fi
   detect_platform
   ensure_uv
   ensure_node
   install_raven
 
   printf '\n'
-  ok "All set! Open a new terminal (or source your shell profile), then run:"
-  printf '\n    \033[1mraven onboard\033[0m    # first-time setup\n'
-  printf '    \033[1mraven\033[0m            # enter the TUI\n'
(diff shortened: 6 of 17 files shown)


raven/cli/_helpers.py
```diff
@@ -66,13 +66,32 @@ def check_provider_credentials(config: Config) -> None:
...
+        return
+
+    # A first run fails this check while naming a provider the user never chose:
+    # with nothing configured, routing falls back to the schema's default model,
+    # whose vendor then gets reported as the thing to go fix. Sending someone who
+    # only has an OpenRouter key to `provider set anthropic` is the wrong errand,
+    # so answer the wizard instead. Both halves are required -- a user who picked
+    # this model, or who has some other provider working, gets the specific
+    # verdict, which for the OAuth families names a sign-in rather than a key.
+    chose_a_model = config.agents.defaults.model != type(config.agents.defaults)().model
```
gloryfromca, line 84: This branch is the PR's second headline fix and it is the only one of the five with no test.

The closest existing test, `test_no_api_key_error_mentions_onboard` in `tests/test_cli_helpers.py`, writes `{}` as the config -- so it does route through exactly this new branch -- but it only asserts `"raven onboard" in f"{summary} {remedy}"`. The old code carried that string in `remedy` and the new code carries it in `summary`, so the test passes identically before and after. Nothing currently pins the behaviour change.

Concretely: drop the `not any(...)` half of the condition and every test in the suite still passes, while a user holding a working OpenRouter key and the untouched schema default model is sent back to `provider set anthropic` -- the precise bug this commit exists to fix. The reverse direction is equally unguarded: widen it and a configured-but-keyless provider loses its specific `<name> needs <field>` verdict.

Two cases in `tests/test_cli_helpers.py` would cover it -- empty config asserts the summary names no vendor, and a config with a credentialed second provider plus the default model asserts the per-provider summary survives.

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. One nonblocking test gap noted inline.

I verified the five concerns against the code and on this machine rather than taking the description on trust:

- **Fixed-delay contract is accurate.** `_execute_job` writes `next_run_at_ms = _compute_next_run(schedule, self._now_ms())` *after* the job body has run (`service.py:566`), and the `every` branch returns `now_ms + every_ms` -- so next-run really is measured from completion. The CONTEXT.md entry and the `--every` help text describe what the scheduler does.
- **The umbrella extra exists and the hint is runnable.** `channels` in `pyproject.toml` fans out to all twelve `channel-*` extras; `uv sync --inexact --extra channels --dry-run` resolves cleanly and reports no removals. `--inexact` is load-bearing beyond channels too (a plain `--extra channels` would evict `sandbox`/`tools` from a dev env).
- **The read-only probe is safe to run from `status`/`doctor`.** I read every adapter's `__init__` plus `ChannelBase`/`Intake`: all are pure attribute assignment, no sockets, no files, no bot construction, and `asyncio.Event()` outside a loop is fine on supported Pythons. Nothing raises `SystemExit` from a constructor, so the `except Exception: continue` really does only swallow diagnosis-is-elsewhere failures. Ran both commands with telegram+matrix enabled and their SDKs absent -- table cell, footer, and the doctor line all render as described.
- **The two-half credential verdict holds up.** With an empty `Config()` no declared provider reads as credentialed, so the new message fires; with `OPENAI_API_KEY` ambient the `any(...)` is satisfied and the specific per-provider verdict is kept. The heuristic is also safe in practice: no wizard path can write the schema default, because anthropic's `default_model` is `claude-sonnet-5`, not `claude-opus-4-5`. `provider=` being unset on the new error is harmless -- it defaults to `""`, and neither the CLI renderer nor the TUI reads it for anything but a passthrough data field.
- **Escaping the Hub id is a no-op on the live path, not a wire regression.** `quote(..., safe="")` leaves the slash-free UUID that `hub_source` puts in `qualified_id` untouched (hyphens are always-safe), and httpx keeps `%2F` on `raw_path` without double-encoding. So the only requests that change shape are the ones that were already addressing the wrong route.
- **Dotted-variant lookup preserves the old precedence.** Exact full key, then exact `/`-stripped key, only then the dotted forms -- and the catalogue is fetched once outside the loop. I could not construct an id where a one-boundary dot lands on a genuinely different model, so the "degrades to the same None" claim stands.
- **Repo rules:** commits are Conventional and ASCII, tests went into the existing `test_*` files rather than new phase-suffixed ones, the new domain term is defined in CONTEXT.md, no assets added.
- **Tests were not weakened to go green.** `test_install_sh_all_set_mentions_onboard` was replaced because its premise inverted, and the replacements assert more than it did (both installers, both branches, plus README parity). `uv run pytest tests/test_channels_manager.py tests/test_provider_rates.py tests/test_cli_onboard_commands.py tests/test_cli_doctor_commands.py tests/test_cli_channel_commands.py tests/test_skill_hub_client.py -q` -> 358 passed, 0 skipped.

Two smaller gaps I am not going to argue about, alongside the inline one: `_id_segment` has no test pinning that a slash id becomes one segment, and nothing asserts the new `channels status` footer / `doctor` line or the added `channels_missing_deps` JSON field -- those are thin renderings over a probe that is tested.

gloryfromca (dismissed):

raven/cli/_helpers.py
gloryfromca: Resolved by 859c4fca. `test_a_working_extra_provider_is_not_a_first_run` and `test_nothing_configured_at_all_names_the_wizard` pin both directions, and I confirmed they are not vacuous: the first fails under a one-half condition (it asserts the summary is the vendor-specific `API key` one, not the first-run sentence), the second fails if the branch is removed. `uv run pytest tests/test_cli_helpers.py -q` -> 28 passed.

gloryfromca (commented):
No blockers; this revision can merge as far as I am concerned. My earlier finding is resolved and I have nothing new.

`859c4fca` is the only delta since I reviewed `21ffecf9`. I re-read just that commit rather than redoing the review:

- **The `get()` switch is a strict superset of the old walk, not a swap.** The risk worth checking was that routing declared field names through `ProvidersConfig.get` could stop resolving one of them and silently drop a configured provider from the `any(...)` -- which would reintroduce the exact bug the branch fixes. It does not: for all 21 declared fields `config.providers.get(name) is getattr(config.providers, name)`, because `canonical_provider_name` is identity on every field spelling. So declared providers behave identically and extras now count as well.
- **A malformed extra value does not crash the gate.** `get()` returns `None` for a non-dict extra, and the new code hands that straight to `credential_status`. `{"providers": {"junk": 5}}`, `"sk-x"` and `[]` all degrade to not-ok and still raise the ordinary `MissingCredentialsError` -- no `AttributeError` escaping the check.
- **Behaviour reproduced:** `{"providers": {"my-vllm": {"apiKey": ..., "apiBase": ...}}}` with the model left at the schema default now yields `anthropic needs an API key -- run \`raven provider set anthropic --api-key <key>\`` instead of the first-run sentence; an empty `providers` still yields the wizard sentence.
- **The two new tests close the gap I raised** and are not vacuous in either direction (detail in the thread).
- Commit message is Conventional, all-ASCII, correctly scoped `(cli)`, and the tests went into the existing `tests/test_cli_helpers.py` per the naming rule.

`uv run pytest tests/test_cli_helpers.py tests/test_cli_onboard_commands.py tests/test_provider_rates.py tests/test_channels_manager.py -q` -> 299 passed, 0 skipped.

gloryfromca (dismissed):

gloryfromca (commented):
No blockers; this revision can merge as far as I am concerned.

`3944f4a0` is tests-only and is the last gap on this PR that anyone had named. I checked the one thing that matters for a test-only commit -- whether the tests are load-bearing -- by mutating the source rather than reading the assertions:

- Dropping `show_next_steps=False` from the gate call makes `test_the_gate_starts_the_wizard_without_its_outro` fail at the assert (`[None] != [False]`), not error out.
- Deleting the early-return block in `_print_next_steps` makes `test_the_outro_flag_swaps_the_panel_for_one_line` fail. It also pins the half that matters for the user: the suppressed path still prints the completion line and the recap, and drops only the command table.

The second test deliberately takes no `tmp_env`, so it renders against whatever config the machine has. I checked that this cannot make it flaky -- run standalone against a deliberately unparseable `~/.raven/config.json` it still passes, because the recap reads defensively and the assertions are on the two panel strings only.

Conventions: `test(cli)` is a declared type, scope is a real subpackage, subject is 49 chars and ASCII, and the cases went into the existing `tests/test_cli_onboard_commands.py` per the naming rule rather than a new file.

`uv run pytest tests/test_cli_onboard_commands.py tests/test_cli_helpers.py tests/test_channels_manager.py tests/test_provider_rates.py tests/test_cli_doctor_commands.py tests/test_cli_channel_commands.py -q` -> 381 passed, 0 skipped. Working tree restored clean after the mutations.

gloryfromca (approved):

PR #310, opened 2026-08-13
fix(*): make everos ownership explicit and stop memory failures from stalling a session

gloryfromca:
## Summary

EverOS memory used to fail in ways that were hard to attribute: a session
would stall for up to a minute, or quietly come up with no memory at all, and
the reason lived in a log file. Underneath, raven could not answer two
questions it kept acting on -- which EverOS root the memories are in, and
whether it is allowed to touch that root's process.

**Ownership becomes a recorded decision.** The root and whether raven may write
to it are stored in `plugins.config["everos-memory"]` instead of being
re-derived at each call site from a path. A root the user runs is read-only:
raven records the address, never edits the config, never starts or stops it.
Enforcement sits on the write primitives, so a new caller cannot opt out.

**Nothing claims to know what it cannot.** A self-managed install records no
root, which is what makes the read-only promise structural. Everything that
reports on memory had to stop asking the filesystem about it: `raven doctor`
names no directory for such an install and takes the retrieval mode from what
the server says about itself, because the fallback root it used to print was
one the user had never chosen, and the roles read out of that root's toml
described an install nobody was using.

**Two ways to get memory, chosen rather than inferred.** Step 4 asks. The
managed path is unchanged in spirit: raven creates a root, configures the
models, runs the server. The self-managed path is two prompts and one health
probe, with no scan and no port default -- the only port raven could suggest is
its own, and accepting it by reflex would point this path at the other path's
address. Only the address is recorded, not a root, so there is no path on disk
a later change could write to by accident.

**The port is a setting, not a constant.** 18791 is a recommendation. When it
is already taken the wizard asks for another instead of failing with nowhere to
go, and stays silent otherwise -- a port screen on every fresh install asks for
a decision almost nobody has to make. A port already held by raven's own
service does not count as taken: a bind test cannot tell a stranger from the
process the step is about to restart into.

**Reconfiguring reaches the process that reads the models.** EverOS builds its
LLM client in the API lifespan, at startup, so a server already running keeps
what it booted with. Writing new models and leaving it running made the whole
reconfigure inert until some later, unrelated restart, with nothing on screen
saying so. Raven's own server for that root is stopped first -- identified from
the lock rather than the pidfile, so a root whose pidfile was lost is not
reported as somebody else's.

**Convergence asks before it acts.** The target address comes from a recorded
port rather than a constant, because comparing against a hardcoded 18791 cannot
tell a port an old raven left behind from one the user set deliberately -- and
[...]

PR #343, opened 2026-08-17
refactor(*): ask who runs everos once and drop discovery-time ownership

gloryfromca:
## Summary

Follow-up to #310 that grew into removing the thing both of its leftover findings
came from. Ownership of an EverOS root was treated as something the wizard could
observe: discovery reported it per candidate, and five nested menus each
re-derived what raven was allowed to do from whatever happened to be on disk.
Every fix in this area has landed on the least-used of those states, and the two
that opened this branch were the same shape again -- a menu that did not consult
ownership, and a leftover root adopted before any menu appeared.

It is a decision, not an observation. Step 4 now asks one question -- raven runs
EverOS, the user runs it, or neither today -- and ownership follows from the
answer, so nothing downstream reads it back off a directory. The managed lane
takes over whatever memory directory it finds, whatever an earlier run recorded
there, and builds its own when there is none. `RootState.owned` is gone.

The found-root question also moved in front of the work it authorises.
Convergence used to stop and restart the service before asking, so an install
that only wanted confirming was moved to another port unprompted. "Use it as it
is" now touches nothing and keeps the address the service already answers on;
relocating is what "reconfigure" means. A refused self-managed address returns to
the lane question instead of ending the step, and skipping is an answer of its
own rather than something reached by backing out of a lane.

Two states have no way forward and are reported rather than worked around: a
directory held by something that serves no HTTP (pid and command line), and a
start that failed (the reason EverOS gave). One memory directory admits one
engine, so there is nothing to offer there.

Also kept from the original branch: `_proc_locks_pid`'s companion port lookup no
longer depends on lsof. `LockHolder.port` falls back to `/proc/net/tcp` matched
against the process's own socket inodes, and the `/proc/locks` branch is asked
first because it can tell a holder from a blocked waiter, which `lsof -t` cannot.
That lookup is now what decides whether a busy directory can be taken over at
all.

Removed: `_enabled_unowned_menu`, `_reuse_unowned_root`, `_converge_owned_root`,
`_adopt_or_move`, `_adopt_running_address`, `_stop_failure_line`, `_same_address`,
`RootState.owned`. `_record_root` and `_adopt_own_root` collapse into
`_adopt_root`. Net about 480 lines lighter. The runtime is untouched: backend,
tools and doctor read the recorded `owned` field, whose meaning is unchanged
(raven may write and start it, or connect only).

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [x] Refactor
- [ ] Other

## Verification

```
uv run pytest tests/ -q
1 failed, 6643 passed, 33 skipped, 13 deselected in 122.58s

uv run pytest tests/test_cli_onboard_commands.py tests/test_everos_discover.py -q
267 passed in 3.18s

uv run ruff check raven/ tests/            # All checks passed
uv run ruff format --check raven/ tests/   # 830 files already formatted
```

The single failure is `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`.
It reproduces with every change on this branch stashed, when that file runs as a
whole rather than as a single node -- ordering pollution inside it, macOS-local,
green in CI on Linux. This branch introduces no failures.

Two test-isolation faults surfaced while rewriting and are fixed in the memory
fixture: a real bind test and real discovery decided which branch a test took, so
a machine already running everos on 18791 met an unscripted prompt and died on
EOF, and every managed-lane test probed the developer's own /health. The onboard
file now runs in 3s instead of 100s.

## Risk

- Step 4 opens with one question and three answers (managed / self-managed /
  skip). The separate keep-or-reconfigure screens for an already-configured
  install are gone; the same three answers cover them.
- Taking over a found root asks before acting. "Use it as it is" no longer stops,
  moves or reconfigures the service, which also means a pre-upgrade port survives
  a reuse -- moving to the standard port is what "reconfigure" does.
- A root recorded as the user's is taken over by the managed lane instead of being
  offered as read-only reuse. Read-only use of an EverOS someone else runs is the
  self-managed lane, which records an address and no root.
- A refused self-managed address no longer turns memory off; it re-asks.

Not verified here: the Windows branch (unchanged, still skips and disables), and
the report path for a real `everos demo` holding the lock (covered with a stubbed
lock holder, not against a live process).

Removing the recorded-root-owned-false state means raven no longer reads a
user-managed root's `everos.toml`, so capability reporting for that setup relies
on `/health` alone, and `make_understand_media_tool` still exports `EVEROS_ROOT`
from the fallback root there -- a finding from #310's review that this branch
does not address either.

Rollback: revert the branch. No config keys are added, removed or renamed; the
`owned` field keeps its meaning and its readers.

## Related Issues

Follow-up to #310. Still open from its review, not addressed here: doctor's
retrieval line cannot tell "server down" from "server too old to report"; the
win32 branch leaves the service state at `UNKNOWN` so every write counts as
dropped; `drain_backend_stores` neither cancels nor counts writes that outlive
its budget before `stop()` reads the total; and `MemoryBackend`'s contract test
gained no assertion for `store() -> bool`, so a backend returning `None` still
passes it.

### DIFF
--- a/raven/cli/onboard_everos.py
+++ b/raven/cli/onboard_everos.py
@@ -1314,31 +1314,6 @@ def _config_everos_role(
         return
 
 
-_LOOPBACK_NAMES = frozenset({"localhost", "127.0.0.1", "::1", "[::1]"})
-
-
-def _same_address(a: str | None, b: str | None) -> bool:
-    """Whether two URLs name the same socket, allowing for spelling.
-
-    The declared address is copied out of the root's toml and the target is
-    built from a recorded port, so the same endpoint routinely arrives as
-    ``127.0.0.1`` on one side and ``localhost`` on the other. Comparing the
-    strings makes a service already on its configured port look like drift, and
-    the wizard then offers to move it onto itself.
-    """
-    from urllib.parse import urlparse
-
-    if not a or not b:
-        return False
-    pa, pb = urlparse(a), urlparse(b)
-    if pa.port != pb.port:
-        return False
-    ha, hb = (pa.hostname or "").lower(), (pb.hostname or "").lower()
-    if ha == hb:
-        return True
-    return ha in _LOOPBACK_NAMES and hb in _LOOPBACK_NAMES
-
-
 def _lock_holder(root: Path | str):
     """The process serving ``root``, or None. Indirected so callers can stub it."""
     from raven.plugin.memory.everos._server import lock_holder
@@ -1486,69 +1461,6 @@ def _retry_or_skip_address() -> str:
     return str(choice)
 
 
-_SWITCH_TO_MANAGED = object()
-
-
-def _enabled_unowned_menu() -> object:
-    """Keep the server you run, move it, or hand the job to raven."""
-    questionary = oc._require_questionary()
-    from raven.cli._styles import RAVEN_STYLE
-
-    slice_ = _recorded_memory_slice()
-    where = slice_.get("base_url") or "?"
-    oc.console.print()
-    oc.console.print(
-        oc._t(
-            f"  [green]v Long-term memory uses the EverOS you run at {where}.[/green]",
-            f"  [green]✓ 长期记忆正在使用你自己运行的 EverOS：{where}。[/green]",
-        ),
-        highlight=False,
-    )
-    action = questionary.select(
-        oc._t("What would you like to do?", "想做什么？"),
-        choices=[
-            questionary.Choice(oc._t("Keep it", "保持不变"), value="keep"),
-            questionary.Choice(oc._t("Point Raven at a different address", "改成别的地址"), value="address"),
-            # The way out. Without it this menu is a one-way door: the managed
-            # path is unreachable and the only exit is editing config.json by
-            # hand. Offered as its own answer rather than as a side effect of
-            # "reconfigure", which is how ownership used to flip unasked.
-            questionary.Choice(
-                oc._t("Let Raven run its own EverOS instead", "改用 Raven 自己运行的 EverOS"),
-                value="managed",
-            ),
-        ],
-        style=RAVEN_STYLE,
-        qmark=oc._QMARK,
-    ).ask()
-    if action is None:
-        raise typer.Exit(1)
-    if action == "managed":
-        # Say what it costs before taking it: their server keeps its data and
-        # keeps running, but raven stops reading it, so the memories behind that
-        # address stop being the ones it recalls.
-        oc.console.print(
-            oc._t(
-                f"  [yellow]! Raven will build its own memory and stop using {where}.[/yellow]\n"
-                "  [dim]That server and its data are untouched -- Raven simply stops reading them.[/dim]",
-                f"  [yellow]⚠ Raven 将建立自己的记忆，不再使用 {where}。[/yellow]\n"
-                "  [dim]那台服务器和它的数据不受影响，只是 Raven 不再读它。[/dim]",
-            ),
-            highlight=False,
-        )
-        return _SWITCH_TO_MANAGED
-    if action == "address" and not _use_self_managed_everos():
-        # A refused address leaves the working one in place rather than
-        # discarding a setup that was fine a moment ago.
-        oc.console.print(
-            oc._t(
-                f"  [dim]Keeping {where}.[/dim]",
-                f"  [dim]继续使用 {where}。[/dim]",
-            )
-        )
-    return None
-
-
 def _use_self_managed_everos() -> bool:
     """Point raven at an EverOS the user runs. Returns False if it is unreachable.
 
@@ -1621,41 +1533,6 @@ def _use_self_managed_everos() -> bool:
     return True
 
 
-def _record_root(root: Any, owned: bool) -> None:
-    """Record which EverOS root is in use and whether raven may write to it.
-
-    Both are decisions rather than derivable facts, and every later reader --
-    the wizard, the runtime, doctor -- consults the record instead of guessing
-    again. Recording ``owned`` is what makes read-only reuse enforceable.
-    """
-    from raven.config.update import set_plugin_config_fields
-
-    set_plugin_config_fields("everos-memory", {"root": str(root), "owned": owned})
-
-
-def _adopt_own_root() -> None:
-    """Move to raven's own root, retracting an address that was not raven's.
-
-    A merging write cannot say "this address no longer applies", and the address
-    recorded for a server the user runs is exactly what ``_ask_managed_port``
-    falls back to when no port is recorded -- so leaving it behind parks raven's
-    own service on the user's port, one screen after promising to stop using it.
-
-    Only retracted when the record says the address is theirs. A managed port
-    the user deliberately moved to is also reachable here, and dropping that one
-    would quietly offer the shipped default again on the next run.
-    """
-    from raven.config.update import set_plugin_config_fields
-    from raven.config.update_everos import default_everos_root
-
-    theirs = _recorded_memory_slice().get("owned") is False
-    set_plugin_config_fields(
-        "everos-memory",
-        {"root": str(default_everos_root()), "owned": True},
-        remove=("base_url", "port") if theirs else (),
-    )
-
-
 def _capability_lines(base_url: str) -> list[str]:
     """One line per capability the running server actually built.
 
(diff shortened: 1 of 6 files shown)


gloryfromca (commented):
Blocking on one thing: the self-managed -> managed handover keeps the user's address in the config, so raven's own server ends up configured on the user's port -- inline at `raven/cli/onboard_everos.py:2114`. Everything else here is nonblocking, and the shape of the change is right.

Reviewed as the author, so I traced the new paths rather than trusting the reasoning in the comments.

**What I covered:** the diff; the callers and surrounding flow (`_step4_memory` ordering, `_ask_managed_port`, `_configured_target_url`, `_converge_owned_root`, `lock_holder`); whether the changed/added tests assert behaviour; backward compatibility of recorded-config shapes; AGENTS.md rules. Not covered: anything on a real Linux host -- the /proc code was read, not run.

**The two fixes themselves hold up.** Gating ahead of discovery is the right place: with a `owned: false` slice the leftover-root test never even reaches `pick()`, which is the point. On the server side the `/proc/net/tcp` field indices check out against a real row (inode at index 9, port as hex in field 1, `0A` for LISTEN); `tcp6`'s header line survives the concatenation but the `0A` test drops it. The `/proc/locks`-first reorder is a genuine improvement -- `lsof -t` does list waiters alongside the holder -- and it is a no-op on darwin, where `/proc/locks` is absent.

**Test state, honestly:**
* `uv run pytest tests/test_everos_server.py` -> 61 passed.
* `uv run pytest tests/test_cli_onboard_commands.py` -> **11 failed, 257 total on this host, and none of it is this PR.** A real EverOS is listening on 18791 on this machine, so `_port_is_free` returns false and `_ask_managed_port` injects a port prompt into 11 tests that never stubbed one. Neutralizing that single host fact (autouse `_port_is_free -> True`) turns the file green: 257 passed. That host-dependence is pre-existing and on untouched lines, so it is not counted against this branch -- but it is worth its own fix, since any developer running everos locally sees 11 red tests.

**Two small notes, neither a blocker:**
1. Both commits carry `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>`, while AGENTS.md 3.3 and every co-authored commit on `main` use `Co-authored-by: Claude (claude-opus-5[1m]) <noreply@anthropic.com>`. Squash collects the trailer, so the non-conforming form is what lands on `main`. Rewriting a message needs your say-so per 3.4, so it stays yours to call.
2. `_proc_locks_pid` matches the inode only, dropping the `MAJOR:MINOR` half of what `/proc/locks` records. That was harmless while lsof answered first on Linux; it is now the primary source, so a file with the same inode number on another filesystem holding a lock returns a stranger's pid, `lock_holder` discards it on the cmdline check, and a healthy server reads as "no holder" -- the same class of misdiagnosis this PR is fixing, arriving through another door. Long odds, and the fix is one line (`lock.stat().st_dev` against `os.makedev` of the two fields). Take it or leave it.

Branch name, commit types and scopes, ASCII, test file placement and the no-assets rule are all fine.

raven/cli/onboard_everos.py
```diff
@@ -2029,6 +2092,27 @@ def _step4_memory(
...
+        # And discovery adds raven's own roots with owned=True unconditionally,
+        # so an abandoned managed root -- the normal leftover after switching --
+        # would be picked and adopted before any menu appeared. A recorded
+        # ownership decision outranks a directory that happens to still exist.
+        if _enabled_unowned_menu() is not _SWITCH_TO_MANAGED:
+            return None
+        # An explicit handover. Recorded here rather than left to the branch that
+        # builds the root, because everything downstream -- _memory_enabled(),
+        # owned_everos_root() -- has to see the new answer first.
+        _record_root(default_everos_root(), owned=True)
```
gloryfromca: **The handover records raven's own root but leaves the user's `base_url` behind, so the managed server inherits the user's port.**

`_record_root` merges, so after this line the slice is `{owned: true, root: <raven root>, base_url: http://127.0.0.1:8000}` -- the address the screen above just promised raven would stop using. `_ask_managed_port` then reads it: `port` was removed on the way in by `_use_self_managed_everos`, so `current` falls back to `urlparse(base_url).port`, i.e. 8000. `_configured_target_url` refuses exactly that fallback on purpose ("Deliberately not falling back to the recorded address"); `_ask_managed_port` does not, and after this handover the recorded address is no longer raven's.

Traced by driving `_step4_memory` with a self-managed slice at `http://127.0.0.1:8000` and answering `managed`:

* **Their server stopped** (they shut it down, then re-ran onboard -- the natural order). Every port looks free, `_ask_managed_port` returns 8000 without a word, and the final slice is `{"owned": true, "base_url": "http://localhost:8000", "port": 8000, "root": <raven root>}`. Raven's managed EverOS is now parked on their port: the next time they start their own server the two collide, and 18791 was never mentioned.
* **Their server still up** (what the handover text explicitly promises: "That server and its data are untouched"). Output is `! Port 8000 is already in use by something else.` -- that "something else" is their own EverOS -- and the port prompt then defaults to `8000`, so an Enter records 8000 and the start a few lines later fails.

The pre-existing `TestReusingAnEverosTheUserManages` case lands the same way (its `http://localhost:8000` is carried into a managed root); it just never asserts on `base_url`, so it stays green.

Smallest fix I see is to drop the stale address in the same write, mirroring how `_use_self_managed_everos` clears `root`/`port` on the way in:

```python
set_plugin_config_fields(
    "everos-memory",
    {"root": str(default_everos_root()), "owned": True},
    remove=("base_url", "port"),
)
```

so `_ask_managed_port` falls through to the 18791 default. The other option -- teaching `_ask_managed_port` to ignore an address that is not raven's -- needs it to know ownership just flipped, which is more moving parts for the same outcome. Your pick.

raven/cli/onboard_everos.py
```diff
@@ -2029,6 +2092,27 @@ def _step4_memory(
...
+        # writing into a root it owns, which by definition it does not. Sharing
+        # the managed menu made "Reconfigure" the only plausible button for
+        # changing the address, and it answered by recording raven's own root,
+        # flipping ownership and overwriting the address, none of it confirmed.
+        #
+        # And discovery adds raven's own roots with owned=True unconditionally,
+        # so an abandoned managed root -- the normal leftover after switching --
+        # would be picked and adopted before any menu appeared. A recorded
+        # ownership decision outranks a directory that happens to still exist.
+        if _enabled_unowned_menu() is not _SWITCH_TO_MANAGED:
```
gloryfromca: Nonblocking: taking the handover asks the same question twice.

The user picks "Let Raven run its own EverOS instead"; the next screen is "Where should long-term memory come from? / Let Raven run EverOS for me / I run my own EverOS". Cause is two lines down: `_record_root(default_everos_root(), owned=True)` leaves `_memory_enabled()` false for a fresh root (no llm role yet), so the `found is None and not _memory_enabled()` block fires the source question. Traced: two selects, the second one offering `['managed', 'self']`.

This is the same shape as the pre-existing `_OWN_ROOT_INSTEAD` path, so it is not new behaviour -- but there the preceding screen asked something else, whereas here the second screen is verbatim the question just answered, which reads as if the answer did not register. If you want it gone the handover has to tell the source question it is already settled (a local flag, or fall straight into the managed build). Fine to leave for later.

tests/test_cli_onboard_commands.py
```diff
@@ -5874,3 +5876,132 @@ def test_creating_a_root_still_honours_a_recorded_address(
...
+    def test_a_managed_install_still_gets_the_model_roles(
+        self, tmp_env: Path, everos_isolated: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """The managed menu is unchanged; only the unowned one is new."""
+        import inspect
+
+        from raven.cli import onboard_everos
+
+        src = inspect.getsource(onboard_everos._step4_memory)
+        assert "_enabled_unowned_menu" in src, "the enabled branch does not consult ownership"
```
gloryfromca: This test asserts on the *source text* of `_step4_memory`, not on behaviour.

Delete the `_config_everos_role` loop outright and it still passes; rename the helper without changing anything and it fails. It also declares `everos_isolated` and `monkeypatch` without using either, and its name plus docstring claim a coverage ("a managed install still gets the model roles") that the assertion does not provide.

`TestReusingAnEverosTheUserManages` already drives the real path and asserts `len(reached) == 4`, so the coverage is not missing -- this file just gains a green line that would not catch the regression it is named after. Either do the same here with an `owned: true` slice, or drop it.

raven/plugin/memory/everos/_server.py
```diff
@@ -399,7 +400,76 @@ def _cmdline_of(pid: int) -> str:
...
+        except ValueError:
+            continue
+        if inode in mine:
+            return port
+    return None
+
+
 def _listening_port(pid: int) -> int | None:
+    """The TCP port ``pid`` listens on, or ``None`` if it serves no HTTP."""
+    port = _lsof_listening_port(pid)
```
gloryfromca, line 466: Nonblocking, and it is this commit's own argument turned around: `_lock_holder_pid` was reordered because "asking lsof first meant that branch never ran on a Linux box that has lsof -- including the one it was validated on". Behind lsof, `_proc_net_listening_port` is in the same position: it runs only on a Linux host with no lsof, so no ordinary dev box or CI runner ever executes it, and both new tests stub out the two /proc readers it depends on (`_proc_net_rows`, `_socket_inodes_of`) -- the parsing is covered, the reads are not.

Nothing is wrong today: both sources agree wherever both work. Reading /proc first would exercise the new path where it is actually developed and skip a subprocess with a 5s timeout. `test_lsof_is_still_used_when_present` pins the current order deliberately, so this is a decision, not an oversight -- flagging it so it is a conscious one.

raven/cli/onboard_everos.py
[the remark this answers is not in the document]
gloryfromca: Agreed, and independently reproduced: driving `_step4_memory` with a self-managed slice and answering `managed` yields two selects, the second offering `['managed', 'self']`. Same conclusion on severity too -- the recovery paths are clean, and the sibling `_OWN_ROOT_INSTEAD` path at 2130 already does this, so it is an extended quirk rather than a regression. It is the same point as my note one line up at 2109.

One correction to "it's not destructive either way": on this exact path it is, just not through the prompt. `_record_root` merges, so the handover leaves the user's `base_url` in the slice, and `_ask_managed_port` derives the managed port from it (`urlparse(base_url).port`, since `port` was removed on the way in). Answering "managed" a second time therefore configures raven's own server on the user's port -- silently on 8000 when their server is stopped, or with `Port 8000 is already in use by something else` plus an `8000` default when it is still up. Details and a fix in the inline note at 2114.

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. The port inheritance I blocked on is fixed, and I re-verified it independently rather than reading the new tests.

**Delta since my last pass:** the branch was rebased onto current `main` and gained one commit, `c7f03d0e`. `git diff 31c26095 7bbf0109` over the four touched files is empty, so the previously reviewed content came through the rebase untouched -- the only new material is `_adopt_own_root`.

**The fix holds.** Re-running my own harness against this head, both variants now land on 18791: their server stopped -> `{"owned": true, "root": <raven root>, "base_url": "http://localhost:18791", "port": 18791}`; their server still listening on 8000 -> same result and no port prompt at all, where before it printed `already in use by something else` and defaulted the prompt to their port.

`owned is False` is the right discriminator, and the else branch is not over-broad -- I checked all three ways into that line:
* the handover (record still says `owned: false`) -> retracts, which is the fix;
* a fresh install (no slice at all, `.get("owned")` is `None`) -> `remove=()`, byte-identical to the old `_record_root`;
* declining to share a discovered root with a managed port the user moved to -> kept, and unit-tested directly.

Mechanics check out too: `remove=()` is safe (`pop(key, None)` per key, and the log line's `if remove` handles the empty tuple), and dropping the `default_everos_root` import from `_step4_memory` leaves no dangling reference -- the only remaining use is inside the new helper, which is also what makes the test's `monkeypatch.setattr(ue, "default_everos_root", ...)` bite.

**The test change is a strengthening, not a weakening.** `test_declining_falls_through_to_ravens_own_root` now stubs `_port_is_free`: the port was never that test's subject, the new `base_url`/`port` assertions pin the outcome that matters, and the stub removes a dependency on whether the host happens to be running an EverOS. That is the same root-cause fix the 11 host-dependent tests below still need.

**Test state on this host:**
* `uv run pytest tests/test_cli_onboard_commands.py tests/test_everos_server.py` -> 11 failed, 310 passed. All 11 are the identical pre-existing host-dependence from my last pass (a real EverOS listening on 18791 here makes `_ask_managed_port` inject an unstubbed prompt); the set of failing names is unchanged by this commit.
* Neutralizing that one host fact (autouse `_port_is_free -> True`) -> `tests/test_cli_onboard_commands.py` 260 passed, i.e. the three new tests pass and nothing else moved. `tests/test_everos_server.py` -> 61 passed on its own.

**Still open, all nonblocking and already argued -- not repeating the detail:** the double-ask after the handover (now acknowledged in a comment in the new test), `test_a_managed_install_still_gets_the_model_roles` asserting on source text, `_proc_net_listening_port` sitting behind lsof, and the inode-only `/proc/locks` match. None of them has to move for this to land.

One bookkeeping note: `c7f03d0e` uses the AGENTS.md 3.3 trailer form, but the two rebased commits still carry `Co-Authored-By: Claude Opus 5 (1M context)`. Squash collects every commit's trailer, so both forms end up on the squash commit. Rewriting messages needs your say-so per 3.4, so it stays your call.

**Covered this pass:** the delta, all three call sites of the new helper, `set_plugin_config_fields` remove semantics, rebase integrity of the previously reviewed content, the new and modified tests, and AGENTS.md. Not covered: a real Linux host -- the /proc code is still read, not run.

raven/plugin/memory/everos/_discover.py
```diff
@@ -113,39 +116,30 @@ def discover() -> list[RootState]:
...
         return state
 
-    slice_ = _recorded_slice()
-    recorded = slice_.get("root")
+    recorded = _recorded_slice().get("root")
     if recorded:
-        root = Path(str(recorded)).expanduser()
-        owned = bool(slice_["owned"]) if "owned" in slice_ else root_is_raven_owned(root)
-        add(root, owned=owned)
+        add(Path(str(recorded)).expanduser())
```
gloryfromca, line 138: **Blocking: the managed lane will take over, write into, and restart an EverOS the config says the user runs.**

`discover()` now adds the recorded root unconditionally and `pick()` returns the first *configured* candidate -- which is the recorded one, since it is added first. When the recorded slice is `{root: <the user's dir>, owned: false}`, the managed lane offers that directory as "a memory directory Raven can take over", and "Reconfigure it (port and models)" writes into it.

Traced end to end with real discovery doing the picking (only `default_everos_root` / `applicable_legacy_root` redirected somewhere harmless, and their `/health` answering):

```
config:  {"root": "<theirs>", "owned": false, "base_url": "http://localhost:8000"}
theirs/everos.toml:  [llm] model = "their-private-model", api_key = "their-secret-key"
                     [api] host = "localhost", port = 8000

  v Found a memory directory Raven can take over
      memory dir  <theirs>
      state       running at http://localhost:8000
  -> answer "redo"

ensure_everos_home       -> <theirs>        # templates written into their directory
configure_everos_env     -> <theirs>
model roles written into -> llm, embedding, rerank, multimodal @ <theirs>
stop_for_reload          -> <theirs>        # their running server signalled
server started at        -> http://localhost:18791
final slice              -> {"root": "<theirs>", "owned": true, "port": 18791}
```

Their `[llm]`/`[embedding]`/`[rerank]`/`[multimodal]` sections and the keys in them are replaced, their process is stopped, and the service is moved off the port they chose. That is the case `EverosRootNotOwnedError` exists for, and the test this commit deleted asserted it must not happen ("adopted the root the user declined to share"). The docstring immediately above this line still promises it: *"An EverOS the user runs is never discovered."*

I read the commit message, so I know the takeover is deliberate -- "the managed lane takes over whatever memory directory it finds, whatever an earlier run recorded there" -- and for a root that is raven's own that is exactly right. My objection is only to the case where the record says the directory is not raven's.

**Reachability, stated plainly.** `_use_self_managed_everos` retracts `root`, so no config written by current raven holds this shape; the migration at `raven/config/loader.py:516-523` fills a missing root with `fallback_everos_root()`, which is raven's own. The exposure is pre-#310 configs, which recorded whatever root was active including the user's -- exactly what the comment on the deleted test said ("A recorded unowned *root* only reaches this code from a config written before self-managed setups stopped recording one"). A bounded population, not a hypothetical one. If you can establish that no such config exists in the wild, this drops to fixing the docstring.

**Two ways out, both small:**

1. **Exclude it in discovery** -- skip the recorded root when the recorded slice says `owned is False`. The managed lane then finds raven's own root or none and builds there, the self-managed lane is untouched, and the docstring above becomes true again. Cost: someone who genuinely wants to hand their directory over has to point the managed lane at a fresh root, which is what the old design implied anyway.
2. **Keep the takeover, make it informed** -- have `_found_root_menu` recognise `owned: false` and say it: this is the directory you told raven you run, reconfiguring replaces the models and keys in it, and the running service will be stopped. Cost: the wizard keeps reading `owned` for routing, which is part of what this refactor set out to delete -- though `_adopt_root` reads it anyway, three lines later, for the address retraction.

I would take the first: one condition restores the invariant, whereas the second leaves a destructive answer one keystroke away. Your call either way.

raven/cli/onboard_everos.py
```diff
@@ -1968,24 +1624,182 @@ def _set_base_url(base_url: str) -> None:
...
+    if state.serving:
+        status = oc._t(f"running at {where}", f"正在 {where} 上运行")
+    elif state.busy_elsewhere:
+        status = oc._t("its data is in use, but not at that address", "数据正被占用，但不在该地址上")
+    else:
+        status = oc._t(f"not running ({where} declared)", f"未在运行（配置声明 {where}）")
+    oc.console.print()
+    oc.console.print(
+        oc._t(
+            f"  [green]v Found a memory directory Raven can take over[/green]\n"
```
gloryfromca, line 1700: Narrower point, and it stands whichever way you resolve the other note: **"Use it as it is" changes hands without saying so.**

`_adopt_root(found.root)` runs before the branch, so on a root the record marks `owned: false` the safest-sounding answer -- the one that genuinely touches nothing *this run* -- still writes `owned: true` against the user's directory. From then on `everos_owned()` is true, so doctor reports the directory as raven's, and the runtime and the next `raven onboard` are free to write to it and to start and stop it. The screen says "Found a memory directory Raven can take over" and never mentions that this is the one raven was told not to touch.

Compounding it slightly: the lane question is now asked cold, with no line stating the current setup. The previous revision opened with "Long-term memory uses the EverOS you run at http://..."; a self-managed user re-running the wizard now sees three lanes with "Let Raven run EverOS for me" first, and two Enters reach this screen. If you keep the takeover design, this is the line where the change of hands should be stated.

gloryfromca (commented):
Blocking: the managed lane takes over, reconfigures and restarts an EverOS the config records as one the user runs -- inline at `raven/plugin/memory/everos/_discover.py:138`. Everything else in this refactor is an improvement, and I would sign off the moment that one case is handled.

Reviewed the whole rewrite rather than the delta, since it replaces the step wholesale (-963/+484, eight helpers and `RootState.owned` gone).

**What I covered:** the new `_step4_memory` and all three lanes; `_memory_source_menu` / `_found_root_menu` / `_use_found_root` / `_adopt_root`; the `discover()` and `pick()` change and every remaining reader of ownership (`doctor_commands.py` via `everos_owned()`, `loader.py` migration, `owned_everos_root`); the deleted tests against the code they covered; AGENTS.md; and whether the fixture changes hide anything. Not covered: a real Linux host, and the runtime/session path beyond confirming it reads the same recorded `owned` field.

**The refactor lands three things I raised, properly:**
* The double-ask is gone -- one lane question, no second screen re-asking it.
* `test_a_managed_install_still_gets_the_model_roles`, the test that asserted on the source text of `_step4_memory`, is deleted rather than patched.
* The host-dependence I flagged is fixed at the root: pinning `_port_is_free` and `discover` in the `everos_isolated` fixture. On this machine, with a real EverOS still listening on 18791, `tests/test_cli_onboard_commands.py + test_everos_discover.py + test_everos_server.py + test_config_update_everos.py` -> **371 passed, 0 failed**, where the same host produced 11 failures two revisions ago. The pin is a default, not a blanket: six tests still patch `_port_is_free` to `False`, so the occupied-port branch is genuinely exercised.

**On the deleted tests.** `_same_address` and `_adopt_running_address` tests went with the functions they tested, which is right. `test_reuse_records_the_address_and_writes_nothing` was replaced by `test_a_root_recorded_as_the_users_is_taken_over_all_the_same`, which asserts the opposite outcome -- that is an honest specification change, pinned by a test, not a test weakened to look green. It is that specification I am disputing in the inline note, not the way it was tested.

**Also checked and clean:** `RootState.owned` has no readers left anywhere (`doctor` reads the recorded field, whose meaning is unchanged); `root_is_raven_owned` is still live in `loader.py` and `everos_owned()`, so nothing became dead code; `_describe` sets `alive=False` whenever no address is declared, so `_use_found_root`'s `str(state.declared_url)` cannot stringify `None`; `_adopt_root`'s retraction still only fires on `owned is False`, and the "managed port the user moved to survives" case keeps its unit test. "Reuse" reaches `_config_everos_role` nowhere, which is the promise the new menu makes.

**Still open from earlier, unchanged and nonblocking:** `_proc_net_listening_port` sitting behind lsof in `_server.py` -- untouched by this commit, already argued, no need to move it here.

0xKT (approved):

PR #284, opened 2026-08-08
feat(agent): bind the model to the conversation, not the process

arelchan:
## Summary

The model was two attributes on a process-wide `AgentLoop`, so there was one answer for everyone: two sessions could not run different models, a switch in one moved all of them, and the "default" was whatever the last switch happened to write to `agents.defaults`. This makes the model a property of the conversation, and -- following the same thread down -- makes the credential that serves it something the user says rather than something Raven infers.

A `ModelBinding` is a model id, the credential that serves it, and how much that model can hold, as one value. `ProviderPool` is the single place deciding which credential a model id pairs with, cached per (vendor, model). `run_turn` resolves the binding for the turn's session and holds it in a `ContextVar` for the whole turn tree; the loop's `provider` / `model` / `context_window_tokens`, the context engine's LLM-backed segments, the skill gate and rewriter, and the consolidator all read that instead of a reference of their own.

| Rule | How it lands |
|---|---|
| Different sessions, different models | A dict of overrides, read once at turn entry |
| A switch moves only the session that asked | Nothing else reads that session's entry |
| A new session starts on the configured default | A session-scoped switch does not write `agents.defaults` |
| A configured subsystem uses its own model+credentials; otherwise it follows the conversation | The factory resolves each pin through the pool, so a holder has either a complete pair or nothing |
| A switch mid-turn takes effect next turn | Free -- the turn holds the binding it entered on |

Detached work inherits the context copy `asyncio` makes at task creation, so a subagent finishes on the model it was spawned under.

### The window belongs to the binding

Sizing came to a head during the rebase. `refresh_context_window` and its cascade assumed the loop has one current model -- which this PR makes false, since two sessions can be on a 200k and a 1M model at once, and one int on the loop cannot answer for both. That is the same mis-sizing #341 removed, arriving through another door.

So the window moved onto `ModelBinding`, beside the credential, and every holder reads the binding of the turn it is running under. `refresh_context_window` is gone: a turn now enters on a binding that already knows its own size, so there is nothing to refresh.

It resolves on first read rather than at construction, because building a provider is what imports LiteLLM -- a lazily built one has not done it yet, so an eager window is the catalogue-miss default for every model, which is the defect `on_built` existed to paper over. A miss is deliberately not cached, so a cold read cannot make the process wrong for its lifetime. An explicit `contextWindowTokens` travels on every binding, so a pinned number survives whatever model a session switches to.

### The provider is a word the user says

[...]

PR #346, opened 2026-08-18
fix(*): keep migration notices off stdout, and stop planting defaults

arelchan:
## Summary

Three follow-ups from #341, which fixed a retired `contextWindowTokens: 65536` capping every upgraded install at 64k. Two are its loose ends; the third is the reason it happened at all.

**The notice no longer prints on stdout.** #341 drains config-migration notices in the `run()` wrapper so every command tells the user, which caught the two whose stdout is a machine-readable document: `raven doctor --json` and `raven import --json`, both advertised for automation. Reproduced -- a config still carrying the retired pin makes `doctor --json` emit JSON with a sentence after it, and `jq` refuses the result. Moved to stderr, where the rest of this class of message already goes (the `ConfigReadError` branch a few lines up in the same wrapper, and the loader's malformed-config warning). That fixes any future `--json` command without it having to know this exists, which allow-listing the two would not. The window is narrow -- the migration fires once per install -- but the run it can spoil is a first run after upgrading, and a CI job is exactly the caller not watching.

**`raven doctor` gains a Config section and a `--fix`.** The premise changed while building it: the migrations run at load, so by the time this command looks, nothing is pending -- they already happened, silently, one launch ago. What is left for a person to ask about is what Raven deliberately will not decide for them. So it reports a `contextWindowTokens` pinned below what the configured model holds, and names what rides on it: the per-turn history budget, when the Curator starts paying for a slow path, and when memory consolidation archives. `--fix` removes the pin.

Reported without `--fix`, never written: that pin is a real configuration for a local model or an endpoint served smaller than the catalogue thinks, and the flag is the user saying which case theirs is. The write carries the file mode across (config.json holds `providers.*.apiKey`) and uses a per-PID temp name -- the same rules the loader's migration writer follows.

One check rather than two. A config naming no provider is the other thing worth asking about, but it reads differently on either side of the explicit-provider rule in #284, so it belongs to that change; a model that routes to nothing is already reported above it.

**The bootstrap stops writing every default into the user's config.** This is the cause of the bug #341 cured. `save_config` dumped the whole `Config()`, so a new install started life with eight kilobytes of settings its owner had never heard of and, worse, with that day's defaults frozen into their file -- after which a default we improve never reaches anyone who already has one. `contextWindowTokens: 65536` got onto disk exactly this way.

It now writes only what differs from the defaults. Lossless on reload, since a value equal to its default reloads as that default. A fresh config drops from 7,985 bytes to 2, and to 634 once the extension blocks are seeded -- what is left is the handful of lines someone actually chose, which is also a config they can read.

That third one is the only change here with a broad blast radius, and it has one visible consequence, which is an improvement: the onboarding picker used to lead with the schema's default model, because the bootstrap had written it into the config and the wizard read it back as "the current model" -- a value nobody had chosen, offered as though somebody had. With nothing written, the wizard leads with the provider's own recommendation instead. The test that pinned the old list now pins the new one and says why.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/                       6650 passed, 1 failed, 33 skipped
                                    (tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare,
                                    which fails identically on github/main when its file runs
                                    together and passes alone on both)
ruff check raven/ tests/            All checks passed
ruff format                         clean
```

Driven by hand as well as by tests, each against a throwaway home:

```
doctor --json on a fossil config   stdout parses as JSON again; the notice is on stderr
doctor    (pin 65536, model 1M)    reports the pin and what is sized against it
doctor --fix                       pin removed, file still 0600, a second run reports nothing
save_config on a fresh Config()    2 bytes; 634 after the extension blocks are seeded
```

New cases cover the stderr split, both doctor paths including the mode preservation, and the save round-trip keeping every value the user chose.

## Risk

The bootstrap change alters what every new install's config file looks like: nearly empty instead of a full dump. Existing configs are untouched -- nothing rewrites them -- and both shapes load identically, so the risk is confined to anyone reading `config.json` as documentation of what can be set. `raven doctor` and the schema remain the answer to that question.

The onboarding picker's default moves from the schema default to the provider's recommendation, as described above.

`doctor --fix` writes only when asked, and only removes a key. Everything else here is output routing.

Rollback: revert the commits. Nothing persists a new format, so a reverted build reads every config written under this one.

## Related Issues

N/A

### DIFF
--- a/raven/cli/doctor_commands.py
+++ b/raven/cli/doctor_commands.py
@@ -15,7 +15,7 @@ import json
 import time
 from dataclasses import asdict, dataclass, field
 from datetime import datetime
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Any, Optional
 
 import typer
 from rich.console import Console
@@ -24,6 +24,8 @@ from raven import __logo__
 from raven.cli._helpers import print_probe_troubleshooting, send_probe
 
 if TYPE_CHECKING:
+    from pathlib import Path
+
     from raven.config.raven import RavenConfig
 
 console = Console()
@@ -113,6 +115,24 @@ class ProbeResult:
 
 
 @dataclass
+class ConfigHealth:
+    """What the config says that the migrations deliberately did not change.
+
+    The migrations run at load, so by the time this command looks there is
+    nothing pending -- they already happened, silently, one launch ago. What is
+    left for a person to ask about is the set of things Raven will not decide
+    on their behalf: a window they pinned that is smaller than the model can
+    hold, and a provider nothing could resolve. Both are legitimate
+    configurations and both are common mistakes, which is exactly why they need
+    somewhere to be asked rather than a rule that guesses.
+    """
+
+    findings: list[str] = field(default_factory=list)
+    fixes: list[str] = field(default_factory=list)
+    applied: list[str] = field(default_factory=list)
+
+
+@dataclass
 class DoctorReport:
     version: int = 1
     config_loaded: bool = False
@@ -122,6 +142,7 @@ class DoctorReport:
     gateway: Optional[GatewayInfo] = None
     memory: Optional[MemoryInfo] = None
     probe: Optional[ProbeResult] = None
+    config_health: Optional[ConfigHealth] = None
 
     def exit_code(self) -> int:
         if self.paths is None or not self.paths.config_exists:
@@ -139,6 +160,73 @@ class DoctorReport:
         if self.memory is not None and self.memory.broken:
             return 2
         return 0
+
+
+def _inspect_config_health(config: Any, *, fix: bool) -> ConfigHealth:
+    """Ask the two questions the migrations refuse to answer for the user.
+
+    ``fix`` is the consent: without it this only reports, because the finding is
+    a value someone may have meant -- a window pinned below the model is a real
+    configuration for a self-hosted endpoint served smaller than the catalogue
+    thinks.
+
+    One check for now. The other candidate -- a config naming no provider --
+    reads differently before and after the explicit-provider rule lands, so it
+    belongs to that change rather than to this one; routing that resolves to
+    nothing is already reported above.
+    """
+    from raven.config.loader import get_config_path, read_raw_or_raise
+    from raven.providers.rates import resolve_context_window
+
+    health = ConfigHealth()
+    defaults = config.agents.defaults
+    pinned = defaults.context_window_tokens
+    model = defaults.model
+
+    if pinned and model:
+        real = resolve_context_window(model)
+        if real and pinned < real:
+            health.findings.append(
+                f"contextWindowTokens is pinned to {pinned:,}, but {model} holds {real:,}. "
+                f"Everything is sized against the pin: the history budget, when the Curator "
+                f"starts paying for a slow path, and when memory consolidation archives."
+            )
+            health.fixes.append("remove agents.defaults.contextWindowTokens so the window follows each model")
+
+    if fix and health.fixes:
+        path = get_config_path()
+        try:
+            raw = read_raw_or_raise(path)
+            raw.get("agents", {}).get("defaults", {}).pop("contextWindowTokens", None)
+            raw.get("agents", {}).get("defaults", {}).pop("context_window_tokens", None)
+            _write_config_preserving_mode(path, raw)
+        except Exception as exc:  # noqa: BLE001 -- reported, never fatal
+            health.findings.append(f"could not write the fix: {exc}")
+        else:
+            health.applied = list(health.fixes)
+            health.fixes = []
+
+    return health
+
+
+def _write_config_preserving_mode(path: "Path", raw: dict) -> None:
+    """Atomic replace that carries the original's mode across.
+
+    ``os.replace`` swaps the inode, so the mode of what lands is the temp
+    file's: a config the user tightened to owner-only (it holds
+    ``providers.*.apiKey``) would come back world-readable. Same rule the
+    loader's migration writer follows.
+    """
+    import json as _json
+    import os as _os
+
+    tmp = path.with_name(f"{path.name}.doctorfix.{_os.getpid()}")
+    tmp.write_text(_json.dumps(raw, indent=2, ensure_ascii=False), encoding="utf-8")
+    try:
+        _os.chmod(tmp, path.stat().st_mode & 0o7777)
+    except OSError:
+        pass
+    _os.replace(tmp, path)
 
 
 def _gather_static_checks() -> DoctorReport:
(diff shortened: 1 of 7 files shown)


0xKT (approved):

raven/cli/doctor_commands.py
```diff
@@ -141,6 +162,73 @@ def exit_code(self) -> int:
...
+    from raven.config.loader import get_config_path, read_raw_or_raise
+    from raven.providers.rates import resolve_context_window
+
+    health = ConfigHealth()
+    defaults = config.agents.defaults
+    pinned = defaults.context_window_tokens
+    model = defaults.model
+
+    if pinned and model:
+        real = resolve_context_window(model)
```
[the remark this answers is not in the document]
gloryfromca: Confirming this independently -- it reproduces, and the failure is not only the theoretical contract break.

With `agents.defaults.model = "openrouter/acme/nonexistent-model-xyz"` and `contextWindowTokens: 32768` in a throwaway home, `_inspect_config_health(load_config(cfg), fix=False)` with `httpx.Client` swapped for a recorder:

```
network calls attempted: [('GET', 'https://openrouter.ai/api/v1/models')]
```

So plain `raven doctor` -- no `--probe` -- reaches `_fetch_openrouter_models`, and that client is built with `timeout=10.0`. Offline or behind a firewall, the documented millisecond-fast default mode now blocks for up to ten seconds, and `--json` is the flag advertised for CI.

Reachable whenever the pin is set and the LiteLLM tier misses, which is exactly the self-hosted / newer-than-the-table case the finding text is written for. Not every model: `_lookup_openrouter_entry` returns early unless the id starts with `openrouter/`.

Worth adding for whoever fixes it: `allow_fetch=False` costs nothing in coverage here. It also turns off the LiteLLM *import*, but `_gather_static_checks` has already called `resolve_max_output_tokens(defaults.model)` by this point, so LiteLLM is in `sys.modules` and `_try_litellm_context_window(allow_import=False)` proceeds exactly as it would with `True`. The only tier lost is the one that should not be running.

raven/cli/doctor_commands.py
```diff
@@ -141,6 +162,73 @@ def exit_code(self) -> int:
...
+    ``os.replace`` swaps the inode, so the mode of what lands is the temp
+    file's: a config the user tightened to owner-only (it holds
+    ``providers.*.apiKey``) would come back world-readable. Same rule the
+    loader's migration writer follows.
+    """
+    import json as _json
+    import os as _os
+
+    tmp = path.with_name(f"{path.name}.doctorfix.{_os.getpid()}")
+    tmp.write_text(_json.dumps(raw, indent=2, ensure_ascii=False), encoding="utf-8")
```
gloryfromca, line 224: [nit] Unlike the loader's writer this says it mirrors, this one does not unlink the temp file when the write fails. That one ends `except OSError: ... tmp.unlink(missing_ok=True)` (`loader.py:258-261`); here nothing cleans up.

Concrete: on ENOSPC or a quota, `tmp.write_text` creates the file, fails partway, and the caller's `except Exception` turns it into a "could not write the fix" finding -- leaving `~/.raven/config.json.doctorfix.<pid>` behind holding a partial copy of the config. The mode is the part that stings: `chmod` runs *after* the write, so the orphan keeps the default umask (0644) while containing `providers.*.apiKey`, sitting next to a config.json the user may have tightened to 0600 -- which is the exact exposure the docstring above exists to prevent. A failing `os.replace` leaves the same thing, fully written.

`tmp.unlink(missing_ok=True)` in an `except OSError:` that re-raises closes it, and makes the "same rule the loader's migration writer follows" claim true.

raven/config/loader.py
```diff
@@ -334,14 +334,26 @@ def save_config(config: Config, config_path: Path | None = None) -> None:
     """
     Save configuration to file.
 
+    Only what differs from the defaults is written. A dump of everything is
```
gloryfromca, line 337: [nit] This makes a sibling comment false. `_persist_migrations` still justifies itself with "``save_config`` is no use here either -- it dumps every default (~8 KB), re-planting the very kind of fossil this migration exists to pull out" (lines 224-226), and after this change it dumps no defaults at all.

The paragraph's other reason still holds and is now the whole justification: the mapping `load_config` migrated has the extension blocks popped, and a config that needed no edit must stay byte for byte. So this is a one-line trim, not a rethink. Left as is, the next reader takes a present-tense comment as documenting current behaviour and walks away with the wrong picture of what `save_config` writes -- the same drift the previous commit in this branch fixed for the watermark comment.

gloryfromca (commented):
Blocking: the zero-network break already flagged at `doctor_commands.py:187` reproduces -- plain `raven doctor` issues a live GET to openrouter.ai. Repro and a note on why the fix is free are in that thread. Everything else here I would merge as it stands.

Covered: the full diff against `origin/main`, the callers of what it changes (`save_config`'s only real caller, `read_raw_or_raise`, the loader's own migration writer, `tui_rpc.methods.config`), backward compatibility of the on-disk config, AGENTS.md conventions, and the tests for weakening.

**What I checked and am satisfied with**

- `save_config(exclude_defaults=True)`: the blast radius really is one caller. `onboard_commands._bootstrap_empty_config` is the only one, and only when the file is absent; every other config write goes through the surgical `update_*` helpers. The round trip is lossless in the way that matters -- the schema's only `model_fields_set` consumer is `should_warn_deprecated_memory_window`, and `memory_window` is `Field(exclude=True)`, so it never survived a dump anyway. `tui_rpc` reads raw keys and falls back to its own `_DEFAULTS`, but its keys (`agent.temperature`, `tui.*`) are not schema paths, so a thinner file changes nothing there either.
- The onboarding picker change is a genuine behaviour change, and the test was repinned with a reason rather than loosened. `anthropic/claude-sonnet-5` is the registry's own `default_model` for that provider (`providers/registry.py:322`), so leading with it is the right answer, not an accident of the diff.
- The stderr move follows the pattern already in the file (`_helpers.py:311` writes the config-path notice to stderr), and the two new cases assert both directions -- stderr carries it, stdout is empty.
- Conventions: commits are conventional, ASCII-clean, and carry `Co-authored-by` with a real model id; the branch name fits section 2.1; no new test files, and the touched ones are the right three per section 5.4.
- Tests, run here rather than taken on faith: `tests/test_cli_doctor_commands.py tests/test_cli_helpers.py tests/test_config_loader.py tests/test_cli_onboard_commands.py` -> 360 passed, 0 skipped. The other four files that touch `save_config` (`status` / `agent` / `gateway` / `cli_config_precedence`) -> 59 passed. Nothing skipped, xfailed, or env-gated.

I also agree with the dead-branch nit at line 582: `_inspect_config_health` is the only producer of `findings`, and neither of its two strings is indented, so that `else` cannot fire.

Two nits of my own inline. And one forward-looking note that is not a finding: a user who legitimately pins a smaller window now gets the `!` on every `raven doctor` run with no way to say "yes, I meant it". Acceptable while the section holds one check, but if it grows, an acknowledge path will be wanted before the section starts being ignored.

PR #349, opened 2026-08-19
fix(*): close the follow-ups the model-binding review left open

arelchan:
## Summary

Six follow-ups from #284's review rounds. Five are defects in what that PR itself added or removed and were deliberately deferred rather than rushed into a diff already under review; the sixth is a check #346 wrote down as belonging here, because it only becomes a real state once the provider is explicit.

**A session's model now survives a restart on every surface, not only in the TUI.** The read half of the persistence hung off `session.resume`, so only the surface that calls that handler got it: a conversation arriving from a channel came back on the configured default with its own choice sitting unread in its own session record. The read moves onto the loop, where every entry point already goes, and the handler's duplicate copy is gone -- one reader instead of two. `has_session_binding` restores first as well, or a session that switched before a restart reports as having inherited the default and the picker stars the wrong row. `clear_session_binding` counts as having consulted the record, so returning a session to the default is not undone by the next read.

**`raven provider use` refuses an unroutable provider before it writes.** A typo exited 0 with the config already changed, then printed a credential warning suggesting `raven provider set <typo> --api-key ...` -- advice that would have created a section for a vendor nothing routes to. `provider set` already applied the right test before writing a section; it is exposed as `ensure_routable_provider` and called first, which is the order the TUI's `/model` already used. Only an unroutable *name* is refused: a known provider with no credentials yet is still written and reported, because picking a model before configuring its key is a normal order to do things in. "No spec of ours" cannot be the typo test either -- Raven carries none for mistral or xai, and refusing on it would block every passthrough vendor the config documents as supported.

**The guard that keeps a model and its provider paired has a home again.** A whole-repo AST scan failing any `set_default_model` without `provider=`, and any raw `agents.defaults.model` write not paired with a provider write, lived in `tests/test_provider_pin.py` and was deleted with that file. Only half its message was obsolete; the rule it enforces became more central, and it is the only place a machine checks it. `test_provider_resolution_invariants.py` is where the other source guards of this kind already live.

**The binding cache is keyed on everything it caches.** The fingerprint covered `config.providers`, but `bind` also copies `contextWindowTokens` onto the binding, so editing that number left a stale binding answering with the old one until a restart -- the "I changed it and nothing happened" shape the window ladder exists to prevent, arriving through the cache instead.

**A gateway call is priced from the gateway's own table.** Who routes a request is not who bills for it. LiteLLM answered first for every id, including one naming a gateway, and the two disagree: measured on the pinned LiteLLM, `openrouter/z-ai/glm-4.6` is filed at 0.40/1.75 per million where OpenRouter's own current table says 0.50/2.00, so every such call was reported a fifth under its real cost. OpenRouter's catalogue now answers ahead of LiteLLM for ids that name it, fresh-only and never fetching: an expired copy is not better evidence than the router's table and would suppress the refetch the expiry exists to trigger, while a blocking fetch would be paid by the turn this runs after. Unchanged for a direct id, which is what once priced a self-hosted deployment at a hosted model's rate.

**`raven doctor` can ask about a provider nothing resolved.** #346 deferred this with its reason: before the explicit-provider rule, the field defaulted to `auto` and blank never happened. Now blank means the load-time migration tried the derivation and had no answer. Two findings -- blank, and a name nothing routes to -- both report-only. Neither is offered to `--fix`, and that is not an omission: only the user knows which vendor they meant to pay, and the migration has already tried. Guessing again under the name `--fix` would be the same guess wearing more confidence.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/                     6477 passed, 43 skipped, 13 deselected   (TERM=dumb)
ruff check / format --check       clean, 832 files formatted
npx tsc --noEmit (ui-tui)         clean
npm test (ui-tui)                 998 passed (87 files)
commitlint + check_commit_messages.py    clean over github/main..HEAD
npm run lint:rpc                  generated.ts in sync
commitlint + check_commit_messages.py   clean over github/main..HEAD
```

Every fix is mutation-checked rather than assumed -- reverting it turns the new test red. That is worth naming for two of them, where writing the test taught something the fix alone did not:

- The doctor check's scope is narrower than it looks. With a configured vendor that serves the model, the migration fills the blank during `load_config` and there is nothing left to report, so the fixture has to be genuinely unresolvable -- which is exactly when the finding is real.
- The first attempt at the pricing tier read the any-age cache and made a stale disk copy outrank a refetch. `test_expired_disk_triggers_refetch` caught it, which is why the tier is fresh-only.

Three review rounds landed on this PR after it opened, each fixed and each verified rather than
taken from the report:

- **Blocking:** the doctor provider check read a leftover `auto` as an unroutable vendor name. That
  is the legacy state the check exists for -- `_migrate_auto_provider` leaves the literal in place
  when it cannot resolve one -- so the user was told to fix a typo they did not make. It now matches
  `Config._match_provider`'s own `forced != "auto"` test, and says the word is retired rather than
  merely ignoring it. Neither neighbouring test used a literal `auto`, which is how the branch
  between them went unentered.
- `_restore_attempted.add` in `clear_session_binding` was inert, and its stated reason did not hold
  for the only caller: `session.delete` unlinks the record before the clear runs. Dropped rather than
  relocated -- marking in `set_session_binding` would claim the record had been consulted when it had
  not. The test that was supposed to guard it is renamed and its docstring now scopes what it does
  and does not pin.
- `_routes_anywhere` caught `Exception` where the sibling check catches `KeyError`, and the new
  refusal printed a different failure marker from the eleven others in its file.

## Risk

Two user-visible changes, both deliberate:

`raven provider use` now exits 1 on a provider name nothing routes to, where it used to exit 0 with the config already written. A script passing a misspelled `--provider` starts failing; a script passing a real one is unaffected, including for a passthrough vendor.

Cost figures change for calls made through OpenRouter on ids where its catalogue and LiteLLM's copy disagree. The new number is the one the gateway bills, so this corrects a systematic under-report rather than introducing a new figure -- but a user comparing today's total to yesterday's will see the step.

Rollback: revert the commits. Nothing here writes a new field or changes a stored shape, so a revert needs no migration.

## Related Issues

N/A -- follow-ups from the review of #284, which is merged.

### DIFF
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -766,11 +766,13 @@ class AgentLoop:
     def clear_session_binding(self, session_key: str) -> None:
         """Drop a session's override so it follows the default again.
 
-        Counts as having consulted the record: without that, the next ask would
-        read the stored model straight back in and undo this.
+        Deliberately does not mark the key as consulted. The one caller is
+        ``session.delete``, which unlinks the record before this runs, so there
+        is nothing left for a later ask to read back in -- and a session that
+        somehow kept its record is better served by re-reading it than by a
+        marking that claims we looked when we did not.
         """
         self._session_bindings.pop(session_key, None)
-        self._restore_attempted.add(session_key)
 
     def _forget_transport_verdicts(self) -> None:
         """Drop the capability verdicts a new provider may answer differently.
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -356,6 +356,9 @@ class AgentLoop:
         self._configured_window = context_window_tokens or None
         self._default_binding = ModelBinding(provider, model or provider.get_default_model(), self._configured_window)
         self._session_bindings: dict[str, ModelBinding] = {}
+        # Keys whose session record has been consulted for a stored model, hit
+        # or miss. See ``_restore_once``.
+        self._restore_attempted: set[str] = set()
         # Resolved lazily on the first tool result that carries an image. Keyed
         # by model, not a single flag: the loop is a long-lived singleton and
         # takes a per-call model (strategies rewrite it, and the model chain
@@ -666,8 +669,42 @@ class AgentLoop:
 
         A new session has no entry, so it starts on the configured default
         rather than on whatever the last session switched to.
+
+        A session whose choice is on disk but not yet in memory is restored here,
+        on first ask. That is what makes the choice outlive a restart on *every*
+        surface: the overrides live in this process, the session record is the
+        only place they survive, and hanging the read off a TUI-only resume call
+        meant a conversation on a channel came back on the default with its
+        choice sitting unread in its own record.
         """
+        binding = self._session_bindings.get(session_key)
+        if binding is not None:
+            return binding
+        self._restore_once(session_key)
         return self._session_bindings.get(session_key, self._default_binding)
+
+    def _restore_once(self, session_key: str) -> None:
+        """Read this session's stored model, at most once per key per process.
+
+        The negative answer is remembered too. Most sessions never switched, and
+        without that this would re-read a record on every turn to learn the same
+        nothing.
+        """
+        if session_key in self._restore_attempted:
+            return
+        self._restore_attempted.add(session_key)
+        sessions = getattr(self, "sessions", None)
+        if sessions is None or self._provider_pool is None:
+            return
+        try:
+            record = sessions.peek(session_key)
+        except Exception as exc:
+            logger.debug("cannot read session {!r} to restore its model: {}", session_key, exc)
+            return
+        metadata = getattr(record, "metadata", None) or {}
+        model = metadata.get("model")
+        if model:
+            self.restore_session_model(session_key, model, metadata.get("provider"))
 
     def session_model(self, session_key: str) -> str:
         """What to show this session's user, which is not the global default."""
@@ -679,19 +716,27 @@ class AgentLoop:
         ``session_model`` cannot answer that -- it falls back to the default,
         so it never returns None. Callers that must distinguish "chose this"
         from "inherited this" ask here.
+
+        Restores first, so a session that switched before a restart answers yes
+        rather than being reported as having inherited the default.
         """
+        if session_key not in self._session_bindings:
+            self._restore_once(session_key)
         return session_key in self._session_bindings
 
     def restore_session_model(self, session_key: str, model: str, provider_name: str | None = None) -> None:
-        """Put a resumed session back on the model it was last switched to.
+        """Put a session back on the model it was last switched to.
 
         Session overrides live in memory, so without this a restart moves every
-        switched session back to the default and the user's choice lasts
-        exactly as long as the process. The model comes from the session
-        record. A model that can no longer be built (a credential since
-        removed) leaves the session on the default rather than failing the
-        resume.
+        switched session back to the default and the user's choice lasts exactly
+        as long as the process. A model that can no longer be built (a credential
+        since removed) leaves the session on the default rather than failing the
+        turn that asked.
+
+        Normally reached through ``_restore_once``, which supplies the stored
+        pair; kept public for a caller that has the pair already.
         """
+        self._restore_attempted.add(session_key)
         pool = self._provider_pool
         if pool is None or not model:
             return
@@ -719,8 +764,13 @@ class AgentLoop:
         self._forget_transport_verdicts()
 
     def clear_session_binding(self, session_key: str) -> None:
-        """Drop a session's override so it follows the default again."""
+        """Drop a session's override so it follows the default again.
+
+        Counts as having consulted the record: without that, the next ask would
+        read the stored model straight back in and undo this.
+        """
         self._session_bindings.pop(session_key, None)
+        self._restore_attempted.add(session_key)
 
     def _forget_transport_verdicts(self) -> None:
         """Drop the capability verdicts a new provider may answer differently.
--- a/tests/test_agent_loop_session_model.py
+++ b/tests/test_agent_loop_session_model.py
@@ -588,8 +588,9 @@ def test_a_restored_session_stays_on_the_default_once_cleared(tmp_path) -> None:
 
     End to end, not a guard on one line: the key is marked both where the record
     is read and where a caller supplies the pair, so removing either alone still
-    leaves this green. ``test_the_stored_model_is_read_once_per_session`` is what
-    pins the marking itself.
+    leaves this green. ``test_the_stored_model_is_read_once_per_session`` pins
+    the marking in ``_restore_once``; the redundant one in
+    ``restore_session_model`` is unpinned, and deleting it passes the suite.
 
     Named for what it does show, and not for a guarantee
     ``clear_session_binding`` does not give: it deliberately does not mark, so a
--- a/tests/test_agent_loop_session_model.py
+++ b/tests/test_agent_loop_session_model.py
@@ -583,9 +583,19 @@ def test_the_stored_model_is_read_once_per_session(tmp_path) -> None:
     assert reads == ["tui:a", "tui:never-switched"]
 
 
-def test_clearing_an_override_is_not_undone_by_the_stored_model(tmp_path) -> None:
-    """Dropping a session back to the default has to stick. Re-reading the
-    record on the next ask would restore the very choice just cleared.
+def test_a_restored_session_stays_on_the_default_once_cleared(tmp_path) -> None:
+    """What makes clearing stick is that the read happens once per key.
+
+    End to end, not a guard on one line: the key is marked both where the record
+    is read and where a caller supplies the pair, so removing either alone still
+    leaves this green. ``test_the_stored_model_is_read_once_per_session`` is what
+    pins the marking itself.
+
+    Named for what it does show, and not for a guarantee
+    ``clear_session_binding`` does not give: it deliberately does not mark, so a
+    session cleared without ever having been read would be read afterwards. No
+    caller creates one -- ``session.delete`` unlinks the record first -- which is
+    why the marking lives in the read rather than in the clear.
     """
     loop = _restorable_loop(tmp_path, {"tui:a": {"model": "claude-sonnet-4-5", "provider": "anthropic"}})
 
--- a/tests/test_agent_loop_session_model.py
+++ b/tests/test_agent_loop_session_model.py
@@ -519,6 +519,95 @@ def test_a_stored_model_that_cannot_be_built_leaves_the_default(tmp_path) -> Non
     assert not loop.has_session_binding("tui:a")
 
 
+def _restorable_loop(tmp_path, stored: dict[str, dict[str, str]]) -> AgentLoop:
+    """A loop whose session store already holds someone's earlier choice."""
+    from raven.config.schema import Config
+    from raven.providers.pool import ProviderPool
+
+    cfg = Config()
+    cfg.agents.defaults.model = "claude-opus-4-5"
+    cfg.providers.anthropic.api_key = "sk-ant"
+    loop = AgentLoop(
+        provider=_Provider("boot"),
+        workspace=tmp_path,
+        model="boot/model",
+        context_config=ContextConfig(),
+        skill_forge_config=SkillForgeConfig(),
+        provider_pool=ProviderPool(cfg),
+    )
+    for key, metadata in stored.items():
+        record = loop.sessions.get_or_create(key)
+        record.metadata.update(metadata)
+        loop.sessions.save(record)
+    return loop
+
+
+async def test_a_channel_turn_restores_the_model_that_session_chose(tmp_path) -> None:
+    """The reason the read moved onto the loop. It used to hang off
+    ``session.resume``, so only the TUI got it: a conversation arriving from a
+    channel came back on the default after a restart, with its own choice
+    sitting unread in its own record. No resume call anywhere in this test.
+    """
+    loop = _restorable_loop(tmp_path, {"whatsapp:alice": {"model": "claude-sonnet-4-5", "provider": "anthropic"}})
+
+    seen: list[str] = []
+
+    async def body(*a, **k):
+        seen.append(loop.model)
+        return None
+
+    await _run(loop, "whatsapp:alice", body)
+
+    assert seen == ["claude-sonnet-4-5"]
+
+
+def test_the_stored_model_is_read_once_per_session(tmp_path) -> None:
+    """Most sessions never switched, and the miss has to be remembered too --
+    otherwise every turn of every unswitched conversation re-reads a record to
+    learn the same nothing.
+    """
+    loop = _restorable_loop(tmp_path, {"tui:a": {"model": "claude-sonnet-4-5", "provider": "anthropic"}})
+    reads: list[str] = []
+    real_peek = loop.sessions.peek
+
+    def counting_peek(key: str):
+        reads.append(key)
+        return real_peek(key)
+
+    loop.sessions.peek = counting_peek
+
+    for _ in range(3):
+        loop.session_model("tui:a")
+        loop.session_model("tui:never-switched")
+
+    assert reads == ["tui:a", "tui:never-switched"]
+
+
+def test_clearing_an_override_is_not_undone_by_the_stored_model(tmp_path) -> None:
+    """Dropping a session back to the default has to stick. Re-reading the
+    record on the next ask would restore the very choice just cleared.
+    """
+    loop = _restorable_loop(tmp_path, {"tui:a": {"model": "claude-sonnet-4-5", "provider": "anthropic"}})
+
+    assert loop.session_model("tui:a") == "claude-sonnet-4-5"
+    loop.clear_session_binding("tui:a")
+
+    assert loop.session_model("tui:a") == "boot/model"
+    assert not loop.has_session_binding("tui:a")
+
+
+def test_has_session_binding_sees_a_choice_that_only_exists_on_disk(tmp_path) -> None:
+    """``model.options`` asks this to decide whether to report the session's own
+    model or the configured default. Answered without the restore, a session
+    that switched before a restart is reported as having inherited the default,
+    and the picker stars the wrong row.
+    """
+    loop = _restorable_loop(tmp_path, {"tui:a": {"model": "claude-sonnet-4-5", "provider": "anthropic"}})
+
+    assert loop.has_session_binding("tui:a")
+    assert not loop.has_session_binding("tui:b")
+
+
 def test_has_session_binding_distinguishes_chosen_from_inherited(tmp_path) -> None:
     """``session_model`` falls back to the default, so it cannot answer this --
     and callers that override a forced provider need the difference.
--- a/raven/cli/doctor_commands.py
+++ b/raven/cli/doctor_commands.py
@@ -162,6 +162,23 @@ class DoctorReport:
         return 0
 
 
+def _routes_anywhere(provider: str) -> bool:
+    """Is this a provider name anything can route to?
+
+    Deliberately generous: it accepts a vendor Raven carries no spec for as long
+    as LiteLLM knows the name -- mistral and xai are supported exactly that way,
+    and reporting them as broken would be worse than saying nothing. What it
+    catches is a name nothing has ever heard of, which is a typo.
+    """
+    from raven.config.update_providers import ensure_routable_provider
+
+    try:
+        ensure_routable_provider(provider)
+    except Exception:
+        return False
+    return True
+
+
 def _inspect_config_health(config: Any, *, fix: bool) -> ConfigHealth:
     """Ask the two questions the migrations refuse to answer for the user.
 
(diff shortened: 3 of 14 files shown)


raven/agent/loop/main.py
```diff
@@ -719,8 +764,13 @@ def set_session_binding(self, session_key: str, binding: ModelBinding) -> None:
...
 
     def clear_session_binding(self, session_key: str) -> None:
-        """Drop a session's override so it follows the default again."""
+        """Drop a session's override so it follows the default again.
+
+        Counts as having consulted the record: without that, the next ask would
+        read the stored model straight back in and undo this.
+        """
         self._session_bindings.pop(session_key, None)
+        self._restore_attempted.add(session_key)
```
gloryfromca: Nonblocking, and the only one of the six changes I could not justify.

This line is unreachable as a fix, and the test presented as its guard does not guard it.

**The test does not cover it.** I removed this line and ran both files: 94 passed, including `test_clearing_an_override_is_not_undone_by_the_stored_model`. It passes either way because `restore_session_model` already adds the key at line 738, so by the time anything calls `clear_session_binding` on a session that *was* restored, the key is in `_restore_attempted` regardless. The test restores first (`session_model("tui:a")`), so it can only ever exercise the already-marked path.

**The docstring's reason does not hold for the only caller.** "the next ask would read the stored model straight back in" needs `peek` to still find a record. The sole caller is `session.delete` (`tui_rpc/methods/session.py:426`), and `SessionManager.delete` calls `self.invalidate(key)` and then unlinks the file *before* the clear runs -- so `peek` afterwards returns None from both the cache and the disk. There is nothing left to read back in.

The one path where the line would bite is a session that got a binding from `set_session_binding` without ever going through `_restore_once` -- `/model` in a lazy session before its first turn, which `_remember_session_model` explicitly calls out as normal. `set_session_binding` does not mark the key, so that session's binding and its `_restore_attempted` state disagree. If that is the case you have in mind, the marking belongs in `set_session_binding` (where the disagreement is created) rather than in `clear_session_binding`, and it would then be testable: switch, clear, ask, assert the default.

As it stands, either move it to `set_session_binding` and write the test, or drop the line and the two docstring lines above it -- a comment explaining a `why` that is not true for any caller is worse than none. Not a blocker: the line is inert, not wrong. But the PR says "every fix is mutation-checked rather than assumed -- reverting it turns the new test red", and for this one it does not.

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. One nonblocking note inline about the sixth change being unreachable and unguarded, and a one-character nit below.

## What I checked

- The diff vs `github/main` (6 commits, 14 files), plus the callers and the code around each change: `AgentLoop.run_turn` / `binding_for_session`, `tui_rpc/methods/config.py::_remember_session_model` and `_set_model`, `session.delete` / `session.branch`, `model_options`, `ProviderPool.bind`, `cli/_helpers.make_provider`, `config/update_providers._provider_schema_cls` / `_litellm_knows`, `model_catalog_cache.load`, `token_wise/pricing.estimate_cost_usd`.
- Repo conventions: commit grammar, scopes, ASCII, `Co-authored-by` trailers (all six clean); test file naming under AGENTS.md 5.1/5.4 (existing files updated, no new files, no phase/ticket suffixes); no new domain term coined, so no CONTEXT-MAP change owed; no assets.
- Backward compatibility of the two behaviour changes the description names.
- Whether tests were weakened, by mutation rather than by reading.

## Tests, actually run

```
TERM=dumb uv run pytest tests/ -q -p no:randomly
6476 passed, 43 skipped, 13 deselected, 1 warning in 152.26s
```

That matches the description exactly. I looked at every skip: 8 are optional-dependency imports (`dingtalk_stream`, `lark_oapi`, `nio`, `botpy`, `slack_sdk`, `telegram`, `wecom_aibot_sdk`), and 33 are data-driven `parametrize` exclusions in `test_provider_catalog.py` / `test_provider_resolution_invariants.py` that pre-date this branch. Nothing here is skipped to look green.

## Mutation checks I ran myself

I reverted three of the six fixes in place and re-ran the tests named as their guards:

| reverted | result |
|---|---|
| `_cache_fingerprint` back to credentials-only | `test_an_edited_window_drops_the_binding_that_carried_the_old_one` FAILS -- real guard |
| tier 0 removed from the `token_rates` ladder | `test_the_gateway_price_wins_over_the_routers_copy` FAILS on `(2e-06, 8e-06) != (4e-06, 1.6e-05)` -- real guard |
| `_restore_attempted.add` removed from `clear_session_binding` | all 94 tests in `test_agent_loop_session_model.py` + `test_tui_rpc_session.py` still PASS |

The third is the inline note. Working tree restored to HEAD afterwards.

## Things I considered and dropped

- **`_cache_fingerprint` overclaims.** Its docstring says "everything a cached binding was built from", but `bind` -> `make_provider` also reads `agents.defaults.model_overrides`, `temperature`, `reasoning_effort` and `llm_call_timeout`, none of which are in the key. I chased this and could not make it fail: `config.set`'s `agent.temperature` writes an `agent.temperature` key, not `agents.defaults.temperature`, so it never reaches `make_provider`; and nothing rebuilds the default binding on a raw file edit of those fields either, so dropping the pool entry would not help. `contextWindowTokens` really is the only one of the group with a live-update path, so the fix is complete for every reachable case -- the docstring is just wider than the key.
- **`_routes_anywhere` pays the ~2s authoritative LiteLLM import in a report-only command**, which is the cost `_litellm_knows` documents as one "a command that only reports configuration should not" pay. But `authoritative=False` would answer False for a vendor newer than the bundled snapshot, i.e. it would tell a user with a working provider that nothing routes to it -- worse than 2 seconds in a diagnostic, and the opposite of what the helper's own docstring argues for. `True` is the right call here.
- **Tier 0 runs `_fresh_openrouter_models()` unconditionally**, before the cheap `openrouter/` prefix test, on a path that runs after every completion. On a miss `model_catalog_cache.load()` is one `path.exists()`; on a hit it fills the module cache once and every later call short-circuits. Not worth a branch.
- **Tier 0 seeds `_OPENROUTER_CACHE` from an expired disk copy.** I checked this cannot suppress the refetch: `_fetch_openrouter_models` TTL-checks the in-process cache at line 265 before the disk copy at 270, so tier 2 still goes to the network. `test_expired_disk_triggers_refetch` covers it.
- **`provider use` with no `--provider`** cannot reach `ensure_routable_provider("")` -- the required-option guard above it already exits 1.

## Nit

`provider_commands.py:606` prints `[red]x[/red]` where the sibling failure three lines up, and the other 16 error prints in the file, use `[red]✗[/red]`. There is precedent for the ASCII form in `onboard_commands.py`, so this is taste rather than a rule -- but the two error paths of one command reading differently is easy to even up while you are here.

## The substance

The five real fixes hold up. Moving the read onto `binding_for_session` is the right place for it -- every entry point goes through `run_turn`, and `test_a_channel_turn_restores_the_model_that_session_chose` exercises the loop with no resume call anywhere, which is exactly the gap. `has_session_binding` restoring first matters more than the description makes it sound: `session.branch` gates the child's inherited binding on it, so before this a fork of a restored session silently dropped to the default. And `_restore_attempted` remembering the miss is what keeps this off the per-turn path for the sessions that never switched -- `test_the_stored_model_is_read_once_per_session` pins the negative, which is the half that is easy to forget.

gloryfromca (dismissed):

arelchan:
Blocker fixed in `01b25ed7`, reproduced first rather than taken from the description:

```
config: {"agents": {"defaults": {"model": "some/unclaimed-model", "provider": "auto"}}}

after load_config, provider = 'auto'
finding: agents.defaults.provider is 'auto', which nothing routes to. ...
         raven provider list  # the names this accepts
```

Exactly as reported, and the diagnosis is right about why it matters: this is the legacy state the check exists for, and reading it as a name tells the user to fix a typo they did not make while the advice that would help goes unsaid. `Config._match_provider` already treats it as unset (`forced != "auto"`, `schema.py:921`); the check now does the same.

One thing added beyond the fix: when the literal is present, doctor says so -- `(your config says `auto`, which is the retired spelling of unset -- it never detected anything)`. Silently treating it as unset would be correct and still leave the user staring at a word in their own file with no idea it stopped meaning anything.

**On how the gap survived**: neither of the two checks around it used a literal `auto` -- one used `""`, the other a misspelled vendor -- so the branch between them was never entered. That is the same shape as the `scope` prop note on #284: a test that exercises the neighbours of a case is not a test of the case. The new test uses the literal, and mutation-checking it (reverting to `if not provider:`) turns it red.

The nit is taken too: `_routes_anywhere` now catches `KeyError` only, matching `provider use`. Your reasoning is the part worth keeping -- a broad catch files a genuine fault in the lookup as an ordinary "unroutable" finding, which reads as the user's problem rather than ours.

```
pytest tests/                  6477 passed, 43 skipped, 13 deselected   (TERM=dumb)
ruff check / format --check    clean, 832 files
commitlint + ASCII check       clean over github/main..HEAD
```

gloryfromca (commented):

Re-reviewing only the delta `699d54de..01b25ed7` -- the substance I went through last round is unchanged.

## The `auto` blocker is genuinely fixed

I checked the reachability claim rather than taking it from the commit message. `_migrate_auto_provider` (`config/loader.py:277-278`) returns False on `if not resolved:` *before* the `defaults["provider"] = resolved` write two lines down, so the literal survives `load_config` -- the bot's diagnosis was right and so is the reproduction. And `schema.py:922` really is `if forced and forced != "auto":`, so `provider == "auto"` mirrors the runtime's own test exactly, including its case-sensitivity. Matching the runtime is the right bar here; a case-insensitive doctor would report a state the runtime does not treat that way.

Mutation-checked myself: reverting `or provider == "auto"` back to `if not provider:` turns `test_doctor_reads_a_leftover_auto_as_unset_not_as_a_typo` red. The test also pins the half that is easy to leave out -- `assert not any("nothing routes to" ...)` -- so it fails a fix that reports both findings at once rather than replacing the wrong one. Working tree restored afterwards.

The extra `(your config says \`auto\` ...)` line is two-space indented, which is the form `_render_human_output` already dims, so it renders as a continuation rather than as a third bullet.

## The `except KeyError` narrowing is safe

I went looking for a currently-reachable non-`KeyError` that the old broad catch was absorbing, because `_inspect_config_health` is called unguarded at `doctor_commands.py:669` -- if one existed, narrowing would turn a report into a traceback. There is none: `_litellm_knows` already funnels every failure of the LiteLLM import into `None`, and the two module-level imports left outside its try (`litellm_provider_names`, `litellm_setup`) are first-party and stdlib-only at module scope. So nothing in that call path raises anything but the two explicit `KeyError`s. The nit was right and the fix does not cost the report-only guarantee.

## Tests, actually run

```
TERM=dumb uv run pytest tests/ -q -p no:randomly
6477 passed, 43 skipped, 13 deselected, 1 warning in 121.55s
```

One more than the previous revision, which is the one new test. Identical skip set to last round -- 8 optional-dependency imports, 33 data-driven `parametrize` exclusions, none of them new and none of them hiding this change.

Commit is conventional, ASCII, 68-char header, `Co-authored-by` present. No new test file, so AGENTS.md 5.1/5.4 still hold.

## Considered and dropped

Doctor strips before comparing (`(... or "").strip() == "auto"`) while `_match_provider` compares the raw value, so `provider: " auto "` diverges. It is not worth raising: the `.strip()` pre-dates this commit, and in the divergent case doctor's message is the *more* useful of the two anyway -- it points at `raven provider use --provider <name>`, which is the advice that helps.

## Still open from last round, both nonblocking

- `raven/agent/loop/main.py:773` -- my note on `clear_session_binding` stands as written; no reply yet, and it does not block.
- `raven/cli/provider_commands.py:606` -- the `[red]x[/red]` / `[red]✗[/red]` mismatch is still there.

gloryfromca (dismissed):

raven/agent/loop/main.py
arelchan: Taken, and dropped rather than moved -- your second option.

Both halves reproduced before acting. Removing the line leaves 94 tests in both files green, and `SessionManager.delete` invalidates the cache and unlinks the file before the clear runs, so `peek` afterwards finds nothing in either place. The docstring was defending against a caller that does not exist.

On moving it to `set_session_binding`: marking there would say the record had been consulted when it had not. This one was idle; that one would be false, and the disagreement it papers over is unreachable anyway -- a session holding a binding returns from `binding_for_session` before the read.

**Checking the test taught me two more things, and my first draft of the fix got both wrong.** The key is marked in *two* places -- `_restore_once` before it reads, and `restore_session_model` when a caller supplies the pair -- so they are redundant with each other and removing either alone still leaves the clearing test green. I wrote a docstring naming `restore_session_model` as the guard, mutation-checked it, and it did not go red; named `_restore_once` instead, mutation-checked again, still not red. The test is an end-to-end statement, not a guard on a line. It now says that and points at `test_the_stored_model_is_read_once_per_session`, which is the one that actually pins the marking.

It is also renamed. `test_clearing_an_override_is_not_undone_by_the_stored_model` promised a guarantee the method deliberately does not give: a session cleared without ever having been read *would* be read afterwards -- which is the better behaviour for a caller that keeps its record, and no caller today is one.

Fixed in `8221ad85`.

tests/test_agent_loop_session_model.py
```diff
@@ -519,6 +519,105 @@ def test_a_stored_model_that_cannot_be_built_leaves_the_default(tmp_path) -> Non
...
+    assert reads == ["tui:a", "tui:never-switched"]
+
+
+def test_a_restored_session_stays_on_the_default_once_cleared(tmp_path) -> None:
+    """What makes clearing stick is that the read happens once per key.
+
+    End to end, not a guard on one line: the key is marked both where the record
+    is read and where a caller supplies the pair, so removing either alone still
+    leaves this green. ``test_the_stored_model_is_read_once_per_session`` is what
+    pins the marking itself.
```
gloryfromca: Nonblocking, and only half a correction -- the rest of this docstring is exactly right.

`test_the_stored_model_is_read_once_per_session` pins the marking in `_restore_once`, not "the marking". I ran both mutations:

- remove `main.py:695` (the mark in `_restore_once`) -> that test FAILS, because `tui:never-switched` gets re-read every turn. Real guard, as the commit message says.
- remove `main.py:739` (the mark in `restore_session_model`) -> **all 94 tests pass.**

So of the two writers this docstring names, one is pinned and the other is not. `_restore_once` marks before it delegates, so the marking inside `restore_session_model` is redundant for every in-repo path -- its only non-test callers are gone, and the two tests that call it directly (`main.py` lines 494 and 516 of this file) do not assert on the marking.

I am not asking you to change that line. Its contract is coherent as documented -- "kept public for a caller that has the pair already", and a caller who supplies the pair genuinely has consulted the record, which is the honesty test `clear_session_binding` failed. The problem is narrower: this docstring exists to warn the next reader that the clearing test is not a guard on a line, and in doing so it hands them a guard that only covers half of what it names. Someone trusting it can delete the `restore_session_model` marking during a refactor and see a green suite -- the same trap, one line over.

Cheapest fix is to say which one: "...`test_the_stored_model_is_read_once_per_session` pins the marking in `_restore_once`; the redundant one in `restore_session_model` is unpinned." That is accurate and costs nothing. Adding a test for the redundant marking would work too, but I would not bother -- naming it as unpinned is the honest and cheaper move.

gloryfromca (commented):

Reviewing only the delta `01b25ed7..8221ad85`. The change is the right resolution of my note and the docstrings are now honest about what the code does and does not promise. One inline nit on a sentence in the new test docstring that overstates its guard by half -- verified by mutation, and nonblocking.

## The drop is correct and safe

`_restore_attempted` now has exactly two writers (`main.py:695`, `main.py:739`) and none in `clear_session_binding`. I re-checked the only caller: `session.delete` runs `mgr.delete(session_key)` before `clear(session_key)`, and `SessionManager.delete` calls `self.invalidate(key)` *then* unlinks -- so `peek` afterwards misses in both the cache and on disk regardless of `delete`'s return value. Nothing can read the choice back in, which is what makes dropping the marking a no-op rather than a behaviour change.

I agree with dropping rather than moving it to `set_session_binding`, and for the reason given: marking there would assert the record had been consulted when it had not. The old line was inert; that one would have been wrong. Trading an idle invariant for a false one is not an improvement.

The new docstring is also right that clearing an unread key would leave it readable afterwards, and right that no caller creates that state. A documented non-guarantee of an unreachable state owes no test.

## Mutation checks I ran

| reverted | result |
|---|---|
| mark removed from `_restore_once` (`main.py:695`) | `test_the_stored_model_is_read_once_per_session` FAILS, `test_a_restored_session_stays_on_the_default_once_cleared` stays green |
| mark removed from `restore_session_model` (`main.py:739`) | all 94 green |

The first row is exactly what the commit message claims. The second is the inline note. Working tree restored to HEAD after each.

## Tests, actually run

```
TERM=dumb uv run pytest tests/ -q -p no:randomly
6477 passed, 43 skipped, 13 deselected, 1 warning in 125.14s
```

Same count and the same skip set as the previous revision -- 8 optional-dependency imports, 33 data-driven `parametrize` exclusions. The test was renamed, not removed, and its assertions are unchanged, so nothing was traded away for the green.

Commit is conventional: `refactor(agent)` is a real subpackage scope, 71-char header, ASCII throughout, `Co-authored-by` present. Test file naming untouched, so AGENTS.md 5.1/5.4 still hold.

## Still open, nonblocking

`raven/cli/provider_commands.py:606` -- the `[red]x[/red]` / `[red]✗[/red]` mismatch.

gloryfromca (dismissed):

gloryfromca (commented):

Delta `8221ad85..e170bb04` is the one-character marker swap I raised, and it is complete: no `[red]x[/red]` left in `provider_commands.py`, and the cross count goes 16 -> 17, so the new refusal now matches the missing-`--provider` refusal three lines above it. Nothing in the tests was coupled to the old literal -- `test_use_refuses_a_misspelled_provider_before_it_writes` asserts on `"antropic" in r.output`, not on the marker -- so this could not have been green by accident.

```
TERM=dumb uv run pytest tests/ -q -p no:randomly
6477 passed, 43 skipped, 13 deselected, 1 warning in 123.66s

ruff check / format --check raven/cli/provider_commands.py    clean
```

Identical counts and skip set to the last three revisions. Commit is conventional: 60-char header, `Co-authored-by` present, and the message describes the glyph in words rather than embedding it, so the message stays ASCII under AGENTS.md 3.1.1 while the source keeps the character it needs.

That closes every item I raised except one, which does not block: `tests/test_agent_loop_session_model.py:592` still names `test_the_stored_model_is_read_once_per_session` as pinning "the marking" when it pins only the one in `_restore_once`. It is a sentence in a docstring, the fix is to name which writer, and it is fine to land as is or to fold in later.

tests/test_agent_loop_session_model.py
arelchan: Taken exactly as suggested, in `f1c52ba3`.

You are right and the framing is the part that stings: this docstring exists to warn the next reader that the clearing test is not a guard on a line, and it then handed them a guard covering half of what it names. The same trap, one line over, written by the person closing it.

Confirmed both mutations independently before changing anything -- removing the mark in `_restore_once` turns `test_the_stored_model_is_read_once_per_session` red, removing the one in `restore_session_model` leaves all 94 green. So it now says which is which:

> ...`test_the_stored_model_is_read_once_per_session` pins the marking in `_restore_once`; the redundant one in `restore_session_model` is unpinned, and deleting it passes the suite.

Took the cheaper of your two options for the reason you gave. The redundant marking stays: its contract is coherent as documented, and a caller supplying the pair genuinely has consulted the record -- which is the honesty test `clear_session_binding` failed and this one passes.

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. Every item from all four rounds is now closed, and I have nothing left open.

Delta `e170bb04..f1c52ba3` is docstring-only -- no code, no assertion changed -- and it says the accurate thing.

One detail worth recording, because the new wording claims more than my note did. I said the redundant marking "can be deleted with the suite green" on the evidence of 94 tests in two files; this docstring says "deleting it passes the suite." So I re-ran mutation B against everything rather than let the stronger claim stand on my narrower check:

```
# _restore_attempted.add removed from restore_session_model
TERM=dumb uv run pytest tests/ -q -p no:randomly
6477 passed, 43 skipped, 13 deselected
```

The whole-suite claim holds -- nothing anywhere else pins that marking. Working tree restored to HEAD afterwards, and `tests/test_agent_loop_session_model.py` is 28 passed at HEAD.

Keeping the redundant marking is the right call for the reason given: a caller who supplies the pair has genuinely consulted the record, which is exactly the property `clear_session_binding` lacked. Redundant and honest is a different thing from inert and false, and the docstring now distinguishes them by name instead of gesturing at "the marking".

Commit is conventional -- 67-char header, ASCII, `Co-authored-by` present.

I checked `docs(test)` against AGENTS.md 3.1's "a top-level subpackage of `raven/`" and am not raising it: neither `commitlint.config.cjs` nor `scripts/commit_lint.py` enforces a scope enum, and the last 400 commits already carry `deps`, `install`, `build`, `readme`, `release`, `tools`, `commitlint` and one prior `test`. The rule is applied loosely here by long practice, so this is precedent rather than a miss.

gloryfromca (approved):

PR #353, opened 2026-08-21
refactor(rpc): rename tui_rpc to rpc for any interactive client

ZuyiZhou:
## Summary

`raven/tui_rpc/` was named after the TUI because the TUI was its only client. It
is now the JSON-RPC surface that any interactive client speaks: the served page
already mounts the same dispatcher, transports and wire vocabulary, and an ACP
agent server mounts them too. The name described the first caller rather than the
module, so this renames it to `raven/rpc/`.

Behaviour is unchanged. The diff is a rename plus one token substitution, across
86 files:

- `raven/tui_rpc/` to `raven/rpc/` (30 modules)
- `tests/test_tui_rpc_*.py` to `tests/test_rpc_*.py` (22), plus 2 under
  `tests/integration/`
- the `tui_rpc` token replaced wherever it appears (395 lines)
- `.github/coverage-baseline.json`: the 29 path keys renamed, every number left
  alone

Every changed line was checked mechanically rather than by eye. Grouping the diff
by file and comparing each removed line, with the substitution applied, against
the added lines gives: files renamed with no content change, files whose content
change is exactly the substitution (reordering included, which absorbs the import
re-sorting ruff applied), 1 new test file, and test files with deliberate
additions. Nothing else.

The wire protocol is untouched. None of the 48 method names in
`ui-tui/rpc-schema/openrpc.json` carried a `tui_rpc` prefix, and the three
`tui_rpc.cli.dispatch` occurrences were loguru log prefixes, not method names.
The only edit to that file is one module path inside `info.description`, which
`ui-tui/scripts/gen-rpc-types.mjs` never reads, so the generated TypeScript is
byte-identical -- confirmed here by `lint:rpc` reporting the generated types in
sync.

Two decisions worth flagging for review:

- **The coverage baseline is not regenerated, only re-keyed.** Both gates read
  `totals` alone (`baseline-check` compares monotonicity, `ratchet` compares live
  totals against the committed ones); neither reads per-file path keys, so those
  29 keys are inert data. Regenerating would have pulled in unrelated drift
  across 78 files and turned a mechanical rename into a line-by-line review.
- **`raven/rpc/__init__.py`'s first docstring line is deliberately left as it
  was.** Updating it is a content change, not a rename, and belongs with the
  follow-up work that touches this module's semantics.

### Rebased onto current `main`

This branch is one commit directly atop `main` (`9da50ac`). It was rebased rather
than merged, per AGENTS.md section 3.5, so the earlier `Merge branch 'main'` commit is
[...]

PR #356, opened 2026-08-22
feat(spine): add turn_id and tool event fidelity to lifecycle events

ZuyiZhou:
## Summary

A turn's lifecycle events now name the turn they belong to, and a tool event carries what a client needs to draw the call.

`turn_id` is the second correlation axis alongside `conversation_id`: the lane is WHERE a turn ran, this is WHICH turn ran. A consumer keyed only on the lane stamps a turn's end with whatever a per-lane slot last held, which is a different turn whenever the runtime submits one of its own onto a busy lane. The lane resolves the id on the way in, so a turn cancelled while still queued can still name itself on its terminal event.

`turn.send` now puts the id it returns onto the request. Without that the lane mints its own, the sink's ownership check never matches what `turn.send` bound, and the active-turn slot is never released, so every later send is refused. The slots are released only by the turn that bound them, for the same reason: a runtime turn ending on a busy lane would otherwise free a queued client turn's bindings and drop its usage before its own end reads it.

A blocked action reaches the client as a `notice` instead of as token deltas. Pushed down the token stream it landed in the buffer holding the model's own prose, so it rendered as the model's answer: run together with whatever had just been narrated, and wearing the answer's copy and branch actions. The blocking tool's own first line rides along as `detail`, because the canned sentence says an operation stopped and never which one.

Every outlet renders that one notice kind, because it replaces the answer rather than accompanying it. `run_turn` marks the reply delivered when it fires, so no surface says the same sentence twice; the terminal and the channel each compose their own copy of it with the `detail`, and a channel that dropped it would end a refused turn in silence -- a refusal and an empty answer are indistinguishable to the person waiting.

An `error` event names the turn that failed, for the same reason `message.complete` does. The lane is shared, so a runtime turn can fail while a client's turn is queued behind it; an ungated failure cleared the queued turn's guard and idled an input the person was still waiting on. The field is optional and omitted rather than blanked, so a connection-level failure and the cancellation a client asked for both pass the guard unchanged. A failure belonging to another turn is still surfaced as a note: that turn may have streamed bytes into the buffer before it died, and hiding its end would leave them the only unexplained thing on screen.

Rendering it in the terminal is not optional: before this event existed the same sentence was at least visible, so a client that received the notice and drew nothing would end the turn in silence. The sentence is written in the component rather than looked up, because this repo has no i18n layer.

`tool.start` carries `blocking`, so a client that clocks the event stream for liveness can stop that clock instead of declaring a call that is merely waiting on a person dead. `tool.complete` carries `metadata` and `diff`: the arguments hold the new content and nothing else, so a client without the diff draws a whole-file write as all additions.

A new gate compares each union variant's payload field names across the schema and the models. The existing union test compares only the set of variants, so a field added to one side alone passed it, while both declarations forbid unknown fields and a validating consumer drops the whole event.

Stacked on #353: this branch is cut from `refactor/rename_tui_rpc_module` because the files it edits only exist at their new paths after that rename. Merge #353 first.

Reviewers: 0xKT, gloryfromca.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run on this head (`a94e647`):

- `uv run --frozen --all-extras pytest -q` -> 7000 passed, 33 skipped, 13 deselected, no failures.
- `make coverage-diff COVERAGE_BASE_REF=origin/refactor/rename_tui_rpc_module` -> 100.00% (72/72 executable changed lines).
- `make coverage-ratchet` -> line 77.30% (+1.74pp), branch 66.57% (+2.53pp), PASS.
- `make lint-python` -> ruff check and format both clean.
- `npm --prefix ui-tui run lint:rpc` -> generated types in sync (the `error` payload's new optional field was regenerated, not hand-edited). `npm run type-check` -> clean. `npm test` -> 92 files, 1094 tests passed.
- Each new test was checked to fail without its fix, by reverting the fix and re-running: the notice branch, the channel outlet's blocked-notice case, the `message.complete` turn guard and the `error` turn guard each turn their own cases red.

## Risk

The behaviour change a user can see is that a blocked action now arrives as runtime prose in its own line rather than inside the assistant's answer. Everything else is additive: three optional payload fields and one new union variant, all with defaults, so a client that ignores them behaves exactly as before.

The `turn_id` change is the one to look at twice. The lane now mints an id when the submitter supplies none, and the sink releases a lane's slots only for the turn that bound them. A mismatch between the two would leave the active-turn slot held and refuse every later send, which is why `turn.send` binding its own id onto the request is part of this change rather than a follow-up.

Rollback is a plain revert: no data migration, no persisted format change, and the wire additions are optional fields.

## Related Issues

N/A

### DIFF
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -77,6 +77,15 @@ _STORE_MAX_INFLIGHT: int = 4
 # Teardown's total budget for letting those writes finish.
 _STORE_DRAIN_BUDGET_S: float = 15.0
 
+
+def _first_line(text: str) -> str:
+    """The one line of a tool error worth putting in front of a person."""
+    for line in str(text or "").splitlines():
+        if line.strip():
+            return line.strip()
+    return ""
+
+
 _ABORTED_ACTION_REPLY = (
     "The operation was not completed, and no alternative method will be attempted. "
     "Would you like me to continue with the remaining parts of the task that do not "
@@ -109,6 +118,7 @@ if TYPE_CHECKING:
     from raven.rpc.question_broker import QuestionBroker
     from raven.sandbox.debug_server import SandboxDebugServer
     from raven.skill_hub import SkillHubClient
+    from raven.spine.events import NoticeKind
     from raven.spine.runner import Drain, Emit, TurnOutcome
     from raven.spine.turn import TurnRequest
     from raven.token_wise.base import UsageSnapshot
@@ -2079,6 +2089,7 @@ class AgentLoop:
         on_reasoning_delta: Callable[[str], Awaitable[None]] | None = None,
         on_tool_event: Callable[[str, dict], Awaitable[None]] | None = None,
         on_episode_start: Callable[[int], Awaitable[None]] | None = None,
+        on_notice: Callable[[NoticeKind, str], Awaitable[None]] | None = None,
         usage_sink: dict[str, Any] | None = None,
         drain: Drain | None = None,
     ) -> tuple[str | None, list[str], list[dict], TurnOutcome]:
@@ -2270,6 +2281,7 @@ class AgentLoop:
 
             if response.has_tool_calls:
                 abort_action = False
+                abort_reason = ""
                 if on_progress:
                     thought = self._strip_think(response.content)
                     if thought:
@@ -2365,6 +2377,11 @@ class AgentLoop:
                         pending_images.extend(attach_blocks)
                     if getattr(result, "abort_action", False):
                         abort_action = True
+                        # The blocking tool's own words, kept for the reader: the
+                        # canned reply below says an operation stopped but never
+                        # which one, so without this the user is told a thing
+                        # happened and given no way to find out what.
+                        abort_reason = _first_line(model_text)
                         # A single assistant message may contain several parallel
                         # tool calls (for example ``rm`` followed by a Python
                         # fallback). Once policy terminates the action, none of
@@ -2400,11 +2417,22 @@ class AgentLoop:
                     # the rejected operation into an equivalent interpreter,
                     # script, or tool call. Finish the turn in runtime code and
                     # expose only the non-destructive continuation question.
-                    # Streaming callers need the explicit callback because no
-                    # final model response exists to generate token deltas.
+                    # The model must read this, so it goes into the history as an
+                    # assistant message -- but it goes to the CLIENT as a notice.
+                    # Pushed down the token stream instead, it arrived as the
+                    # model's own prose: glued to whatever the model had just
+                    # narrated (nothing separates two segments in one buffer),
+                    # dressed in the answer's copy and branch actions, and always
+                    # in English no matter what language the turn was in.
+                    from raven.spine.events import NoticeKind as _NoticeKind
+
                     messages = self.context.add_assistant_message(messages, _ABORTED_ACTION_REPLY)
                     final_content = _ABORTED_ACTION_REPLY
-                    if on_token_delta is not None:
+                    if on_notice is not None:
+                        await on_notice(_NoticeKind.ACTION_BLOCKED, abort_reason)
+                    elif on_token_delta is not None:
+                        # A channel with no notice outlet still has to say
+                        # something, and silence is the worse failure.
                         await on_token_delta(_ABORTED_ACTION_REPLY)
                     break
 
@@ -2634,6 +2662,7 @@ class AgentLoop:
         on_reasoning_delta: Callable[[str], Awaitable[None]] | None = None,
         on_tool_event: Callable[[str, dict], Awaitable[None]] | None = None,
         on_episode_start: Callable[[int], Awaitable[None]] | None = None,
+        on_notice: Callable[[NoticeKind, str], Awaitable[None]] | None = None,
         usage_sink: dict[str, Any] | None = None,
         origin: Origin | None = None,
         drain: Drain | None = None,
@@ -2878,6 +2907,7 @@ class AgentLoop:
             on_reasoning_delta=on_reasoning_delta,
             on_tool_event=on_tool_event,
             on_episode_start=on_episode_start,
+            on_notice=on_notice,
             usage_sink=usage_sink,
             drain=drain,
         )
@@ -3156,6 +3186,7 @@ class AgentLoop:
                         tool_call_id=info["tool_call_id"],
                         name=info["name"],
                         arguments=info["arguments"],
+                        blocking=bool(info.get("blocking")),
                         display=info.get("display"),
                     )
                 )
@@ -3166,8 +3197,13 @@ class AgentLoop:
                         tool_call_id=info["tool_call_id"],
                         result_preview=info["result_preview"],
                         truncated=info["truncated"],
+                        metadata=info.get("metadata"),
+                        diff=info.get("diff"),
                     )
                 )
+
+        async def on_notice(kind: NoticeKind, detail: str) -> None:
+            await emit(Notice(kind=kind, detail=detail or None))
 
         async def on_progress(text: str, tool_hint: bool = False) -> None:
             # Keep the progress/tool-hint distinction so an outlet can gate each on
@@ -3260,6 +3296,7 @@ class AgentLoop:
                 on_reasoning_delta=on_reasoning if stream else None,
                 on_tool_event=on_tool,
                 on_episode_start=on_episode if stream else None,
+                on_notice=on_notice,
                 usage_sink=usage_sink,
                 origin=req.origin,
                 drain=drain,
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -78,6 +78,35 @@ _STORE_MAX_INFLIGHT: int = 4
 _STORE_DRAIN_BUDGET_S: float = 15.0
 
 
+_FILE_CHANGE_MAX_CHARS = 512 * 1024
+
+
+def _file_change_payload(change: Any) -> dict[str, Any] | None:
+    """One write as a plain mapping, or ``None`` when there is nothing to send.
+
+    Flattened here rather than passed as the dataclass: ``spine.events`` is
+    deliberately free of the tools package, and a mapping is also what goes on
+    the wire two hops later.
+
+    ``before`` is preserved as ``None`` when the file did not exist, because a
+    client renders a creation differently from a rewrite -- so this cannot use a
+    "falsy means absent" shortcut, an empty file having the same emptiness.
+    """
+    if change is None:
+        return None
+    after = getattr(change, "after", None)
+    path = getattr(change, "path", None)
+    if not isinstance(after, str) or not isinstance(path, str) or not path:
+        return None
+    before = getattr(change, "before", None)
+    if len(after) + len(before or "") > _FILE_CHANGE_MAX_CHARS:
+        return None
+    payload: dict[str, Any] = {"path": path, "after": after}
+    if before is not None:
+        payload["before"] = before
+    return payload
+
+
 def _first_line(text: str) -> str:
     """The one line of a tool error worth putting in front of a person."""
     for line in str(text or "").splitlines():
@@ -2369,10 +2398,13 @@ class AgentLoop:
                                 # Client-only, straight off the ToolOutput the
                                 # registry built: never shown to the model, and
                                 # ``getattr`` because a tool may still return a
-                                # bare str, which the registry wraps without
-                                # either field.
+                                # bare str, which the registry wraps without any
+                                # of these three.
                                 "metadata": getattr(result, "metadata", None),
                                 "diff": getattr(result, "diff", None),
+                                # Alongside the diff, for a surface that renders
+                                # the change itself rather than a unified form.
+                                "file_change": _file_change_payload(getattr(result, "file_change", None)),
                             },
                         )
                     model_text, blocks, attach_blocks = self._route_result_images(
--- a/raven/agent/loop/main.py
+++ b/raven/agent/loop/main.py
@@ -2328,6 +2328,12 @@ class AgentLoop:
                                 "arguments": tool_call.arguments,
                                 # Tool-authored call label; None -> UI derives one.
                                 "display": _tool.display_call(tool_call.arguments) if _tool else None,
+                                # The same flag the registry reads to skip its
+                                # timeout: this call waits on a human, so it has
+                                # no deadline and may emit nothing while it runs.
+                                # A client that clocks the stream needs to know
+                                # before the wait, not after it.
+                                "blocking": bool(_tool.blocking_interaction) if _tool else False,
                             },
                         )
                     if tool_call.name == "exec":
@@ -2360,6 +2366,13 @@ class AgentLoop:
                                 "tool_call_id": tool_call.id,
                                 "result_preview": preview,
                                 "truncated": len(display_src) > 200,
+                                # Client-only, straight off the ToolOutput the
+                                # registry built: never shown to the model, and
+                                # ``getattr`` because a tool may still return a
+                                # bare str, which the registry wraps without
+                                # either field.
+                                "metadata": getattr(result, "metadata", None),
+                                "diff": getattr(result, "diff", None),
                             },
                         )
                     model_text, blocks, attach_blocks = self._route_result_images(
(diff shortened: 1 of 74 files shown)


raven/agent/loop/main.py
```diff
@@ -3240,6 +3276,7 @@ async def _route_deep_research(kind: str, text: str) -> None:
                 on_reasoning_delta=on_reasoning if stream else None,
                 on_tool_event=on_tool,
                 on_episode_start=on_episode if stream else None,
+                on_notice=on_notice,
```
gloryfromca, line 3354: [P1] Suppress the fallback Text after emitting the notice

`on_notice` does not set `streamed`, while `_process_message` still returns `_ABORTED_ACTION_REPLY`. The post-call boundary therefore emits `Text(reply_content)`, and `TuiOutlet` maps that straight back to `token.delta`. A direct `run_turn(..., stream=True)` denial produced both `Notice(ACTION_BLOCKED)` and `Text(_ABORTED_ACTION_REPLY)`, so the runtime sentence still appears in the model answer and is also shown as a notice. Mark this terminal reply as already delivered or otherwise suppress the boundary `Text` when the notice path is used.

raven/agent/loop/main.py
```diff
@@ -3136,6 +3166,7 @@ async def on_tool(phase: str, info: dict[str, Any]) -> None:
                         tool_call_id=info["tool_call_id"],
                         name=info["name"],
                         arguments=info["arguments"],
+                        blocking=bool(info.get("blocking")),
```
gloryfromca, line 3241: [P1] Populate the fidelity fields at the producer

These `info.get` calls always take their defaults in real turns. `_run_agent_loop` constructs the start callback dictionary with only id/name/arguments/display, and the complete dictionary with only id/preview/truncated; `ToolResult`/`ToolOutput` also exposes neither metadata nor diff. A direct blocking-tool probe emitted a start dictionary without `blocking` and a complete dictionary without `metadata` or `diff`, so the wire always receives `blocking: false`, `metadata: null`, and `diff: null`. Populate the start value from the tool and extend the result contract/producer for metadata and diff; cover the real loop callback rather than only synthetic `TuiOutlet` events.

raven/rpc/spine.py
```diff
@@ -243,18 +275,20 @@ def _drop(conversation_id: str) -> None:
...
                 "completion_tokens": 0,
                 "total_tokens": 0,
             }
-            _drop(event.conversation_id)
-            await outlet.emit_complete(event.conversation_id, turn_id, usage)
+            # Read before _drop pops the register.
+            _drop(event.conversation_id, owns=_owns_lane(event.conversation_id, event.turn_id))
+            # The ending turn's own id, never the lane slot's current value: the
+            # slot may hold a client turn that has not started yet.
+            await outlet.emit_complete(event.conversation_id, event.turn_id, usage)
```
gloryfromca, line 337: [P1] Correlate completions before clearing the TUI turn

Sending the ending id is not sufficient because `chatStream.onMessageComplete` ignores `payload.turn_id` and unconditionally clears `state.turnId` and finalizes the current buffer. I reproduced the intended busy-lane case by running a SUBAGENT turn on `tui:default`, calling `turn.send` while it was running, then releasing the system turn; the wire order was `message.start(client)`, `message.complete(runtime)`, `message.complete(client)`. The first completion therefore clears the queued client's guard and finalizes the wrong buffer even though the ids differ. Gate completion state changes by turn id while preserving the system reply, or restructure start/completion ordering so one turn cannot terminate another.

gloryfromca (commented):
Blocking: three lifecycle/tool-event paths still break their advertised correlation and fidelity; see the inline notes.

I reviewed the live head against `github/refactor/rename_tui_rpc_module`, including the complete diff, AGENTS.md and the Runtime/TUI domain contracts, callers through delivery and UI consumption, the single-commit history, backward compatibility, and whether tests were weakened. I found no weakened tests; the gaps are that the new tests stop at synthetic callback/outlet shapes instead of exercising the real producer and final client state.

Verification:
- `uv run pytest tests/test_agent_loop_approval.py tests/test_rpc_schema_match.py tests/test_rpc_spine.py tests/test_spine_events.py tests/test_spine_scheduler_lane.py tests/test_rpc_tool_events.py tests/test_rpc_turn_send.py tests/test_rpc_turn_cancel.py -q` -> 232 passed.
- `npm test -- --run src/__tests__/chatStream.test.ts src/__tests__/turnControllerEpisodes.test.ts` -> 2 files, 20 tests passed.
- `npm run type-check` -> passed.
- `npm run lint:rpc` -> generated RPC types are in sync.
- Direct runtime probes reproduced all three inline failures.

raven/agent/loop/main.py
ZuyiZhou: Fixed in 53f989c. `on_notice` now marks the reply delivered when the kind is `ACTION_BLOCKED`, and the boundary checks that alongside `streamed`. Keyed on the kind rather than on "a notice fired": the other four accompany the answer, so gating on any of them would silence a turn that merely reported progress -- `test_a_normal_turn_still_gets_its_closing_text` is the discriminator.

`text_sink` still receives the sentence. The cron read-back and the gateway delivery path read it there, and a channel with no notice outlet is the case the `on_token_delta` fallback already covers.

raven/agent/loop/main.py
ZuyiZhou: Fixed in 53f989c. `blocking` now comes from `_tool.blocking_interaction` -- the same flag the registry reads to skip its timeout -- and `ToolResult`/`ToolOutput` carry `metadata` and `diff` through the registry boundary, so the producer reads them off the result rather than defaulting. An error drops both, as it already drops `blocks`.

The writing tools that produce a diff are in the ACP commit upstack (`_unified` plus `FileChange` on write_file and edit_file), so I did not add a second implementation one PR lower; what this commit owes them is a live path. Tests go through `_run_agent_loop`: `test_a_blocking_tool_says_so_on_its_start_event` with its negative twin, `test_metadata_and_diff_reach_the_complete_event`, and `test_a_tool_returning_bare_text_publishes_neither_field`.

raven/rpc/spine.py
ZuyiZhou: Fixed in 53f989c, in `chatStream.onMessageComplete`. A completion whose `turn_id` differs from the turn being watched no longer clears `state.turnId` or finalizes the buffer; usage still lands, since it is session-cumulative and the property of neither turn. A completion arriving with no turn being watched still finalizes, so a replayed or resumed stream cannot lose its content.

The system reply is preserved by leaving it in the buffer: the owner`s completion commits everything streamed, including what arrived while the runtime turn was ending. `ignores a completion for a turn other than the one it is watching` fails without the guard -- verified by reverting it.

raven/agent/loop/main.py
gloryfromca: Confirmed that the TUI duplicate is gone, but the fallback claim does not hold for non-streaming channels. `run_turn` always passes its internal `on_notice` callback to `_process_message`, including through `AgentTurnRunner(stream=False)`, so `_process_message` never reaches `elif on_token_delta`. The new `replaced_by_notice` flag then suppresses the closing `Text`, while `ChannelOutletAdapter` deliberately eats `Notice`. I drove a denied exec through Scheduler -> AgentTurnRunner(stream=False) -> DeliveryHub -> ChannelOutletAdapter; the turn ended normally and `channel.sent` was empty. Please preserve `Text` for outlets that cannot render `ACTION_BLOCKED` notices, or make notice capability explicit.

raven/agent/loop/main.py
gloryfromca: Confirmed on `53f989c1`. The real `_run_agent_loop` producer now gets `blocking` from the tool, `ToolResult`/`ToolOutput` carries `metadata` and `diff` through the registry, errors strip those client-only fields, and the focused producer/registry tests pass. This resolves my finding.

raven/rpc/spine.py
gloryfromca: The successful-completion guard is effective, but lifecycle correlation is still missing on failure. `TurnFailed` carries `turn_id` internally, yet `TuiOutlet.emit_error` sends no id and `chatStream.onError` unconditionally clears `state.turnId`. I reproduced a runtime system turn failing while a client turn was queued: `message.start(client)`, `error(no turn_id, internal)`, `message.complete(client)`. That foreign error clears the queued client's guard just as the prior foreign completion did. Please propagate and gate failure identity, or otherwise keep a runtime turn's failure from terminating active client state.

gloryfromca (commented):
Blocking: tool fidelity and successful-completion correlation are fixed, but blocked actions are silent on non-streaming channels and foreign runtime failures still clear queued client turns.

I reviewed the `39129f0e..53f989c1` fix delta and the complete target-range change, including AGENTS.md and the Runtime/TUI domain contracts, callers through the delivery outlets and UI terminal handlers, commit history, backward compatibility, and whether tests were weakened. No tests were weakened; the new tests cover the corrected producer and successful-completion paths but omit the two failing outlet/failure shapes described in my replies. The merge-tree against the fetched target tip is clean.

Verification:
- `uv run pytest tests/test_agent_loop_approval.py tests/test_rpc_schema_match.py tests/test_rpc_spine.py tests/test_spine_events.py tests/test_spine_scheduler_lane.py tests/test_rpc_tool_events.py tests/test_tool_registry_execute.py tests/test_rpc_turn_send.py tests/test_rpc_turn_cancel.py tests/test_channels_outlet.py tests/test_cli_gateway_spine.py -q` -> 278 passed.
- `npm test -- --run src/__tests__/chatStream.test.ts src/__tests__/turnControllerEpisodes.test.ts` -> 2 files, 22 tests passed.
- `npm run type-check` -> passed.
- `npm run lint:rpc` -> generated RPC types are in sync.
- Direct channel probe: denied action ended normally but `ChannelOutletAdapter` delivered no message.
- Direct busy-lane failure probe: `message.start(client)`, uncorrelated `error(runtime)`, then `message.complete(client)`.

raven/agent/loop/main.py
ZuyiZhou: Fixed in 79eb3c2, and you are right that the fallback claim did not hold -- `run_turn` wires `on_notice` unconditionally, so the channel path never reaches `elif on_token_delta`.

I took the second option you offered rather than the first: `ChannelOutletAdapter` now renders `ACTION_BLOCKED` instead of eating it. That is what `NoticeKind` already says about this kind -- it replaces the answer, so an outlet that renders nothing else should still render it -- and it is what the ACP surface upstack already does. The channel adapter was the only outlet dropping it. Preserving the `Text` instead would have put the sentence back in the model`s voice on any outlet that renders both, which is the defect the notice path exists to prevent.

The detail is the message, with a sentence of its own when a tool aborts without a readable line; written in the outlet rather than looked up, the same way the terminal writes its own copy. Two tests: `test_a_blocked_action_is_not_eaten` plus the no-detail case in `tests/test_channels_outlet.py`, and `test_a_denied_action_still_reaches_a_channel_that_cannot_draw_notices`, which drives a real denial through `run_turn(stream=False)` into the adapter and asserts exactly one message -- it fails with the outlet change reverted.

raven/rpc/spine.py
ZuyiZhou: Fixed in 79eb3c2. `ErrorEventPayload` carries an optional `turn_id`, `emit_error` takes it, the sink passes `event.turn_id`, and `chatStream.onError` gates on it the same way `onMessageComplete` does.

Optional and omitted rather than blanked: a connection-level failure belongs to no turn, and the cancellation this client asked for belongs to the turn it is watching, so both fall through the guard unchanged -- `turn.cancel` is left alone. Wire sync done in all four places (models, openrpc, regenerated types, and the match test).

Tests: `test_a_failure_names_the_turn_that_failed` drives the busy-lane shape through `_make_tui_sink` with the lane slot holding the queued client turn, `test_a_failure_with_no_turn_to_name_omits_the_field` pins the absence, and the vitest pair asserts the foreign failure leaves `busy` set while an untagged one still surfaces. Verified both by reverting the guard.

raven/agent/loop/main.py
gloryfromca: Confirmed that the channel is no longer silent and that the duplicate delivery is gone; the direct `run_turn(stream=False)` test exercises the right path. The outlet still replaces the controlled blocked-turn response with the tool error whenever `detail` is present, though, so I left a current-head inline note on that remaining regression.

raven/rpc/spine.py
gloryfromca: Confirmed that the wire now carries the failing runtime turn`s own id and the guard preserves the queued client turn`s busy state; the focused producer/schema/TUI tests pass. The guard currently returns before surfacing that failure at all, so I left a current-head inline note on the resulting error-data loss.

raven/channels/outlet.py
```diff
@@ -37,4 +53,8 @@ async def deliver(self, out: Deliverable) -> None:
             await self._channel.send(out.source.chat_id, out.content)
         elif isinstance(out, MediaOut):
             await self._channel.send(out.source.chat_id, "", media=[m.path for m in out.media])
-        # StreamDelta / Reasoning / ToolEvent / Notice: eaten — render-can't path.
+        elif isinstance(out, Notice) and out.kind is NoticeKind.ACTION_BLOCKED:
+            detail = (out.detail or "").strip()
+            await self._channel.send(out.source.chat_id, detail or _BLOCKED_FALLBACK)
```
gloryfromca: [P1] Preserve the blocked-turn response when detail is present

`detail` is the tool's first error line; it was added to say which operation was blocked, not to replace the runtime's controlled final response. This `detail or fallback` makes the normal exec-denial path send only `Error: User denied this command or the approval request expired`, so channel users no longer learn that no alternate will be attempted or get the prompt to continue with the safe remainder. The TUI correctly renders its blocked-turn sentence plus `detail`. Please compose the channel's blocked-turn sentence with the detail too, and make the direct denial test assert that content rather than only the send count.

ui-tui/src/app/chatStream.ts
```diff
@@ -237,6 +282,15 @@ const onError = (
...
 ): void => {
   const { reason, message, code, detail } = ev.payload
+  // Same correlation the completion path needs, for the same reason: a turn the
+  // runtime submitted can fail while this client's turn is queued behind it on
+  // the same lane, and an ungated failure clears the queued turn's guard and
+  // idles an input the user is still waiting on. A failure with no turn_id --
+  // a connection-level one, or the cancellation this client asked for -- is
+  // this client's business by construction, so it falls through.
+  if (state.turnId && ev.payload.turn_id && ev.payload.turn_id !== state.turnId) {
+    return
```
gloryfromca, line 310: [P1] Do not discard a correlated foreign failure

The id guard must preserve `state.turnId`, but returning here suppresses the failure before `sys` sees it. Runtime turns share this conversation stream: one can emit token deltas and then throw, leaving those partial bytes in `turnController`; the queued client's later completion then commits them while this guard has hidden the only indication that the producing turn failed. Surface or mark the foreign error without clearing the watched turn's state (or isolate/discard its failed output); correlation should not turn the terminal event into silence. The new test currently locks in the loss with `expect(...).toBe(false)`.

gloryfromca (commented):
Blocking: the channel fallback drops the controlled blocked-turn response, and foreign failures are correlated by being hidden instead of surfaced; see the inline notes.

I reviewed the new `79eb3c2` delta after separating the target rebase with range-diff. I covered AGENTS.md and the applicable CONTEXT vocabulary, the full target diff and new commit, callers through the agent loop, scheduler, outlets, and TUI, history/rebase equivalence, backward compatibility of the optional wire field and generated schema, and whether tests were weakened. The earlier channel-silence and queued-turn state-corruption failures are fixed, but the fixes leave two user-visible errors.

Verification: `git diff --check` clean; merge-tree against `github/refactor/rename_tui_rpc_module` clean; 283 focused Python tests passed; the focused chatStream suite passed 19/19; generated RPC types are in sync; TypeScript type-check passed. A direct channel probe produced only `Error: User denied this command or the approval request expired`, confirming the first regression.

raven/channels/outlet.py
ZuyiZhou: Fixed in 275fbe1. You are right that `detail` says which operation was stopped and nothing else -- sending it alone reduced a denied turn to the tool`s error line, with no word that no alternative would be attempted and no offer to continue with the safe remainder.

The outlet now composes its own sentence with the detail, the way the terminal composes `NOTICE_TEXT[kind]` with it, and the sentence is the response rather than a fallback for a missing detail (the no-detail case sends the sentence alone, with no dangling second half). The direct denial test asserts the content: `test_a_denied_action_still_reaches_a_channel_that_cannot_draw_notices` now requires both halves, and the outlet test pins the exact composed string.

ui-tui/src/app/chatStream.ts
ZuyiZhou: Fixed in 275fbe1 -- correlating it should not have made it silent, and the point about the partial bytes is the sharper half: a runtime turn can stream deltas into this buffer and then throw, and the watched turn`s completion commits them as if they were its answer.

The guard now says so through `sys` -- "error in another turn on this session: ..." with the same first-line detail the normal path shows -- and still changes no turn state, so the queued turn keeps its guard and its input. I took the surface-or-mark option rather than isolate-or-discard: per-turn buffers are a real change to `turnController`, and this note is what makes those bytes explainable in the meantime. Say the word if you would rather see the isolation done here instead.

The test now asserts the note is surfaced rather than locking in the loss.

PR #357, opened 2026-08-22
feat(rpc): add a reusable engine assembly and a working-directory guard

ZuyiZhou:
## Summary

Two pieces a transport that is not a socket needs before it can serve turns, plus the parameter that lets it pick its own delivery channel.

`build_rpc_stack` is the engine half of the socket path's startup: an AgentLoop, the three prompt brokers, a `SubscriptionEmitter`, the spine scheduler, the method registrations and a teardown, with the transport left to the caller.

The socket path deliberately does NOT call it, and the reason is in the module docstring: `RpcServer` is constructed FROM the dispatcher and only then can provide `send_frame`, while this function creates the dispatcher itself. Inverting that changes the startup path every terminal session depends on and buys nothing until a second socket transport exists. So the duplication is bounded and the reason is written down rather than left for somebody to rediscover.

What this assembly does not wire is listed in the same docstring rather than left as a diff to interpret: no DAG progress sink, sub-agent delivery sink or MCP event sink, because nothing here sets them; no direct-chat target map; and no per-connection scoping of the approval and question brokers, because one process serves one client here, so a broadcast IS the one client.

The channel is a parameter now, and it has to reach BOTH collaborators. The delivery outlet is registered under the channel name and `turn.send` stamps it on every turn it submits as `source.channel`; the hub routes a deliverable by that name, so giving it to one side and not the other loses every turn with no error anywhere. A test pins that. The default is the value both sides independently hardcoded before, so existing callers are unchanged.

`approval_responder` replaces the transport shell approvals are asked over, without touching the classification, the one-command scope, or the absence of any always-allow state. The local broker is still built either way, because `approval.respond` is registered from it.

`workdir.validate_override` is the guard a caller needs before running anything in a directory somebody else named. The refused answers are agent home itself, any ancestor of it, and three of its subtrees, each for a different reason written at the check. The ancestor case is the sharp one: the realistic ancestor is the instance data directory holding provider keys and tokens, and a per-turn checkpoint that adds everything under the working directory would commit them into a shadow repo. Paths are resolved before comparison, so a symlink into a protected subtree is refused too.

Only the guard is here. The per-channel default roots and the per-turn binding that belong to a full working-directory feature are left out, because nothing on this side asks where a turn should run yet.

Both modules were landed with their tests as the only caller, so the transport that uses them stayed reviewable on its own.

### #359 is now part of this branch

`feat(acp): serve the agent client protocol over stdio` (#359) was squash-merged into this branch after its approval, so this PR now carries the ACP agent surface as well: `raven acp` speaking newline-delimited JSON-RPC over stdio against the pinned v1 schema, and the shell-approval and turn-correlation work that came with it. `build_rpc_stack` and `workdir.validate_override` are what it mounts, which is why it landed here rather than on `main`.

One conflict came out of that merge and is resolved in this branch: `ErrorEventPayload.turn_id` was declared by both PRs -- twice in `raven/rpc/models.py` and twice in `ui-tui/rpc-schema/openrpc.json`, which Python and JSON both accept silently. The field now comes from #356, one PR lower, and this branch declares it once. The generated TypeScript was regenerated rather than hand-edited.

Stacked on #356, which is itself stacked on #353. Merge order: 353, 356, then this.

Reviewers: 0xKT, gloryfromca.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run on this head (`c7640ea`), which includes the merged #359:

- `uv run --frozen --all-extras pytest -q` -> 7734 passed, 33 skipped, 13 deselected, no failures.
- `uv run pytest tests/integration/ -m integration -q` -> 32 passed (the real-binary ACP suites, which are deselected by default).
- `make coverage-diff COVERAGE_BASE_REF=origin/feat/spine_turn_correlation` -> 99.56% (2035/2044 executable changed lines).
- `make coverage-ratchet` -> line 78.26% (+2.70pp), branch 68.16% (+4.12pp), PASS.
- `make lint-python` -> ruff check and format both clean.
- `npm --prefix ui-tui run lint:rpc` -> generated types in sync. `npm run type-check` -> clean. `npm test` -> 92 files, 1094 tests passed.
- The teardown regression was checked to fail without its fix, and it crosses a real worker thread: the previous cancel-only code leaves the spawn side effect landing after `stop()`. A second case drives the rejected shape under `asyncio.run` in a subprocess, recording both that ordering and the fact that the process waits for the worker regardless -- which is why the wait is not on a timer.

## Risk

Two new modules and three new keyword arguments, all defaulted. No existing call site changes behaviour: the channel default is the value both sides already hardcoded, and `approval_responder=None` keeps the locally built broker.

The one thing to look at is the teardown ordering in `build_rpc_stack`: the memory backend is drained before it is stopped, because stopping without draining loses what the last turn wrote, and the stop is what releases the embedded index lock the next process needs. Each step is guarded on its own so a failure in one does not skip the rest; a test drives a teardown where cron stop, the spine teardown and the backend stop all raise, and asserts the drain is still reached.

Rollback is a plain revert. Nothing imports either module in production yet, so a revert cannot break a caller.

## Related Issues

N/A

### DIFF
--- a/raven/rpc/bootstrap.py
+++ b/raven/rpc/bootstrap.py
@@ -1,0 +1,235 @@
+"""Reusable RPC stack assembly for a transport that owns its own frame sink.
+
+The engine half of ``tui_commands._run_rpc_server_until_done``: an AgentLoop,
+the three prompt brokers, a SubscriptionEmitter, the spine scheduler and the
+method registrations, minus any transport. The caller supplies ``send_frame``,
+so a transport that is not a socket (a stdio protocol, a WebSocket) gets the
+same engine without reimplementing the wiring.
+
+Why the socket path does not call this, though its shape came from there:
+``RpcServer`` is constructed FROM the dispatcher and only then can provide
+``send_frame``, while this function creates the dispatcher itself. Inverting
+that is a change to the startup path every terminal session depends on, and it
+buys nothing until a second socket transport exists. So the duplication is
+deliberate and bounded, and the reason is written here rather than left for
+somebody to rediscover.
+
+What a caller does NOT get here, and why -- so that anyone reconciling this file
+with a fuller assembly elsewhere has the list rather than a diff to interpret:
+a DAG progress sink, a sub-agent delivery sink and an MCP event sink (nothing on
+this side sets them), the direct-chat target map (no direct chat here), and
+per-connection scoping of the approval and question brokers (one process serves
+one client here, so a broadcast IS the one client; scoping matters only where
+several sockets share a dispatcher).
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass
+from typing import Any
+
+from loguru import logger
+
+SendFrame = Callable[[dict[str, Any]], Awaitable[None]]
+
+
+@dataclass
+class RpcStack:
+    """Everything a transport needs back from the assembly.
+
+    ``build_error`` is latched rather than raised: a bad provider config must
+    still let the client connect and be told what is wrong, which is why the
+    factory closure below re-raises it on first use instead.
+    """
+
+    dispatcher: Any
+    emitter: Any
+    agent_loop: Any
+    build_error: Any
+    teardown: Callable[[], Awaitable[None]]
+    # The ask_user / deep-research broker. A caller that answers questions its
+    # own way (a protocol with an elicitation method of its own) needs the handle
+    # to rebind or wrap it.
+    question_broker: Any = None
+
+
+async def build_rpc_stack(
+    send_frame: SendFrame,
+    *,
+    channel: str = "tui",
+    approval_responder: Any = None,
+) -> RpcStack:
+    """Assemble dispatcher + engine wired to ``send_frame``.
+
+    Must run inside the event loop that will serve requests: cron start and the
+    spine scheduler both bind to the running loop.
+
+    ``channel`` names the delivery channel this stack's turns run on, and it has
+    to reach BOTH collaborators. ``build_tui`` registers its outlet under this
+    name and ``register_turn_methods`` stamps it on every turn it submits as
+    ``source.channel``; the hub routes a deliverable by that name, so passing it
+    to one and not the other loses every turn with no error anywhere. It is its
+    own channel per surface and not a shared name because session keys are
+    prefixed with it and the session listing filters on that prefix -- sharing
+    would put one surface's sessions in another's picker.
+
+    ``approval_responder`` replaces the transport shell approvals are asked
+    over. The broker built here emits ``approval.request`` on the same
+    ``send_frame``, which is right for a client that implements that method and
+    useless for one that does not. Only the transport is replaced: the
+    classification, the one-command scope and the absence of any always-allow
+    state all stay where they are. The local broker is still constructed either
+    way, because ``approval.respond`` is registered from it and a caller
+    overriding the responder is not necessarily removing that method.
+    """
+    from raven.cli.tui_commands import (
+        _build_cron_callback_spine,
+        _build_tui_agent_loop,
+        _fanout_cron_missed,
+    )
+    from raven.rpc.approval_broker import ApprovalBroker
+    from raven.rpc.confirm_broker import ConfirmBroker
+    from raven.rpc.dispatcher import Dispatcher
+    from raven.rpc.errors import RpcError
+    from raven.rpc.methods import register_aligned_methods_except_system
+    from raven.rpc.methods import turn as turn_module
+    from raven.rpc.methods.system import register_system_methods
+    from raven.rpc.question_broker import QuestionBroker
+    from raven.rpc.spine import build_tui
+    from raven.rpc.subscriptions import SubscriptionEmitter
+
+    dispatcher = Dispatcher()
+    emitter = SubscriptionEmitter(send_frame=send_frame)
+    # Three brokers rather than one: a shell approval is not a conversational
+    # confirmation. It binds one exact command to one turn, has dual deadlines,
+    # and always fails closed when the transport disappears.
+    confirm_broker = ConfirmBroker(send_frame=send_frame)
+    approval_broker = ApprovalBroker(send_frame=send_frame)
+    question_broker = QuestionBroker(send_frame=send_frame)
+
+    # Eager, so a bad provider config surfaces at connect rather than on the
+    # first turn; latched, so the client still gets to connect and be told.
+    build_error: RpcError | None = None
+    agent_loop = None
+    try:
+        agent_loop = _build_tui_agent_loop()
+    except RpcError as e:
+        build_error = e
+
+    # Late-bound now that the tool registry exists. deep_research goes through
+    # the loop rather than the tool, so a tool built later by a mid-session
+    # enable inherits the broker too.
+    if agent_loop is not None:
+        if (ask_tool := agent_loop.tools.get("ask_user")) is not None and hasattr(ask_tool, "set_broker"):
+            ask_tool.set_broker(question_broker)
+        agent_loop.set_deep_research_broker(question_broker)
+
+    def _agent_loop_factory():
+        if agent_loop is not None:
+            return agent_loop
+        if build_error is not None:
+            raise build_error
+        return None
+
+    turn_scheduler = None
+    turn_ids: dict[str, str] = {}
+    turn_teardown = None
+    if agent_loop is not None:
+        from raven.cli._cron_handler import make_on_cron_job
+
+        # The spine is built before cron is wired: a reminder submits a CRON turn
+        # through this scheduler, captured non-streaming and read back so the
+        # wrapper can fan it out as a cron.delivered event.
+        cron_readback: dict[str, str] = {}
+        turn_scheduler, _turn_hub, turn_ids, turn_teardown = build_tui(
+            agent_loop,
+            emitter,
+            channel=channel,
+            on_turn_end=turn_module.clear_active,
+            readback_texts=cron_readback,
+            approval_responder=approval_responder or approval_broker,
+        )
+        agent_loop.subagents.set_submit(turn_scheduler.submit)
+        if agent_loop.cron_service is not None:
+            base_on_cron = make_on_cron_job(
+                submit=turn_scheduler.submit,
+                readback_texts=cron_readback,
+                default_channel=channel,
+                cron_service=agent_loop.cron_service,
+            )
+            agent_loop.cron_service.on_job = _build_cron_callback_spine(base_on_cron, emitter)
+            # on_job must be wired before start(), or an immediately-firing job
+            # has no callback.
+            await agent_loop.cron_service.start()
+            # start() dropped past-due one-shot reminders on this runner's
+            # partition. A transport that reaches the runtime through here rather
+            # than through tui_commands would otherwise collect the drops and
+            # never tell anyone.
+            if agent_loop.cron_service.last_startup_drops:
+                await _fanout_cron_missed(emitter, drops=agent_loop.cron_service.last_startup_drops)
+
+    register_system_methods(dispatcher)
+    register_aligned_methods_except_system(
+        dispatcher,
+        emitter=emitter,
+        agent_loop_factory=_agent_loop_factory,
+        approval_broker=approval_broker,
+        confirm_broker=confirm_broker,
+        question_broker=question_broker,
+        scheduler=turn_scheduler,
+        turn_ids=turn_ids,
+        build_error=build_error,
+        default_channel=channel,
+    )
+
+    if agent_loop is not None and agent_loop.backend is not None:
+        # Backgrounded because it may spawn a server and take tens of seconds;
+        # a first render must not wait on the memory path.
+        async def _start_backend() -> None:
+            try:
+                await agent_loop.backend.start()
+            except Exception:
+                logger.exception("rpc: memory backend start failed; continuing with degraded memory path")
+
+        asyncio.create_task(_start_backend())
+
+    async def teardown() -> None:
+        # Pending UI waits are released before the transport goes away.
+        # Cancelling an approval is denial, which keeps a disconnect fail-closed;
+        # an ordinary confirm keeps its configured default.
+        confirm_broker.cancel_all()
+        approval_broker.cancel_all()
+        if agent_loop is not None and agent_loop.cron_service is not None:
+            try:
+                agent_loop.cron_service.stop()
+            except Exception:
+                logger.exception("rpc: cron stop failed; continuing shutdown")
+        if turn_teardown is not None:
+            try:
+                await turn_teardown()
+            except Exception:
+                logger.exception("rpc: turn spine teardown failed; continuing shutdown")
+        # Drain before stop, and both before returning: the drain flushes what a
+        # turn wrote and has not persisted, and the stop releases the embedded
+        # index lock the next process needs. Stopping without draining loses the
+        # last turn's memory writes silently.
+        if agent_loop is not None and agent_loop.backend is not None:
+            try:
+                await agent_loop.drain_backend_stores()
+                await agent_loop.backend.stop()
+            except Exception:
+                logger.exception("rpc: memory backend stop failed; continuing shutdown")
+
+    return RpcStack(
+        dispatcher=dispatcher,
+        emitter=emitter,
+        agent_loop=agent_loop,
+        build_error=build_error,
+        teardown=teardown,
+        question_broker=question_broker,
+    )
+
+
+__all__ = ["RpcStack", "SendFrame", "build_rpc_stack"]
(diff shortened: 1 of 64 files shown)


raven/rpc/bootstrap.py
```diff
@@ -0,0 +1,235 @@
...
+    if agent_loop is not None and agent_loop.backend is not None:
+        # Backgrounded because it may spawn a server and take tens of seconds;
+        # a first render must not wait on the memory path.
+        async def _start_backend() -> None:
+            try:
+                await agent_loop.backend.start()
+            except Exception:
+                logger.exception("rpc: memory backend start failed; continuing with degraded memory path")
+
+        asyncio.create_task(_start_backend())
```
gloryfromca: [P1] Retain and settle the backend-start task before teardown returns

This fire-and-forget task can still be inside `backend.start()` when `teardown()` drains and stops the backend. Because its handle is discarded, teardown then returns while start remains live. I reproduced the sequence with a gated backend: immediately after teardown the events were `['start-enter', 'drain', 'stop']`; releasing the start gate afterward appended `'start-finish'`. A client that connects and closes while EverOS is still starting can therefore leave the service or adapter live after the stack reports itself closed, including retaining the lock the next process needs. Keep the task and make teardown settle it before stopping/returning, with a regression test for start overlapping teardown.

raven/rpc/bootstrap.py
```diff
@@ -0,0 +1,235 @@
...
+
+    async def teardown() -> None:
+        # Pending UI waits are released before the transport goes away.
+        # Cancelling an approval is denial, which keeps a disconnect fail-closed;
+        # an ordinary confirm keeps its configured default.
+        confirm_broker.cancel_all()
+        approval_broker.cancel_all()
```
gloryfromca, line 210: [P2] Release pending question waits along with the other brokers

The stack constructs and exports `question_broker`, but teardown cancels only confirm and approval waits. A direct probe that started `question_broker.await_question(..., timeout_s=60)` and then awaited `stack.teardown()` left the question task unfinished (`done() == False`), despite the comment above promising pending waits are released. On a transport disconnect outside an active Spine turn, that ask can remain alive until the broker's 600-second default timeout. Call `question_broker.cancel_all()` here and cover the pending-question teardown path.

gloryfromca (commented):
Blocking: teardown does not fully close the resources this assembly owns; see the two inline notes.

I reviewed head `a6481f777c07` against `github/feat/spine_turn_correlation`. I covered the repository rules (`AGENTS.md`, `CONTEXT-MAP.md`, and the relevant Runtime terms), the complete diff, the existing socket assembly, affected callers through dispatch and teardown, history, backward-compatible defaults, and the added and nearby tests. The tests were not weakened; the changed tests are additive.

Verification:

- `uv run --frozen pytest -q tests/test_agent_workdir.py tests/test_rpc_bootstrap.py tests/test_rpc_turn_send.py tests/test_rpc_turn_subscribe.py tests/test_rpc_turn_cancel.py tests/test_cli_tui_bootstrap.py` -> 81 passed.
- `uv run --frozen --all-extras pytest -q` with inherited HTTP proxy variables unset -> 6768 passed, 1 failed, 33 skipped, 13 deselected. The sole failure, `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`, reproduces identically on target head `39129f0e`.
- Focused Ruff check and format check passed; `git diff --check` passed.
- Direct async probes reproduced both teardown findings.

I found no other blocker or plain error.

raven/rpc/bootstrap.py
ZuyiZhou: Fixed in 707a152. The task is held in `backend_start` and settled before the stop: cancelled rather than awaited, because `start` may take tens of seconds and teardown is the exit path, and `MemoryBackend.stop` is contractually safe after a failed start (it cleans up partial-init state), so cancel-then-stop releases whatever a half-finished start created. Running the two concurrently would have `stop` racing it.

`test_teardown_settles_a_start_still_in_flight_before_it_stops` uses a gated backend and asserts the recorded order is `[start-enter, stop]`; it then releases the gate and asserts nothing more is appended, so the task is settled rather than merely unobserved.

raven/rpc/bootstrap.py
ZuyiZhou: Fixed in 707a152 -- `question_broker.cancel_all()` sits with the other two, so the comment above them is true of all three now.

One note on the test: it asserts `waiting.done()` immediately after teardown instead of awaiting the task. `await_question` returns its default on cancellation too, so `asyncio.wait_for` would have passed with no fix at all -- the timeout doing the releasing. I measured that before settling on the shape.

raven/rpc/bootstrap.py
gloryfromca: Thanks for adding the regression test, but the production path refutes the assumption that cancel-plus-await settles startup. `EverosBackend.start()` awaits `asyncio.to_thread(_start_server_if_unlocked, base_url)` in `raven/plugin/memory/everos/_server.py`; cancelling the asyncio task does not stop a worker thread that is already running. I reproduced that shape on `707a152`: teardown left `['thread-enter', 'stop']`, then releasing the worker after teardown appended `'thread-side-effect'`. In the real path that side effect is the config write, `Popen`, and pidfile write; cancellation also prevents control from returning to `on_proc`, so the backend need not retain the new child before `stop()`. The gated async fake cooperates with cancellation and therefore cannot cover this production branch. Please make teardown wait until startup's synchronous side-effecting portion is quiescent, or add equivalent backend coordination, and drive the regression through a blocking `to_thread` start.

raven/rpc/bootstrap.py
gloryfromca: Verified on `707a152`. `question_broker.cancel_all()` is present, the focused test passes, and a pending question now finishes at teardown with its fail-safe default. The `done()` assertion also avoids the false green that a timeout-driven cancellation would create. This finding is resolved.

gloryfromca (commented):
Blocking: cancelling the backend-start task still does not settle EverOS startup's running `to_thread` worker, so startup side effects can occur after teardown and stop.

I reviewed head `707a1522a151` against current target `53f989c1`. The original feature patch is range-diff identical to the previously reviewed patch; the only PR-owned delta is `707a1522`. I covered the repository rules, complete diff, production EverOS startup and memory lifecycle contracts, affected callers through teardown, history, backward compatibility, and whether the tests were weakened. The tests are additive.

The question-broker finding is resolved and I replied on that thread. The backend-start finding remains open: `EverosBackend.start()` crosses `asyncio.to_thread(_start_server_if_unlocked)`, and cancelling/awaiting the outer task does not stop a running worker. A production-shaped probe recorded `thread-enter`, then `stop` during teardown, then a worker side effect after teardown returned. I replied in the existing finding thread rather than duplicating it.

Verification:

- Focused affected suite -> 83 passed.
- `uv run --frozen --all-extras pytest -q` with inherited HTTP proxy variables unset -> 6779 passed, 1 failed, 33 skipped, 13 deselected. The sole failure, `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`, reproduces identically on current target head `53f989c1`.
- Focused Ruff and format checks passed; `git diff --check` passed.

gloryfromca (commented):
Blocking: the backend-start teardown race remains unchanged; see the existing thread.

This head folds the ACP-over-stdio change into the branch, but it does not modify either `raven/rpc/bootstrap.py` or `raven/plugin/memory/everos/_server.py`. The prior production-path failure therefore still applies: cancelling and awaiting the task that is blocked in `asyncio.to_thread(...)` does not stop the worker, so teardown can call `stop()` and return before the startup thread performs its config, process-start, and pidfile side effects. I have not duplicated that inline finding.

I reviewed the repository rules (`AGENTS.md`/`CLAUDE.md` and both runtime/TUI context glossaries), the target diff and this head's incremental ACP/shared-runtime changes, their callers and commit history, backward-compatibility paths, and the tests for weakening. I found no additional reportable defect.

Verification on `0326604fc48f`:
- Focused Python suites: 839 passed, 32 deselected.
- TUI: 1,000 passed; TypeScript type-check and generated-RPC consistency check passed.
- `make check-large-files`: passed.
- Full Python suite: 7,486 passed, 1 failed, 33 skipped, 13 deselected. The sole failure is `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`, previously reproduced on the exact unchanged target commit `53f989c1`; it is not introduced by this revision.

raven/rpc/bootstrap.py
ZuyiZhou: Fixed in 51f2e0a -- you were right that cancel-plus-await settles the task and not the thread; `await asyncio.to_thread(...)` raises on cancel while the worker runs on, which is why `on_proc` never fires and the backend does not even hold the child it just spawned.

Teardown now waits on the start it finds in flight before stopping, and only cancels if the wait expires. The bound is sized for the uninterruptible part -- a config write and a `Popen` -- rather than for a cold boot: this is the exit path, and a person quitting should not be held for the health poll of a server they are leaving. Past the bound what is likeliest still running is that poll, and stranding a poll costs nothing.

The regression test crosses a real worker thread, per your note that the async gated fake cannot reach this branch: it asserts teardown does not return while the worker is mid-spawn, then that `["thread-enter", "thread-side-effect", "stop"]` is the order. It fails on the previous cancel-only code. One detail worth flagging: it polls the ready flag instead of calling `threading.Event.wait`, because that blocks the loop which has to run for `to_thread` to be scheduled at all -- waiting there deadlocks the thing being waited for. A second case pins the bound as a bound, and I dropped the broad `except Exception` around the await since `_start_backend` swallows its own exceptions and an unreachable guard reads as a case somebody handled.

Note on the branch: #359 was squash-merged into this branch while I was working, so I replayed that squash commit onto the corrected base rather than force-pushing over it. The resolution was the duplicated `ErrorEventPayload.turn_id` -- the field now comes from #356, one PR lower, so the squashed commit stops declaring it a second time (it was declared twice in both `models.py` and `openrpc.json`, which Python accepts silently). The rebuilt tree is byte-identical to the tree I ran the full suite against. I also rewrote the squash commit trailer, which had arrived carrying `claude-opus-5[1m]`.

gloryfromca: The ordinary completion case is fixed, but the timeout branch still reproduces the original ordering and does not provide the claimed exit bound. On `39263dea`, I drove the current assembly with the new `_ThreadStartBackend`, set `_BACKEND_START_SETTLE_S=0.05`, released its worker after 0.40s, and ran the whole coroutine under `asyncio.run()`. `stack.teardown()` returned after 0.051s with `["thread-enter", "stop"]`; then the worker appended `"thread-side-effect"`, and `asyncio.run()` itself did not return until 0.59s. The runner waits for default-executor threads during loop shutdown, so cancelling the task cannot make this process exit sooner—it only moves the unavoidable wait until after `stop()` while still skipping `on_proc`.

Please either await `backend_start` through completion (which does not lengthen actual `asyncio.run()` process exit versus this code), or coordinate the spawn worker separately so teardown waits for that worker before stopping while remaining free to cancel only the later health poll. The regression needs to cover the timeout with a blocked `to_thread` worker and assert process-level completion/order; the current timeout test uses the cancellation-cooperative async fake and therefore cannot catch this branch.

gloryfromca (commented):
Blocking: the backend-start timeout still permits `stop()` before the running worker's spawn side effect, and it does not bound process exit.

I replied on the existing teardown thread with the current-head reproduction rather than duplicating the finding. With `_BACKEND_START_SETTLE_S=0.05`, the real assembly and its new `to_thread` fake produced `['thread-enter', 'stop']` when teardown returned; the worker later appended `'thread-side-effect'`. Because both shipped client paths ultimately use `asyncio.run()`, loop shutdown then waited for that default-executor worker anyway. Cancelling after the timeout therefore buys no earlier process exit while preserving the lost `on_proc` handoff and stop-before-side-effect race.

I reviewed the repository rules (`AGENTS.md`/`CLAUDE.md` and the Runtime/TUI context glossaries), the complete current target diff, callers through process exit, history/range-diff, backward compatibility, and whether tests were weakened. The rewritten feature and ACP commits are patch-equivalent apart from the new teardown fix, corrected attribution, and removal of fields now supplied by the updated target; I found no additional reportable defect.

Verification on `39263dea31a6`:
- Focused lifecycle/ACP/workdir suite: 99 passed.
- TUI: 1,094 passed; TypeScript type-check and generated-RPC consistency passed.
- Focused Ruff and format checks passed; `git diff --check` and `make check-large-files` passed.
- Full Python suite: 7,672 passed, 1 failed, 33 skipped, 13 deselected. The sole failure, `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`, reproduces on the exact current target `79eb3c29`.

raven/rpc/bootstrap.py
ZuyiZhou: Fixed in 03ef3f0 by taking your first option: the timeout and the cancel are gone, and teardown awaits `backend_start` through completion.

Your `asyncio.run` measurement is what settles it -- the runner waits on the default executor at loop shutdown, so a worker mid-spawn holds the process open either way, and a timer only moves that same unavoidable wait to after `stop()`, which is the one place it does damage. The start is bounded by its own budget instead: `ensure_everos_server` polls health for ten seconds and gives up early on a child that has already died, so awaiting cannot hang the exit indefinitely.

Two regressions now. The thread one asserts teardown does not return while the worker is mid-spawn and that the order is `["thread-enter", "thread-side-effect", "stop"]`; it fails on the previous cancel-only code. The second drives the rejected shape in a real subprocess under `asyncio.run`, because loop shutdown is the subject and pytest`s loop outlives the test -- it records both halves of your argument in one run: the side effect landing after the stop, and the process still not exiting until the worker finished. The cancellation-cooperative fake is no longer used to cover this branch; it now pins the ordinary case, where teardown blocks until the start completes.

I also refreshed this PR`s description: it still described #359 as the next PR, and its Verification section predated the merge. It now records this head`s numbers and the `ErrorEventPayload.turn_id` duplicate the merge produced.

gloryfromca: Verified on `c7640ea9`. Reverting just the await to the cancel-then-stop shape fails both regressions -- `test_teardown_waits_for_the_part_of_start_it_cannot_interrupt` with `teardown must not return while the worker is mid-spawn`, and the gated-backend case with `['start-enter', 'stop']` instead of `['start-enter', 'start-finish', 'stop']`. Restored, the file's 17 tests pass. This finding is resolved.

I also checked the claim the timer's removal rests on, since an unconditional await is only safe if the start is bounded: `_start_server_if_unlocked` takes the startup lock with `blocking=False` and returns on `LockTimeoutError`, `ensure_everos_server` polls health against a 10s budget and gives up early on a dead child, and the unowned branch's `probe_capabilities` carries its own HTTP timeout. So there is no path where `start()` parks forever and teardown parks with it.

One residual, not a request and not something this site can fix. `asyncio.run`'s shutdown calls `_cancel_all_tasks`, which cancels `backend_start` directly out of `all_tasks()` -- so a Ctrl-C that lands while the worker is mid-spawn still gets the old ordering, and `asyncio.shield` here would not change that. Your own argument covers the cost: the runner waits on the executor either way, so the thread still finishes; it just finishes after `stop()`. Worth knowing about, not worth widening this PR for.

gloryfromca (commented):

Re-review of `c7640ea9`. Two things moved since `39263de`: the teardown fix in `364fdcd`, and the merge of #359, which brings the whole `raven/acp` tree into this PR's diff. I reviewed and approved #359 on its own PR and did not re-read those 20k lines here; what I looked at this round is the fix and the seam it lands on.

**The teardown fix.** Correct, and for the stated reason. Reverting only the await to the cancel-then-stop shape fails both regressions; restored, the 17 tests in `tests/test_rpc_bootstrap.py` pass. The premise the removed timer rested on holds too: `_start_server_if_unlocked` takes its lock with `blocking=False`, `ensure_everos_server` polls a 10s budget and gives up early on a dead child, and the unowned branch's `probe_capabilities` has its own HTTP timeout -- so an unconditional await cannot park teardown indefinitely. Details in the thread.

**Tests not weakened.** `test_teardown_settles_a_start_still_in_flight_before_it_stops` was rewritten rather than relaxed: its expectation inverted along with the behaviour (`['start-enter', 'start-finish', 'stop']` now, and it must not return early), and it still fails against the old code. `_ThreadStartBackend` crosses a real worker thread because the asyncio-gated fake cooperates with cancellation and cannot reach the branch, which is the right call. The subprocess case asserts a property of `asyncio.run` rather than of this repo, so it is documentation more than regression cover -- fine where it is; a real subprocess in `tests/` has plenty of precedent here, so no naming objection under section 5.2.

**Coverage.** Covered: the delta itself, the teardown callers (`acp/server.py`, `cli/acp_commands.py`), the EverOS start path the fix reasons about, AGENTS.md sections 1/5, and whether the tests were softened to pass. Not covered: the #359 content, reviewed separately.

Ran `uv run pytest tests/test_rpc_bootstrap.py tests/test_rpc_spine.py tests/test_acp_server.py tests/test_acp_methods.py tests/test_acp_stdio.py tests/test_cli_acp_commands.py` -- 214 passed.

gloryfromca (approved):

PR #359, opened 2026-08-22
feat(acp): serve the agent client protocol over stdio

ZuyiZhou:
## Summary

`raven acp` is now an agent an editor can launch: it negotiates, mints and reloads sessions, runs a turn, streams the reply and its tool calls, asks before acting, and answers a cancel while a turn is in flight.

Four commits, each one concern: the wire framing, the lane-to-session mapping, the shell approval families this surface needs, and the protocol layer itself.

**The process model is a self-contained engine**, not a client of a resident gateway. An ACP agent IS a stdio child of the editor, so there is nothing to mount onto unless a gateway happens to be running, and an agent that only works when another service is up is not an agent an editor can launch.

Notes on the parts that are not obvious:

- Every inbound frame is handled in its own task, because `session/prompt` is suspended for as long as the turn takes and `session/cancel` has to be read during it. Shutdown answers instead of cancelling: on EOF the pending prompts are settled as cancelled and the handlers return through their own code, so a finite input (a scripted client, a smoke test) still gets its replies.
- A prompt is never answered with a JSON-RPC error. A failed turn is explained as message content and then ends with a stop reason. Measured from the other side on another implementation: an error in reply to a turn-shaped request makes clients tear the whole turn down.
- The schema is vendored and pinned, and its blindness is measured rather than assumed: `additionalProperties` appears 118 times and is true in every one, `method` is an unconstrained string, and `params` accepts null. Validating a frame against the union therefore accepts an invented method name, so the tests validate payloads against their own definition and check method names against the metaschema separately.
- Questions take one of two paths by what the client declared. With elicitation forms, a question with no preset answers can come back as text. Without them it rides a permission request and a synthesised tool call, which the payload marks as synthesised rather than dressing up as a real one. The case that has no answer at all, a free-text question to a client with no elicitation, is recorded in the compatibility matrix instead of being papered over.
- Secrets are redacted before anything reaches the client, including the value under a key whose name says it is a secret. A credential file's path is deliberately NOT redacted: a path is not a secret, and hiding it would hide that the agent read credentials at all.
- Six command families ask before running on this surface. What that does not reach is written down: a sub-agent builds its own tool with its own policy and no responder, so a delegated command still runs unprompted.

**The write tools now publish what they changed.** Until now the `diff` field declared on `tool.complete` had no producer here, so it was always null. The file tools return a `ToolResult` carrying both the unified diff and the two file versions, and the registry, the loop and the outlet carry them through. The protocol needs the second form specifically: its diff content block is `{path, newText, oldText}`, which a rendered diff cannot fill. A reply's files reach the client the same way, as a `media` event carrying paths rather than bytes.

**Divergences from the same layer in the sibling runtime, each measured rather than assumed:**

- The compatibility matrix is corrected for this build in two rows. A model change during a turn is not refused, because `config.set` here has no busy guard. A sub-agent's output is untagged, because there is no direct sub-agent chat for the tag to distinguish.
- The self-verification suite is not included. It drives the ACP client's own capability probe, which this repo does not have, and porting a client subsystem to run one test file is the tail wagging the dog.
- The `session/load` smoke test is not included either, and the reason is worth knowing: nothing here can point a spawned agent at a throwaway session store. `RAVEN_HOME` steers tracing and the terminal runtime root only, and the config path has no override at all, so that test would either read the developer's real store or prove nothing. Load and replay keep their unit coverage; what is lost is the leg that read an actual file.
- The working-directory guard is checked against the configured home for the same reason, rather than against a temporary one.

Known fidelity losses, all in `docs/specs/2026-08-21-acp-agent-compatibility.md`: `tool_call_update` always reports completed because the event carries no success flag; `rawInput` is not sent because it carries a command line verbatim; there is no plan because its priority field is required and has no source here.

Stacked on #357, which is stacked on #356, which is stacked on #353. Merge order: 353, 356, 357, then this.

Reviewers: 0xKT, gloryfromca.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run --frozen --all-extras pytest -q` -> 7375 passed, 19 failed. The same 19 that fail on the base branch: `test_provider_rates.py` (15) and `test_agent_loop_usage_sink.py` (4), pricing-table and network-cache tests unrelated to this change. Zero new failures.
- `uv run --frozen python scripts/coverage_gate.py diff --base-ref feat/rpc_engine_assembly` -> 100.00% (1718/1718 executable changed lines). The first run was 99.30%; the twelve uncovered lines were the framing module's public API and the oversized-diff branch, and cases were added for both.
- `uv run --frozen python scripts/coverage_gate.py ratchet` -> line +2.03pp, branch +3.08pp, PASS.
- `raven/acp/` unit coverage: 100 percent line and 100 percent branch across 540 tests.
- Integration against the real binary, run with `-m ""`: `test_acp_stdio_smoke.py` (6) and `test_acp_adversarial_smoke.py` (26) pass.
- `uv run --frozen ruff check raven/ tests/` and `ruff format --check` -> clean.
- `cd ui-tui && npm run lint:rpc` -> generated.ts in sync. `npm run type-check` -> exit 0. `npm test` -> 998 tests passed.
- `raven acp --help` resolves through the real CLI app.

## Risk

A new CLI surface plus three optional wire fields and one new event variant, all defaulted, so a client that ignores them behaves as before.

Two changes reach beyond the new surface and deserve a second look. The file tools now return a `ToolResult` where they returned a string: the model-facing text is unchanged and the registry already accepted both shapes, but every caller of `write_file` and `edit_file` goes through that boundary. And shell approval gains six command families, though only for a surface that registers them; the terminal registers none and behaves exactly as before.

One dodge in the approval walk stays open and is pinned by a test: a command runner (`timeout 60 npm install`) hides its inner command. Verified that the same wrapper already hides a delete from the existing check, so the new gate is no weaker than the surface it joins, and no stronger.

Rollback is a plain revert: no data migration and no persisted format change.

## Related Issues

N/A


gloryfromca (commented):
Blocking: model switching is unusable, supported image payloads are rejected below the documented limit, and the new subprocess smoke test can invoke the operator's real provider; see the inline notes.

I covered the repository rules and canonical Runtime terms, the complete target-range diff and commit history, callers through their final contracts, backward compatibility of the shared RPC/spine/TUI paths, and whether the tests bypassed newly reachable branches.

Verification:
- `uv run pytest`: 7225 passed, 43 skipped, 13 deselected, 1 failed. The failure is `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`, and I reproduced it on the exact target commit `a6481f777c07`, so it is unrelated to this change.
- Focused ACP/approval/RPC tests: 811 passed.
- ACP subprocess integration: 4 passed, then the concurrency case timed out after reading the operator's real config; the identical case passed with an isolated home.
- TUI generated-schema check and type-check passed; TUI tests: 998 passed.
- `git diff --check` and `make check-large-files` passed.

raven/acp/config_options.py
```diff
@@ -0,0 +1,193 @@
...
+async def set_model(call: Call, *, session_id: str, value: Any) -> None:
+    """Apply a model selection, letting the runtime's own refusals through.
+
+    ``session_id`` is passed so ``config.set`` can refuse a switch during that
+    session's turn. Swapping the provider under a running request is the failure
+    the guard exists for, and this layer must not route around it.
+    """
+    if not isinstance(value, str) or not value:
+        raise ValueError("the model value must be a non-empty string")
+    await call("config.set", {"key": MODEL_OPTION_ID, "value": value, "session_id": session_id})
```
gloryfromca: **[P1] Forward the provider required by `config.set`**

Every model selection currently fails before applying: the target branch's `config.set` contract requires a separate non-empty `provider` field (`raven/rpc/methods/config.py:304-320`), but this sends only `key`, `value`, and `session_id`. I reproduced the call with `anthropic/claude-opus-5`; it raises `ConfigValidationError: config.set model needs a provider`. Please carry the selected provider through this mapping and cover the real dispatcher contract rather than only the stub. The refresh should also call `model.options` with this `session_id`, otherwise a successful session-scoped switch would still return the installation default as `currentValue`; the process-wide/mid-turn description above is stale against the same target contract.

raven/acp/stdio.py
```diff
@@ -0,0 +1,228 @@
...
+from typing import Any, BinaryIO
+
+from raven.agent.acp.protocol import AcpProtocolError, decode, encode
+
+# One ACP frame is not a line of text. ``session/prompt`` carries images and
+# embedded resources, and base64 inflates by about 4/3, so a 4 MB screenshot
+# arrives as a single frame of roughly 5.5 MB. The 1 MiB cap in
+# ``rpc/server.py`` is sized for the TUI's own traffic and would reject input
+# this protocol is specified to accept.
+MAX_FRAME_BYTES = 8 * 1024 * 1024
```
gloryfromca: **[P1] Make the frame cap accommodate the advertised image limit**

This cap prevents the 20 MiB image limit in `methods.py` from being reached. A 7 MiB raw image (well below that limit) becomes a 9,786,855-byte JSON/base64 request and `read_frames` rejects it at 8,388,608 bytes before `session/prompt` sees it. Please either size the wire cap for a 20 MiB image plus base64/JSON overhead, or lower the method/documented limit consistently, with a boundary test that passes through framing.

tests/integration/test_acp_adversarial_smoke.py
```diff
@@ -0,0 +1,344 @@
...
+from tests.acp_stub_client import raven_binary, stub_client
+
+pytestmark = pytest.mark.integration
+
+
+@pytest.fixture
+def home(tmp_path):
+    if not raven_binary().exists():
+        pytest.skip(f"raven console script not installed at {raven_binary()}")
+    return {"RAVEN_HOME": str(tmp_path / "home")}
```
gloryfromca: **[P1] Isolate the config before starting the real ACP child**

`RAVEN_HOME` only redirects tracing/runtime data; `get_config_path()` still reads `Path.home() / '.raven/config.json'` (as this file itself notes later). On this run the fixture loaded the operator's OpenRouter configuration, sent the `hello` prompt to a real provider, and timed out after 60 seconds. The identical scenario returned immediately once `HOME` was isolated. This makes the supposedly credential-free test nondeterministic and can spend or expose a developer's live credentials; point the child at an isolated home/config before spawning it.

raven/acp/config_options.py
ZuyiZhou: Fixed in ecfe130. `set_model` now sends `provider`, looked up in the same catalogue the value was offered from rather than split off the string -- splitting would have named `anthropic` for a model a gateway serves, which is what the runtime`s separate field exists to prevent. A tie goes to the provider in use, and a value no configured provider offers is refused rather than guessed at, since inventing one spends a credential nobody chose.

`model.options` is asked with the `session_id` too, threaded through `_config_options` so `session/new` and `session/set_config_option` both scope it. And both stale descriptions are corrected against the target contract: the switch is session-scoped, not process-wide, and a mid-turn switch is accepted -- the turn in flight keeps the binding it started on and the choice lands on the next one. The compatibility matrix rows say the same.

Covered against the dispatcher contract rather than the stub`s shape: `test_the_write_names_the_provider_the_runtime_requires`, `test_the_provider_is_looked_up_rather_than_split_off_the_value` (openrouter serving `anthropic/claude-haiku-4-5`), the tie case, and the refusal.

raven/acp/stdio.py
ZuyiZhou: Fixed in ecfe130. `MAX_FRAME_BYTES` is 32 MiB now, sized off `methods.MAX_IMAGE_BYTES` rather than off a plausible screenshot: 20 MiB of image is 26.7 MiB of base64, and the remainder is the envelope. The arithmetic is written down at the constant.

One image at the ceiling, not several -- the prompt limit is per image and this cap is per frame, so they meet at exactly one, and a prompt carrying two is refused at the frame boundary. That is now stated in the comment and in the compatibility matrix rather than discovered.

`test_an_image_at_the_advertised_ceiling_fits_through_framing` builds a real `session/prompt` frame at `MAX_IMAGE_BYTES` and passes it through `read_frames`; it fails at the old cap. Its twin pins the two-image refusal, including the resynchronisation.

tests/integration/test_acp_adversarial_smoke.py
ZuyiZhou: Fixed in ecfe130, and thank you for catching it -- the suite advertised itself as credential-free while it was not.

Three things, not one. `RAVEN_HOME` moves tracing and runtime data; `get_config_path` reads `Path.home()/.raven/config.json` whatever it says; and a provider key in the environment is read before either. Both integration files now set `HOME` and `USERPROFILE` (the one `Path.home()` reads on Windows) and empty every provider key, read off `providers.registry.PROVIDERS` so a provider added later is isolated by the same fixture rather than quietly re-arming the file.

`test_a_cwd_containing_the_agent_home_is_refused` needed the same correction: it derived its path from this process`s `load_config()`, which named a directory the isolated child no longer has, so the guard did not fire. It now derives it from the child`s own `HOME`, whose `.raven/workspace` is the agent home it will refuse for.

The full suite runs in 53s locally now -- 32 passed, no timeout.

gloryfromca (commented):

I re-reviewed the current revision against `feat/rpc_engine_assembly` and verified that all three prior blockers are resolved on their real paths: model changes now reach `config.set` with the catalogue-selected provider and session-aware read-back, an image at the 20 MiB method ceiling passes through framing, and both real-child integration suites isolate the user config and inherited provider keys. I also checked the force-update with `range-diff`, including its conflict resolution against the target's newer tool metadata/diff work.

Coverage included the repository rules (`AGENTS.md`, `CONTEXT-MAP.md`, and the relevant Runtime terms in `CONTEXT.md`), the current diff and fix delta, callers through dispatch and subprocess teardown, relevant history, backward compatibility, and the test changes for weakening. I found no new issue worth raising.

Verification:
- `uv run pytest tests/test_acp_config_options.py tests/test_acp_methods.py tests/test_acp_stdio.py -x`: 154 passed.
- `uv run pytest tests/integration/test_acp_adversarial_smoke.py tests/integration/test_acp_stdio_smoke.py -m integration -x`: 32 passed, including the previously timing-out real prompt path.
- `uv run pytest`: 7,249 passed, 43 skipped, 13 deselected, with one unrelated local failure in `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`; the failing source and test are unchanged across the target move and the same failure was reproduced on the target revision during the prior review.
- `npm run lint:rpc`, `npm run type-check`, and `npm test -- --run`: passed; 1,000 TUI tests passed.
- `git diff --check` and `make check-large-files`: passed.
- Current GitHub checks, including Python unit tests, TUI checks, lint, repository files, and commit messages: passed.

gloryfromca (approved):

PR #361, opened 2026-08-23
feat(tui): fill the prompt block and drop the cover after the first turn

LivXue:
## Summary

Two presentation defects in the TUI chat transcript. They ship together because both
change what the virtualized transcript has to reserve rows for, and reviewing the row
arithmetic once is safer than twice.

**The cover was never a screen.** It is transcript row 0 -- a message with `kind: 'intro'`
that `appLayout` renders as the wordmark plus the session panel -- and `capHistory` pinned
it at index 0, so it was exempt even from the history cap. Measured against the real
renderer it occupies 35 terminal rows, while `estimatedMsgHeight` guessed 9. So the first
answer was appended *below* a full screen of cover art instead of replacing it, and
scrolling back up always found it again.

`hideIntroAfterFirstTurn` drops it from the view once a `user` or `assistant` row exists.
The row itself stays in `historyItems`, because the late `session.info` event patches
itself onto that row, and `/export` and the slash handlers read the session's own list
rather than whatever is on screen. The trigger is deliberately a turn and not "any other
row": startup notices go through `sys()` as `role: 'system'` and slash output is
`kind: 'panel'`, so gating on those would mean a host that warns about its credentials on
boot never shows a cover at all.

Both the rows and the layout's copy come from the one filtered array. `appLayout` compares
`row.index` against indices it derives from `transcript.historyItems`, so filtering one and
not the other would move the inter-turn separator and the todo panel one row off.

**The person's own message now renders on a filled background,** so a turn reads as an
inserted card rather than one more line of prose. `userBg` is a curated literal per tier
and scheme, the way the rest of the reduced-tier palette is carried, and is skinnable
through `ui_user_bg`. Below 256 colors there is no shade between black and brightBlack, so
the fill is skipped and the prompt chevron carries the row alone. The two padding rows are
drawn at every tier so `estimatedMsgHeight` can reserve a row count without reading the
terminal's color capability; it now reserves four rows for a user message rather than two.

One testing note worth carrying forward: `ink-testing-library` brings its own reconciler
and silently drops both foreground colors and background fills, so a fill is only
observable through `renderSync` writing real escape codes to a stream. An early version of
these tests passed against nothing.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

Mixed in practice: the cover half is a fix, the filled block is new behaviour.

## Verification

Run in a clean worktree at this branch, with `packages/hermes-ink` built from this base
rather than borrowed, so the render assertions exercise the Ink this branch ships:

- `npm run type-check` -- clean.
- `npx eslint src/` -- 0 errors, 22 warnings, all pre-existing. `useMainApp.ts` carries two
  of them before and after the change (same rules, shifted 12 lines by the inserted
  comment), confirmed by linting the pristine file from `main` and diffing.
- `npx vitest run --no-file-parallelism` -- 87 files, 996 passed, 13 skipped, 0 failed.
- `scripts/check_commit_messages.py main..HEAD` -- exit 0.
- Each of the three mechanisms reverted **in isolation** to prove the tests are not
  theatre: neutering `hideIntroAfterFirstTurn` reddens the two cover tests; removing only
  the `backgroundColor` prop while keeping the padding reddens the fill test; changing the
  reserved rows from four back to two reddens both height tests.

Suites that exercise the changed code directly: `messages.test.ts` (the filter, and the
block rendered through `renderSync`), `virtualHeights.test.ts` (the reserved rows),
`theme.test.ts` (the token per tier, and the fill gate).

`ui-tui/CONTEXT.md` gains **Prompt Block** and **Cover**, since both are named in the code
and neither existed as a documented term.

## Risk

Visible change to every chat transcript: a user message grows from three rows to five (two
of them the filled padding), and the cover stops being visible after the first turn.
Nothing is removed from the session -- the cover row is still in `historyItems`, so
`/export` and session save are byte-identical.

Rollback is the revert of this one commit; there is no migration, no persisted state, and
no wire-format change. A skin that dislikes the shade can set `ui_user_bg` without a code
change.

Known gaps, disclosed rather than left for review to find:

- The index alignment between `virtualRows` and the layout's `historyItems` is held by
  construction and by a comment, not by a test; there is no existing harness that mounts
  `useMainApp` against a gateway.
- `theme.ts` cites `scripts/gen-color-palettes.mjs` as the provenance of the reduced-tier
  palettes and a `docs/tui-color-problem/tokens.md` for the tier-2 values. Neither holds
  today: that script's embedded source palette is a stale green one, its algorithm maps
  `#161b22` to `ansi256(16)` where the file has `ansi256(234)`, and the doc does not exist.
  The values are hand-curated, and the ones added here follow the existing neighbours. Both
  stale comments are left alone rather than widen this diff.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the complete nine-file diff and covered the repository rules (`AGENTS.md`, `CONTEXT-MAP.md`, and the TUI glossary), the transcript/rendering callers and relevant history, backward compatibility, and whether the tests weakened existing assertions. I traced the visible/raw history split through `useMainApp`, `appLayout`, virtual row heights, sticky prompts, live todos, session create/resume/compress, late `session.info`, and slash/raw-history consumers. The indices stay aligned while the retained cover row remains available to session-side consumers. I also checked the prompt fill through theme resolution, skin fallback, terminal color tiers, wrapping, and reserved-height accounting.

Validation on this head:
- Focused changed suites: 3 files, 63 tests passed.
- Full TUI suite: 87 files, 1009 tests passed.
- `npm run type-check --prefix ui-tui`: passed.
- `npm run lint --prefix ui-tui`: 0 errors, 22 warnings on untouched lines.
- Commit-message check and `git diff --check`: passed.
- Merge tree against latest fetched `main` (`0d2f2c25`): clean.

gloryfromca (approved):

gloryfromca (commented):
No blockers; this revision preserved the reviewed patch and the landed tree is clean.

This was a revision-specific delta review, not a repeat of the earlier feature review. `git range-diff` reports the feature commit unchanged, and the old/new stable patch IDs are identical. The absorbed `main` changes have no path overlap with the nine feature files; I rechecked the adjacent virtual-history identity contract, repository-rule impact, callers, backward compatibility, and test assertions, and found no new failure path or weakened test. The `b562f332` tree is also identical to the squash-merged `main` tree at `22b71765`.

Validation on the combined tree:
- Focused feature plus virtualization suites: 4 files, 65 tests passed.
- Full TUI suite: 90 files, 1046 tests passed.
- `npm run type-check --prefix ui-tui`: passed.
- Current-head GitHub checks: all passed.

PR #362, opened 2026-08-23
feat(*): add end-to-end trajectory debugging and regression workflows

forrestjgq:
## Summary

Add an end-to-end trajectory workflow for capturing, packaging, reporting, replaying, and permanently guarding Raven agent runs.

- Give every tracing span an attempt id, with explicit multi-turn attempt scopes and a single-turn fallback. Add append-only pass/fail/infra verdicts, trajectory pins, rotation-transparent lookup, and canonical attempt resolution.
- Add `raven trajectory` commands to list, label, pin, save, report, replay, and minimize attempts. Saved bundles are self-contained and include rewritten artifact references, session history when available, verdicts, and a manifest.
- Add three-layer report redaction: exact replacement of known config and environment secrets, pattern-based credential removal, and a residual scan for human review. Reports are produced as local tarballs through a pluggable uploader interface.
- Add deterministic replay through the live agent harness with recorded model replies and stubbed tool results. Replay never executes real tools, uses a temporary workspace, suppresses its own tracing, and reports strict or warning-mode divergences under narrow normalization rules.
- Add minimized and redacted trajectory cassettes, a YAML assertion DSL, and pytest discovery for permanent harness regression cases. Assertions can require a specific first divergence and verify the fixed live message or tool request at that point.
- Harden all externally derived paths and replacement directories, including canonical ids, bundle output paths, cassette destinations, artifact traversal, and escaping symlinks.
- Change streamed response aggregation to preserve a missing upstream finish reason as `unknown` instead of fabricating `stop` or `tool_calls`, keeping recorded trajectories faithful to the provider response.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run --frozen pytest tests/ -q --ignore=tests/integration`: 6862 passed, 33 skipped, 13 deselected, and 1 unrelated existing theme-rendering test failed. The branch does not modify that test, the theme implementation, or dependency manifests.
- `uv run --frozen pytest tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare -q`: the same unrelated theme-rendering failure reproduced in isolation.
- `uv run --frozen ruff check .`: passed.
- `uv run --frozen ruff format --check .`: 937 files already formatted.
- `make check-large-files`: passed.

## Risk

- The new CLI and trajectory formats are additive. Existing spans gain an `attempt.id` attribute, and readers retain compatibility with older spans by falling back to the trace id.
- Report and cassette payloads can contain complete model and tool text. Redaction runs on copies, excludes unverifiable binary files, records its actions, and requires residual review before sharing or committing data.
- Replay is isolated from real tool execution, the real workspace, session persistence, and tracing. Its normalization deliberately covers only values regenerated on every run so real behavior changes remain visible.
- The only intentional existing-path behavior change is the `unknown` fallback for a missing streamed finish reason. Production consumers were checked for dependence on the previous synthesized values.
- Rollback is to revert this PR. Existing trajectory state files and directories are additive and can remain unused after rollback.

## Related Issues

N/A


raven/trajectory/replay.py
```diff
@@ -0,0 +1,953 @@
...
+    A reference that is still absolute (the source file was missing at pack
+    time) or that does not exist in the bundle yields ``None`` — the caller
+    decides whether that is fatal.
+    """
+    if not isinstance(ref, str) or not ref:
+        return None
+    path = Path(ref)
+    if path.is_absolute():
+        return None
+    resolved = bundle_dir / path
```
gloryfromca: `_load_artifact` rejects absolute paths but accepts traversal and symlink escapes. A bundle can set `llm.output.artifact_path` to `../../secret.json`; `load_recording()` reads that file before the cassette validator runs, while `trajectory replay` never calls the validator at all. I reproduced this with a temporary bundle and got `{'content': 'LOCAL-ONLY-SECRET', 'finish_reason': 'stop'}` from a JSON file outside the bundle. This makes replaying a shared bundle an arbitrary local JSON read. Please resolve the reference and require it to remain under the bundle's `artifacts/` directory before `is_file()` or `read_text()`.

raven/trajectory/redact.py
```diff
@@ -0,0 +1,564 @@
...
+
+def _apply_exact(text: str, secrets: list[KnownSecret], counts: dict[str, int]) -> str:
+    # One pass over the text with all variants as an alternation, longest
+    # first (re picks the first alternative at a position, so longest-first
+    # gives longest-match). Sequential str.replace re-scanned its own output:
+    # a short value (a 1-char junk env key, say "k") then shredded the "k"
+    # inside placeholders inserted for earlier, longer secrets.
+    variants: dict[str, str] = {}
+    for secret in secrets:
+        for variant in _variants(secret.value):
```
gloryfromca, line 422: Exact credential names intentionally have no length floor, and this loop adds their raw value to a global substring replacement. With `DEEPSEEK_API_KEY=k` (a value the new end-to-end test itself installs), redacting the checked-in reproduction cassette made 108 replacements: `session.key` became `session.[REDACTED:env.DEEPSEEK_API_KEY]ey` and `marker` became `mar[REDACTED:env.DEEPSEEK_API_KEY]er`. Strict replay then diverged at call 2 and the regression's `name_equals: marker` check failed. Thus a local short credential silently changes the semantics of generated cassettes instead of merely sanitizing them. Please prevent low-specificity exact values from being applied as raw substrings, or otherwise preserve the parsed payload semantics.

raven/trajectory/replay.py
```diff
@@ -0,0 +1,953 @@
...
+    def _to_response(output: dict[str, Any], index: int) -> LLMResponse:
+        tool_calls = [
+            ToolCallRequest(
+                id=tc.get("id") or f"replay-{index}-{j}",
+                name=tc.get("name") or "",
+                arguments=tc.get("arguments") if isinstance(tc.get("arguments"), dict) else {},
+            )
+            for j, tc in enumerate(output.get("tool_calls") or [])
+        ]
+        return LLMResponse(
```
gloryfromca, line 596: The replayed `LLMResponse` drops `thinking_blocks` (and the existing trace output payload does not record it), even though AgentLoop uses that field to choose its empty-response recovery branch and persists it into subsequent assistant messages. A direct check with an empty response carrying Anthropic-style `thinking_blocks` chose `RecoveryAction.PREFILL` live, while the response reconstructed here chose `RecoveryAction.RETRY`. Real extended-thinking trajectories therefore take a different harness path during replay and can produce a false divergence before reaching the bug under test. Please carry this field through trace output, cassette minimization, reconstruction, and request comparison.

raven/trajectory/store.py
```diff
@@ -0,0 +1,145 @@
...
+        return {}
+    return data if isinstance(data, dict) else {}
+
+
+def pin(id_: str, *, reason: str = "", state_dir: Path | None = None) -> None:
+    """Protect an attempt/trace id from any future purge. Idempotent."""
+    if not id_:
+        raise ValueError("id is required")
+    current = pins(state_dir)
+    current[id_] = {"reason": reason, "ts": datetime.now(timezone.utc).isoformat()}
```
gloryfromca: The registry read happens before `atomic_replace` acquires its lock, so the lock serializes only the final replacements, not this read-modify-write transaction. I started 24 processes on a barrier, each pinning a distinct id; all exited successfully but only 2 ids remained in `pins.json`. `unpin` has the same lost-update shape. Since a pin is the promised never-purge retention state, concurrent CLI/report activity can silently discard it. Please hold one cross-process lock across the read, mutation, and replace.

gloryfromca (commented):
Blocking: this revision can read files outside a replay bundle and has three trajectory fidelity/retention failures; see the inline notes.

I covered the repository rules and canonical trajectory terms, the full diff, callers and tracing/session/provider contracts, commit history and backward compatibility, and whether tests were weakened. In particular, the traversal test verifies that the outside file is unchanged, but not that it was never read.

Verification: `uv run pytest tests/test_agent_loop_stream.py tests/test_cli_smoke.py tests/test_cli_trajectory_commands.py tests/test_tracing_api.py tests/test_trajectory_bundle.py tests/test_trajectory_cassette.py tests/test_trajectory_redact.py tests/test_trajectory_regressions.py tests/test_trajectory_replay.py tests/test_trajectory_store.py` passed (223 tests); `make check-large-files` passed; `git diff --check github/main...HEAD` passed.

raven/trajectory/redact.py
forrestjgq: Fixed in 0ac8336. Exact values shorter than six characters now use token-boundary matching instead of global substring replacement. A standalone JSON value such as k is still redacted, while identifiers and values such as session.key, marker, and ok remain unchanged. The regression test asserts both the preserved payload semantics and the standalone secret replacement.

raven/trajectory/replay.py
forrestjgq: Fixed in 0ac8336. thinking_blocks is now recorded in the trace output artifact, retained by cassette minimization, reconstructed on ReplayProvider responses, and included alongside reasoning_content in canonical assistant-message comparison. Regression coverage verifies the trace artifact, cassette round trip, reconstructed response, and request mismatch behavior.

raven/trajectory/store.py
forrestjgq: Fixed in 0ac8336. The new atomic_update primitive holds the existing cross-process sidecar lock across the complete read, mutation, and atomic replacement. Both pin and unpin now use that transaction instead of reading before atomic_replace acquires the lock. Spawned multiprocessing barrier tests verify that concurrent distinct pin and unpin operations do not lose updates.

raven/trajectory/replay.py
forrestjgq: Fixed in 0ac8336. Artifact references now go through one resolver shared by replay and cassette. It rejects traversal, paths outside artifacts/, and symlink escapes before is_file() or read_text() can access the target; absolute dangling source references retain the existing missing-artifact behavior. The new load_recording regression tests replace Path.read_text with a guard that fails if the outside file is ever read, covering both traversal and symlink cases.

raven/trajectory/replay.py
gloryfromca: Verified on 0ac8336. thinking_blocks is present in the trace payload, retained by minimization, reconstructed on the response, and compared in assistant messages. In the original direct reproduction, both the live and reconstructed thinking-only response now choose RecoveryAction.PREFILL. This resolves my finding.

raven/trajectory/store.py
gloryfromca: Verified on 0ac8336. The new update callback runs under the same sidecar lock for the read and replacement. Re-running my 24-process barrier reproduction retained all 24 distinct pins, and the concurrent unpin test also passes. This resolves my finding.

raven/trajectory/redact.py
gloryfromca: Verified on 0ac8336. Re-running the original one-character-secret reproduction leaves session.key and marker intact; strict replay completes and the checked-in regression has no failures. The new test also confirms that a standalone JSON value of k is still redacted. This resolves my finding.

raven/trajectory/replay.py
gloryfromca: Verified on 0ac8336. The shared resolver rejects traversal and symlink escapes before artifact loading reaches is_file or read_text. My original ../../secret.json reproduction now raises ValueError, and the guarded-read tests cover both escape forms. This resolves my finding.

gloryfromca (commented):

I verified all four fixes against their original reproductions and found no new issue in the 34c47f403e50..0ac833636ec5 delta. This pass covered the repository rules and canonical trajectory contracts, the fix diff, affected callers and shared atomic-I/O consumers, commit history and backward compatibility, and whether the new tests weakened the original failure conditions.

Verification: the expanded affected suite passed (318 tests); make check-large-files passed; both diff checks passed; the 24-process pin reproduction retained all 24 ids; current GitHub checks pass or skip.

gloryfromca (approved):

Issue #297, opened 2026-08-11
DeepSeek thinking mode still drops reasoning_content on v0.1.11 when model is missing from litellm's model map (deepseek/deepseek-v4-flash)


fatinghenji:
## Summary

#89 fixed the Curator/history-trimmer projection stripping `reasoning_content`, and that fix is confirmed present in v0.1.11 (`curator.py:319`, `history_trimmer.py:43-44` allowlists). However, the **same DeepSeek error still reproduces on latest `main` (v0.1.11, `9e2079b`)** when the configured model is **not in litellm's model map** — in our case `deepseek/deepseek-v4-flash`.

```text
Error calling LLM: litellm.BadRequestError: DeepseekException -
{"error":{"message":"The reasoning_content in the thinking mode must be passed back to the API.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
```

## Reproduction (4 crashes in one evening, 2026-08-11)

1. Configure Raven with `deepseek/deepseek-v4-flash` (thinking enabled server-side by DeepSeek).
2. Run any tool-heavy agent task (`raven agent -m "..."`).
3. The loop dies mid-task with the error above. Observed:
   - mid-session, after ~10 min of tool calls (fresh session, first user turn)
   - twice more after resuming with `-c`
   - once on the **first LLM call of a brand-new CLI session** in the same workspace (prior-session assistant history in context)

Verified the #89 fix is in the running tree:

```python
# raven/context_engine/segments/curator.py:319
allowed = {"role", "content", "tool_calls", "tool_call_id", "name", "reasoning_content", "thinking_blocks"}
# raven/context_engine/history_trimmer.py:43-44 -> "reasoning_content", "thinking_blocks"
```

and `git rev-list --count 9e2079b..origin/main` = 0 (nothing newer upstream).

## Root-cause lead: litellm doesn't know this model is reasoning-capable

```python
>>> import litellm  # litellm 1.85.0 (Raven's pinned dependency)
>>> litellm.get_model_info("deepseek/deepseek-v4-flash")
Exception: This model isn't mapped yet. model=deepseek/deepseek-v4-flash, custom_llm_provider=deepseek.
>>> litellm.get_model_info("deepseek/deepseek-reasoner").get("supports_reasoning")
True
```

[...]

PR #363, opened 2026-08-24
fix(providers): replay empty DeepSeek reasoning on tool continuations

guix4ever:
## Summary

Normalize outgoing DeepSeek V4 tool continuation messages so an assistant tool call retains an empty `reasoning_content` field when the upstream response contained no reasoning tokens.

- Apply the normalization at the LiteLLM Provider boundary for streaming and non-streaming requests.
- Cover direct DeepSeek routing and DeepSeek models routed through OpenRouter.
- Leave persisted history, non-thinking DeepSeek models, and other Provider payloads unchanged.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_litellm_provider_stream.py tests/test_agent_loop_stream.py tests/test_subagent_manager.py -x -q`: 58 passed.
- `uv run ruff check .`: passed.
- `uv run ruff format --check .`: 920 files already formatted.
- `npx --yes --package=@commitlint/cli@21.1.0 commitlint --from origin/main --to HEAD --config commitlint.config.cjs`: passed.
- `PYTHONPATH=. uv run python scripts/check_commit_messages.py origin/main..HEAD`: passed.
- `PYTHONPATH=. uv run python scripts/check_large_files.py origin/main..HEAD`: passed.
- `git diff --check origin/main...HEAD`: passed.
- No user-facing documentation or screenshots are needed for this wire-format fix.

## Risk

- The behavior change is limited to DeepSeek V4 requests that carry tools and replay assistant tool calls without a `reasoning_content` key.
- Existing non-empty and explicit null reasoning values remain unchanged, as do non-thinking DeepSeek models and other Providers.
- Rollback is to revert this change; no data migration or persistent format change is involved.

## Related Issues

#297


raven/providers/litellm_provider.py
```diff
@@ -424,6 +436,7 @@ async def chat(
         original_model = model or self.default_model
         model = self._resolve_model(original_model)
         extra_msg_keys = self._extra_msg_keys(original_model, model)
+        ensure_tool_reasoning = bool(tools) and self._requires_tool_reasoning_replay(original_model, model)
```
gloryfromca: Non-blocking: the `bool(tools)` half of this gate is narrower than the rule #297 actually established, and it buys nothing.

The reporter's bisect variant B was "same payload, `tools` param removed" -> still a 400. DeepSeek's own doc says a request that once carried `tools` must replay `reasoning_content` in *all* subsequent requests. What decides the rejection is the message shape -- an assistant with pending `tool_calls` and no `reasoning_content` key, followed by a tool result at the end of the array -- not whether this particular request declares tools.

I could not find a live Raven path that this breaks today, which is why it is not a blocker. The one `tools=None` call site that replays tool-bearing history is `_synthesize_final_on_exhaustion` (`raven/agent/loop/main.py:2034`), and it appends a user message after the tool result, which is bisect variant G -> accepted. No TokenWise strategy empties the tool list either.

So the gate is safe now and wrong in principle: the day someone adds a tools-less continuation that ends on a tool result, this comes back as a 400 that looks unrelated to this PR. Dropping `bool(tools) and` costs one dict key on messages that already carry `tool_calls`. If you keep it, a line saying it is deliberate would help the next reader, since the issue explicitly ruled tools out as the trigger.

raven/providers/litellm_provider.py
```diff
@@ -396,8 +398,18 @@ def map_id(value: Any) -> Any:
...
                 clean["tool_call_id"] = map_id(clean["tool_call_id"])
+            if ensure_tool_reasoning and clean.get("role") == "assistant" and clean.get("tool_calls"):
+                clean.setdefault("reasoning_content", "")
         return sanitized
 
+    def _requires_tool_reasoning_replay(self, original_model: str, resolved_model: str) -> bool:
+        """Return whether the wire model requires reasoning keys on tool continuations."""
+        for model in (original_model, resolved_model):
+            upstream_model = self._strip_gateway_prefix(model).lower()
+            if upstream_model.startswith("deepseek/deepseek-v4-"):
```
gloryfromca, line 409: Non-blocking scope note: this matches direct DeepSeek and OpenRouter, and cannot match AiHubMix -- which is in Raven's own onboarding catalog and fronts DeepSeek's official API.

AiHubMix's spec sets `strip_model_prefix=True` and `via_driver="openai"` (`raven/providers/registry.py:254`), so a stored `aihubmix/deepseek-v4-pro` resolves to `openai/deepseek-v4-pro`, and `_strip_gateway_prefix` -- which strips the gateway's `model_prefix`, i.e. `openai/` -- leaves `deepseek-v4-pro`. Neither argument starts with `deepseek/`, so the replay never fires. The same applies to an OpenAI-compatible provider pointed at `api.deepseek.com`, where the id is stored under the custom provider's prefix.

That is not a regression -- those routes are broken today too -- and I have not measured whether AiHubMix relays DeepSeek's 400 verbatim, so I am not asking you to widen the match blind. Just worth knowing the issue stays open for those users. If you want the cheap version, matching `"deepseek-v4-" in upstream_model` instead of `startswith("deepseek/deepseek-v4-")` would cover them, at the cost of also firing for self-hosted `deepseek-ai/DeepSeek-V4-*` on SiliconFlow etc., where the extra key is inert anyway.

gloryfromca (commented):
No blockers; this can merge as far as I am concerned. Two scope notes below, neither of which is a bug today.

The fix matches the contract the reporters actually established in #297: DeepSeek thinking mode rejects only an *absent* `reasoning_content` key on the assistant message that issued the pending tool calls -- any value passes, including `""` (bisect variant H). `setdefault` is the right primitive here: a real reasoning string, or an explicit `null`, both survive untouched, and the key is only added on the wire, so persisted history and the archive keep the shape they had. Injecting it on every assistant-with-tool_calls rather than just the last one is harmless and cheaper to reason about than tracking which one is pending.

**What I checked**

- The diff, its callers, and `_sanitize_messages`' ordering. `reasoning_content` is already in `_ALLOWED_MSG_KEYS`, so an existing value survives `_sanitize_request_messages` and `setdefault` is a no-op on it. `_sanitize_empty_content` runs first and only touches `content`, so nothing races.
- Whether `reasoning_content: None` could reach the sanitizer and defeat `setdefault`. `build_assistant_message` (`raven/utils/helpers.py:252`) only writes the key when the value is not `None`, and the curator / history-trimmer allowlists copy keys rather than synthesise them -- so the key is either absent or a real string. The Risk section's "explicit null values remain unchanged" holds.
- Model resolution for both id shapes. A bare `deepseek-v4-pro` misses on `original_model` but matches on `resolved_model` after `wire_model` prefixes it; `openrouter/deepseek/deepseek-v4-*` matches after `_strip_gateway_prefix`. Checking both arguments is what makes that work.
- Backward compatibility: gated on model and on `tools`, so no other provider's payload changes. Confirmed by the negative test over `deepseek/deepseek-chat` and `openai/gpt-4o`.
- Tests are not weakened. I reverted `litellm_provider.py` to `origin/main` with the new tests in place: 4 of the 6 new cases fail with `KeyError` on `reasoning_content`, so they genuinely pin the behaviour rather than passing vacuously. With the fix: `uv run pytest tests/test_litellm_provider_stream.py tests/test_agent_loop_stream.py tests/test_subagent_manager.py tests/test_provider_resolution_invariants.py -q` -> 219 passed, 30 skipped (the skips are that file's pre-existing per-provider parametrize guards -- "not a gateway" etc. -- not anything this PR touches).
- AGENTS.md: commit grammar and scope are fine, the message is ASCII, the tests went into the existing `test_litellm_provider_stream.py` with its "Covers:" docstring updated rather than a new file, and no assets are added. The hardcoded model prefix in the provider is consistent with the neighbouring `_extra_msg_keys` / `_ANTHROPIC_EXTRA_KEYS` precedent, so I am not asking for it to move into the registry.

One thing I could not verify: the OpenRouter row is inferred rather than measured. `capabilities.py` is explicit that its own tables list only what was tested against a live key. Sending `reasoning_content` to OpenRouter is very likely inert, but if you have a key handy it is worth one real call before merging.

gloryfromca (dismissed):

gloryfromca:

```bash
git fetch upstream main
git push --force-with-lease
```

PR #370, opened 2026-08-28
feat: interactive trajectory browser with attempt merge and split

forrestjgq:
## Summary

`raven trajectory` had eight id-oriented subcommands; using them meant
copying ids out of `list` by hand. This PR changes the attempt data model
and adds the human face, in four phases:

1. **Attempt definitions** (`attempts.json` sidecar): an attempt id equals
   the trace id unless a definition groups several traces under one minted
   id. Merge creates a definition (absorbing prior definitions and legacy
   groups as aliases, so verdicts/pins recorded under old ids stay
   visible); split deletes it, migrating pins down to members - naturally
   undoable, span logs stay append-only. Pin migration is linearized under
   a fixed lock order (attempts > pins); every failure and concurrency
   interleaving over-protects, never under-protects. Legacy logs
   (span-level `attempt.id`) stay addressable and mergeable, but cannot be
   split (their grouping lives in append-only spans).
2. **Write-side removal**: the span-level attempt mechanism (the
   begin/end/current_attempt trio and the `attempt.id` span attribute,
   shipped in v0.1.13) is deleted; `attempt.id` is now a reserved
   attribute key stripped from caller input, so `attempts.json` is the
   sole grouping source by mechanism. Readers keep resolving legacy logs;
   zero data migration. This is a breaking API change, declared in the
   BREAKING CHANGE footer below.
3. **CLI adaptation and reader defect tolerance**: `list` folds definition
   members into one row and surfaces verdicts through the alias set; new
   `merge`/`split` subcommands wrap the data layer thinly; `unpin` clears
   definition, aliases, members, and the literal id in one transaction.
   Two whole crash surfaces are closed along the way: Rich markup
   injection (a legal id may contain `[/red]`; every dynamic value is
   escaped), and JSON-legal but type-broken records (per-line UTF-8
   decoding, required-field string validation, non-object attribute
   containers) now degrade per record/field instead of killing commands.
4. **Interactive browser**: a bare `raven trajectory` on a TTY opens a
   three-screen questionary flow (sessions -> attempts -> actions: save /
   report / minimize / verdict / pin|unpin / split, plus multi-select
   merge). Menus and action messages never show a session key, trace id,
   or attempt id - only artifact paths carry ids. Every action triggers a
   full rescan (report bundles and auto-pins before confirming, so even an
   aborted action changed state); data-layer errors surface as one fixed
   id-free message; aggregation reads one snapshot per refresh and
   deduplicates records into logical spans keyed by (traceId, spanId), so
   a root turn's checkpoint+final pair counts once.

CONTEXT.md gains the Attempt Definition term and the final-state Attempt
wording.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_cli_trajectory_browse.py tests/test_cli_trajectory_commands.py tests/test_trajectory_store.py tests/test_trajectory_bundle.py tests/test_tracing_api.py -q` -> 241 passed
- `uv run pytest tests/ -q --ignore=tests/integration` -> 6838 passed; the 11 failures + 20 errors match main's pre-existing environment-specific set node-id for node-id (cron timezone cases, everos/config root-permission cases, viewer probe, theme), compared as sorted FAILED/ERROR sets against a baseline recorded before this branch
- `uv run ruff check .` and `uv run ruff format --check .` -> clean

- [x] User-facing docs or screenshots are updated when needed (CONTEXT.md terms)

## Risk

- Bare `raven trajectory` changes behavior: it opens the browser on a TTY
  (previously the help page) and exits 2 with a hint when non-interactive.
  Scripts are unaffected: the subcommand check runs before the TTY gate.
- New spans no longer carry `attempt.id`, and `trace.begin_attempt` /
  `trace.end_attempt` / `trace.current_attempt` are removed. These shipped
  in v0.1.13, so an integration calling them raises AttributeError right
  after upgrading - an explicit, immediately visible failure rather than a
  silent one. Migration: group attempts after recording with
  `merge_attempts()` (library) or `raven trajectory merge` (CLI); reader
  fallback keeps existing logs addressable with zero data migration.
- `attempts.json` is the only new mutable state; deleting it reverts every
  attempt to single-turn semantics. Span logs remain append-only.
- Rollback: revert the squash commit; no data migration either way. The one
  known crash window (process death between merge's two file writes) leaves
  a harmless over-protective pin, documented in the store module.

- [x] Security impact considered (markup-injection and path-escape surfaces
  closed; ids treated as untrusted in every renderer)
- [x] Backward compatibility considered (legacy span-attribute logs stay
  addressable, mergeable, and listable)

## Related Issues

#362 (reference only: this PR removes the write-side attempt mechanism
introduced there and shipped in v0.1.13). No issue is closed.

BREAKING CHANGE: trace.begin_attempt, trace.end_attempt, and
trace.current_attempt (shipped in v0.1.13) and the span-level attempt.id
attribute are removed. Group attempts after recording instead:
merge_attempts() in raven.trajectory, or `raven trajectory merge` in the
CLI. Existing logs carrying a span-level attempt.id stay addressable
through the reader fallback; no data migration is needed.

### DIFF
--- a/raven/tracing/trace.py
+++ b/raven/tracing/trace.py
@@ -50,7 +50,6 @@ class Span:
         "kind",
         "trace_id",
         "span_id",
-        "attempt_id",
         "_parent",
         "_start",
         "_attrs",
@@ -78,13 +77,11 @@ class Span:
         chat_id,
         start,
         source=None,
-        attempt_id=None,
     ):
         self.name = name
         self.kind = kind
         self.trace_id = trace_id
         self.span_id = span_id
-        self.attempt_id = attempt_id
         self._parent = parent
         self._session_key = session_key
         self._channel = channel
@@ -177,7 +174,6 @@ class Span:
                     session_key=self._session_key,
                     channel=self._channel,
                     chat_id=self._chat_id,
-                    attempt_id=self.attempt_id,
                     start_time=self._start,
                     end_time=_spans.now_iso(),
                     status_code=self._status_code,
@@ -197,7 +193,6 @@ class _NoopSpan:
     trace_id = ""
     span_id = ""
     name = ""
-    attempt_id = ""
     invocation_source = None
 
     def set(self, *_a, **_k):
@@ -257,9 +252,6 @@ def span(
         session_key = session_key if session_key is not None else (cur.session_key if cur else None)
         channel = channel if channel is not None else (cur.channel if cur else None)
         chat_id = chat_id if chat_id is not None else (cur.chat_id if cur else None)
-        # Children inherit the tree's attempt id; a root span resolves it from
-        # the session's open attempt, else this trace is a single-turn attempt.
-        attempt_id = (cur.attempt_id if cur else None) or _ctx.current_attempt(session_key) or trace_id
         span_id = _ctx.new_span_id()
         handle = Span(
             name,
@@ -272,7 +264,6 @@ def span(
             chat_id=chat_id,
             start=_spans.now_iso(),
             source=cur.source if cur else None,
-            attempt_id=attempt_id,
         )
         handle.set(attributes, **kw)
         # A detached span is a leaf marker: it does NOT become the active parent,
@@ -290,7 +281,6 @@ def span(
                 session_key=session_key,
                 channel=channel,
                 chat_id=chat_id,
-                attempt_id=attempt_id,
             )
         )
     except Exception:  # noqa: BLE001 — open must never break the host
@@ -318,7 +308,6 @@ def span(
                         session_key=handle._session_key,
                         channel=handle._channel,
                         chat_id=handle._chat_id,
-                        attempt_id=handle.attempt_id,
                         start_time=handle._start,
                         end_time=_spans.now_iso(),
                         status_code=handle._status_code,
@@ -432,38 +421,6 @@ def instrument(name: str, *, kind: str | None = None, detached: bool = False, se
 def current() -> Any | None:
     """The active trace context (or None). Exposed for adopters that need it."""
     return _ctx.current()
-
-
-def begin_attempt(session_key: str, attempt_id: str | None = None) -> str:
-    """Open a multi-turn attempt for ``session_key``.
-
-    Every span of every turn opened while the attempt is active carries its id
-    as ``attempt.id`` — the stable trajectory address a task spanning several
-    turns groups under. Without an open attempt each turn is its own
-    single-turn attempt (``attempt.id`` = trace id). No-throw; returns the id.
-    """
-    try:
-        return _ctx.begin_attempt(session_key, attempt_id)
-    except Exception:  # noqa: BLE001 — tracing must never break the host
-        _log.debug("tracing: begin_attempt failed", exc_info=True)
-        return attempt_id or ""
-
-
-def end_attempt(session_key: str) -> str | None:
-    """Close the open attempt for ``session_key`` (no-op if none). No-throw."""
-    try:
-        return _ctx.end_attempt(session_key)
-    except Exception:  # noqa: BLE001
-        _log.debug("tracing: end_attempt failed", exc_info=True)
-        return None
-
-
-def current_attempt(session_key: str | None) -> str | None:
-    """The open attempt id for ``session_key``, or None. No-throw."""
-    try:
-        return _ctx.current_attempt(session_key)
-    except Exception:  # noqa: BLE001
-        return None
 
 
 def enabled() -> bool:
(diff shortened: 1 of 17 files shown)


raven/tracing/trace.py
```diff
@@ -434,38 +423,6 @@ def current() -> Any | None:
     return _ctx.current()
 
 
-def begin_attempt(session_key: str, attempt_id: str | None = None) -> str:
```
gloryfromca, line 437: `v0.1.13` already published `trace.begin_attempt`, `trace.end_attempt`, and `trace.current_attempt`: commit `0d2f2c25` is an ancestor of the `v0.1.13` tag, and that tag's `CONTEXT.md` documents `trace.begin_attempt(session_key)`. Removing these functions means an integration that works on v0.1.13 now raises `AttributeError` immediately after upgrading; the associated removal of `Span.attempt_id` and emitted `attempt.id` also changes the released tracing contract. Please preserve a compatibility/deprecation path or handle this explicitly as a versioned breaking change instead of relying on the PR's `never released` premise.

gloryfromca (commented):
Blocking: the released tracing attempt API is removed without a compatibility path; see the inline note.

I reviewed the full diff, surrounding callers, commit and release history, backward compatibility, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md and CONTEXT.md terminology, and the changed tests. I found no evidence that tests were weakened merely to obtain a green result.

Verification: `uv run pytest tests/test_cli_trajectory_browse.py tests/test_cli_trajectory_commands.py tests/test_trajectory_store.py tests/test_trajectory_bundle.py tests/test_tracing_api.py -q` completed all assertions with 241 passed, then the Python 3.13.2 process exited 139 during interpreter shutdown. The same command's implicated file on `github/main` completed 36 passed and then exited 139, so I treated that shutdown crash as pre-existing rather than a finding on this PR.

raven/tracing/trace.py
forrestjgq: Confirmed - you are right, and the PR's premise was wrong. Verified locally: `0d2f2c25` is an ancestor of the `v0.1.13` tag, that tag ships all three functions in `raven/tracing/trace.py`, and its CONTEXT.md documents `trace.begin_attempt(session_key)`. The "never released" claim only checked in-repo callers, not the release history.

We are handling this explicitly as a versioned breaking change rather than a compatibility shim, for one main reason: the old semantics cannot actually be preserved. New spans no longer carry `attempt.id` (that is the point of this PR), so a shim could not make `begin_attempt`'s returned id address anything - a single-turn attempt cannot even be represented (definitions require >= 2 member traces), and a near-miss shim would leave callers' verdicts/pins silently dangling on ids that resolve to nothing. An immediate `AttributeError` with a clear migration path (group after recording via `merge_attempts()` or `raven trajectory merge`) is diagnosable in minutes; the silent variant is not.

The PR description is updated accordingly: the wrong "unreleased" wording is corrected in Summary/Risk/Related Issues, and a `BREAKING CHANGE:` footer now declares the removal with the migration path (the squash commit body inherits it). Logs already carrying span-level `attempt.id` remain addressable, mergeable, and listable through the reader fallback - no data migration.

gloryfromca: Accepted. This resolves my finding: the PR now accurately identifies the v0.1.13 surface as released, explicitly treats its removal as breaking, provides the concrete post-recording migration, and places the `BREAKING CHANGE:` footer in the squash body. I agree that a compatibility shim which returns an unresolvable id would be worse than the explicit versioned break. No blocker remains from this thread.

Verification rerun: the relevant suite completed 241 assertions successfully, then the Python 3.13.2 process exited 139 during the same pre-existing interpreter-shutdown crash already reproduced on `github/main`.

forrestjgq:
```
Verification rerun: the relevant suite completed 241 assertions successfully, then the Python 3.13.2 process exited 139 during the same pre-existing interpreter-shutdown crash already reproduced on github/main.
```
so my work is done and no more blocking issues, correct?

gloryfromca: Correct. As far as my review is concerned, the versioned-breaking-change documentation and migration path resolve the only blocker I raised, and no blocking issue remains on this unchanged revision.

I reran the relevant suite at `1cb24a4b8dba`: all 241 tests passed, after which Python 3.13.2 exited 139 in the interpreter-shutdown crash already reproduced on `github/main`; that baseline-only exit does not block this PR.

gloryfromca (approved):

PR #380, opened 2026-09-05
feat(trajectory): tester-facing bug report flow in raven trajectory

forrestjgq:
## Summary

Tester-facing bug report flow inside the `raven trajectory` browser
(and a scriptable `report-bug` subcommand): pick an attempt, type a
one-line description, confirm once, and get a stable report id with a
local shippable package.

Key decisions, in the order the phases landed:

- **Two artifacts.** A machine-local `record.json` (lifecycle: `draft`
  -> `local_ready` | retryable `failed`, crash recovery for interrupted
  drafts, 26 structural invariants on load) and the only exportable
  artifact `<report-id>.tar.gz` - sanitized problem metadata plus an
  embedded Trajectory Report in its existing layout. Everything the
  package will contain is frozen under `snapshot/export/` (tree digest,
  verified before and after every tar run) *before* the confirmation
  screen, so the text the user approves, the first packaging run, and
  any retry use the same bytes.
- **Safety gates.** User input goes through the same known-value /
  pattern / residual redaction as the trajectory; the merged signals
  classify the report as `clean` (one confirmation), `needs_review`
  (separate authorization; `--yes` never implies it), or `blocked`
  (private-key hit; no flag can bypass it). Two sanitization layers
  strip absolute paths (generic POSIX/Windows/UNC token parser shared
  with the pre-export assertions), tar member headers are anonymized on
  both layers, and `RAVEN_BUGREPORT_REQUIRE_REVIEW` adds an
  organization-policy review trigger.
- **Completeness that cannot drift.** The deliverable tree is judged by
  an offline warn-mode mock replay probe plus explicit contract checks
  (`validate_recording`, corrupt span lines, format version), mapping to
  `complete`/`degraded`/`unreplayable` with reasons; unexplained probe
  failures degrade to `unknown` instead of masquerading as evidence
  damage. The probe runs with a new `rates_offline()` context so no
  pricing/catalog/vision-warm path can reach the network.
- **Compatibility.** `raven trajectory report` (including `--yes`) and
  every existing browser action keep their exact behavior and outputs;
  the new flow writes only under `<trace-state>/bugreports/`. All new
  functions (`member_traces`, `classify_redaction`, the sanitizer set,
  `evaluate_completeness`, `diagnose_history_cut`, `validate_recording`,
  `rates_offline`) are pure additions.

Two pre-existing gaps found and fixed along the way: the sanitizer's
placeholder path for missing artifacts made delivered trajectories fail
[...]

PR #381, opened 2026-09-07
feat(trajectory): adjudicate redaction findings per item in bug reports

forrestjgq:
## Summary

Bug reports previously classified redaction into clean / needs_review /
blocked, where a private-key hit refused to produce a report and the
needs_review confirmation showed at most 5 masked samples pointing at a
file buried inside the frozen snapshot. This PR folds blocked into a
per-item review flow:

- Residual findings become review items, adjudicated by token value:
  keep as-is, or replace with [REDACTED:user-confirmed] (all four
  spelling variants replaced and verified gone, tar member names
  included - files carrying a redacted value are renamed with the same
  replacement their references got). Each item leads with the token and
  its occurrence count, groups identical context lines (every hit
  highlighted), and lists every location as semantic source plus
  file:line ("model call #2 input - artifacts/...json:57"); decisions
  are single-letter input (k/r/c) with no default. The user never has
  to open a file, and can still locate everything.
- A private-key hit no longer blocks the report: the content is already
  replaced, the reporter acknowledges the risk per item, and the package
  carries an explicit security notice.
- One risk-worded confirmation replaces the old create-then-review
  double prompt; --yes never implies the risk authorization. The
  summary shows a Reviewed tally (kept / redacted / acknowledged) with
  short masked tokens, keeping completed decisions apart from the
  warnings that still need consent.
- Records and packages gain additive redaction fields security_notices
  and user_decisions (with a masked_token short identifier; absent on
  old records, validated when present); the persisted classification
  enum is unchanged, so old and new readers stay compatible both ways.
- CLI: --keep-findings / --redact-findings (mutually exclusive) decide
  suspected items per kind for scripts; a run missing flags fails before
  any content change, listing every item and every missing flag at once.
- Provider-generated call_ tool-call ids and the fixed usage field
  names (completion_tokens_details, prompt_tokens_details; exact
  literals only) join the residual-scan exemptions.
- Untouched: the freeze pipeline (digests, byte-identical retries, crash
  recovery), the completeness probe, path sanitization, tar header
  anonymization, the expert command raven trajectory report, and the
  record state machine.

Note: this branch is stacked on feat/trajectory_bug_report (#380); the
diff includes those commits until #380 merges, after which this branch
will be rebased onto main.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_trajectory_redact.py tests/test_trajectory_review.py tests/test_trajectory_bugreport.py tests/test_cli_trajectory_browse.py tests/test_cli_trajectory_commands.py -q` -> 380 passed
- `uv run pytest tests/ --ignore=tests/integration` -> 7156 passed (the 11 failures / 19 errors are pre-existing environment issues on this machine - missing tzdata and similar - identical on the base branch)
- `uv run --frozen --extra dev ruff check raven/ tests/` -> passed
- `uv run --frozen --extra dev ruff format --check raven/ tests/` -> 861 files already formatted
- `make check-large-files` -> passed

## Risk

User-visible changes: the blocked refusal wall is gone (replaced by an
informed acknowledgment with an in-package notice), the review
confirmation is a single risk-worded prompt instead of two, and
non-interactive review runs now require an explicit --keep-findings or
--redact-findings for suspected items (previously --accept-risk alone
shipped them as kept). Rollback: revert the squash commit; the new
metadata fields are additive, so already-produced packages and records
keep loading either way.

## Related Issues

N/A

### DIFF
--- a/tests/test_trajectory_replay.py
+++ b/tests/test_trajectory_replay.py
@@ -577,6 +577,25 @@ async def test_run_replay_drives_the_loop_and_executes_nothing(tmp_path, monkeyp
     assert trace.enabled(), "suppression must not outlive the replay"
 
 
+async def test_run_replay_leaves_no_skill_watcher_threads(tmp_path) -> None:
+    """The loop's ContextBuilder starts a skill file watcher; a replay must
+    stop it, or repeated probes leak daemon threads that crash the process at
+    interpreter shutdown (found by PR review on the completeness probe)."""
+    import threading
+
+    bundle = _make_bundle(
+        tmp_path,
+        llm_calls=[(None, _llm_output(content="done"))],
+        turns=[{"content": "go", "channel": "cli", "chat_id": "direct"}],
+    )
+
+    report = await run_replay(bundle, mode="warn")
+
+    assert report.complete
+    watchers = [t for t in threading.enumerate() if t.name == "SkillFileWatcher" and t.is_alive()]
+    assert watchers == [], "a replay must not leave skill watcher threads running"
+
+
 async def test_run_replay_flags_unconsumed_recording(tmp_path) -> None:
     """A harness that never asks for the rest of the recording diverged too:
     strict halts (exit path), warn records a non-fatal divergence."""
@@ -813,6 +832,7 @@ async def test_end_to_end_record_save_replay_with_real_tracer(tmp_path, monkeypa
             max_iterations=5,
             restrict_to_workspace=True,
         )
+        loop.context.skills.stop_file_watcher()
         loop.tools.register(_MarkerTool(marker))
         # Source identity and session key agree, as they do in every real
         # channel turn (the session key defaults to "<channel>:<chat_id>").
@@ -870,6 +890,7 @@ async def test_end_to_end_replay_after_harness_change_diverges(tmp_path, monkeyp
             max_iterations=5,
             restrict_to_workspace=True,
         )
+        loop.context.skills.stop_file_watcher()
         loop.tools.register(_MarkerTool(tmp_path / "marker"))
         await loop._process_message(
             TurnRequest(
@@ -924,6 +945,7 @@ async def test_end_to_end_streamed_recording_replays_through_the_stream_path(tmp
             max_iterations=5,
             restrict_to_workspace=True,
         )
+        loop.context.skills.stop_file_watcher()
         loop.tools.register(_MarkerTool(marker))
         streamed_tokens: list[str] = []
 
@@ -1011,6 +1033,7 @@ async def test_end_to_end_replay_restores_pre_attempt_history(tmp_path, monkeypa
             max_iterations=5,
             restrict_to_workspace=True,
         )
+        loop.context.skills.stop_file_watcher()
         loop.tools.register(_MarkerTool(marker))
 
         def req(text: str) -> TurnRequest:
@@ -1075,6 +1098,7 @@ async def test_end_to_end_replay_restores_history_when_first_input_repeats(tmp_p
             max_iterations=5,
             restrict_to_workspace=True,
         )
+        loop.context.skills.stop_file_watcher()
         loop.tools.register(_MarkerTool(marker))
 
         def req() -> TurnRequest:
--- a/tests/test_trajectory_replay.py
+++ b/tests/test_trajectory_replay.py
@@ -580,7 +580,8 @@ async def test_run_replay_drives_the_loop_and_executes_nothing(tmp_path, monkeyp
 async def test_run_replay_leaves_no_skill_watcher_threads(tmp_path) -> None:
     """The loop's ContextBuilder starts a skill file watcher; a replay must
     stop it, or repeated probes leak daemon threads that crash the process at
-    interpreter shutdown (found by PR review on the completeness probe)."""
+    interpreter shutdown. Measured as a delta: other suites in the same
+    process may hold their own watcher threads."""
     import threading
 
     bundle = _make_bundle(
@@ -588,12 +589,15 @@ async def test_run_replay_leaves_no_skill_watcher_threads(tmp_path) -> None:
         llm_calls=[(None, _llm_output(content="done"))],
         turns=[{"content": "go", "channel": "cli", "chat_id": "direct"}],
     )
+    before = {t.ident for t in threading.enumerate() if t.name == "SkillFileWatcher"}
 
     report = await run_replay(bundle, mode="warn")
 
     assert report.complete
-    watchers = [t for t in threading.enumerate() if t.name == "SkillFileWatcher" and t.is_alive()]
-    assert watchers == [], "a replay must not leave skill watcher threads running"
+    leaked = [
+        t for t in threading.enumerate() if t.name == "SkillFileWatcher" and t.is_alive() and t.ident not in before
+    ]
+    assert leaked == [], "a replay must not leave new skill watcher threads running"
 
 
 async def test_run_replay_flags_unconsumed_recording(tmp_path) -> None:
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -653,6 +653,26 @@ never re-collects.
 _Avoid_: confusing it with the Trajectory Report it embeds — the Package wraps
 one and adds the problem metadata envelope.
 
+**Review Item** (`raven/trajectory/review.py`):
+One decision the user must make before a Bug Report Package may ship, built by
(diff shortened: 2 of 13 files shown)


raven/trajectory/completeness.py
```diff
@@ -0,0 +1,243 @@
...
+def _probe(tree: Path) -> list[str]:
+    """Unreplayable reasons from one warn-mode mock replay of the tree.
+
+    Only explicitly recognized outcomes are classified here: the entry
+    rejection (``ValueError``) and fatal divergences. Anything else raises to
+    the caller — an unexplained probe failure is an evaluation failure, not
+    evidence damage.
+    """
+    try:
+        report = asyncio.run(run_replay(tree, mode="warn"))
```
gloryfromca, line 141: Blocking: each completeness probe leaks a live SkillFileWatcher, causing repeated probes and the changed test suite to segfault at process shutdown.

`run_replay` constructs an `AgentLoop`, whose default `ContextBuilder` starts the Rust-backed watcher, but neither replay nor this caller stops it. This is reproducible with `uv run pytest tests/test_trajectory_completeness.py::test_complete_sample_probes_clean -q`: it prints `1 passed` but exits 139; after its two replay calls, `threading.enumerate()` shows two live `SkillFileWatcher` threads. Overriding `ContextBuilder` with `start_watcher=False` makes the same reproducer exit 0. The interactive trajectory browser can invoke this probe repeatedly, so the probe needs to avoid starting the watcher or tear the loop's watcher down.

gloryfromca (commented):
Blocking: stop the completeness probe from leaking SkillFileWatcher threads and crashing at process shutdown.

I reviewed the full `github/main...HEAD` diff and the surrounding bundle, replay, provider, pricing, store, CLI, and watcher call paths. I also checked the branch history, backward-compatible record validation, test changes for weakening, and the repository rules in `AGENTS.md`/`CLAUDE.md` plus the canonical trajectory terms in `CONTEXT-MAP.md` and `CONTEXT.md`. The per-item decision, redaction, path-sanitization, frozen-package, retry, and CLI authorization flows otherwise look internally consistent.

Verification: `uv run pytest tests/test_trajectory_redact.py tests/test_trajectory_review.py tests/test_trajectory_bugreport.py tests/test_trajectory_completeness.py tests/test_cli_trajectory_commands.py tests/test_cli_trajectory_browse.py tests/test_provider_rates.py tests/test_token_wise_pricing.py -x` reached `499 passed in 16.38s`, then segfaulted and exited 139. The isolated completeness reproducer also prints `1 passed` and exits 139; preventing `ContextBuilder` from starting its watcher makes that same two-replay reproducer exit 0.

raven/trajectory/completeness.py
forrestjgq: Fixed in bf934b0. The stop lives at the run_replay level rather than only in the probe: right after the AgentLoop is built, its ContextBuilder's skill watcher is stopped via the catalog's stop_file_watcher() (a replay feeds recorded content only, so the watcher has nothing to do), which covers the completeness probe, the replay CLI, and the regression runner in one place. The end-to-end recording tests in test_trajectory_replay.py stop the watcher on their five directly-built loops too (their leak pre-existed on the base branch), and a new regression test asserts run_replay leaves no SkillFileWatcher threads alive. Verification: your isolated reproducer (test_complete_sample_probes_clean) now exits 0, the full 8-file command reaches 499 passed with exit 0, and test_trajectory_replay.py alone is 32 passed exit 0 (previously 139).

gloryfromca: Blocking: remove the transient PR-review provenance from the new test docstring before this revision merges.

The watcher leak itself is resolved by bf934b0. Stopping the watcher in `run_replay` covers the completeness probe and the other replay callers, the original isolated reproducer now exits 0, and the full changed-area run completes normally with 531 passing tests. The remaining blocker is separate from this thread and is marked on the new test docstring.

tests/test_trajectory_replay.py
```diff
@@ -577,6 +577,25 @@ async def test_run_replay_drives_the_loop_and_executes_nothing(tmp_path, monkeyp
     assert trace.enabled(), "suppression must not outlive the replay"
 
 
+async def test_run_replay_leaves_no_skill_watcher_threads(tmp_path) -> None:
+    """The loop's ContextBuilder starts a skill file watcher; a replay must
+    stop it, or repeated probes leak daemon threads that crash the process at
+    interpreter shutdown (found by PR review on the completeness probe)."""
```
gloryfromca: Blocking: remove the transient PR-review provenance from this test docstring.

The parenthetical `found by PR review` is historical task provenance, not part of the behavior being specified. `AGENTS.md` section 1.1 explicitly forbids repository comments that reference a PR or describe transient task context because they become stale after the task. The rest of the docstring already explains the regression completely, so please drop only that parenthetical.

gloryfromca (commented):
Blocking: remove the transient PR-review provenance from the new regression-test docstring.

The watcher lifecycle fix is correct at the shared `run_replay` layer: I checked the new delta, replay/completeness/browser callers, watcher stop semantics, branch history, backward compatibility, test changes for weakening, and the repository rules in `AGENTS.md`/`CLAUDE.md` plus the trajectory terminology in `CONTEXT-MAP.md` and `CONTEXT.md`. I found no other issue in this revision.

Verification: the original reproducer now completes with `1 passed` and exit 0. `uv run pytest tests/test_trajectory_redact.py tests/test_trajectory_review.py tests/test_trajectory_bugreport.py tests/test_trajectory_completeness.py tests/test_trajectory_replay.py tests/test_cli_trajectory_commands.py tests/test_cli_trajectory_browse.py tests/test_provider_rates.py tests/test_token_wise_pricing.py -x` completed with 531 passed in 21.03s and exit 0.

tests/test_trajectory_replay.py
forrestjgq: Fixed in e363812: the parenthetical is gone, and the same cleanup was applied to three sibling docstrings that carried the same kind of transient wording (review-verified trap / reviewed-verified leak in test_trajectory_review.py, test_trajectory_bugreport.py, test_cli_trajectory_browse.py) - the behavior descriptions stand on their own now. The same commit also fixes the CI failure this test caused under the full-suite process: the assertion was global, counting watcher threads leaked by other suites in the same process; it now snapshots SkillFileWatcher threads before the replay and asserts the replay added none. Verified with a leaking suite run first in the same process (tests/test_decision_consumer.py + tests/test_trajectory_replay.py -> 41 passed, exit 0) and the replay file alone (32 passed, exit 0).

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed: the transient provenance wording is gone, and scoping the watcher assertion to thread identities created during this replay preserves the regression check without depending on unrelated suite-owned watchers. The 531 changed-area tests pass with exit 0.

gloryfromca (commented):

Both prior findings are resolved. I reviewed the new delta and rechecked its replay/completeness/browser callers, watcher lifecycle semantics, branch history, backward compatibility, test changes for weakening, and the repository rules in `AGENTS.md`/`CLAUDE.md` plus the trajectory terms in `CONTEXT-MAP.md` and `CONTEXT.md`. The delta-scoped watcher assertion remains capable of detecting a thread introduced by `run_replay`, while tolerating pre-existing watchers owned by other suites.

Verification: the four directly touched test files completed with 290 passed in 10.80s. The full changed-area command completed with 531 passed in 17.45s; both exited 0. A full `uv run pytest -x` run stopped after 1750 passed on `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`; that test also fails in isolation, and neither it nor its implementation is changed by this branch, so I did not attribute it to this revision.

gloryfromca (commented):

This is a history-only refresh: `git diff e363812a253c HEAD` is empty, so the reviewed source and test tree is unchanged. I refreshed the target ref and rechecked the actual `github/main...HEAD` diff and rewritten history against the repository rules; caller behavior, backward compatibility, architectural fit, and test strength remain as previously reviewed.

Verification: `uv run pytest tests/test_trajectory_completeness.py::test_complete_sample_probes_clean tests/test_trajectory_replay.py::test_run_replay_leaves_no_skill_watcher_threads -q` completed with 2 passed.

0xKT (approved):

PR #384, opened 2026-09-10
feat(agent): add Serply as a web_search provider

googio:
## Summary

`web_search` has only ever spoken to Serper: the endpoint, the `SERPER_API_KEY` fallback, the capability row, and the error text were all hard-wired to it. This adds Serply (https://serply.io) as a second provider behind a single new config field, `tools.web.search.provider`, which accepts `serper` (the default) or `serply`.

Key decisions:

- Nothing changes for existing deployments. The default is `serper`, the Serper request and its answer box / knowledge graph rendering are untouched, and every existing test still passes unmodified.
- The chosen provider decides which environment variable stands in for the configured key (`SERPER_API_KEY` or `SERPLY_API_KEY`). `WebSearchTool.is_configured` takes the provider, so the registration gate in the agent loop and the sub-agent manager, `capabilities.is_configured`, and `configured_from` all keep answering the same question the same way, which is the invariant `tests/test_tool_capabilities.py` pins.
- `raven doctor` rows follow the provider through a small `capabilities.for_deployment` helper, so a Serply deployment without a key is told to set `SERPLY_API_KEY` and go to serply.io rather than serper.dev. The `Capability` table itself is unchanged.
- The provider travels the same way the key already does: a `web_search_provider` keyword on `AgentLoop` and `SubagentManager`, filled by the three CLI entry points from config. It defaults to `serper`, so external callers of either constructor are unaffected.
- Serply results are normalised to the Serper shape (`title`, `link`, `snippet`) so the rendering loop is shared. Serply has no answer box or knowledge graph; those lines are simply absent.
- The two benchmark READMEs mention the switch next to the existing `SERPER_API_KEY` note.

Serply support is optional; behavior is unchanged when `tools.web.search.provider` is left at its default.

Disclosure: I work with Serply. Happy to adjust scope, naming, or drop this entirely if it isn't a direction you want for the project.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
uv run --frozen --python 3.12 --extra dev ruff check raven tests            -> All checks passed!
uv run --frozen --python 3.12 --extra dev ruff format --check raven tests   -> 861 files already formatted
uv run --frozen --python 3.12 --all-extras pytest -q                         -> 1 failed, 7437 passed, 33 skipped
  (the failure, test_cli_theme.py::test_bold_accent_renders_styled_not_bare, fails the same way on
   upstream/main on this machine and is unrelated; upstream/main before this change gave
   3 failed, 7430 passed, 33 skipped, the two extra being TUI tests that fail intermittently here)
npx commitlint --from upstream/main --to HEAD --config commitlint.config.cjs -> exit 0
```

New tests, all with the network replaced by a stub:

- `tests/test_agent_loop_web_tools.py`: a Serply deployment registers `web_search` from `SERPLY_API_KEY` alone while a Serper deployment ignores that variable; the unconfigured error names the chosen provider and its variable; the Serply request goes to `https://api.serply.io/v1/search/q=...&num=N` with the `X-Api-Key` header and the configured proxy, and renders like the Serper branch.
- `tests/test_tool_capabilities.py`: with `provider = "serply"`, only `SERPLY_API_KEY` counts, and the table agrees with the loop in both directions.
- `tests/test_cli_doctor_commands.py`: the `web_search` doctor row reports `SERPLY_API_KEY` / `https://serply.io` for a Serply deployment, in both the JSON and the text output.

Also exercised once against the live Serply API with a real key: `WebSearchTool(provider="serply").execute("EverMind Raven agent harness", count=2)` returned two organic results rendered in the existing format.

## Risk

- Security: the Serply key is sent only to `api.serply.io` over HTTPS in a request header, exactly as the Serper key is today, and the query is URL-encoded with `quote_plus`. The request goes through the same `httpx.AsyncClient(proxy=...)` as the Serper branch, so deployments that route search through a proxy keep doing so.
- Backward compatibility: `provider` defaults to `serper`; the new constructor keywords default to `serper`; `WebSearchTool.is_configured` keeps its one-argument form working. No config migration is needed. An unknown provider value is rejected at config load by the `Literal` type.
- Rollback: revert the single commit. No data or config format is changed, and a config that already says `"provider": "serply"` would fail validation on the reverted schema, which is the loud failure rather than a silent fallback.

## Related Issues

N/A


gloryfromca:

```bash
git fetch upstream main
git push --force-with-lease
```

PR #392, opened 2026-09-14
fix(*): give the frontend copy one vocabulary and one home

LivXue:
## Summary

The frontend carried user-visible copy in three broken states at once: English
literals in the source, which a Chinese reader simply saw in English; catalogue
entries whose English and Chinese said different things, so one of the two was
wrong wherever they disagreed; and Chinese text punctuated as English. This
routes the literals through the shared catalogue, gives each concept one
spelling in each language, and punctuates the Chinese as Chinese.

One hundred and two catalogue keys are new, all with both languages and all
referenced. No key is removed.

**The vocabulary decision.** The catalogue shipped both `Subagents` and
`Sub-agents`, and the Chinese carried two different words for the same thing.
Everything settles on the term
`CONTEXT.md` defines: its entry is spelled **Subagent**, and its Deep Research
entry quotes `"Subagent"` as the word that entry must not be confused with, so
that is the registry spelling AGENTS.md section 6 points at. The hyphenated form
was the catalogue's majority, which is why an earlier commit on this branch
merged the pair the wrong way; it is corrected here rather than left as the
majority-but-wrong choice. Nine English strings and two Chinese ones move, plus
nine more in the hotkey sheet and four slash commands that live outside the
catalogue entirely.

**Two shapes needed more than a lookup.**

- `SECTION_TITLES` fed `buildSubagentView`, a pure function whose returned lines
  are asserted as data. Localizing it there would have made a shaping function
  read process-global locale state, so the header now carries a `section` tag and
  the view names it. Two assertions move from the English literal to the tag.
- `STARTUP_MESSAGES` is built at import, which is before `setLocale` runs, so
  holding text there would have frozen the boot line in English whatever the
  config said. It holds keys and resolves at render.

**Left in English on purpose.** `line === 'analyzing tool output...'` and
`g.label.startsWith('Delegate Task')` in `thinking.tsx` match values arriving
from the backend. They are predicates, not copy, and translating them would
break the match.

**Disclosed, not fixed.** `modelPicker.tsx` (25 literals) and `agentsOverlay.tsx`
(14) are still English. Both are large enough that finishing them is its own
change, and neither regressed here: they were English before this branch and are
English after. Separately, the effort-tier descriptions mirrored in ui-web's demo
layer and in `shell/tier.test.ts` come from `raven/config/schema.py`, whose
Chinese translation in `raven/i18n/zh.py` still uses the retired word. Editing
the frontend copies alone would replace one disagreement with a worse one,
between the page and the server, so that is left for a `raven/` change.

**Review round.** One finding, non-blocking, fixed. `gui.panel.more_down` was
added with its upward twin and wired at one of its five call sites, so a scrolled
list drew the hint above its rows from the catalogue and the one below from an
English literal -- Chinese above and English below in the same frame, on any picker
holding more rows than a page. The other four sites now call it. The English locale
is unaffected, because the catalogue entry and the literal it replaces are the same
string, which is also why no existing assertion moved.

Its guard derives the set from the sources rather than listing the four call sites:
every component calling `more_up` must call `more_down` as often, and no component
may hardcode a paging hint. A list would hold today's pickers and let the next one
repeat the omission, which is the shape this defect already took once. Both
assertions were run against mutants -- reverting one site to the literal reddens
both, deleting one hint without hardcoding it reddens only the pairing one.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Gates, all from the branch head:

```
npm run lint:i18n --prefix ui-tui     generated catalogue is up to date
make check-commits                    exit 0
make check-source-language            exit 0
make check-large-files                exit 0
npx tsc --noEmit  (ui-web)            exit 0
npx tsc --noEmit  (ui-tui)            exit 0
```

`ui-tui`, `npx vitest run --no-file-parallelism`, against a detached worktree at
the then merge base `5b7566b49` run the same way:

```
base     11 failed | 2061 passed | 13 skipped (2072)   0 unhandled
branch   10 failed | 2049 passed | 13 skipped (2072)   0 unhandled
branch    9 failed | 2050 passed | 13 skipped (2072)   0 unhandled
```

Failing ids diffed rather than counted: **0 introduced**. Nine failures are
shared by the base and both branch runs -- four `dumpScreen`, two
`intensity transitions`, one slab-rule and two typing-burst tests, none of them
touched here. Three more are timing flakes rather than results: two `ModelPicker`
login tests fail on the base and pass on the branch, and one `NewInstancePicker`
refusal test failed in the first branch run, did not recur in the second, and
passes three times out of three in isolation. Its assertion waits on a string the
test's own mock throws, which this branch does not touch.

Re-measured after the rebase onto `56c79edf5`: 144 files, 9 failed, 2,052 passed
and 13 skipped, 0 unhandled. The same nine, and their ids diffed against the new
base run the same way are identical -- 0 introduced. The three timing flakes above
did not recur.

Independently re-run by the reviewer on a different machine at
`fcfb7d26e595` and again at `525adc7dcd78`: 143 files passed, 2,059 passed and
13 skipped, and `ui-web` 103 files and 1,678 tests passed. That is this
branch's own count plus exactly the nine, and the two storage tests green as
well -- so both sets are artifacts of the machine this branch was developed on
rather than anything a reviewer needs to carry.

`ui-web`, `npx vitest run --no-file-parallelism` with a `--localstorage-file`:

```
2 failed | 1676 passed (1678)   0 unhandled
```

Both failures are the two `when storage refuses` storage tests, which fail on an
untouched tree on this machine and are unrelated to this change. The unhandled
count is read separately because a broken hand-built harness in
`ui-web/scripts/*.test.mjs` surfaces there and not in the failure count.

Rendered rather than reasoned about, at `locale=zh` and `locale=en`: the
sub-agent hub with an empty roster and with one row, and the skills hub error
line. The hub previously drew one Chinese line in five; it is now whole in
Chinese, and byte-identical in English apart from the spelling.

### CI, and what is red

Two jobs were red on this branch's own account and both are green now.

`TUI checks` failed on eslint: every file that gained the catalogue import
appended it after the relative imports, which perfectionist/sort-imports
rejects. Nine errors, all introduced here; eslint is clean at `origin/main`.

`pre-commit diff` failed on `end-of-file-fixer`, not on anything this branch
wrote: `ui-web/src/live/110-connections.js` ends `};` followed by a blank
line, on `origin/main` as well, and the hook runs on changed files. Touching
that file to route its toast copy through the catalogue was enough to surface
it. Fixed here, with the hook's own one-byte fix.

Worth recording for whoever reads this job next: the `eslint (ui-tui)` and
`prettier (ui-tui)` hooks in that same run reported Passed in about a third of
a second each, while the tree had twenty-six files failing `prettier --check`.
They did not run. Both entries begin `[ -d ui-tui/node_modules ] || exit 0`,
and the pre-commit job's only install step is `make install-deps`, which is
`uv sync` -- no `npm ci` for any workspace. So those two hooks are inert in CI
and fire only on a developer's machine. eslint is covered anyway by
`TUI checks`; prettier is enforced nowhere in CI, which is how twenty-six
files drifted. #394 paid off twenty-six of them and left four, which are this
branch's; with those formatted here, `prettier --check` passes over the whole of
`ui-tui/src` for the first time.

Three further jobs were red for the base's own reasons, not this branch's:
`page checks`, `unit (py3.12 / ubuntu-latest)` and `Windows self-upgrade` failed
identically on `origin/main` itself at the then merge base `5b7566b49`, in its
own CI run `34744930015`. All three have since been fixed on `main`, and this
branch is rebased onto `56c79edf5`, where every check is green.

## Risk

User-visible, and the point of the change: a reader on `zh` now sees Chinese on
the sub-agent hub, the skills hub, the session picker, the instance picker, the
boot line and the agents-overlay transcript, where they previously saw a mix.

English changes only where a concept had two spellings, plus the pager, casing
and separator normalizations from the earlier commits. No behaviour, no RPC, no
schema, no dependency changes. Every edit is copy or a copy lookup, except the
`DisplayLine` header tag described above, which is internal to `subagentsHub.tsx`
and has no consumer outside it.

Rollback is reverting the branch; nothing here is stateful or migrated.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the complete change from the PR's actual base (5b7566b49) through this head, then followed the changed catalogue lookups through the TUI locale implementation, the web shell bridge, their callers, and the shaping tests. I also read the commit history to distinguish deliberate scope decisions from omissions.

Coverage included AGENTS.md/CLAUDE.md, CONTEXT-MAP.md and the TUI domain vocabulary; the full diff; callers and history; backward compatibility of the catalogue fallback/interpolation paths; test changes for weakening; and the TUI/web architecture boundary. The test edits only update copy expectations and do not relax behavior. No RPC, schema, dependency, or persistence behavior changes here.

Verification on this head:

- `npm run lint:i18n` in `ui-tui`: passed; generated catalogue is current.
- `npm run type-check` in both `ui-tui` and `ui-web`: passed.
- `npm test -- --no-file-parallelism` in `ui-tui`: 143 files passed; 2,059 tests passed and 13 skipped.
- `npx vitest run --no-file-parallelism` in `ui-web`: 103 files and 1,678 tests passed.
- Source-language and large-file repository gates: passed.
- `git diff --check`: passed.

I also inspected the currently red CI jobs. They report automatic import-order/EOF formatting, a page-checker mismatch already present at the base, and an unrelated Windows upgrade failure in untouched code; none yielded a manual review finding under this review's stated scope.

gloryfromca (commented):

Delta review from `0f7700b4ae05`: the new commit only sorts the catalogue imports and applies the repository formatter to lines widened by the existing lookups. The one token-level change, dropping optional parentheses from a single-argument arrow function, is behaviorally identical. No tests or assertions changed in this revision.

For this revision I checked the applicable AGENTS.md/CLAUDE.md and context vocabulary, the delta and cumulative diff, affected callers and commit history, backward compatibility, test integrity, and the existing TUI/web architecture boundary. None is altered by the formatting commit.

Verification on `fcfb7d26e595`:

- `npm run lint`: passed with 0 errors (23 pre-existing warnings).
- `npm run lint:i18n`: passed.
- Prettier check over all 11 files in the new commit: passed.
- `npm test -- --no-file-parallelism`: 143 files passed; 2,059 tests passed and 13 skipped.
- GitHub's TUI check now passes on this head.

gloryfromca (commented):

Delta review from `fcfb7d26e595`: this revision removes one extra trailing newline from `ui-web/src/live/110-connections.js`. The file still ends with the required single newline, and no executable token, copy, test, caller contract, or compatibility behavior changes.

For this revision I checked the applicable repository rules and context, the delta and cumulative diff, the file's callers and commit history, backward compatibility, test integrity, and the frontend architecture boundary. There are no new assertions to weaken and no architecture impact.

Verification on `525adc7dcd78`:

- Repository `end-of-file-fixer` hook on the changed file: passed.
- `npm run type-check` in `ui-web`: passed.
- `npx vitest run --no-file-parallelism` in `ui-web`: 103 files and 1,678 tests passed.
- `git diff --check`: passed.
- GitHub's pre-commit diff and TUI checks pass on this head.

gloryfromca (commented):

This revision is a patch-identical rebase onto `c87622da4`: `git range-diff` maps all 14 old commits one-for-one to the rewritten commits, so it introduces no new PR delta. I re-read the cumulative `github/main...HEAD` diff and its test changes, checked the callers/catalogue generation and compatibility with the new base, and checked history for the rebase equivalence. I also checked the applicable repository rules and domain context; those files are unchanged, and the revision adds no architecture or asset concerns. The tests still assert the same behaviors while updating localized output, rather than weakening coverage.

Local verification: `ui-tui` catalogue check and type-check passed; its full suite passed (2,059 passed, 13 skipped). `ui-web` type-check passed; its full suite passed (1,678 passed). `git diff --check github/main...HEAD` passed. GitHub's frontend, page, RPC-contract, lint, repository, commit-policy, bridge, upgrade, and wheel checks are green; the Python unit job was still running when reviewed.

ui-tui/src/components/skillsHub.tsx
0xKT, line 1: The new `gui.panel.more_down` key is wired at one of its five call sites, so a
scrolled list renders its two paging hints in two different languages.

This PR adds both halves of the pair to `i18n/messages.json` -- neither key
exists at the base commit c87622da:

    gui.panel.more_up    en "^ {n} more"   zh "^ <chinese>"
    gui.panel.more_down  en "v {n} more"   zh "v <chinese>"

`more_up` is called at all five sites. `more_down` is called at one:

    ui-tui/src/components/subagentsHub.tsx:1034   up   uiText('gui.panel.more_up', ...)
    ui-tui/src/components/subagentsHub.tsx:1055   down uiText('gui.panel.more_down', ...)   <- the pair, done

The other four keep the literal they had before:

    ui-tui/src/components/newInstancePicker.tsx:151 up   catalogue
    ui-tui/src/components/newInstancePicker.tsx:191 down literal "<arrow> {n} more"
    ui-tui/src/components/sessionPicker.tsx:206     up   catalogue
    ui-tui/src/components/sessionPicker.tsx:243     down literal
    ui-tui/src/components/skillsHub.tsx:220         up   catalogue
    ui-tui/src/components/skillsHub.tsx:239         down literal
    ui-tui/src/components/skillsHub.tsx:258         up   catalogue
    ui-tui/src/components/skillsHub.tsx:278         down literal

Effect, zh locale, any picker holding more rows than one page: the hint above
the rows is Chinese and the hint below them is English, in the same frame.

Why this is attributable to the PR rather than pre-existing: before it, all ten
lines were the same hardcoded English, so the frame was at least internally
consistent. Localizing one half of each pair is what splits them.

subagentsHub.tsx:1034/1055 is the control -- same PR, same key pair, same
component shape, wired both ways -- so the intended shape is not in question,
and the fix is to call `gui.panel.more_down` at the four sites above.

Not blocking: no behaviour, data or API surface changes, and the English locale
is unaffected. Filing it so the dead-on-arrival half of the key does not outlive
the PR that introduced it.

Verified on head 6e2b8a2f in a clean worktree: `npm ci` from each tree's own
lockfile, then ui-tui `npm run lint:i18n` (up to date), `npm run type-check`
(clean), `npm test` (143 files, 2072 tests, all pass), and ui-web
`npm run type-check` (clean), `npm test` (103 files, 1678 tests, all pass).

gloryfromca: No blockers; suggestions only, and they are marked inline.

Confirmed as written: the four downward paging hints remain hardcoded English and are reached whenever the corresponding list exceeds one page, so the Chinese locale can show mixed-language paging. The finding explicitly marks this as nonblocking, and I carry it as the one outstanding suggestion; it changes my stance from clean to nonblocking without holding the merge. Focused verification passed: catalogue check and type-check were clean, and the related component tests passed (106/106).

gloryfromca (commented):

The newly reported paging-copy mismatch is confirmed: four TUI views use the localized upward hint but retain the English downward literal. It is explicitly nonblocking and has no behavioral, data, or API impact, so this revision remains mergeable with that named suggestion outstanding. I replied on the existing thread and left it open because it belongs to another reviewer.

Focused verification: TUI catalogue generation and type-check passed; `newInstancePicker` and `subagentsHub` passed 106/106 tests.

ui-tui/src/components/skillsHub.tsx
LivXue: Fixed in aef53df3c. Confirmed at the head before fixing: five `uiText('gui.panel.more_up')` calls, one `uiText('gui.panel.more_down')`, four hardcoded literals, exactly the sites you list.

The four now call it. Rather than four more assertions, the guard derives the set from the sources: every component under `ui-tui/src/components` calling `more_up` must call `more_down` as often, and none may hardcode a paging hint. A list holds today's pickers and lets the next one repeat the omission, which is the shape this defect already took once.

Both halves are load-bearing, checked separately:

  revert sessionPicker to the literal    both assertions red
  delete its down hint, no literal       only the pairing assertion red

Verified at the new head: `prettier --check` at the lockfile's 3.8.3 clean (not the version `npx` fetches), `lint:i18n` up to date, `type-check` clean, ui-tui 144 files with 9 failed / 2052 passed / 13 skipped. Those nine are the paint, intensity and typing-burst tests; run the same way at `origin/main` they fail with a byte-identical id set, so 0 introduced. They do not reproduce on your machine or on chandler's, so they are this box's rather than anything to carry.

The branch is rebased onto 56c79edf5 and the shas above are post-rebase. Worth noting alongside #394: the four files it left are these, so `prettier --check` now passes over the whole of `ui-tui/src`.

gloryfromca (commented):

The new tip fixes the outstanding paging-copy suggestion at all four omitted call sites, and the added guard detects both an unmatched catalogue pair and a reintroduced hardcoded paging hint. The other 14 PR commits are patch-identical after the rebase. I covered the cumulative diff, affected callers, commit history/rebase equivalence, compatibility with the new base, test strength, and the applicable repository/domain/architecture rules; the rules and context files are unchanged, no tests were weakened, and I found no new defect.

Local verification: TUI lint completed with 0 errors (23 existing warnings), RPC and i18n generation checks passed, type-check passed, and the full suite passed with 2,061 tests passed and 13 skipped across 144 files. `git diff --check github/main...HEAD` passed. All GitHub checks are green, including the TUI job, four Python shards, and coverage gates. The existing review thread belongs to another reviewer, so I left its resolution to them.

ui-tui/src/components/skillsHub.tsx
0xKT: Verified at aef53df3, and the guard is better than the fix I asked for.

I re-ran the two checks you name rather than taking them:

  sessionPicker's down hint back to the literal   -> both assertions red
  its down hint deleted, no literal               -> only the pairing one red

So both halves carry weight; neither is along for the ride. The derived set
is the right call for the reason you give: a list would have held today's
four and let the fifth picker repeat the omission, which is how this arrived.
The `expect(scrolling.length).toBeGreaterThan(0)` line is what stops the
whole thing passing vacuously if the components move, and I checked it is
there.

What else I ran at that head: `lint:i18n` up to date, `type-check` clean, and
the full ui-tui suite 144 files / 2074 tests, all passing. That is +1 file and
+2 tests against the 143 / 2072 I measured at 6e2b8a2f, which accounts for
exactly the new guard and nothing else. Your nine reds do not appear here,
which matches your reading of them as that box's.

I also checked the rebase did not carry anything else in: `git range-diff
6e2b8a2f...aef53df3` marks all fourteen of the commits I reviewed as
unchanged, with three upstream commits and this fix as the only additions. So
the rest of my pass still speaks for this head.

Closing this one.

0xKT (approved):

PR #394, opened 2026-09-14
chore(ui-tui): format the files the pre-commit prettier hook has been failing on

LivXue:
## Summary

Thirty files under `ui-tui/src` do not satisfy the repo's own prettier config
on `main`. This formats twenty-six of them.

Nobody noticed because the gate that should have caught it does not run. The
hook is declared like this:

```yaml
- id: prettier-ui-tui
  entry: bash -c '[ -d ui-tui/node_modules ] || exit 0; cd ui-tui && npx prettier --check "src/**/*.{ts,tsx}"'
  files: '^ui-tui/src/.*\.(ts|tsx)$'
```

Two things follow. The entry ignores the paths pre-commit hands it and checks
the whole glob, so on a developer's machine it fails over files the commit
never touched. And in CI it never gets that far: the `pre-commit diff` job's
only install step is `make install-deps`, which is `uv sync` with no `npm ci`
for any workspace, so `ui-tui/node_modules` does not exist and the guard exits
0. On #392 the hook reported Passed in about a third of a second while
twenty-six files were failing `prettier --check`. The sibling
`eslint (ui-tui)` hook is inert for the same reason, though eslint is covered
anyway by the `TUI checks` job. Prettier is covered by nothing, which is how
thirty files drifted.

So this is paying off real drift behind a gate that has been reporting green,
not chasing a red CI job. Expect no CI status to change here.

The remaining four -- `core.ts`, `session.ts`, `agentsOverlay.tsx` and
`modelPicker.tsx` -- are formatted by #392, which was already editing them for
copy reasons. Splitting the set this way means the two branches share no file
and can land in either order without a conflict.

The hook itself is left alone. Making the entry honour its file list, or
giving the CI job the `npm ci` that would let it run at all, are both changes
to a shared gate and a separate argument from paying off what it has been
quietly accumulating.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other

Formatting only. `chore` rather than `refactor`: nothing is restructured.

## Verification

Reproducible rather than hand-made. Against this base, running

```
npx prettier --write <the 26 paths>
```

yields this tree byte for byte, and running it a second time changes nothing
further -- the diff is idempotent under the tool that produced it.

```
npx prettier --check "src/**/*.{ts,tsx}"   4 files left, and they are #392's
npx tsc --noEmit -p tsconfig.json          exit 0
npm run lint                               exit 0 (23 warnings, same as main)
```

Full `ui-tui` suite, `npx vitest run --no-file-parallelism`, against a
detached worktree at the base `5b7566b49` run the same way:

```
base     11 failed | 2061 passed | 13 skipped (2072)   0 unhandled
branch    9 failed | 2050 passed | 13 skipped (2072)   0 unhandled
```

Failing ids diffed rather than counted: **0 introduced**. The nine are the same
nine the base fails; the two the base fails and this branch does not are the
`ModelPicker` login tests, which are timing flakes rather than a result.

Independently re-run by the reviewer on a different machine at
`4ce89884f25c`: 143 files, 2,059 passed and 13 skipped, all green. That is this
branch's count plus exactly the nine it fails here, so those nine are an
artifact of the machine this branch was prepared on, not a repository state a
reviewer needs to carry.

Every hunk is prettier re-wrapping a line or normalizing a quote to avoid an
escape -- `'the agent\'s arm'` becomes `"the agent's arm"`, the same string. No
identifier, no literal value and no test name is altered, which is what the
type-check and the unchanged failing-id set are evidence for.

## Risk

None to behaviour: no runtime file changes meaning, and the suite is unchanged
against the base. The risk is to other open branches -- anything with pending
edits inside these twenty-six files will see conflicts on rebase, and the fix
there is to rerun prettier on the conflicted file rather than to resolve hunk
by hunk.

The three jobs red here -- `page checks`, `unit (py3.12 / ubuntu-latest)` and
`Windows self-upgrade` -- are red on `origin/main` itself, run `34744930015`
at this branch's base `5b7566b49`. The failing job sets are identical.

Rollback is reverting the single commit.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the complete `github/main...HEAD` diff and confirmed every changed file is byte-for-byte the configured Prettier output from the base revision. No values, control flow, assertions, public interfaces, caller contracts, or RPC boundaries change, so backward compatibility and the TUI architecture are unaffected; the tests were reformatted without being weakened.

I also checked the applicable repository guidance in AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and ui-tui/CONTEXT.md, plus the commit history and PR record. The four untouched files still reported by the full-tree Prettier check are the explicitly documented, non-overlapping portion already carried by #392, not a regression introduced here.

Local verification: `npm run lint` exited 0 with 23 existing warnings and no errors; `npm run type-check` exited 0; `npm test -- --run` passed 143 files and 2,059 tests with 13 skipped. GitHub's TUI checks also passed.

gloryfromca (commented):

This revision is a clean rebase onto the newer `main`: `git range-diff` shows the reviewed formatting commit is unchanged, and the full `github/main...HEAD` diff remains the same 26-file patch. I rechecked the applicable AGENTS.md/CLAUDE.md and TUI context constraints, the diff and history, callers, backward compatibility, architecture boundaries, and test assertions; there is still no semantic change or weakened test.

Post-rebase verification: all 26 changed files are byte-identical to configured Prettier output from the new base; `npm run lint` exited 0 with 23 warnings and no errors; `npm run type-check` exited 0; and `npm test -- --run` passed 143 files and 2,059 tests with 13 skipped. The full-tree Prettier check continues to name only the same four untouched files assigned to #392.

0xKT (approved):

PR #398, opened 2026-09-14
feat(*): give the default models a page of their own

KevinHuSh:
## Summary

Moves what a conversation defaults to onto a page of its own, under Agent, and adds three model choices beside it. The providers page keeps the accounts and their model lists, and is renamed Model providers to say what it is now.

Five rows on the new page:

- Agent model -- the default binding, moved here with how it answers.
- Quick model -- the short errands beside a conversation, today the call that names a new session.
- Translate model -- reserved. Nothing calls it yet, and the row says so.
- Painting model -- the image tool's model and quality, moved out of the tool row it was folded inside.
- Embedding model -- what knowledge bases are embedded and searched with. New.

Each of the three new rows writes a model and the provider serving it, because an id alone does not name a credential: `openrouter` + `anthropic/claude-haiku-4-5` and `anthropic` + `claude-haiku-4-5` are both valid and bill different accounts. That is the shape every subsystem pin in the config already states.

A defect fixed on the way. `sessionTitle.model` had no provider half, so a pinned naming model was sent on the conversation's provider and key -- exactly the mis-pairing the pin convention exists to prevent. It now binds the pair through the provider pool and follows the conversation's model when the pair cannot be built, reported once. A config that sets only a model keeps working, since the pool derives the vendor from the id.

The embedding pin is the first time raven decides where a knowledge base embeds. Set, it wins; unset, the EverOS endpoint answers exactly as before, so no existing install changes behaviour. Provider credentials resolve through a new public helper in `update_providers`, which is already the one place that knows what address a request goes out on, rather than a second answer to that question.

Two decisions worth a reviewer's attention:

- Painting is not a pin. The image tool carries its own key and address instead of borrowing a provider account, so giving it a provider half would mean re-plumbing the media stack. The existing model/quality picker moved as it is, and the row says how it is reached.
- Each slot's offer is filtered by what the slot is for, using the same bucketing the backend does (`registry_data.kind_of`). A picker that lists every model for every slot is how an embedding id lands in a painting slot; for the embedding slot the cost of that is a vector store that cannot be searched. Providers that are not connected are left out, since picking from one fails on the next call rather than at the click.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

`make coverage`: full suite green.

`npm test --prefix ui-web` -> 103 files, 1681 tests. `npm test` in ui-tui -> 143 files, 2072 tests. Type-checks, `lint:rpc`, `lint:i18n` and ui-web's `gen:check` all clean, and `ruff check` / `ruff format --check` pass.

New tests cover the embedding pin (a complete pair wins, half a pair does not, an uncredentialed provider falls back, no pin behaves as before), the quick pin (both halves reach the pool, a refused pin follows the conversation), and the slot filtering. Each was confirmed to fail with its fix removed rather than assumed to cover.

`check-page` fails on this branch. It is the two-versus-three script block mismatch already failing on main, fixed in a separate PR; repeating the fix here would collide when both land.

## Risk

User-visible: the settings panel gains a Default models entry under Agent, Models is renamed Model providers, and the agent model, how-it-answers and painting controls are reached from the new page rather than where they were.

Behaviour: an install that sets none of the new pins runs exactly as before. The one behaviour change is the session-title fix above, which drops a pin that cannot be paired instead of sending it on the conversation's credential.

`settings.set` gains six writable keys. The whitelist is pinned wholesale by a test so this had to be a deliberate edit: the pins name a model and a provider the deployment has already credentialed, carry no address or key of their own, and reach no endpoint the operator did not already choose.

Rollback: revert the squash commit. No migration, and no config is rewritten.

## Related Issues

N/A


raven/config/raven.py
```diff
@@ -1486,6 +1534,8 @@ class RavenConfig(_Base):
     runtime: RuntimeConfig = Field(default_factory=RuntimeConfig)
     tracing: TracingConfig = Field(default_factory=TracingConfig)
     session_title: SessionTitleConfig = Field(default_factory=SessionTitleConfig)
+    translate: TranslateConfig = Field(default_factory=TranslateConfig)
```
gloryfromca, line 1537: Blocking: saving either new extension block makes the whole config unloadable.

`load_raven_config()` invokes the base loader before extracting extension blocks, but `translate` and `knowledge` were not added to `loader.EXTENSION_KEYS`. After `settings.set` writes either top-level block, the base `Config` rejects it as an extra; I reproduced both as `extra_forbidden` in a temporary config. Consequently the next ordinary `load_config()` / `load_raven_config()` fails, rather than merely ignoring the new pin. Please register these blocks in the loader contract and cover a file round trip.

ui-web/src/features/settings/SettingsPage.tsx
```diff
@@ -1835,14 +1839,123 @@ function ModelTuning({ s }: { s: SettingsState }): JSX.Element {
...
+  providerKey: string
+  foot?: string
+  s: SettingsState
+}): JSX.Element {
+  const [nl, say] = useNl()
+  const model = String(V(s.snap.raw, modelKey, '') || '')
+  const provider = String(V(s.snap.raw, providerKey, '') || '')
+  const groups = s.snap.providers
+    .filter((p) => p.on)
+    .map((p) => ({ p, models: (p.configured ?? p.models).filter((m) => kindOf(p.labels?.[m]) === kind) }))
```
gloryfromca: Blocking: ordinary connected providers lose all default-model choices here.

The live bridge always supplies `configured: p.configured_models || []`, so this nullish fallback selects the empty configured array and never reaches `p.models`. Normal onboarding writes the provider credential and default model but does not populate `providers.<slug>.models`; for that common state, the new page says `Add a provider first` and offers no quick/translation choice even though `model.options` contains the curated models. The new test avoids this production shape by omitting `configured`.

ui-web/src/features/settings/store.ts
```diff
@@ -430,3 +430,18 @@ export function reset(): void {
...
+   and writing only one leaves a state the backend reads as "no pin" -- so a
+   failed second write rolls nothing back but is reported, rather than leaving
+   the page claiming a pin the config does not hold. */
+export async function writePin(
+  modelKey: string,
+  providerKey: string,
+  model: string,
+  provider: string,
+): Promise<WriteOutcome> {
+  const first = await write(modelKey, model)
```
gloryfromca: Blocking: the model/provider pair can be persisted in a mismatched state.

When switching from provider A to B, the first RPC commits B's model and reloads it; if the second RPC is refused or the connection drops, A's old provider remains. The runtime then sees a complete pin and can send B's model through A's credential, exactly the half-pair the binding architecture is meant to prevent. Reporting the second failure does not undo the first write, so this pair needs an atomic persistence path (or equivalent restoration on failure).

gloryfromca (commented):
Blocking: register the new config blocks, expose usable model choices, and persist model/provider pins atomically.

I reviewed the full `github/main...HEAD` diff, AGENTS.md and CONTEXT-MAP.md, the config-loader and provider-binding callers, the live settings bridge, relevant history, backward compatibility, and the changed tests. I found no weakened tests, but the new UI test fixture does not reproduce the live `configured` field described inline.

Verification: `uv run pytest tests/test_knowledge_embedding.py tests/test_rpc_console.py tests/test_rpc_session_naming.py -q` (151 passed); `uv run pytest tests/test_config_raven_loader.py tests/test_config_raven_sections.py tests/test_config_loader.py -q` (121 passed); `uv run pytest tests/test_config_update_providers.py tests/test_rpc_model.py -q` (250 passed); `npm test --prefix ui-web -- --run src/features/settings/SettingsPage.test.tsx` (104 passed); `npm run type-check --prefix ui-web` (passed); source-language, large-file, and `git diff --check` checks passed. A temporary-config reproduction still fails with `extra_forbidden` for `translate` and `knowledge`.

raven/config/raven.py
gloryfromca: Blocking: model/provider pin persistence still has to change before merge.

This specific issue is fixed: both extension blocks are now in `EXTENSION_KEYS`, and the new file-level and RPC round-trip tests cover the failure I reproduced. Closing this thread.

ui-web/src/features/settings/SettingsPage.tsx
gloryfromca: Blocking: model/provider pin persistence still has to change before merge.

This specific issue is fixed: an empty live `configured` array now falls back to the provider offer, and the updated test uses the production-shaped `configured: []` fixture. Closing this thread.

ui-web/src/features/settings/store.ts
gloryfromca: Blocking: persist each model/provider pin as one indivisible operation.

The compensating write does not settle this thread. If the provider write fails because the WebSocket connection dropped, the rollback uses that same unavailable connection and its result is ignored, leaving the mismatched pair in place. There is also an all-success path: two rapid changes can interleave their independent writes as model B, model C, provider C, provider B, ending with model C paired to provider B. The new test permits the third call after selectively failing only the provider key, so it covers neither failure. The persistence boundary still needs to guarantee the pair rather than attempt a best-effort repair.

gloryfromca (commented):
Blocking: persist each model/provider pin as one indivisible operation.

The loader-registration and provider-offer blockers from the prior revision are fixed and their threads are resolved. The attempted model rollback remains best-effort: it cannot recover through the same dropped connection, and concurrent picker changes can interleave successful per-key writes into a mismatched pair. I left that thread open with the concrete sequences.

I reviewed the delta from `15bbc0301282`, the full `github/main...HEAD` diff where affected, AGENTS.md and CONTEXT-MAP.md, relevant loader/RPC/provider/UI callers and history, backward compatibility, and test strength. I found no weakened tests and no additional issue worth raising in the documentation, knowledge-error, layout, or provider-offer changes.

Verification: `uv run pytest tests/test_knowledge_embedding.py tests/test_rpc_console.py tests/test_rpc_settings.py tests/test_rpc_session_naming.py tests/test_config_raven_loader.py tests/test_config_raven_sections.py tests/test_config_loader.py tests/test_config_update_providers.py tests/test_rpc_model.py -q` (547 passed); `npm test --prefix ui-web -- --run src/features/settings/SettingsPage.test.tsx src/features/knowledge/KnowledgePage.test.tsx` (139 passed); `npm run type-check --prefix ui-web` (passed); source-language, large-file, and `git diff --check` checks passed.

ui-web/src/features/settings/store.ts
gloryfromca: Blocking: preserve valid snake_case config when writing model-pin blocks.

The issue in this thread is fixed: the settings UI now sends one complete pair, validation happens before the write, and one locked `atomic_update` persists it, so neither connection failure nor concurrent saves can split the pair. Closing this thread; the separate alias-merge regression is raised on the new backend path.

raven/rpc/methods/console.py
```diff
@@ -586,6 +586,13 @@ async def settings_set(params: dict, *, agent_loop_factory=None) -> dict:
     raise ConfigValidationError(f"key not writable via settings.set: {key}")
 
 
+#: Keys whose value is an object covering only part of the block it names, so
+#: the write merges instead of replacing. ``sessionTitle`` carries enabled, the
+#: timeout and the width gate beside its pin, and a replacing write would drop
+#: every one of them.
+_MERGED_KEYS = frozenset({"tools.media.image", "sessionTitle", "translate", "knowledge"})
```
gloryfromca, line 594: Blocking: block writes must update existing snake_case aliases instead of creating duplicate camelCase fields.

The Raven config models explicitly accept both spellings, but this literal merge does not normalize the existing block or leaves. With a valid `session_title` block, writing `sessionTitle` creates both root aliases and `load_raven_config()` rejects `session_title` as an extra. Likewise, a valid `knowledge` block containing `embedding_model` / `embedding_provider` is merged with `embeddingModel` / `embeddingProvider`, and both old leaves become `extra_forbidden`. I reproduced both through `settings_set`; the next config load fails. Please preserve the supported snake_case form when applying the atomic pair, with regression coverage for these existing configs.

gloryfromca (commented):
Blocking: preserve valid snake_case extension config when writing model-pin blocks.

The previous atomicity blocker is fixed and its thread is resolved: each picker now sends a complete pair that is validated and persisted in one locked update. The new block-level path has one backward-compatibility regression, marked inline.

I reviewed the delta from `afde872ea870`, the affected portions of `github/main...HEAD`, AGENTS.md and CONTEXT-MAP.md, the settings writer and Raven loader callers, relevant history, backward compatibility, and test strength. No tests were weakened, and I found no other issue worth raising in the atomic-write/UI changes.

Verification: `uv run pytest tests/test_knowledge_embedding.py tests/test_rpc_console.py tests/test_rpc_settings.py tests/test_rpc_session_naming.py tests/test_config_raven_loader.py tests/test_config_raven_sections.py tests/test_config_loader.py tests/test_config_update_providers.py tests/test_rpc_model.py -q` (552 passed); `npm test --prefix ui-web -- --run src/features/settings/SettingsPage.test.tsx src/features/knowledge/KnowledgePage.test.tsx` (139 passed); `npm run type-check --prefix ui-web` (passed); source-language, large-file, and `git diff --check` checks passed. I separately reproduced the alias regression with temporary `session_title` and snake_case knowledge configs.

raven/rpc/methods/console.py
gloryfromca: No blockers; this can merge as far as I am concerned.

This is fixed. `_as_written` now follows the existing spelling at both the root block and nested field levels, so valid snake_case configs are updated rather than duplicated. The new tests cover both reproductions, the single-leaf path, existing camelCase, and a newly created block; the focused suite passes. Closing this thread.

gloryfromca (commented):

The remaining snake_case compatibility blocker is fixed: block and leaf writes follow the spelling already present, while new blocks use the canonical camelCase form. I closed the final thread.

I reviewed the delta from `73ea13f8815b`, the affected `github/main...HEAD` writer and loader paths, AGENTS.md and CONTEXT-MAP.md, relevant callers and history, backward compatibility, the atomic-persistence architecture, and test strength. No tests were weakened, and I found no new issue meeting the round-four bar.

Verification: `uv run pytest tests/test_knowledge_embedding.py tests/test_rpc_console.py tests/test_rpc_settings.py tests/test_rpc_session_naming.py tests/test_config_raven_loader.py tests/test_config_raven_sections.py tests/test_config_loader.py tests/test_config_update_providers.py tests/test_rpc_model.py -q` (557 passed); `npm test --prefix ui-web -- --run src/features/settings/SettingsPage.test.tsx src/features/knowledge/KnowledgePage.test.tsx` (139 passed); `npm run type-check --prefix ui-web` passed; source-language, large-file, and `git diff --check` checks passed.

gloryfromca (commented):

The upgrade-marker import now respects the kernel boundary, and the page gate correctly validates the existing three-script artifact. The prior default-model review remains clean.

Named follow-up (nonblocking): the doctor test's `runner.invoke(..., env={"COLUMNS": "300"})` runs after the module-level Rich console has captured its width, so the intended stabilization is ineffective. I reproduced the changed pair against the repository plugin source: one passed and `test_doctor_answers_where_the_memories_are` still failed because `mem-root` wrapped; setting `COLUMNS=300` before pytest/import made that same case pass. This is a pre-existing checkout-depth test flake with an environment-level escape, not a newly introduced runtime defect, so it does not meet the round-five blocking bar. The assertions themselves were not weakened.

I reviewed the three-commit delta against the already-reviewed `github/main...HEAD` change, plus the affected callers and history, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime kernel/config boundary. Verification: `uv run pytest tests/test_updates_install_guard.py -q` (24 passed); `npm run --prefix ui-web build`, `uv run python ui-web/build.py`, and `node ui-web/scripts/check-page.mjs` passed; source-language, large-file, and `git diff --check` checks passed. The initial combined Python run reported those 24 passes plus 97 setup errors because the optional `raven_everos` distribution is not installed in this environment; rerunning the changed doctor cases through `plugins-dist/everos-memory` produced the one real wrapping failure described above.

gloryfromca (commented):

The new delta safely contains provider-section validation failures and replaces resolver stand-ins with real-path coverage. I found no regression in those changes, and no tests were weakened.

Named follow-up (nonblocking): an ordinary API-key-only OpenAI setup is offered in the embedding-model picker, but `resolve_provider_credentials("openai")` now has an explicit test requiring `None` until the user also supplies `api_base`. A complete OpenAI embedding pin therefore falls back to the EverOS endpoint instead of applying the selected model. An explicit API base or the working EverOS fallback provides a way out, so this does not meet the late-round blocking bar. The previously recorded doctor-width test follow-up also remains nonblocking and unchanged.

I reviewed `69e3b76c8c21...ef89a0cb2302` against the full `github/main...HEAD` context, the credential resolver and embedding callers, provider registry semantics and history, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime model-binding boundary. Verification: `uv run pytest tests/test_config_update_providers.py tests/test_knowledge_embedding.py tests/test_rpc_settings.py -q` (190 passed); `npm test --prefix ui-web -- --run src/features/settings/SettingsPage.test.tsx` (106 passed); source-language, large-file, and `git diff --check` checks passed.

gloryfromca (commented):

The rebase onto the updated `main` is sound. The upgrade-marker and page-gate fixes are now supplied by the base, and the fixed-width Rich console patch resolves the previously reproduced doctor-test failure. No new findings. The previously named API-key-only OpenAI embedding-pin limitation remains the sole nonblocking follow-up.

I reviewed `ef89a0cb2302...6e35d5bb1a28` and the resulting `github/main...HEAD` diff, affected callers and history, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime model-binding and kernel/config boundaries. No tests were weakened. Verification: the two doctor reproductions passed; `uv run pytest tests/test_config_update_providers.py tests/test_knowledge_embedding.py tests/test_rpc_settings.py tests/test_updates_install_guard.py -q` passed (214 tests); the settings/knowledge UI suites passed (139 tests); TypeScript type-check, source-language, large-file, and `git diff --check` checks passed.

gloryfromca (commented):

The merge of the latest `main` is clean and introduced no conflict-resolution changes in the feature paths. No new findings. The previously named API-key-only OpenAI embedding-pin limitation remains the sole nonblocking follow-up.

I reviewed `6e35d5bb1a28...bc433f6499a7` and the resulting `github/main...HEAD` diff, merge history, affected callers, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime model-binding and kernel/config boundaries. No tests were weakened. Verification: the combined resolver, knowledge, settings, upgrade-marker, and MCP synchronization suites passed (266 tests); the settings/knowledge UI suites passed (139 tests); TypeScript type-check, source-language, large-file, and `git diff --check` checks passed.

gloryfromca (commented):

The merge of the latest `main` is clean: the PR's production diff is unchanged, there are no combined conflict-resolution hunks, and no new findings emerged. The previously named API-key-only OpenAI embedding-pin limitation remains the sole nonblocking follow-up.

I reviewed `bc433f6499a7...6eda4fa3e021` and the resulting `github/main...HEAD` diff, merge history, affected callers, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime model-binding boundary. No PR tests were weakened. Verification: the resolver, knowledge, settings, and overlapping upgrade-marker suites passed (215 tests), with four non-failing wall-clock warnings from the new base fixture; the settings/knowledge UI suites passed (139 tests); TypeScript type-check, source-language, large-file, and `git diff --check` checks passed.

gloryfromca (commented):

The latest merge is sound. The provider-catalog expansion composes with the default-model filtering, and the conflict resolution drops only knowledge-store code for which upstream now supplies equivalent error-detail handling. A dry-run merge with the current `github/main` is also clean. I found no new issues. The previously named API-key-only OpenAI embedding-pin limitation remains the sole nonblocking follow-up.

I reviewed `6eda4fa3e021...7b43f7dc77b9` and the resulting `github/main...HEAD` diff, merge resolution and history, affected callers, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md and the Runtime model-binding boundary. No PR tests were weakened. Verification: the settings, credential resolver, knowledge, RPC, and expanded-provider backend suites passed (1,282 passed, 103 capability-dependent skips); settings/knowledge UI suites passed (155 tests); TypeScript type-check, source-language, large-file, and `git diff --check` checks passed. The two repository gates were run through their canonical scripts because `make` is unavailable in this environment.

gloryfromca (commented):

This revision cleanly merges the latest `main`; Git records no conflict-resolution hunks, and every feature file is unchanged from the last reviewed head. The upstream playbook and design-session changes do not overlap the default-model implementation. No new findings. The previously named API-key-only OpenAI embedding-pin limitation remains the sole nonblocking follow-up.

I reviewed `7b43f7dc77b9...23a192bbbec3` and the resulting `github/main...HEAD` diff, merge history, affected callers, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime model-binding boundary. No PR tests were weakened. Verification: the focused config, credential-resolver, knowledge, RPC, and upgrade suites passed (356 tests; four non-failing wall-clock warnings); settings/knowledge UI suites passed (155 tests); TypeScript type-check, source-language, large-file, and `git diff --check` checks passed.

gloryfromca (commented):

The test-only delta removes the live-model cache race without weakening coverage: the cache-hit case now owns a long TTL, while the added controlled-clock case separately proves both the pre-expiry hit and post-expiry refresh. Production code is unchanged, and a dry-run merge with current `github/main` is clean. No new findings. The previously named API-key-only OpenAI embedding-pin limitation remains the sole nonblocking follow-up.

I reviewed `23a192bbbec3...452c33100bfa` and the resulting `github/main...HEAD` diff, the cache caller and history, latest-main idle-gate integration, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime model-binding boundary. Verification: the complete RPC-model suite passed (195 tests); source-language, large-file, and `git diff --check` checks passed. The branch still carries the older warning-only wall-clock reporter, which listed timings but did not fail the run; latest `main` replaces that metric with the idle-time gate this test targets.

tests/test_rpc_model.py
```diff
@@ -822,6 +828,56 @@ def probe(name: str, *, timeout_s: int) -> dict:
...
+
+        def monotonic(self) -> float:
+            return self.now
+
+    clock = _Clock()
+    monkeypatch.setattr("raven.config.update_providers.test_provider", probe)
+    monkeypatch.setattr(model_module, "time", clock)
+    model_module._LIVE_MODEL_CACHE.clear()
+
+    await model_options({})
```
gloryfromca: Blocking: keep the new cache-expiry coverage below main's strict idle-time ceiling.

This test calls the full `model_options()` provider-catalogue path three times. Running just this test with the exact CI gate (`uv run pytest tests/test_rpc_model.py::test_the_live_model_cache_is_asked_again_once_it_has_expired -q --idle-ceiling-strict`) reports `1 passed` but fails the session at 5.63 seconds idle versus the 3-second ceiling. `main` applies that strict gate to every shard, so the asserted behavior is green while the PR's CI remains red.

This meets all three late-round blocker conditions: this revision introduces the test, the gate runs in ordinary PR CI, and a shard that fails after its assertions pass has no working merge path without changing the test. Please retain deterministic pre/post-expiry coverage without making three complete catalogue walks.

gloryfromca (commented):
Blocking: make the new cache-expiry test comply with main's strict idle-time gate.

The merge itself is clean and I found no production-code regression, but the test added in the preceding revision still makes the CI configuration inherited here fail. The previously named API-key-only OpenAI embedding-pin limitation remains nonblocking.

I reviewed `452c33100bfa...88b86c7f7d10` and the resulting `github/main...HEAD` diff, merge resolution and history, the cache caller, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime model-binding boundary. No existing assertion was weakened, but the new test is incompatible with the target branch's gate. Verification: the full RPC-model file reached 195 passing assertions but the strict session failed; the isolated new test reproduced the failure at 5.63 seconds idle against the 3-second ceiling. Source-language, large-file, and `git diff --check` checks passed.

tests/test_rpc_model.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

This is fixed. The expiry test now exercises the cache-owning `_provider_models` path directly, retains the before-expiry cache-hit and after-expiry re-probe assertions, and the exact isolated command passes under `--idle-ceiling-strict`. Closing this thread.

gloryfromca (commented):

The cache-test blocker is fixed and its thread is closed: the test now exercises the cache-owning path directly, keeps deterministic assertions on both sides of expiry, and passes the exact isolated strict-gate reproduction. I found no new issues. The previously named API-key-only OpenAI embedding-pin limitation remains the sole nonblocking follow-up.

I reviewed `88b86c7f7d10...4c2c6dafd232` and the resulting `github/main...HEAD` diff, the cache caller and history, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime model-binding boundary. Verification: the isolated changed test passed under `--idle-ceiling-strict`; the full RPC-model file reached 195 passing assertions, although the local strict session remained nonzero because five unchanged target-branch tests measured 3.18-3.33 seconds idle; source-language, large-file, and `git diff --check` checks passed. No test was weakened or skipped.

gloryfromca (commented):

The new autouse fixture removes LiteLLM's hidden Hugging Face context-window request from unit tests while preserving the production fallback result (`None`). The strict RPC-model suite that previously exposed the random waits is now green, and the provider-rate/LiteLLM coverage remains green. I found no new issues. The previously named API-key-only OpenAI embedding-pin limitation remains the sole nonblocking follow-up.

I reviewed `4c2c6dafd232...4ffc1fe6a7b6` and the resulting `github/main...HEAD` diff, the patched LiteLLM caller and signature, affected tests and history, backward compatibility, test strength, AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Runtime model-binding boundary. Verification: `tests/test_rpc_model.py` passed under `--idle-ceiling-strict` (195 tests); provider-rate and LiteLLM setup suites passed under the same gate (91 tests); source-language, large-file, and `git diff --check` checks passed. No test was weakened or skipped.

tests/conftest.py
```diff
@@ -610,6 +610,55 @@ def _no_openrouter_network(tmp_path):
...
+    Answering "not reachable" is what those tests already expect: a row is
+    configured by holding an address, not by a probe that answered. The two
+    files that assert on a live answer patch this themselves, and their patch
+    lands after this one.
+    """
+    from raven.config import update_providers
+
+    monkeypatch.setattr(
+        update_providers,
+        "test_provider",
```
gloryfromca: Blocking: scope the provider-probe stub so deliberate probe tests still exercise the real function.

Because this fixture is autouse, `tests/test_config_provider_passthrough.py::test_testing_an_unregistered_provider_reports_instead_of_crashing` imports `test_provider` only after this replacement is installed. It therefore receives the lambda's two-field result rather than testing the real unregistered-provider path, and the default test run fails at `result["status"]` with `KeyError: 'status'`. I reproduced it with `uv run pytest tests/test_config_provider_passthrough.py::test_testing_an_unregistered_provider_reports_instead_of_crashing -q --idle-ceiling-strict`. Please keep incidental `model.options` probes offline without globally replacing the function for tests whose subject is the probe itself.

gloryfromca (commented):
Blocking: keep the global provider-probe fence from replacing deliberate probe tests.

I reviewed the new revision against the current `main`, the repository rules, the full PR diff, the changed fixture's callers and import timing, relevant history, backward-compatibility behavior, test integrity, and the Runtime/UI boundary. The new autouse fixture causes a default unit test to fail under the strict CI invocation; details are inline. No other new finding survived verification.

Focused verification: `uv run pytest tests/test_config_provider_passthrough.py::test_testing_an_unregistered_provider_reports_instead_of_crashing -q --idle-ceiling-strict` fails with `KeyError: 'status'`. `git diff --check github/main...HEAD` passes.

This meets all three late-round blocking conditions: this commit introduces it, the ordinary default suite reaches it, and the required CI run has no working path past the failure.

tests/conftest.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

This is fixed. The fence now replaces only `_probe_models_endpoint` when no test transport is supplied, so `test_provider` still constructs and returns its real status vocabulary, while explicit mock transports continue through the production request path. The original strict reproduction passes, as do all 127 provider-probe unit tests and all 195 RPC-model tests under the strict idle ceiling. Closing this thread.

gloryfromca (commented):

The provider-probe blocker is fixed and its thread is resolved. I reviewed the delta against current `main`, the repository rules, the full change record, the fixture's callers and import timing, relevant history, backward compatibility, test integrity, and the Runtime/UI boundary. The previously recorded nonblocking follow-up remains unchanged; I found no new merge blocker or suggestion in this revision.

Verification:
- `uv run pytest tests/test_config_provider_passthrough.py::test_testing_an_unregistered_provider_reports_instead_of_crashing -q --idle-ceiling-strict` - 1 passed
- `uv run pytest tests/test_config_update_providers.py -q --idle-ceiling-strict` - 127 passed
- `uv run pytest tests/test_rpc_model.py -q --idle-ceiling-strict` - 195 passed
- source-language and large-file gates for `github/main..HEAD` - passed

0xKT (approved):

PR #399, opened 2026-09-14
fix(ui-web): let the roster start the conversation an instance needs

Handsome-wzw:
## Summary

On the new-task screen the sub-agent roster drew a live `+` on every addressable
agent and did nothing at all when one was pressed. An instance is created inside
a conversation and a draft has no id, so `startInstance` returned false before
its first await -- correctly, and to nobody: no card, no toast, no change on the
row.

The roster is the right place for that entry point, so the button now does the
whole thing the reader is asking for: start the conversation, then put the
instance in it. Pressing it mints the session, files it on the rail, opens the
direct chat, and leaves the reader inside it.

The promotion is the composer's, not a second one. It has always run on the
first message sent and was written inline in `liveSend`; it is now
`openConversation`, which both callers share. What only a send needs -- the
turn's owner, the naming request, the row's preview line -- stayed with the
send, and the turn's owner still lands at the same instant it used to, through a
hook called while the pointer moves rather than after the staged settings have
gone up.

One predicate decides the offer and the action. A page with no conversation and
no way to make one draws no button at all, the way a stateless agent already
draws none -- rather than a button that refuses on click.

Three defects found while writing this, each its own commit:

- **A race the new caller brings.** `draft` stays raised across the
  `session.create` round trip, so a second caller landing inside it mints a
  SECOND conversation for one visit to the new-task screen. The in-flight
  promotion is now held and shared, and released when it ends.
- **A swallowed failure.** `mine()` read the conversation the create was filed
  under, which is still empty when the promotion itself is what failed -- so the
  card that exists to report it stayed blank. It now falls back to the
  conversation the press was made in.
- **Sends bypassing an unfinished promotion** (gloryfromca, review on this PR).
  `promote` moves the pointer and lowers `draft` before awaiting the staged
  model, tier and permission writes and the subscription, so a send landing
  there dispatched at once -- starting the first turn on a half-made
  conversation, under settings the reader chose and did not get. `sendOnSession`
  now joins an in-flight promotion, and `openConversation` asks about that
  promotion BEFORE it reads `draft`: with the flag read first the joiner was
  answered "there is a conversation, carry on" and walked past the same window
  one layer down.

Stale claims this made false, corrected in their own commit:
`080-overrides.js` promised "nothing is written to disk until the first
message"; `palette.ts` and `deskStore.ts` each described the promotion as what
the first message does. `shell/resume.ts` names the first message as its example
of an arrival no opener runs for and is deliberately unchanged -- the new
promotion moves the same pointer and is covered by the same line.

**CI.** Three jobs fail and all three fail identically on the merge base. `main` at
`5b7566b4` fails exactly these and nothing else, in run `34744930015`:

```
Windows self-upgrade            red here, red on main @ 5b7566b4
page checks                     red here, red on main @ 5b7566b4
unit (py3.12 / ubuntu-latest)   red here, red on main @ 5b7566b4
```

One for one, nothing extra on either side. `page checks` was reproduced locally
on an untouched worktree checked out at `5b7566b4`, built the way the workflow
builds it, with the same message: `check-page: expected exactly two <script>
blocks`. Not read off the log -- built and run.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
cd ui-web && npx vitest run            104 files / 1691 tests, all pass
cd ui-tui && npx tsc --noEmit -p .     clean
```

Thirteen new cases, each mutation-checked -- reverted the line it covers and
watched it go red:

- remove the shared hold -> two conversations minted, red
- never release the hold -> the next draft answered with the last id, red
- move the caller hook after the staged settings -> order assertion red
- promote when not a draft -> the already-open case red
- `mine()` back to reading only the created key -> the failure card red
- the old refusal on a draft -> four cases red
- drop the `promoting` guard in `sendOnSession` -> the send lands early, red
- read `draft` before `promoting` again -> the send lands early, red

The review finding was reproduced before it was fixed: the new case put
`turn.send` at log index 8, ahead of `subscribe` at 9.

Driven through the real entry points. The live layer is plain script, so
`openConversation`, `promote`, `sendOnSession` and `dispatchSend` are sliced out
of `080-overrides.js` and run -- the pattern
`features/rail/session-delete.test.ts` already uses on this same file. The
roster half is driven through the button.

Run on a real gateway, not only in tests: pressed `+` from the new-task screen
for Raven-Oncall, Raven-Research and Raven-Design. Each minted a conversation,
registered the instance against it, opened the direct chat, and left the roster
listing the instance under the new conversation when the pane was closed. The
desk palette came back on the same tab it was opened from.

No user-facing doc covers the roster button; the behaviour change is described
in the code comments the sweep required.

## Risk

Security: nothing new reaches the wire. The added seam calls `session.create`,
which the composer already called on every first message, with the same params.

Backward compatibility: `startConversation` is optional on `ComposerSource`, so
a source without it (the demo canvas) draws no roster button rather than
failing. The settled-send path is byte-for-byte what it was when no promotion is
in flight.

Rollback: the four commits are independent and revert cleanly; reverting the
first alone restores the old refusal.

Known gap, deliberately not fixed here: `liveSend` sends its first message to
`sessionCurrent()` rather than to the id the promotion returned. The two differ
only when the reader opened another conversation inside the promotion, and
making them agree is a fix about where a raced first message lands. Marked in
the code.

## Related Issues

N/A


ui-web/src/live/080-overrides.js
```diff
@@ -363,6 +368,77 @@ const mediaOf = (text) => {
...
+async function promote(preview, atPointer) {
+  /* Taken before the first await, so a caller reporting on THIS conversation
+     reads the view as it stood when the promotion began rather than whatever
+     the reader has opened since. */
+  const gen = viewGen;
+  const r = await rpc.call('session.create', {});
+  wsSetRoot(r.info && r.info.cwd);
+  const s = { id: r.session_id, title: T('gui.new_task'), last: preview || T('gui.sess.not_started'),
+    when: T('gui.sess.just_now'), at: Math.floor(Date.now() / 1000), run: null, live: true, persisted: false };
+  sessionRows().unshift(s); sessionSet(s.id); draft = false;
```
gloryfromca, line 429: Blocking: keep sends from bypassing this promotion after the pointer moves.

This sets `draft = false` before the awaited staged model/tier/permission writes and `subscribe` complete. If the roster started the promotion, the main composer remains usable; a send in this interval takes `liveSend`'s `!draft` branch and issues `turn.send` immediately instead of joining `promoting`. That can start the first turn before the draft-selected settings are applied (and before promotion setup has completed), breaking the ordering this extraction is meant to preserve. The new concurrency test only joins while `session.create` is still pending, so it does not exercise this post-pointer window.

gloryfromca (commented):
Blocking: keep main-composer sends behind an in-flight roster promotion until it has finished.

I found one concrete race, marked inline. I covered the full `github/main...HEAD` diff; the affected roster, composer, session-pointer, workspace, and RPC subscription callers; relevant history and the previous first-send behavior; backward compatibility; test intent and whether coverage had been weakened; and the repository rules in `AGENTS.md` and `CONTEXT-MAP.md`. The composer remains the owner of conversation promotion, so I found no separate architecture-boundary issue.

Verification:
- `npm test -- src/features/composer/open-conversation.test.ts src/features/subagents/SubagentsPage.test.tsx` - 97 passed
- `npm test` - 104 files and 1,690 tests passed (the run printed existing happy-dom aborted-fetch/network diagnostics)
- `npm run type-check` - passed
- `npm run build` - passed
- `uv run python scripts/check_source_language.py github/main...HEAD` - passed

Handsome-wzw:
Thanks - the race is real and I reproduced it before fixing it.

**Reproduced.** A new case in `open-conversation.test.ts` enters the post-pointer window deliberately: it starts the promotion, lets `session.create` resolve (which is what moves the pointer and lowers `draft`), asserts `isDraft()` is already false, then sends. Before the fix `turn.send` landed at log index 8, ahead of `subscribe:made-1` at 9. You were right that the earlier concurrency test never entered this window - it joined while `session.create` was still pending, which is before the flag goes down.

**Fixed in `3465fb69`, and it took two changes.** The obvious one is `sendOnSession` joining an in-flight promotion instead of dispatching. That alone did not work: `openConversation` read `draft` before it read `promoting`, so a caller arriving after the pointer moved was answered "there is a conversation, carry on" and walked straight past the setup that had not finished - the same window, one layer down. The promotion is now asked about first, and the flag second.

The guard sits on `sendOnSession` rather than on its caller so a second way in cannot bypass the wait; `liveSend` is one line either way. Joined rather than queued, because `openConversation` answers exactly when the setup is complete, which is exactly when the send is safe.

Both halves are mutation-checked: dropping the guard, and putting the flag back in front of the promotion, each turn the new case red.

`ui-web` tsc clean, `npx vitest run` 104 files / 1691 tests pass, `ui-tui` tsc clean, `check_commit_messages.py` and `check_source_language.py` exit 0.

I also rewrote the description onto `.github/pull_request_template.md`, since you raised that on #400 and it applies here too.

Still outstanding here: the AGENTS.md 3.3 trailer. I have the corrected commits built locally (`Co-authored-by: Claude (claude-opus-5) <noreply@anthropic.com>`, trees byte-identical to these) but landing them needs a force-push, which my environment blocks without the repo owner authorising it. Asking now; will push as soon as I have it.

Co-authored-by: Claude (claude-opus-5) <noreply@anthropic.com>

ui-web/src/live/080-overrides.js
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in `3465fb69a261`: `openConversation` now joins `promoting` before consulting `draft`, and `sendOnSession` waits for that promotion before dispatching. The added regression test enters the post-pointer window and verifies `turn.send` occurs only after all staged writes and subscription setup. I reran the full web suite and related checks successfully, so this thread is settled.

gloryfromca (commented):

The new revision fixes my prior promotion-order blocker: a main-composer send now waits for a roster-owned promotion even after the pointer has moved, and the regression test covers that exact post-pointer interval. I found no new issue worth raising after re-reading the full `github/main...HEAD` diff.

I covered the repository rules in `AGENTS.md` and `CONTEXT-MAP.md`; the full diff and new delta; roster, composer, session-pointer, workspace, and RPC callers; relevant history and backward compatibility; test intent and whether coverage was weakened; and the composer ownership boundary.

Verification:
- `npm test` - 104 files and 1,691 tests passed (with the existing happy-dom aborted-fetch/network diagnostics)
- `npm run type-check` - passed
- `npm run build` - passed
- `uv run python scripts/check_source_language.py github/main...HEAD` - passed

Handsome-wzw:
Reopening this one myself: the branch carried an AGENTS.md 1.1 violation that your no-blockers read did not catch, and I would rather not land it.

Section 1.1 forbids a source comment that references a PR or Issue, and one that preserves transient task context. Two comments here did that:

- `open-conversation.test.ts` opened the promotion-order case with `Reported by gloryfromca on #399, and reproduced here before the fix` - both halves at once.
- `SubagentsPage.test.tsx` opened the new-task case with `The reported bug:` - the transient-context half on its own.

Both are fixed in `948934c5`. The explanations stay, because they are the parts that do not go stale; only the framing goes, so each reads as a statement about the code rather than as a record of a review.

I found these by sweeping this branch for the class after the identical violation was caught on #401, rather than by being told about it here. #400 is clean.

The head is now `948934c5`. Sorry for the extra round - please re-read when convenient.

`ui-web` tsc clean, `npx vitest run` 1691 tests pass, `check_commit_messages.py` and `check_source_language.py` exit 0. The AGENTS.md 3.3 trailer is still the one item I cannot land without an authorized force-push; the rewritten commits are rebuilt on top of this new head and their tree is identical to the published branch.

gloryfromca (commented):
Blocking: correct the three legacy Claude co-author trailers to the AGENTS.md section 3.3 format before merge.

The `948934c5` cleanup correctly removes both review/transient-context references from source comments; I missed those section 1.1 violations in the prior round. I found no new code defect after re-reading the cleanup delta and full `github/main...HEAD` diff.

The author has already recorded the remaining blocker: the first three Claude-assisted commits still use `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>` instead of the required `Co-authored-by: Claude (claude-opus-5) <noreply@anthropic.com>`. The source tree is ready, but this hard repository rule remains unresolved pending the authorized history rewrite.

I covered AGENTS.md comment and commit rules, `CONTEXT-MAP.md`, the full diff and new delta, affected callers and history, backward compatibility, test integrity, and the composer ownership boundary.

Verification:
- `npm test` - 104 files and 1,691 tests passed (with the existing happy-dom aborted-fetch/network diagnostics)
- `npm run type-check` - passed
- `npm run build` - passed
- `uv run python scripts/check_source_language.py github/main...HEAD` - passed
- `uv run python scripts/check_commit_messages.py github/main..HEAD` - passed; this checker does not reject the legacy trailer shape

gloryfromca (commented):
Blocking: correct the three legacy Claude co-author trailers to the AGENTS.md section 3.3 format before merge.

I reviewed the rebased revision. `git range-diff` shows all five PR commits are patch-identical to the previously reviewed series, now based directly on current `github/main`. I rechecked the repository rules, PR diff, callers and history, backward compatibility, architecture boundary, and whether tests were weakened; the earlier functional conclusion still holds and I found no new code defect.

The standing repository-rule blocker remains: commits `6875d712d`, `811d63848`, and `36fcc420a` still use `Co-authored-by: Claude Opus 5 <noreply@anthropic.com>` rather than the exact required `Co-authored-by: Claude (<model-id>) <noreply@anthropic.com>` form. The two newer commits use the required form.

Verification on `1723324f0023`:
- `npm test`: 104 files, 1,691 tests passed
- `npm run build`: passed
- `uv run python scripts/check_commit_messages.py github/main..HEAD`: passed
- `uv run python scripts/check_source_language.py github/main`: passed

Handsome-wzw:
Fixed in `9e10ef26`. All five commits now carry the section 3.3 form.

You named `6875d712d`, `811d63848` and `36fcc420a`, and those were exactly the three — the raw trailer read `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>` where the rule asks for `Co-authored-by: Claude (<model-id>) <noreply@anthropic.com>`. Rewritten by message filter over `github/main..HEAD`, so only the messages moved:

```
1:  6875d712 ! 1:  7ed44de8  fix(ui-web): let the roster start the conversation an instance needs
2:  811d6384 ! 2:  6e509bfa  fix(ui-web): share one promotion between callers that race for it
3:  36fcc420 ! 3:  2ffc3662  docs(ui-web): a message is no longer the only thing that mints a session
4:  57b27bfe = 4:  2d5f4aa7  fix(ui-web): hold a settled send behind an unfinished promotion
5:  1723324f = 5:  9e10ef26  docs(ui-web): keep review attribution out of two source comments
```

`git diff <pre-rewrite> HEAD` is empty, so the trees are identical and no patch changed — the three marked `!` differ only in their trailer line, and the two you said already complied are untouched.

On the rebase in the revision before this one, since it is what unblocked the three red jobs and the cause is worth recording: the branch was cut at `5b7566b4`, where `ui-web/scripts/check-page.mjs` asserted **two** `<script>` blocks. The page build has since grown the asset-digest block and the gate moved to **three**, so the branch was running its own stale gate against a page that now produces three — `check-page: expected exactly two <script> blocks`, red with nothing wrong in the change itself. Rebasing onto current `main` takes the current gate with it; locally `python3 ui-web/build.py && node ui-web/scripts/check-page.mjs` now answers `OK (1,397,555 bytes, 3 scripts parse)`.

Gates on this head: ui-web `gen:check` (168 methods) / `tsc` clean / `vitest` 1691 passed; `ruff check` and `ruff format --check` both pass; `pytest -k "ui or page or rpc_schema"` 2031 passed; `check_commit_messages.py` and `check_source_language.py` exit 0.

gloryfromca (commented):

The message-only rewrite resolves the standing AGENTS.md section 3.3 blocker. `git range-diff` shows exactly the three requested trailer corrections, the old and new heads have the identical tree, and all five commits now carry `Co-authored-by: Claude (claude-opus-5) <noreply@anthropic.com>`.

I rechecked the repository rules and history. Because the source tree is identical, the completed review of the diff, callers, backward compatibility, architecture boundary, and test strength remains applicable; no test was weakened and no new code issue was introduced.

Verification on `9e10ef2635c5`:
- `uv run python scripts/check_commit_messages.py github/main..HEAD`: passed
- `npm test`: 104 files and 1,691 tests passed

The only review thread I opened remains resolved.

0xKT (approved):

PR #400, opened 2026-09-14
fix(ui-web): open composer panels clear of the card, not of the chip

Handsome-wzw:
## Summary

The permission modes (`#permPop`), the session tier (`#tierPop`) and the model
picker (`.mpick`) all hang off a chip in the composer's bottom bar, and all
three measured their clearance from that chip. A chip is INSIDE the card, so
"just above the chip" is over the line the reader is typing on.

Measured on the running page: the card stands at 436..562 and the permission
panel's lower edge landed at 512, covering the field, the attachment row and
anything staged in them -- while the whole reason the panel was opened is to
change how the next message runs.

The anchor still decides which side of the window a panel hangs from. What it no
longer decides is the height. `clearance(anchor)` answers with the composer card
when the anchor is on one, and with the anchor itself when it is not -- which is
the model picker opening from a settings row and wanting exactly what it had.

The picker's downward fallback moves with it. A window too short to fit the
popover above dropped it eight pixels under the CHIP, which is over the bar the
chip sits on; it now goes under the card.

One helper rather than the fix written out three times: the three sites were
already three copies of one placement, and the next panel anchored on the
composer should not have to rediscover which box it has to clear.

A second commit pins the selector. `clearance` holds `.dock-in` as a string and
`page.html` holds the markup, and nothing else tied the two together: rename the
wrapper and all three panels quietly go back to clearing their chip -- a layout
regression that no type-checker sees and no other test covers. A test now reads
the real `page.html` and asserts the class on the element holding the field and
the three chips, rather than grepping for it.

Checked and left alone: `.csheet` and `.dsheet` sit in the `sheets` rack, a
sibling ABOVE `.dock-in` in `page.html`, so they never overlapped it. `#slashPop`
is CSS-positioned inside `.dock-in` with `bottom: calc(100% + 8px)`, which
already clears the whole card.

**CI.** Three jobs fail and all three fail identically on the merge base. `main` at
`5b7566b4` fails exactly these and nothing else, in run `34744930015`:

```
Windows self-upgrade            red here, red on main @ 5b7566b4
page checks                     red here, red on main @ 5b7566b4
unit (py3.12 / ubuntu-latest)   red here, red on main @ 5b7566b4
```

One for one, nothing extra on either side. `page checks` matters most to name
here, since this is a page change. It was reproduced locally on an untouched
worktree checked out at `5b7566b4`, built the way the workflow builds it, with
the same message: `check-page: expected exactly two <script> blocks`. Not read
off the log -- built and run.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
cd ui-web && npx vitest run            104 files / 1686 tests, all pass
cd ui-tui && npx tsc --noEmit -p .     clean
```

Six new cases. happy-dom measures every box as zero, so each element that
decides a placement is handed the rect it has on the running page -- the card at
436..562, the chip on its bottom bar at 518. Every case was mutation-checked by
reverting `clearance` to the anchor and watching the number it asserts come back
as the old one:

- perm: 190 (fixed) vs 272 (chip-anchored), red
- tier: 190 vs 272, red
- picker above: 128px vs 210px, red
- picker below: 194px vs 171px, red
- the `page.html` guard: renamed the wrapper in the fixture, red

Driven through the real entry points -- `perm.open()`, `tier.open()` and
`store.open(anchor)` -- not through the placement function directly.

Run on a real gateway as well: all three panels open clear of the card. The
permission panel measures 192..433 against a card at 436, and the model picker
77..431.

No user-facing doc describes panel placement.

## Risk

Security: none. The change is geometry; nothing new reaches the wire or storage.

Backward compatibility: an anchor that is not on the composer card resolves to
itself, so the settings-page model picker lands exactly where it did. Every
other panel is unchanged.

Rollback: two commits, both revert cleanly and independently.

Known gap, pre-existing and not touched: none of the three panels reposition on
resize or scroll. A window resized while a panel is open leaves it where it was
opened.

## Related Issues

N/A


gloryfromca (commented):
Blocking: restore the required PR template and replace both nonconforming Claude co-author trailers.

The code change itself looks sound. I traced the composer DOM/CSS geometry, all three changed placement callers, the settings-page model-picker fallback, the other `.dock-in` use, and the relevant history. The helper clears the real composer card while preserving non-card anchoring, and the tests exercise the changed entry points without weakening existing assertions. I also checked backward compatibility and the UI/runtime architecture boundary; neither raises a code finding.

Two written repository requirements still need correction:

- The PR description does not use the exact headings and checklists from `.github/pull_request_template.md`. AGENTS.md section 3.7 requires `Summary`, `Type`, `Verification`, `Risk`, and `Related Issues`, with only satisfied boxes checked. The current prose can be retained under those headings.
- Both commits end with `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>`. AGENTS.md section 3.3 requires `Co-authored-by: Claude (<actual-model-id>) <noreply@anthropic.com>`. Please replace the trailer in each commit with the actual session model ID in that exact format.

Verification on this head:

- `npm test -- src/shell/popover.test.ts src/shell/perm.test.ts src/shell/tier.test.ts src/features/model/ModelPicker.test.tsx`: 4 files, 74 tests passed.
- `npm test`: 104 files, 1686 tests passed (exit 0; happy-dom emitted its existing aborted/localhost fetch diagnostics).
- `npm run gen:check`: passed.

The page-check and Windows self-upgrade jobs are also red on the current `main` head, and this branch does not change their inputs. Locally the page build succeeds, then the unchanged `check-page.mjs` rejects the three script blocks produced by the unchanged assembler because it still expects two. I did not attribute those pre-existing CI failures to this change.

Handsome-wzw:
Both repository requirements confirmed against the tree, not taken on trust.

**PR template.** `.github/pull_request_template.md` carries `Summary` / `Type` / `Verification` / `Risk` / `Related Issues`, and AGENTS.md 3.7 requires them. The description is rewritten onto those exact headings with the prose kept under them, only satisfied boxes checked, and the unchecked one (`User-facing docs or screenshots`) explained rather than left bare. Done on #399 as well, since the same rule applies there.

**Trailer.** AGENTS.md 3.3 requires `Co-authored-by: Claude (<model-id>) <noreply@anthropic.com>` with the actual session model id. Mine is `claude-opus-5`, so both commits were wrong in format and in shape.

I have the corrected commits built locally - each message rewritten, and `git diff <old> <new>` empty on both branches, so only the messages moved. Landing them needs a force-push, which my environment blocks without the repo owner authorising it. I have asked; I will push the moment I have it, and will report back here.

Noting for the record that I did not find a code finding to argue with here - your read of the helper and the non-card anchoring matches mine, and thank you for separately confirming the page-check and Windows self-upgrade failures are pre-existing on `main`. I reached the same conclusion by building an untouched worktree at `5b7566b4` and running `check-page.mjs` there.

Co-authored-by: Claude (claude-opus-5) <noreply@anthropic.com>

gloryfromca:
Blocking: push the corrected commit trailers and finish aligning the PR body with section 3.7.

Thanks for checking both requirements directly. I verified that the five required template headings and their checklists are now present. The blocker is not settled yet, for the reasons your comment partly anticipates:

- The remote head is still `389436753a8d`; both published commits still carry the old `Co-Authored-By: Claude Opus 5` trailer. The locally rewritten commits will satisfy this part once the authorized force-push lands.
- The rewritten description adds a separate `## CI` section, although section 3.7 says material without a template section must be appended to `Summary`.
- The description now ends with a Claude co-author trailer, although section 3.7 explicitly says to keep that trailer in the commits and not duplicate it in the PR description.

Please fold the CI note into `Summary`, remove the PR-body trailer, and push the corrected commits when authorized. I will reassess the new head when it appears.

Verification on the current head: `npm test -- src/shell/popover.test.ts src/shell/perm.test.ts src/shell/tier.test.ts src/features/model/ModelPicker.test.tsx` passed all 74 tests in 4 files.

Handsome-wzw:
Both description points fixed; the trailer is the one still waiting on authorization.

- The `CI` section is gone as a heading. Section 3.7 says material the template has no section for is appended to `Summary`, so the note now closes `Summary` as a `**CI.**` paragraph. Five headings, in template order, nothing else.
- The Claude co-author trailer is off the PR body on both PRs. 3.7 is explicit that it belongs in the commits and duplicating it here is the error - my mistake, and the same edit went onto #399, which had it too.

Still pending: the corrected commit trailers. The remote head is unchanged because landing them needs a force-push my environment refuses without the repo owner authorizing it, and I have asked. The rewritten commits are built and waiting - `git diff` between the published branch and the rewritten one is empty on both, so only the messages move. I will push the moment I have it and say so here.

Noted for the next read: the CI paragraph still names `main @ 5b7566b4` and run `34744930015`. Those are evidence for the claim that the three red jobs are pre-existing rather than bare hash references, which is how I read the 3.7 prohibition - say the word if you read it differently and I will cut them to prose.

gloryfromca:
Blocking: the corrected co-author trailers still need to be pushed.

Verified: the PR description now has exactly the five template headings in order, the CI context is part of `Summary`, and the duplicate PR-body trailer is gone. Those two description points are settled.

The published head remains `389436753a8d`, and both remote commits still expose the old trailer, so I agree that only the authorized history update remains. Targeted verification still passes: 4 files, 74 tests.

Handsome-wzw:
One of the two is real and fixed in `129e283f`; the other I cannot reproduce against the current description, and I would rather show you why than change something that already conforms.

**Trailers — real, fixed.** Both commits ended `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>` where section 3.3 asks for `Co-authored-by: Claude (<model-id>) <noreply@anthropic.com>`. Rewritten by message filter over `github/main..HEAD`, so only the messages moved — `git diff <pre-rewrite> HEAD` is empty, both trees identical, no patch changed.

**Template — the description already matches.** Diffing its headings and checklist lines against `.github/pull_request_template.md`, with the tick marks normalised so only structure is compared:

```
diff <(grep -E "^## |^- \[" .github/pull_request_template.md | sed 's/\[x\]/[ ]/') \
     <(grep -E "^## |^- \[" <the PR body>            | sed 's/\[x\]/[ ]/')
→ no output
```

All five headings in order, and the three checklists item-for-item: `Type` with only `Fix` ticked, `Verification` with the two that hold ticked and the docs line left unticked, `Risk` with all three ticked. If it did not look like this when you reviewed it, it has been corrected since; either way I do not think there is anything left to change here, and I would rather say so than edit it to look like I acted.

**On the three red jobs** — your reading was exactly right, and rebasing is the whole remedy. The branch was cut at `5b7566b4`, where `ui-web/scripts/check-page.mjs` asserted **two** `<script>` blocks; the assembler has since grown the asset-digest block and the gate moved to **three**, so the branch was running its own stale gate against a page that now produces three. This revision is rebased onto current `main`, which brings the current gate with it: locally `python3 ui-web/build.py && node ui-web/scripts/check-page.mjs` now answers `OK (1,392,248 bytes, 3 scripts parse)`, and `page checks` and `Windows self-upgrade` are green on this head.

Gates on `129e283f`: ui-web `gen:check` (168 methods) / `tsc` clean / `vitest` 1686 passed; `ruff check` and `ruff format --check` both pass; `check_commit_messages.py` and `check_source_language.py` exit 0.

gloryfromca:

Your evidence holds. `git range-diff` shows the two patches unchanged across the rebase and only the two co-author trailers corrected; the live PR description also matches the template headings and checklists. The previous blocker is resolved.

One nonblocking metadata follow-up was created by the rebase: the `Summary` still says three jobs fail on the merge base at `5b7566b4`, but the current base is `56c79edf` and every check on this head is green. That paragraph is now stale, and section 3.7 also excludes internal commit-hash references from the PR description. Please remove or refresh it. At this late round it does not meet the blocker bar and does not stop merge from my side.

Verification: the targeted 4 files / 74 tests passed; type-check and `gen:check` passed; the served-page build passes with 3 scripts; commit-message checking passed; all GitHub checks are green.

gloryfromca (commented):

The former blocker is fixed. The rebased patches are equivalent to the revision already reviewed, both commits now carry the required `Co-authored-by: Claude (claude-opus-5) <noreply@anthropic.com>` trailer, and the PR body retains the required template structure. I rechecked the project rules, complete branch diff, callers, history via range-diff, backward compatibility, test strength, and UI/runtime architecture; no code finding remains.

Verification: targeted Vitest passed 4 files / 74 tests; TypeScript type-check and RPC generation check passed; the production build and assembled-page check passed with 3 scripts; commit-message checking passed; `git diff --check` passed; all GitHub checks are green.

One nonblocking metadata follow-up is recorded in the reply above: refresh or remove the stale CI paragraph that still describes the old base and includes an internal base SHA. It does not stop this revision from merging.

0xKT (approved):

PR #401, opened 2026-09-14
feat(ui-web): retire an instance from the roster it is listed in

Handsome-wzw:
## Summary

Retiring a sub-agent instance was reachable only from the standalone sub-agent
panel, and the desk's roster is where the list actually lives. One graph run
leaves six rows there and the reader had no way to clear any of them: the row
opened the instance and offered nothing else.

The compact roster row now carries the same control the standalone row already
had. The same verb and the same warning, drawn in a second place rather than
invented again -- `store.forgetInstance`, with `gui.ws.instance_forget_note` on
the control, so the reader is told before the click that an ACP agent's own
session goes with the row (stopping a running turn) and that the records stay on
disk.

A cross rather than a bin, for that last reason: a bin promises a deletion this
does not do.

Standing rather than revealed on hover, which is the rule the stylesheet already
records above `.agent-new` -- a reader who does not know a control is there will
not hover looking for it. Faint on a 29px row until the row is under the pointer,
or the two controls read as one. `stopPropagation`, or dismissing a row would
also open it.

A second commit draws the row's key boundary. An instance row is a focusable
button holding a real button, and a keydown on the inner one bubbles: the row's
handler called `preventDefault()`, which cancels that button's own native
activation, and then opened the pane -- so a keyboard user aiming at the retire
control got the one thing they were trying not to do, while the mouse worked
fine. The boundary is the one `shell/setuprow.tsx` already draws. Applied to
both branches of `InstanceRowView`: the flat list has carried a Remove button
since before the compact row existed and draws the same handler, so the defect
was already there and reachable from the standalone panel.

Nothing behind the UI changed. `subagents.instance.forget`, `DS.agents.instanceForget`
and `store.forgetInstance` already existed and are untouched; only the second
place that calls the last of them is new.

expected exactly two <script> blocks`. Not read off the log -- built and run.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
cd ui-web && npx vitest run            103 files / 1679 tests, all pass
```

Three new cases, all driven through the rendered controls rather than the store
verb, and each mutation-checked.

The retire path:

- remove the button -> red
- drop `stopPropagation` -> red, and it also reds the existing
  `removes a row without opening it`, which is the guard that already covered
  this on the standalone row

It asserts the whole path: the control carries the warning copy and the removal
label, the click reaches `instanceForget` with that row's agent and handle, the
row leaves the list, and the instance is not opened by the same click.

The keyboard path, one case per row shape:

- drop the guard on the compact row -> red
- drop the guard on the standalone row -> red, independently

Each asserts that a keydown on the button survives uncancelled and the row does
not open, paired with a keydown on the ROW still opening it -- without that
second half the first is satisfied by a row that answers no keys at all.

Checked on the running page against a real session holding five instances: the
control renders on every row, carries `gui.ws.instance_forget_note`, and stands
at reduced opacity until the row is hovered.

Then driven end to end on a real gateway: an instance was created, retired from
the roster with this control, and confirmed gone from the server's own registry
file rather than only from the list. The click did not open a pane, and the desk
stayed up.

No user-facing doc describes the roster's controls.

## Risk

Security: nothing new reaches the wire. The call is the one the standalone panel
has always made, with the same two arguments.

Backward compatibility: `store.forgetInstance` returns early when the source has
no `instanceForget`, so a source without it is a control that does nothing rather
than a crash -- the demo canvas installs its own fixture implementation and is
covered. The standalone row is untouched.

Rollback: two commits, both revert cleanly and independently.

Found while fixing this and deliberately NOT fixed here, because it is a
different feature and would make this PR two topics:
`ui-web/src/features/workspace/WorkspacePage.tsx` around line 122 has the same
defect. The `.chghd` row is `role="button"` with an `onKeyDown` that
`preventDefault()`s, and it contains the `.chgm` actions button -- so focusing
that button and pressing Enter toggles the row instead of opening the menu. I
scanned every `role="button"` row in `ui-web/src` for the combination
(preventDefault in the row handler plus a nested button) and that is the only
other hit. Queued as its own change.

User-visible behaviour change: a row can now be retired from the desk. The
action was already possible from the standalone panel, so no capability is new
-- only its reach. It is not confirmed before it runs, which matches the
standalone control; if reviewers want an arm-then-confirm here, say so and I
will add it to both rather than leave the two disagreeing.

## Related Issues

N/A


ui-web/src/features/subagents/SubagentsPage.tsx
```diff
@@ -138,6 +150,23 @@ export function InstanceRowView({ it, onOpen = store.openInstanceRow, compact =
...
+
+          Standing rather than revealed on hover, for the reason the CSS above
+          `.agent-new` records: a reader who does not know a control is there
+          will not hover looking for it. `stopPropagation`, or dismissing a row
+          would also open it. */}
+      <button
+        className="inst-retire"
+        title={t('gui.ws.instance_forget_note')}
+        aria-label={t('gui.ws.instance_forget')}
+        onClick={(e) => { e.stopPropagation(); store.forgetInstance(it) }}
```
gloryfromca, line 175: Blocking: keyboard activation of the new retire control opens the instance instead of retiring it.

The button's Enter/Space `keydown` bubbles into the compact row's handler at lines 137-142. That handler calls `preventDefault()` and `open()`, which cancels the button's native keyboard activation and opens the pane; stopping propagation only in the later `click` handler cannot stop this earlier event. As a result, a keyboard user who focuses this control cannot perform the advertised removal. The same nested-action failure is already guarded in `SetupRow` by ignoring key events whose target is not the row; please establish that event boundary here as well and cover keyboard activation in the regression test.

gloryfromca (commented):
Blocking: the new roster retire control must work from the keyboard without opening the instance.

I reviewed the full diff and both `AgentList` call sites, traced `forgetInstance` through the live RPC adapter and server implementation, checked the relevant component history, and covered the repository rules in `AGENTS.md`/`CLAUDE.md` plus the domain and architecture routing in `CONTEXT-MAP.md` and `CONTEXT.md`. I also checked backward compatibility, confirmed the tests add coverage without weakening existing assertions, and verified the web UI continues to use its source/store boundary rather than reaching around it. One concrete interaction defect remains inline.

Verification: `npm test --prefix ui-web` (103 files, 1679 tests passed), targeted `SubagentsPage.test.tsx` (86 tests passed), `npm run gen:check --prefix ui-web`, `npm run type-check` from `ui-web/`, `npm run build` from `ui-web/`, and `git diff --check` all passed.

Handsome-wzw:
You are right, and it was worse than a missing affordance - reproduced before fixing.

`fireEvent.keyDown` on the retire button came back cancelled, which is exactly what takes the native activation away: the row handler runs first, `preventDefault()`s, and opens the pane. So a keyboard user aiming at the control got the one thing they were trying not to do, while the mouse worked fine.

Fixed in `36ca4182` with the boundary you pointed at - `if (e.target !== e.currentTarget) return`, the same one `shell/setuprow.tsx` draws, whose comment describes this identical failure.

**Applied to both branches of `InstanceRowView`, not just the new one.** The flat list has carried a Remove button since before the compact row existed and draws the same handler, so the defect was already there and reachable from the standalone panel. Leaving it would have made that the one row where the keyboard still loses, with the fix sitting one function away.

Two new cases, one per row shape, each mutation-checked by removing that row guard alone. Each asserts the keydown on the button survives uncancelled and the row does not open, paired with a keydown on the ROW still opening it - without the second half the first is satisfied by a row that answers no keys at all.

`ui-web` tsc clean, `npx vitest run` 1681 tests pass, `ui-tui` tsc clean, `check_commit_messages.py` and `check_source_language.py` exit 0.

**One more of the same class, found and deliberately not fixed here.** I scanned every `role="button"` row in `ui-web/src` for the combination that produces it - `preventDefault()` in the row handler plus a nested real button - and there is exactly one other hit: `ui-web/src/features/workspace/WorkspacePage.tsx` around line 122, where the `.chghd` row contains the `.chgm` actions button, so focusing it and pressing Enter toggles the row instead of opening the menu. That is a different feature and folding it in would make this PR two topics, so it is queued as its own change. Say the word if you would rather have it here.

ui-web/src/features/subagents/SubagentsPage.tsx
gloryfromca: Blocking: remove the transient reviewer and pull-request attribution from the new regression-test comment.

The event boundary in `36ca4182` fixes the keyboard behavior from this thread, and the paired cases preserve both row shapes while confirming the rows themselves still respond to Enter. I verified the targeted behavior and full web suite, so this finding is settled and I am closing its thread. A separate inline finding identifies the remaining repository-rule issue.

ui-web/src/features/subagents/SubagentsPage.test.tsx
```diff
@@ -292,6 +292,80 @@ describe('subagents island, the list', () => {
...
+    })
+    expect(forgotten).toEqual([['hermes', 'one']])
+    /* Gone from the list, and the row did NOT open -- dismissing a row must not
+       also be a click on it. */
+    expect(store.getState().instances.map((r) => r.handle)).toEqual(['two'])
+    expect(store.getState().open).toBeNull()
+  })
+
+  it('lets the keyboard reach the retire control instead of opening the row', async () => {
+    /* Reported by gloryfromca on #401. The row is a focusable button holding a
```
gloryfromca: Blocking: remove the reviewer and pull-request reference from this regression-test comment.

`AGENTS.md` section 1.1 explicitly forbids source comments that reference a PR or Issue and comments that preserve transient task context. `Reported by gloryfromca on #401` does both, so this otherwise-correct change violates a hard repository constraint and is subject to rejection or reversion. The enduring explanation of the nested key-event failure is useful; only the review attribution needs to leave the source.

gloryfromca (commented):
Blocking: remove the transient reviewer and pull-request attribution from the regression-test source comment.

The keyboard defect from the previous review is fixed: both `InstanceRowView` variants now restrict row activation to the row's own key events, the paired regression cases preserve the nested controls, and I replied on and resolved my original thread.

For this revision I reread the complete `github/main...HEAD` diff and the prior-head delta, checked both row variants and `AgentList` callers, traced the store/live-RPC/server path, reviewed the relevant history, and covered `AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md`, and the applicable `CONTEXT.md` terms and boundaries. I checked backward compatibility, confirmed no tests were weakened, and confirmed the web UI still respects its source/store architecture boundary. One new hard-rule violation remains inline.

Verification passed: `npm test --prefix ui-web` (103 files, 1681 tests), `npm run type-check --prefix ui-web`, `npm run build --prefix ui-web`, `npm run gen:check --prefix ui-web`, the source-language and commit-message Python checks through `uv run`, and `git diff --check`.

Nonblocking follow-up recorded from the author's note: the pre-existing `.chghd`/`.chgm` nested keyboard-action issue in `WorkspacePage.tsx` is intentionally deferred to a separate change.

Handsome-wzw:
Confirmed against the rule and fixed in `c8bed22c`.

AGENTS.md 1.1 forbids a source comment that references a PR or Issue **and** one that preserves transient task context. `Reported by gloryfromca on #401` is both at once. The enduring half stays - a row is a focusable button holding a real button, the inner one's keydown bubbles, and the row's `preventDefault` cancels its activation - now told as what the code does rather than as a record of who found it.

**I then swept my other two branches for the same class rather than waiting to be told twice**, and #399 had it: `open-conversation.test.ts` carried `Reported by gloryfromca on #399`, and `SubagentsPage.test.tsx` there opened a comment with `The reported bug:`, which is the transient-context half of 1.1 on its own. Both are fixed on that branch and pushed - it had already been cleared, so the violation would otherwise have gone to `main` under a no-blockers review. #400 was clean.

The grep I used, for the record, so the next sweep can repeat it: reviewer names, `Reported by`, `#39x`/`#40x`, `reviewer`, and task framing (`the reported`, `this PR`, `ticket`) over `git diff github/main...HEAD -- ui-web/src raven/`, restricted to added lines.

`ui-web` tsc clean, `npx vitest run` 1681 tests pass, `ui-tui` tsc clean, `check_commit_messages.py` and `check_source_language.py` exit 0.

Thank you for closing the keyboard thread.

ui-web/src/features/subagents/SubagentsPage.test.tsx
gloryfromca: No blockers; this can merge as far as I am concerned.

`c8bed22c` removes the reviewer/PR attribution and preserves the durable explanation of the event boundary. The full web suite and validation checks pass, so this finding is settled and I am closing the thread.

gloryfromca (commented):

The new commit settles the remaining `AGENTS.md` source-comment violation without changing behavior or weakening the keyboard regression coverage. I rechecked the complete diff and revision delta, the two row variants and their callers, the store/RPC path and relevant history, backward compatibility, tests, and the repository's documented domain and architecture boundaries. Both threads I opened are now resolved.

Verification passed: `npm test --prefix ui-web` (103 files, 1681 tests), web type-check, production build, generated-client check, source-language check, commit-message check, and `git diff --check`.

Named nonblocking follow-up: the author has deferred the pre-existing `.chghd`/`.chgm` nested keyboard-action issue in `WorkspacePage.tsx` to a separate change.

gloryfromca (commented):

This revision is a rebase onto the current `main`: `git range-diff` reports all three PR commits as patch-identical, and the current PR diff remains limited to the same three web UI files. I rechecked the full current diff, callers and event boundary, repository rules and architecture constraints, backward compatibility, test integrity, and the resolved review record; the newer base introduces no conflict or regression. Both threads I opened remain resolved.

Verification passed on the rebased tree: `npm test --prefix ui-web` (103 files, 1681 tests), web type-check, production build, generated-client check, source-language check, commit-message check, and `git diff --check`. GitHub reports the revision mergeable.

The previously named `.chghd`/`.chgm` keyboard issue in `WorkspacePage.tsx` remains an explicitly deferred, nonblocking follow-up.

0xKT:
Not a finding and not a blocker -- the defect below is not yours. Filing it as
a note because this PR is what makes it easy to hit, and you are the person
standing closest to it.

`store.forgetInstance` drops the row from view, fires the call, and swallows
whatever comes back:

    set({ instances: state.instances.filter(...) })
    instDrawn = ''
    void src.instanceForget(it.agent, it.handle).catch(() => { /* the refresh restores it */ })

I checked what "the refresh restores it" is worth, and measured rather than
read. Driving the real store with an `instanceForget` that rejects:

    view keeps:    ['two']
    server keeps:  ['one', 'two']

So the row is gone from the screen, the instance is still there, and nothing
says so. The reader's next act is to believe it worked.

The recovery is real but it is not automatic. `refreshInstances` has exactly
two production callers -- `shell/resume.ts:82`, when a URL intent opens a named
instance, and `SubagentsPage.tsx:224`, the list's own mount effect. There is no
heartbeat. So the row comes back when the reader leaves this panel and returns,
which is not a thing they will do, because as far as they know the retire
worked.

WHY THIS IS NOT YOUR FINDING. `forgetInstance` is byte-identical at
56c79edf and your diff does not touch store.ts at all. The empty catch and the
comment above it both predate you.

WHY I AM WRITING IT ON THIS PR ANYWAY. You changed how reachable it is. At
56c79edf the file had ONE call site for it; here it has three, and the new one
sits in the list your own comment describes as where "one run leaves six rows".
A silent failure behind a control nobody could find is a different thing from a
silent failure behind a control that is now standing on every row of the list
people actually use. The defect did not move; its exposure did.

Every other write in that store already does better -- the create path reads
`data.detail` because "message is the wire CODE" and a refused create carries a
sentence the reader can act on; the direct-send path rolls its own optimistic
state back on failure. This one call is the only empty catch among them.

Nothing here argues against merging. The retire control is the right feature in
the right place, the keyboard boundary you fixed alongside it is a real fix with
two tests that fail independently (I reverted each half separately and each
reddened only its own), and the warning text is carried rather than reworded.

If you want the smallest thing that closes it: `refreshInstances(true)` in that
catch would put the row back within one call instead of one navigation, without
needing a toast or touching the shape of anything.

0xKT (approved):

PR #403, opened 2026-09-14
test(cli): pin the doctor console so the memories path stays whole

gloryfromca:
## Summary

`test_doctor_answers_where_the_memories_are` has been red on the `unit` job on
every push since the history rewrite, and green on GitLab and on every mac. The
test is not environment-flaky; the failure is deterministic and the trigger is
path length.

rich hard-wraps the value column at the console width and breaks mid-token. At
the default width of 80, an 81-character memory root splits as `mem-roo` + `t`,
so `assert "mem-root" in r.stdout` fails even though doctor printed the root
correctly. On the CI runner `tmp_path` sits under `/tmp` and lands on exactly 81
characters; a mac resolves the same basetemp to `/private/tmp` and gains eight
characters, which moves the wrap point off the token.

Pin the module-level console instead of the ambient width, the way
`tests/test_cli_plugin_commands.py` already does for the same trap. That holds
whatever the terminal reports and does not depend on when rich reads its size.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other (tests only)

## Verification

Single-process runs, at a basetemp chosen so the root is 81 characters again
(`--basetemp=/private/tmp/aaaaaaaaaaaaaaaaaaaaaaaaaaa -n 0`). The pin is
load-bearing, not decoration:

```
console pinned to width=80   -> 1 failed    assert 'mem-root' in r.stdout
console pinned to width=300  -> 1 passed
pytest tests/test_cli_doctor_commands.py -q  -> 97 passed
```

Note for anyone re-running this: the wrap point depends on the resolved
`tmp_path`, so the reproduction needs `-n 0`. Under the default `-n 4` xdist
inserts a `popen-gwN` segment, the root becomes 91 characters, the wrap lands
clear of the token and the test passes whatever the width is.

## Risk

Test-only change. No shipped code is touched. The assertion is unchanged, so it
still fails if doctor stops printing the root; only the renderer's width is now
fixed rather than inherited.

## Related Issues

N/A


gloryfromca:
### Review history carried over from GitLab MR !669

This is a transcript of the review that ran on the GitLab side, pasted here so the
reasoning is not lost in the move. It is **not** a review performed on this PR: the
quoted verdicts were produced by the bots on GitLab, and I am only relaying them.

This PR's head `487bca8a5` is one of the revisions reviewed there, so the verdicts below
apply to exactly this tree.

<details><summary><b>mrbot</b> &mdash; 2026-09-14 08:45 UTC</summary>

> No blockers; this can merge as far as I am concerned.
>
> The direct console replacement is verified and removes the environment and import-order dependency entirely. One nuance for the record: Rich re-reads `COLUMNS` at render time only while `Console._width` remains unset; if `COLUMNS` exists when `Console()` is constructed, `__init__` stores that value in `_width`, and a later invoke-level override cannot move it. Also, `-n 0` is needed to reproduce the exact 81-character case, although the earlier default-xdist run on this runner still hit a split at its longer path. The new fixed-width console handles both cases.

</details>

<details><summary><b>mrbot</b> &mdash; 2026-09-14 08:46 UTC</summary>

> No blockers; this can merge as far as I am concerned.
>
> The replacement of `doctor_commands.console` with `Console(width=300)` fixes this finding. With ambient `COLUMNS=80`, the single-process CI-shaped reproduction now passes, and the complete doctor module is green. This thread is settled.

</details>

<details><summary><b>mrbot</b> &mdash; 2026-09-14 08:46 UTC</summary>

> No blockers; this can merge as far as I am concerned.
>
> Revision `487bca8a597c` directly pins `doctor_commands.console`, resolving the earlier width-order dependency without weakening the assertion. The adjusted 300-column helper also remains sufficient for its existing browser-path tests; the full module covers both callers.
>
> Verification:
> - `COLUMNS=80 uv run pytest tests/test_cli_doctor_commands.py -k doctor_answers_where -n 0 --basetemp=/tmp/aaaaaaaaaaaaaaaaaaaaaaaaaaa -q` - 1 passed, 96 deselected
> - `uv run pytest tests/test_cli_doctor_commands.py -q` - 97 passed
> - `git diff --check origin/main...HEAD` - passed
>
> Review coverage included the full current diff, delta from the previously reviewed revision, doctor console behavior, related fixed-console test patterns, commit history and MR discussion, backward compatibility, test strength, and the repository guidance in `AGENTS.md`, `CLAUDE.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`. The branch's `CONTEXT.md` remains the controlling architecture document; this test-only diff changes no runtime package, named seam, or enforcement roster, so none of the ten runtime architecture sections is substantively in scope. The earlier blocker thread has been closed and resolved.

</details>

gloryfromca (commented):

The fixed module-level console removes the ambient-width and import-order dependency while preserving the existing `mem-root` assertion. The change from 400 to 300 columns in the shared browser-path helper remains sufficient for both callers, as covered by the full doctor module. There is no runtime or backward-compatibility change.

Verification:
- `COLUMNS=80 uv run pytest tests/test_cli_doctor_commands.py -k doctor_answers_where -n 0 --basetemp=/tmp/aaaaaaaaaaaaaaaaaaaaaaaaaaa -q` - 1 passed, 96 deselected
- `uv run pytest tests/test_cli_doctor_commands.py -q` - 97 passed
- `PYTHONPATH=. uv run --frozen --python 3.13 python scripts/check_source_language.py github/main...HEAD` - passed

The current page-check and Windows self-upgrade failures are in untouched areas: the former reports the existing generated-page script-count mismatch, and the latter reports a missing `pydantic` dependency in its upgrade fixture environment. Neither is caused by this test-only diff; the unit job was still running when reviewed.

Review coverage included the full diff, the doctor renderer caller, both users of the fixed-width helper, the analogous plugin-console pattern, commit history and the carried-over discussion, backward compatibility, assertion strength, and `AGENTS.md`, `CLAUDE.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`. No runtime architecture seam or domain term changes in this test-only diff.

0xKT (approved):

PR #405, opened 2026-09-14
feat(ui-web): show how long an instance has been on its turn

Handsome-wzw:
## Summary

A sub-agent pane header said what the instance was and what mode it ran at, and
nothing about whether it was moving. The main transcript puts an elapsed clock
beside a running card; this is the same number, in the same spelling
(`shell/duration.ts`) and on the same self-stopping tick, so a reader who has
learned one has learned both.

**The number is the running turn's, not the instance's lifetime.** That is the
decision this change turns on, so it is worth stating plainly: the lifetime was
the easy number and the wrong one -- a different fact wearing the same clothes,
and a pane header beside a live status is exactly where the two would be
confused.

The turn's start had to be published, because nothing already on the row could
stand in for it:

- `updatedAtMs` was the tempting field -- always present, looks like a time --
  but every registry write stamps it. A binding commit and a graph-origin write
  both move it, so a clock counting from it dates the row rather than the turn,
  and would run on an instance answering nothing at all.
- `createdAtMs` is the lifetime, which is the number this deliberately is not.

So `turnStartedAtMs` comes off the activity the backend is collecting
(`run_activity.live_instance`), which is the only place a turn's start exists,
and it is stamped only while a turn is running. Absent is the signal rather than
zero: the reader draws a clock when the field is there and nothing when it is
not, so an instance that finishes between two polls stops counting instead of
freezing on its last number.

A second commit corrects *which* instant that is. The activity's
`started_at_ms` is when collection opened, and a spawn builds its activity
before waiting on `hold_handle` -- so a spawn queued behind a five-minute direct
turn on the same handle opened its pane claiming five minutes of work it had not
done. The stamp is now taken in `watching_instance`, the single place both
registration paths funnel through and the first instant the run is the one the
instance is answering, and only when the slot is actually taken: a lane that
lost the race stays unstamped, because absent is what says an instance is
answering nothing and a loser must not look like a worker. The manager already
states this rule one line from its own acquisition -- it writes `running` "only
now ... until the lock is held the run is still the pending spawn() reported" --
so the clock now agrees with the status.

That is a new field rather than a move, because `started_at_ms` has two other
readers that mean it as written: the stall watcher's floor, and the live prompt
row's date. The second is worth stating -- for a queued spawn the question
really was asked before the wait, so the prompt keeps the earlier stamp while
the clock takes the later one, and the two now say two different true things.

`useTick` moves to `shell/tick.ts` rather than being written a second time. It
was already the transcript's, with five callers there; this is the sixth and
wanted the same hour-long ceiling and the same self-stopping interval.

The two running rows in the demo canvas's fixtures gain the field and the
settled ones deliberately do not, so the demo shows both halves of the rule
rather than a clock on finished work.

expected exactly two <script> blocks`. Not read off the log -- built and run.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
cd ui-web && npx tsc --noEmit -p .          clean
cd ui-web && npm run gen:check              generated.ts matches the contract (168 methods)
cd ui-web && npx vitest run                 104 files / 1683 tests, all pass
cd ui-tui && npx tsc --noEmit -p .          clean
cd ui-tui && node scripts/gen-i18n.mjs --check   catalogue up to date
pytest tests/test_rpc_instances.py tests/test_rpc_schema_match.py   439 pass
```

`tests/test_rpc_schema_match.py`), and `ui-web/src/rpc/generated.ts` is
regenerated from it (held equal by `npm run gen:check`).

Seven new cases, every one mutation-checked by reverting the line it covers:

Server:

- stamp zero instead of omitting when nothing runs -> red
- publish `started_at_ms` instead of `turn_started_at_ms` -> red. The fake
  carries both, with different values, so the case pins which field is read
  rather than merely that a number appears.
- stamp the turn at construction instead of at slot acquisition -> red, two
  cases. That mutation is the reported defect exactly.
- stamp a lane that lost the race for the slot -> red

Client:

- fall back to `updatedAtMs` when no turn is running -> red. This is the case
  that matters: a first attempt asserted the right field was *preferred*, which
  a fallback satisfies, so it caught nothing. It now drives a row carrying only
  the row stamp.
- drop the sub-second hold-back -> red
- drop the accessible label -> red

No user-facing doc describes the pane header.

## Risk

Security: the new field is a timestamp already held in process for the session
being asked about, published on a method that already answers only for that
session.

Backward compatibility: the field is optional everywhere. A client that does not
read it is unaffected; a server too old to send it makes the clock absent, which
is the same state as "answering nothing" and already the drawn-nothing case. The
`useTick` move is a relocation, not a change -- the transcript's five callers get
the identical hook.

Rollback: two commits. The second reverts on its own and restores only the
earlier (wrong) instant; reverting both removes the contract field, which no
other consumer reads.

Deliberately not done here: the standalone sub-agent panel's own detail header
(`.sahd`) does not get the clock. The desk pane is where the user asked for it
and where the header already carries the mode chip; adding a second copy is a
one-line follow-up once this shape is agreed, and doing both now would put the
same number in two headers before anyone has read either.

## Related Issues

N/A


raven/rpc/methods/instances.py
```diff
@@ -308,14 +308,39 @@ async def instances_list(
...
+    on the activity the backend is collecting, which is the only place it exists.
+
+    Absent rather than zero when nothing is running, and absent is the whole
+    signal: a reader shows a clock when the field is there and nothing when it
+    is not, so an instance that finished between two polls stops counting
+    instead of freezing on its last number.
+    """
+    for row in rows:
+        live = run_activity.live_instance(session_key, str(row.get("agent") or ""), str(row.get("handle") or ""))
+        began = getattr(live, "started_at_ms", None) if live is not None else None
```
gloryfromca: Blocking: publish the start of the instance's actual turn, excluding time spent queued for its handle.

For spawned work, `RunActivity` is constructed in `manager.py` before the task waits on `hold_handle`, and that same object is registered with `watching_instance` only after the lock is acquired. The existing queue regression test establishes that the spawn is still pending during this wait. Once it becomes current, this line therefore exposes the earlier collection timestamp: after a five-minute direct turn on the same handle, the new spawn's pane can immediately say it has been working for five minutes even though its turn just began. Please make the timestamp used here reflect the post-lock acquisition time (or otherwise expose that actual start).

gloryfromca (commented):
Blocking: the elapsed clock must start when a queued spawn actually acquires the instance.

I reviewed `github/main...b451e4d40163`, the activity and handle-lock lifecycle, the RPC schema and generated client, the instance polling and pane data flow, the extracted transcript timer and relevant history. I also checked backward compatibility (the wire field is optional), confirmed tests were added rather than weakened, and covered the repository rules in AGENTS.md/CLAUDE.md plus the Runtime/UI boundary and canonical terms in CONTEXT-MAP.md and CONTEXT.md.

Verification:
- `uv run pytest tests/test_rpc_instances.py -x`: 95 passed
- `uv run pytest tests/test_subagent_direct_chat.py::test_a_spawn_queued_behind_a_direct_chat_stays_pending_and_leaves_the_slot -x`: 1 passed
- `npm test -- --run src/features/subagents/TurnClock.test.tsx src/features/workspace/DeskSurface.test.tsx`: 26 passed
- `npm test -- --run src/features/transcript/TranscriptPage.test.tsx`: 156 passed
- `npm run type-check && npm run gen:check`: passed; generated client matches 168 methods
- The `make check-source-language` wrapper was unavailable because `make` is not installed; its underlying `scripts/check_source_language.py github/main..HEAD` command passed.

Handsome-wzw:
Confirmed in the code before I touched it, and your reading is exact. Fixed in `9a65298f`.

`watching_instance` registers *inside* `hold_handle`, so the activity `live_instance` hands back was built before the wait - and the manager states the very rule I had broken one line from its own acquisition: it writes `running` "only now ... until the lock is held the run is still the pending spawn() reported". The status already excluded the queue; my clock did not.

The stamp now happens in `watching_instance`, which is the single place both registration paths funnel through (the `collecting(instance=...)` path opens its block through it too) and the first instant the run is the one the instance is answering. **Only when the slot is actually taken** - a lane that lost the race stays unstamped, because absent is what says an instance is answering nothing, and a loser must not look like a worker.

A new field rather than moving `started_at_ms`, because that one has two other readers that mean it as written: the stall watcher's floor (`dag_runner.py:151`), and the live prompt row's date (`instances.py` `_live_turns`). The second is worth naming - for a queued spawn the question really *was* asked before the wait, so the prompt keeps the earlier stamp while the clock takes the later one. The two now say two different true things rather than one of them lying.

Mutation-checked, and the first of these is your defect exactly:

- stamp at construction instead of at acquisition -> two cases red
- stamp a lane that lost the race -> red
- publish `started_at_ms` instead of `turn_started_at_ms` -> red. I also strengthened that RPC case: the fake now carries both fields with different values, so it pins which one is read rather than merely that a number comes out.

`pytest` over activity, instances, schema-match and direct-chat: 565 passed. `ui-web` tsc clean, `gen:check` matches 168 methods, `npx vitest run` 1683 pass, `ui-tui` tsc clean, `check_commit_messages.py` and `check_source_language.py` exit 0.

Thank you for the `test_a_spawn_queued_behind_a_direct_chat_stays_pending_and_leaves_the_slot` pointer - that is what made the window concrete rather than theoretical.

raven/rpc/methods/instances.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed fixed in `9a65298f3d95`: `watching_instance` now stamps `turn_started_at_ms` only after it successfully claims the instance slot, and the RPC publishes that timestamp rather than the earlier collection time. The added tests cover both queue exclusion and a lane that fails to take the slot. The relevant backend and web suites pass.

gloryfromca (commented):

The queued-spawn timing defect is fixed: the instance-specific stamp is now taken only when `watching_instance` claims the slot, while the earlier collection timestamp remains separate. I re-read the full `github/main...HEAD` diff and the revision delta, including every activity registration caller, the handle-lock history, RPC schema/generated client, instance polling and pane integration, and the shared transcript timer. I also rechecked AGENTS.md/CLAUDE.md and the relevant CONTEXT-MAP.md/CONTEXT.md architecture terms and boundaries. The optional wire field remains backward compatible, and tests were strengthened rather than weakened.

Verification:
- `uv run pytest tests/test_subagent_activity.py tests/test_rpc_instances.py tests/test_subagent_direct_chat.py::test_a_spawn_queued_behind_a_direct_chat_stays_pending_and_leaves_the_slot -x`: 100 passed
- `npm test -- --run src/features/subagents/TurnClock.test.tsx src/features/workspace/DeskSurface.test.tsx src/features/transcript/TranscriptPage.test.tsx`: 182 passed
- `npm run type-check && npm run gen:check`: passed; generated client matches 168 methods
- `scripts/check_source_language.py github/main..HEAD` through the repository's uv environment: passed

I replied to and resolved the prior thread because the finding is settled.

gloryfromca (commented):

This revision restores the TUI generated client that the shared OpenRPC contract requires. I reviewed the new delta and rechecked it against the full `github/main...HEAD` change, the TUI generator and consumers, relevant history, the resolved timing thread, and the Runtime/TUI RPC boundary described by CONTEXT-MAP.md and ui-tui/CONTEXT.md. I also checked the applicable AGENTS.md/CLAUDE.md rules, backward compatibility (the generated field is optional), and that tests were not weakened. No runtime TUI behavior changes in this commit.

Verification for the new delta:
- `npm run lint:rpc && npm run type-check`: passed; generated client is in sync
- Relevant TUI tests covering `InstanceRow` consumers: 82 passed across 5 files
- Repository source-language check: passed

The thread I opened remains resolved.

gloryfromca (commented):

This head is a rebase of the already-reviewed three commits onto newer `main`, not a changed PR patch. I compared the old and new ranges with `git range-diff` (all three commits are equal), and the aggregate patches have the same SHA-256. I also re-read `github/main...HEAD` and checked the previously covered callers, compatibility, test strength, AGENTS.md/CLAUDE.md rules, and CONTEXT-MAP.md plus Runtime/TUI architecture boundaries for interaction with the new base. No conflict-resolution change or new finding emerged, and the prior thread remains resolved.

Post-rebase verification:
- Backend activity/RPC/queued-spawn tests: 100 passed
- Web clock, workspace, and transcript tests: 182 passed; type-check and RPC generation check passed
- TUI `InstanceRow` consumer tests: 82 passed; type-check and RPC generation check passed
- Repository source-language and diff checks: passed

gloryfromca (commented):

This revision is another rebase onto newer `main`. Zero-context comparison shows the PR's actual added and removed lines are unchanged; the range difference is only updated base context, including preservation of `main`'s corrected Chinese punctuation beside the new message. I rechecked `github/main...HEAD`, the callers and history, backward compatibility, test strength, AGENTS.md/CLAUDE.md rules, and the relevant CONTEXT-MAP.md Runtime/UI/TUI boundaries against the new base. No fourth-round blocker criteria are met, and the prior thread remains resolved.

Current-base verification:
- Backend activity/RPC/queued-spawn tests: 100 passed
- Web clock, workspace, and transcript tests: 182 passed; type-check and RPC generation check passed (172 methods)
- TUI `InstanceRow` consumer tests: 82 passed; type-check, RPC generation, and i18n generation checks passed
- Repository source-language and diff checks: passed

gloryfromca (commented):

This is a base-only rewrite: `git range-diff` reports all three PR commits equal, and zero-context comparison shows no semantic PR change. I rechecked the full diff and its callers against the newer base, plus backward compatibility, test strength, AGENTS.md/CLAUDE.md rules, and the applicable CONTEXT-MAP.md Runtime/UI/TUI boundaries. No late-round blocker condition is present, and the prior thread remains resolved.

Current-base verification:
- Backend activity/RPC/queued-spawn tests: 100 passed
- Web feature tests: 182 passed; type-check and 176-method RPC generation check passed
- TUI consumer tests: 82 passed; type-check, RPC generation, and i18n generation checks passed
- Repository source-language and diff checks: passed

0xKT (approved):

PR #406, opened 2026-09-14
fix(cli): work that belongs to no session still has to show

LivXue:
## Summary

The tracing viewer keys its entire payload on a session id, and three readers drop a span that has none. On the measured store that is **683 of 1,956 spans (35%)**, dropped silently: every `plugin.load` ever recorded, every cron heartbeat, and the titles generated after a turn ended. A day whose only activity was scheduled reads as a day with no activity at all -- which is how three consecutive days (09-12 ... 09-14) came to look empty in the panel while 154 spans sat on disk.

A span with neither `session.id` nor `session.key` now gets a session derived from its own `startTime`, one per calendar day (`background:<day>`). The derivation is **span-local on purpose**: the whole-corpus reader in `server.js` and the per-file shard index are already pinned against each other by test, and any derivation consulting global state would let them elect different ids for the same span. A day is the grain because per-trace would shatter a year of timer ticks into a session each. A span that has a key but no id is untouched -- the identity election resolves that one, and still drops it when it cannot.

`SCHEMA` goes 2 -> 3. The sidecar records nothing about the code that built it, so leaving it at 2 would let every index already on disk keep answering with the old contents: the fix would ship and change nothing for the history it was written for. The comment above `SCHEMA` already describes that failure from the last time it bit.

**This reverses a deliberate decision, and reviewers should weigh it as such.** `server.js` carried an explicit rationale for the drop (added in 7b45527b5): *"there is nowhere in a session-oriented view to show it, and on the measured store there are 34,644 of them."* Both halves are addressed rather than ignored -- the per-day session is the place to show it, and the per-day grain is what keeps the count in the same order of magnitude as a real session. That comment is updated rather than left contradicting the code.

`docs/specs/2026-08-25-tracing-viewer-sharded-reads-design.md` records the narrowing: its constraint "a span cannot be attributed to a session by looking at the span" is a statement about spans carrying a `session.key`. A span with neither has no corpus fact to recover, which is why attributing it span-locally is sound rather than a shortcut past the constraint.

## Review round

Three findings, one commit each, all filed as non-blocking and all fixed.

**The comparison tests could not see a reader disagreeing.** Both tests that hold
the sharded reader against the whole-corpus one ran on a fixture whose every span
carried both identity fields, so neither ever built a background session. Changing
the day grain on one side alone left all 21 green. The fixture now carries a
sessionless span, and the same change fails both comparisons.

**The panel opened on a background day.** The list is newest-first and the page
takes the first row; background work is emitted continuously, so the newest record
in a store is usually one of those. This branch therefore changed the row the panel
opens on, a default nobody chose. The row now declares `isBackground` and the page
prefers a conversation. The declaration is on both readers, which the field-for-field
comparison between a list row and a whole-corpus session then holds together.

**A sessionless trace across UTC midnight had no single owner.** Its spans land in
two day rows, and the trace-id lookup returned whichever row iteration reached
first. It now returns the earliest owner, which is the single-owner rule the
trace-id union already states.

The day grain is deliberately unchanged. Deriving it per trace would make one bucket
per sessionless span: measured on a real store, 770 sessionless spans carried 770
distinct trace ids, which takes one file's pair count from 3 to 341 -- avoiding that
is why the grain is a day. So the straddling trace's spans stay split across two
rows and only the ownership is settled. Measured incidence of a straddle on that
store is zero, so its test forces the state rather than reporting a failure anyone
has seen.

The review also named the `undated` fallback as reached by no test. It is reachable rather than defensive -- `normalizeSpan` passes `startTime` through as it finds it -- so it now has one, asserting the pool exists and the span arrives through it. No writer emits an undated span and none is on the measured store.

`SCHEMA` is untouched by this round: nothing here changes what a sidecar holds.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/integration/test_tracing_viewer_e2e.py -m integration     24 passed  (5 new)
pytest tests/test_tracing_api.py tests/test_cli_tracing_commands.py \
       tests/test_tracing_compact.py tests/test_tracing_artifact_v2.py 131 passed
ruff check / ruff format --check                                       clean
scripts/check_commit_file.py                                           exit 0
node --check server.js shard-index.js                                  clean
```

Against the real store (span logs copied to an isolated dir; artifacts untouched):

| | sessions | distinct spans reachable |
|---|---|---|
| before | 4 | 1,266 |
| after | 9 | 1,950 |

The five new per-day sessions match the previously measured per-day drops exactly: 188 / 341 / 48 / 48 / 59. All three readers agree (`/api/data`, `/api/sessions`, `/api/llm-calls`), and `/api/sessions/background:2026-09-14` returns 200 with 59 traces.

**Five new tests, each watched failing first.** The second of the original pair matters more than it looks: with `SCHEMA` left at 2 it fails because a sidecar written by the previous build keeps hiding the spans -- the fix would have shipped and done nothing for existing archives. Of the three added this round, the reader-comparison one was run against the review's own mutant both ways (21 green before the fixture change, 2 failed after), the trace-owner one against a revert of its fix, and the undated-pool one against a build with the `undated` fallback removed.

Rebased onto `56c79edf5` and re-verified there: `git range-diff` maps the branch one-for-one, the merge commit the previous revision carried is flattened away, and the counts above are from the new base.

## Risk

- [x] Security impact considered -- no new input is parsed; the derived id is built from a span's own `startTime`, already validated upstream.
- [x] Backward compatibility considered -- the `SCHEMA` bump invalidates existing sidecars, which are a cache: they rebuild on first read, no manual cleanup. Session ids in the `background:` namespace are new values in an existing field; no consumer is keyed on the set of ids.
- [x] Rollback path is clear -- revert restores the previous behaviour; sidecars written at schema 3 are rejected by the old code and rebuilt at 2.

### Two things a reviewer should know

1. **CI does not execute these tests.** `norecursedirs` keeps `tests/integration` out of the default scope and `addopts` carries `-m "not integration and not e2e"`; `ci.yml` runs exactly one integration file (`test_cli_upgrade_real_uv.py`). The 21 tests here -- the 19 that already existed and the 2 added -- only run locally. Unchanged by this PR, but it means the regression net is manual.
2. **A known display rough edge, deliberately not fixed here.** `ui/app.js` truncates with `shortId(sessionId, 18)`, and `background:2026-09-14` is 21 chars, so the five cards read `background:2026-09...` -- distinguishable by the start/update times on the card and by the full id in the tooltip, but not by the title. Fixing it means either widening a presentation constant or shortening the id; the UI has no test harness, so it is left as a separate call rather than changed blind.

## Related Issues

N/A


LivXue:
### CI: three red checks, all pre-existing on the base commit

This PR branches from `5b7566b49`, and CI on **that exact commit** (run [34744930015](https://github.com/EverMind-AI/Raven/actions/runs/34744930015), main, 2026-09-13) already failed the same three jobs, with byte-identical errors:

| check | this PR (`b5ae4ecf3`) | base `5b7566b49` on main |
|---|---|---|
| `unit (py3.12)` | `FAILED tests/test_cli_doctor_commands.py::test_doctor_answers_where_the_memories_are` -- `assert 'mem-root' in ...` | identical test, identical assertion, identical captured output |
| `Windows self-upgrade` | `FAILED tests/integration/test_cli_upgrade_real_uv.py::test_running_uv_tool_replaces_itself_in_custom_directories` -- `ModuleNotFoundError: No module named 'pydantic'` | identical |
| `page checks` | `check-page: expected exactly two <script> blocks` | identical |

In each job the failure listed above is the only one, on both sides.

The three sit in `raven/cli` doctor output, the Windows uv self-upgrade path, and the `ui-web` page build. This PR touches `raven/cli/tracing_viewer/{server.js,shard-index.js}`, `tests/integration/test_tracing_viewer_e2e.py` and one spec file, none of which any of those three reads.

main's CI has been red since 2026-09-11 (`fcf469ae6`); the last green run was 2026-09-09 (`fd2a860d9`).

Everything that does exercise this change is green: `pre-commit diff`, `python lint`, `commit messages`, `pull request title`, `repository files`, `TUI checks`, `bridge checks`, `ui rpc contract`, `kernel wheel smoke`.

Worth repeating from the description: the 21 tests in `test_tracing_viewer_e2e.py` -- the 19 that already existed and the 2 added here -- are **not run by CI at all** (`norecursedirs` excludes `tests/integration`, `addopts` carries `-m "not integration and not e2e"`, and `ci.yml` runs only `test_cli_upgrade_real_uv.py` from that directory). They were run locally against this head: 21 passed.

gloryfromca (commented):

I reviewed `github/main...HEAD`, the surrounding normalization, identity election, whole-corpus and sharded readers, the session-detail/trace-owner/LLM-call callers, the UI's encoded session route, the span writer, and the relevant sharding history. The synthetic ID is applied only when both session identity fields are absent, both read paths use the same derivation, keyed-span election remains unchanged, and schema 3 correctly forces persisted schema-2 sidecars to rebuild. I found no backward-compatibility or ordinary-use failure.

I also checked the repository rules in `AGENTS.md`/`CLAUDE.md` and the terminology/architecture routing in `CONTEXT-MAP.md` and `CONTEXT.md`. The test filename and placement conform, no dependency or forbidden-asset change is present, source-language and large-file gates pass, and the CLI-local implementation does not cross an architecture boundary. The existing cross-reader test was updated to assert the intended newly reachable span rather than weakened, and the added cases cover both whole-corpus reachability and stale-sidecar rebuilding.

Verification on Python 3.12:
- `uv run --python 3.12 pytest tests/integration/test_tracing_viewer_e2e.py -m integration -x`: 21 passed.
- `uv run --python 3.12 pytest tests/test_cli_tracing_commands.py tests/test_tracing_api.py tests/test_tracing_artifact_v2.py tests/test_tracing_compact.py -x`: 131 passed.
- `scripts/check_large_files.py github/main..HEAD`, `scripts/check_source_language.py github/main..HEAD`, and `git diff --check github/main...HEAD`: passed.

The three current CI failure signatures were independently matched against run 34744930015 on the exact base commit: the doctor assertion, Windows missing-pydantic error, and page script-count error are all already present there and do not touch this change's paths.

gloryfromca (commented):

This head moved only by merging the now-current `main` at `4f5d7dac5`. After fetching that target, `git diff github/main...HEAD` is the same four-file feature diff reviewed at `b5ae4ecf3fbf`; the merge contains no conflict resolution or feature-side change, so the earlier clean assessment still holds.

For this revision I rechecked the full diff, repository rules in `AGENTS.md`/`CLAUDE.md`, terminology and architecture routing in `CONTEXT-MAP.md`/`CONTEXT.md`, affected normalization/index/read paths and callers, relevant history, backward compatibility, and that the tests still strengthen rather than weaken coverage. No dependency, asset, language, naming, or architecture constraint is newly violated.

Verification on Python 3.12:
- `uv run --python 3.12 pytest tests/integration/test_tracing_viewer_e2e.py -m integration -x`: 21 passed.
- `uv run --python 3.12 pytest tests/test_cli_tracing_commands.py tests/test_tracing_api.py tests/test_tracing_artifact_v2.py tests/test_tracing_compact.py -x`: 131 passed.
- `uv run python scripts/check_large_files.py github/main..HEAD`, `uv run python scripts/check_source_language.py github/main..HEAD`, and `git diff --check github/main...HEAD`: passed.

The target-branch fixes for the three previously red jobs are incorporated by the merge; on this head Windows self-upgrade and page checks are green, with the unit job still running when reviewed.

tests/integration/test_tracing_viewer_e2e.py
0xKT, line 1: The PR's headline property -- the two readers derive the same background
session id -- is pinned by no test. A drift between them leaves the whole
suite green while the panel breaks.

The code as written is correct: both sides call the one exported
`backgroundSessionId`. This is a hole in the guard, not a bug in the feature.

WHAT I RAN, in a worktree at head f78653ee merged onto 09a5f7f8, restoring
the tree between each run (`git status --porcelain` empty afterwards):

  0. BASELINE, untouched
     bash board/pyt.sh tests/integration/test_tracing_viewer_e2e.py -m integration
     -> 21 passed

  1. CONTROL -- revert the whole-corpus derivation to the base line
     server.js:216  ->  sessionId: attrs['session.id'] || null,
     -> 2 failed, 19 passed
        FAILED test_llm_calls_match_the_whole_corpus_reader
        FAILED test_a_span_with_no_session_is_still_reachable
     So the suite does reach that expression. A mutant surviving below is
     not a selection that never runs.

  2. DRIFT MUTANT -- keep the feature, change only the GRAIN on one side
     server.js:216  shardIndex.backgroundSessionId(span.startTime)
                 -> shardIndex.backgroundSessionId(String(span.startTime || '').slice(0, 7))
     -> 21 passed        <-- the two readers now disagree, suite fully green

The join that breaks is server.js:894: buildSessionDetail takes the row from
the shard derivation and then filters spans by `span.sessionId === sessionId`
from the whole-corpus derivation. Spell the id differently and the row is
listed with a span count while its detail page holds nothing.

WHY NO TEST SEES IT. The two tests that compare the readers,
test_session_list_matches_the_whole_corpus_reader (:338) and
test_session_detail_matches_the_whole_corpus_reader (:353), both run on
_sessions_across_two_files (:269). Every span in that fixture is built by
_span() (:71-85), which always writes session.id and session.key, so the
fixture contains no sessionless span and no background session is ever
constructed in either comparison. The three tests that do involve a
sessionless span assert a shape instead of comparing:
test_a_span_with_no_session_is_still_reachable (:634) reads only /api/data,
test_a_sidecar_from_the_previous_schema_does_not_hide_background_spans (:654)
reads only /api/sessions, and test_llm_calls_match_the_whole_corpus_reader
(:400) compares the SET OF spanIds (:413-421), never the sessionId each
reader assigned.

That fixture's own docstring is the rule this misses:

    Carries events of both shapes on purpose. One is attributed by session
    id, the other only by session key -- and the key-only shape is the one a
    reader can drop silently, which is exactly what happened until a review
    caught it.

This PR adds a THIRD shape a reader can drop or spell differently, and does
not add it to the fixture built to carry exactly those shapes. The file
header (:25-29) says the same thing directly: "the disagreement is silent, so
the tests below compare them rather than asserting a shape."

THE FIX, WHICH I RAN BOTH WAYS. One sessionless span in the shared fixture,
in the active log alongside the existing three:

    sessionless = _span("", "span-sessionless", start="2026-08-01T00:00:00+00:00")
    sessionless["attributes"] = {"span.type": "tool"}
    _write_spans(
        logs / "audit-spans.log",
        [late, _span("other-session", "span-other"), _span("third-session", "span-third"),
         sessionless],
    )

  against the PR's own code   -> 21 passed        (not a false red)
  against the drift mutant    -> 2 failed, 19 passed
       FAILED test_session_list_matches_the_whole_corpus_reader
       FAILED test_session_detail_matches_the_whole_corpus_reader (HTTP 404)

So one line in the fixture turns the invariant the PR is built on into
something the suite holds.

SECOND UNCOVERED BRANCH, same class: the `undated` fallback at
shard-index.js:85 is reached by no test. It cannot fire on this machine's
store -- I scanned 75,488 records across 16 files and found 0 sessionless
traces with an unusable startTime -- but it is the branch that would pool
every undated background span into one session if it ever did.

INTRODUCED BY THIS PR, not pre-existing: at 09a5f7f8, shard-index.js:111 is
`if (!id && !key) continue;` and server.js:212 is
`sessionId: attrs['session.id'] || null,`. No span could reach either reader
without the shared election, so there was no second derivation to compare.

Not blocking: the shipped code is correct, so nothing is broken for a user
today. Filed so the invariant the feature rests on gets a guard while the
feature is still fresh -- a drift introduced later would be silent.

raven/cli/tracing_viewer/shard-index.js
0xKT, line 1: The day is derived per SPAN, not per trace. Two consequences, both measured
on this machine's real store; one is live today, the other is latent.

The grain itself is argued in the code, and I agree with the argument:
"a day is the grain because per-trace would shatter a year of timer ticks
into a session each" (shard-index.js:79-80). Neither item below asks for a
different grain -- they are the two edges that choice leaves, and the design
note this PR adds mentions neither.

(1) LIVE TODAY: the dashboard's opening row is no longer the user's most
    recent conversation.

    buildSessionList sorts by updatedAt descending (server.js:870) and the
    page selects sessions[0] when nothing is selected (ui/app.js:1808,
    `sessions.find(...) || sessions[0]`). Neither line is changed by this PR;
    the PR changes their input.

    Measured read-only over ~/.raven/traces, 16 files, 75,488 records:

      newest span with a session    2026-09-07T05:04:51  (session cli:c)
      newest span with NO session   2026-09-09T17:33:33

    The newest record in the store is sessionless, and now it has a row. So
    `background:2026-09-09` takes slot 0 and the panel opens on a background
    day instead of the conversation it opened on before. Sessionless spans
    are emitted continuously -- in this store 12,230 llm.call, 6,544
    tool.call, 4,696 session.turn, 3,969 memory.recall, 3,676 plugin.load --
    so this is the steady state, not a one-off.

    No rule forbids it and a user can click past it. Raising it because it
    is a user-visible default nobody chose, and no test covers row ordering.

(2) LATENT: a sessionless trace that straddles UTC midnight is split into
    two sessions, and the trace-owner jump resolves it to one half.

    Each span gets its own `startTime.slice(0, 10)`, so a parent and child of
    one trace on opposite sides of midnight land in different
    `background:<day>` rows. findTraceOwner (server.js:985-991) returns on
    the FIRST row whose visibleTraceIds holds the id, so the subagent-run to
    parent-turn jump the page makes (ui/app.js:809, :822) reaches one half
    and the other becomes unreachable from it.

    The rule is written on the sibling path in the same file
    (shard-index.js:345-347):

        Union across files, not a sum: one trace can straddle a rotation, and
        it is visible if any of its spans is. Used to decide whether a
        session has anything worth showing, and to find which session owns a
        trace id.

    That union exists so a straddling trace keeps ONE owning session. The
    per-span day grain re-introduces the split on a different axis.

    Incidence today is zero, and I measured it rather than assuming:

      distinct sessionless traceIds                 31,299
      of those, spanning more than one day               0
      of those, landing in the `undated` bucket          0
      sessionless spans starting in the 23:00 or
        00:00 UTC hour                               1,395

    So nothing is broken now. The window is narrow but it is real, and it
    widens with any sessionless work that runs longer than a timer tick.

INTRODUCED BY THIS PR for both: at 09a5f7f8, shard-index.js:111 is
`if (!id && !key) continue;` and server.js:212 is
`sessionId: attrs['session.id'] || null,`. A sessionless trace was dropped
whole by both readers, so it had no row to sort into slot 0 and no pair of
rows to be split across. Nothing regressed; the new feature is half-delivered
for this shape.

Not blocking either way.

tests/integration/test_tracing_viewer_e2e.py
LivXue: Fixed in 13b9341c6 (the fixture) and bb7abf642 (the undated pool). I reproduced your experiment before taking it, and it lands exactly as you describe:

  mutant + old fixture      21 passed              <- the suite is blind to the drift
  mutant + new fixture      2 failed, 19 passed
        FAILED test_session_list_matches_the_whole_corpus_reader
        FAILED test_session_detail_matches_the_whole_corpus_reader
  no mutant + new fixture   21 passed              <- not a false red

One deviation from your patch, same semantics. The file already carries a `_sessionless_span()` helper at :630 whose body is your two lines -- `_span` with the id argument dead, then `attributes` replaced wholesale -- so the fixture calls that instead of open-coding it. It writes `name="llm.call"` / `span.type="model"` where yours wrote `tool.call` / `tool`; neither is in `INVISIBLE_SPAN_NAMES`, so the span is visible either way. Because the shape is mine rather than yours, the three runs above are against my version. I did not carry your numbers forward.

The fixture docstring gains the third shape. That docstring is the rule the omission broke, so leaving it enumerating two would set the next reader up the same way.

On the second uncovered branch: I closed it rather than leaving it. The `undated` fallback is reachable rather than defensive -- `normalizeSpan` passes `startTime` through as it finds it, so a record without one reaches the derivation -- and a test now asserts the pool exists and the span arrives through it. Removing the `|| 'undated'` reddens exactly that test and nothing else. You are right that no store holds one; none on mine either, which is why the test states that reachability is what it pins.

raven/cli/tracing_viewer/shard-index.js
LivXue: Both taken, one commit each: 9fad1e393 for the opening row, 32bf567b1 for the trace owner.

(1) The default. The row now declares `isBackground` and the page prefers the first row that is not one, rather than matching the `background:` prefix in `ui/app.js` -- the prefix is the shard index's to mint, so recognizing it belongs beside the minting. Both readers carry the flag, and the field-for-field comparison between a list row and a whole-corpus session is what then holds them together: dropping it from `buildSessionList` alone reddens that comparison and the new test, which is how I checked it was pinned rather than merely present.

Disclosed: `ui/app.js` has no test harness, so the preference itself is code-read only. What is tested is the flag on both readers, and the ordering condition it exists to work around.

(2) The straddle. I did not change the grain, for the reason your own note gives it, and I measured before deciding. On my store 770 sessionless spans carry 770 distinct trace ids -- one span each -- so deriving the day per trace means a bucket per trace: one file's pair count goes from 3 to 341. That is the shattering the day grain was chosen to avoid.

What I fixed instead is the single-owner rule your note cites from shard-index.js:345-347. `findTraceOwner` returned whichever row iteration reached first; it now returns the earliest owner, the half that started the trace. The test forces the state rather than waiting for it -- it writes the later half first, because the previous code answers correctly whenever the earlier half happens to be indexed first. Reverting the fix makes it report `background:2026-08-02` where it wants `background:2026-08-01`.

What that leaves, stated plainly: the spans stay split across the two day rows, so the owner's detail shows the earlier half only. The jump lands somewhere stable instead of somewhere arbitrary; it does not reunite the trace. Reuniting it is the grain change, and I did not judge the measured incidence -- zero straddling traces on either of our stores -- to justify that cost. Say so if you read it the other way and I will take it.

gloryfromca (commented):

Since my last review, the feature was rebased onto current `main` and gained four focused follow-ups. The shared comparison fixture now carries a sessionless span, so both the session-list and detail comparisons exercise the background-id derivation. Background rows now carry an additive `isBackground` field on both readers, and initial UI selection prefers a real conversation while preserving an explicit background selection and retaining the all-background fallback. Cross-midnight trace ownership now chooses the earliest matching day deterministically without changing the documented per-day grouping. The final commit reaches and asserts the `background:undated` fallback. These changes address both open review threads in code; those threads were opened by another reviewer, so I have left resolution to them.

I reviewed the delta from `f78653eec0c6`, the complete `github/main...HEAD` diff, affected UI and API callers, identity/index behavior, relevant history, backward compatibility, and whether the new tests could pass while the intended paths stayed unexercised. I also rechecked `AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md`/`CONTEXT.md`, test naming and placement, source language, dependencies, assets, and architecture boundaries. I found no plain error or new constraint violation.

Verification at `bb7abf642427` on Python 3.12:
- `uv run --python 3.12 pytest tests/integration/test_tracing_viewer_e2e.py -m integration -x`: 24 passed.
- `uv run --python 3.12 pytest tests/test_cli_tracing_commands.py tests/test_tracing_api.py tests/test_tracing_artifact_v2.py tests/test_tracing_compact.py -x`: 131 passed.
- `node --check raven/cli/tracing_viewer/ui/app.js`: passed.
- `uv run python scripts/check_large_files.py github/main..HEAD`, `uv run python scripts/check_source_language.py github/main..HEAD`, and `git diff --check github/main...HEAD`: passed.

An intermediate run on `32bf567b1ce7` also passed all 23 integration tests but emitted four warning-only wall-clock overruns on unchanged tests; the final 24-test run above emitted no such warnings. CI was still running when reviewed.

tests/integration/test_tracing_viewer_e2e.py
0xKT: Verified at bb7abf64. I re-ran my own mutant against your fixture rather than
reading the matrix you posted, and it lands where you said.

  drift mutant (server.js side to the month grain)
     before your fix  21 passed          <- the suite was blind to it
     now              3 failed, 21 passed
        test_session_list_matches_the_whole_corpus_reader
        test_session_detail_matches_the_whole_corpus_reader
        test_a_background_row_says_so_on_both_readers

  remove the `|| 'undated'` fallback
     1 failed, 23 passed
        test_a_sessionless_span_with_no_start_time_is_still_reachable

So both halves are closed, and the second one reddens exactly the test it
should and nothing else.

Your deviation is better than my patch and I would not want it back. Reusing
`_sessionless_span()` instead of open-coding the two lines means the fixture
and the dedicated tests build the same shape, and `llm.call` / `span.type=model`
carries the same visibility as my `tool.call` did. Not carrying my numbers
forward was the right call too -- they were measured against a different span.

Adding the third shape to the fixture docstring is the part I would have
missed. That docstring is the rule my finding cited, and leaving it
enumerating two would have set the next reader up exactly as this one was.

Closing.

raven/cli/tracing_viewer/shard-index.js
0xKT: Verified at bb7abf64, both halves, by mutating each fix rather than taking
the claim.

  (1) drop `isBackground` from buildSessionList only, leave the other reader
      -> 2 failed, 22 passed
         test_session_list_matches_the_whole_corpus_reader
         test_a_background_row_says_so_on_both_readers

  (2) findTraceOwner back to returning the first row iteration reaches
      -> 1 failed, 23 passed
         test_a_sessionless_trace_across_midnight_has_one_owner

Both bite, and each reddens only what it should.

On the default: putting the flag beside the minting rather than matching the
`background:` prefix in the page is the right seat, for the reason you give --
the prefix is the shard index's to mint. Carrying it on both readers is what
lets the field-for-field comparison hold it, which is the difference between
present and pinned. Your disclosure that the preference itself in ui/app.js is
code-read only is noted and I am not asking for more; there is no harness
there to ask it of.

On the straddle: not changing the grain is the right decision and you measured
before making it. 770 sessionless spans carrying 770 distinct trace ids means
a per-trace day would shatter one file's pairs from 3 to 341, which is the
cost the day grain was chosen to avoid -- my note argued the same direction
and did not ask for the grain to move. Fixing the single-owner rule instead is
the narrower change that actually closes what I raised.

The test forcing the state by writing the later half first is the part worth
naming. The old code answers correctly whenever the earlier half happens to be
indexed first, so a test that did not force the order would have passed
against the bug.

Closing.

0xKT:
Not a blocker and not a finding -- an edge in the new owner rule, for
whenever that code is next open. I am recording it rather than filing it
because I cannot show it is reachable on a real store.

`findTraceOwner` now picks the earliest owner:

    if (!owner || (row.startedAt || '') < (owner.startedAt || '')) owner = row;

`row.startedAt` starts as `null` (shard-index.js:339) and is only set from
`pair.minTime` (:348), so a row whose spans carry no parseable time keeps it
null. `null || ''` is the empty string, which sorts before every ISO stamp,
so such a row wins the comparison:

    rows: background:undated    startedAt=null
          background:2026-09-16 startedAt=2026-09-16T00:00:01+00:00
    picked: background:undated

A row with no start time is not the half that started the trace; it is the
half whose start is unknown. The comment above the loop says the rule wants
"the half that started it", and this is the one input where the code and that
sentence part ways.

Why I am not filing it: it needs one trace with a span in `background:undated`
and another in a dated background day. When I measured your store's shape
during the first pass -- 31,299 distinct sessionless trace ids across 16 files
-- zero landed in the undated pool and zero straddled a day at all. So I can
name the input but not show it happening.

Treating a null as "later than everything" rather than "earlier" would close
it, and the test you already added for the straddle is the place it would go.
Entirely your call whether it is worth the line.

0xKT (approved):

PR #408, opened 2026-09-14
feat(subagent): let a person put one instance on a model of their own

Handsome-wzw:
## Summary

A sub-agent pane could say how hard an instance tries and not what does the
trying. The effort chip has had a model-shaped hole beside it since it shipped,
and closing it turns out to need no new protocol: every acp agent this host
drives already advertises its whole model menu on `session/new`, and already
accepts a switch. Raven was reading neither.

**The read was looking at the wrong key.** `models.availableModels` is older
material that none of these agents send. The stable surface is one
`configOptions` entry carrying `category: "model"` -- which is what raven's own
ACP server has served all along. A survey of live handshake journals came back
empty against the old key while every one of them carried a full menu under the
new one, grouped by provider with short display names beside long
provider-qualified ids.

**The write did not exist at all.** `session/set_model` is not in the stable
schema, so the channel is `session/set_config_option` with `configId: "model"` --
the same method raven answers as a server, which is what makes this contract
symmetric rather than guesswork. It is pushed on every route into a session, not
only on creation, for the reason `_set_mode` records beside it: the agent binds
the choice to a session id it holds in memory, and the pool relaunches that
process whenever the launch key changes.

The override is held per instance on the manager, keyed and unpersisted exactly
as a mode is: which model answers a conversation is a judgement made while
having it, not a property of the record.

### What is deliberately not copied from the mode chip

`inherited`. A cleared mode falls through to the session's tier, which this host
chose and can therefore name, so that chip draws the inherited rung. A cleared
model falls through to whatever the agent picked for itself, which this host
cannot see: the `currentValue` an agent reports belongs to the throwaway session
a capability probe opened, not to this instance. `_read_session_modes` already
states the rule for its own `currentModeId` and the model reader follows it --
the probe's current value is not recorded, `null` means "the agent's own", and
nothing is invented. Drawing the probe's value would put a model on screen this
conversation may never have run on.

### Scope

Builtin rows are out, and stay out here. Their model is baked into the backend at
`build_builtin_backend(row, build)`, which receives no session and no handle, and
`raven/providers/binding.py` `resolve` takes a `ModelBinding` rather than an id
precisely because "a model id without a credential is what mis-pairs one
vendor's key with another's endpoint". That is a different change with a
different hazard, and pretending otherwise would put a control on a row where it
cannot work.

expected exactly two <script> blocks`.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/ (targeted: rpc, subagent, acp, registry, backend, manager)   5838 pass
pytest tests/test_rpc_schema_match.py tests/test_rpc_instances_declare_their_errors.py   348 pass
cd ui-web && npx tsc --noEmit -p .        clean
cd ui-web && npm run gen:check            generated.ts matches the contract (169 methods)
cd ui-web && npx vitest run               1688 tests, all pass
cd ui-tui && npx tsc --noEmit -p .        clean
cd ui-tui && node scripts/gen-i18n.mjs --check   catalogue up to date
```

`tests/test_rpc_schema_match.py`), and `ui-web/src/rpc/generated.ts` is
regenerated from it (held equal by `npm run gen:check`).

25 new cases across four layers, every one mutation-checked by reverting the
line it covers:

Reader and transport:

- the stub agent gained the real `configOptions` shape, including a
  non-model entry beside the model one and a choice with no value; the snapshot
  case asserts neither becomes a menu item
- send the non-stable `session/set_model` -> red
- never push on a new session -> red
- push when no model was asked for -> red

Manager:

- accept any invented id -> red (two cases)
- key the override per agent instead of per instance -> red

RPC:

- drop the ghost-handle guard -> red
- let a write degrade to empty instead of refusing -> red

Chip:

- no empty-menu guard -> red
- label from the raw id instead of the short name -> red (three cases)
- send the label instead of the value -> red
- drop the group separators -> red

One test was weak and was strengthened rather than kept: the stub logged
`session/set_config_option` only on success, so a case asserting that *no* frame
was sent passed against a mutation that sent an unrequested one. The stub now
logs on entry.

No user-facing doc describes the pane header's controls.

## Risk

Security: no credential handling is added or moved. An acp agent owns its own
provider and its own keys; raven sends it an id the agent itself advertised, and
the agent refuses anything it will not write. This is exactly why the builtin
half is out of scope -- that one would touch `ModelBinding`, and this does not.

Backward compatibility: every new field is optional and every new verb is
optional on its source. An agent that advertises no model option gets no chip and
no frame; a server too old to answer `subagents.instance.set_model` makes the
chip draw nothing, which is the same state as "no menu" and already handled. The
new `session_model` argument travels through `optional_keyword`, so a backend
that never declared it is not handed it -- the idiom this repo already uses for
exactly that. `AgentMeta` gained its field last, after `owns_watched_work`,
because several callers build it positionally and its own docstring says so.

Rollback: two commits, both revert cleanly. Nothing outside the sub-agent
surface reads any of it.

Known gap, deliberately not closed: a capability snapshot cached before this
change carries no `modelChoices`, so an agent probed under an older build shows
no menu until its snapshot is refreshed. That is the existing staleness path
(`snapshot_fingerprint` digests the launch config, not the advertised menu) and
the same is already true of `available_modes`; closing it for models alone would
leave the two disagreeing.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the full `github/main...HEAD` diff, the ACP capability/session path, all manager dispatch callers, routing and DAG dispatch, the RPC wire schema and generated client, the web control, relevant history, backward-compatible optional keyword handling, test changes, and the applicable AGENTS.md/CONTEXT.md architecture and source/asset rules. I found no merge blocker and no weakened test.

Named nonblocking follow-up (kept here rather than as an open thread): `raven/agent/subagent/dag_runner.py` continues a named instance with `mode_for`, but has no corresponding model resolver and sends no `session_model` to `agent_backend.run`. A model selected in the instance pane therefore applies to direct-chat and spawn continuations, but a DAG node reusing the same handle runs with the ACP agent's own model. This change introduces the gap, but at this fourth-round bar it is not blocking because the instance remains usable on the selected model through direct chat or spawn.

Verification: `uv run pytest tests/test_rpc_instances.py tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_rpc_schema_match.py -x` (790 passed); `npm test -- --run src/features/subagents/InstanceModel.test.tsx` (10 passed); `npm run type-check` (passed); `npm run gen:check` (passed, 169 methods); direct invocations of `scripts/check_source_language.py github/main..HEAD` and `scripts/check_large_files.py github/main..HEAD` (passed). The Make wrapper itself was unavailable in this environment (`make: command not found`), so I ran the exact underlying checks from the Makefile.

raven/acp_client/acp_agent.py
```diff
@@ -1517,8 +1523,54 @@ async def request_or_busy(method: str, params: dict[str, Any]) -> Any:
...
+        ``session/set_model`` is not in the stable schema, and an agent waiting
+        for it would never be asked to switch.
+
+        Never fatal, like the mode beside it. An agent that offers no such option
+        answers invalid-params and one that will not take the value answers with
+        its own code; either way the task still runs on the agent's own model,
+        and failing the run would be a worse outcome than running it on a model
+        the caller did not pick.
+        """
+        if not model:
```
gloryfromca: Blocking: clearing an instance model override does not restore the ACP session to the agent's own model.

After an operator selects model B, `session/set_config_option` binds the remote ACP session to B. Selecting `Agent's own` removes only the manager's `_instance_models` entry; the next turn passes `session_model=None`, and this early return sends no reset to the resumed session. The RPC response and chip therefore say `model: null` / `Agent's own` while the agent continues answering on B. This is the new control's ordinary set-then-clear path, and the same conversation has no way back to an agent-controlled choice. Please make clear restore the advertised remote behavior and cover set -> clear -> next turn in a regression test.

gloryfromca (commented):
Blocking: clearing an instance model override does not restore the ACP session to the agent's own model.

The fourth-round blocker bar is met: this behavior is introduced by this change, the set-then-Agent's-own path is ordinary use of the new control, and there is no working way to return that same conversation to an agent-controlled choice.

I covered the repository rules in AGENTS.md/CLAUDE.md and the relevant Runtime/TUI context vocabulary; the complete `github/main...HEAD` diff and the delta from `d18b1607b8ad`; manager, ACP, RPC, generated-client, and web callers; commit history; backward compatibility; test additions/deletions; and the RPC-only UI architecture boundary. The new revision correctly restores the missing TUI generated client and only reformats the Python lines it says it does.

Verification:
- `uv run pytest tests/test_rpc_instances.py tests/test_subagent_acp.py tests/test_subagent_manager.py -q`: 444 passed.
- `uv run pytest tests/test_rpc_schema_match.py -q -o addopts=''`: 346 passed.
- `npm test --prefix ui-web -- --run src/features/subagents/InstanceModel.test.tsx`: 10 passed.
- Web RPC generation check and type-check: passed.
- TUI RPC and i18n generation checks and type-check: passed after installing its locked local dependencies.
- Ruff check and format check for the changed Python files: passed.
- Source-language check for `github/main...HEAD`: passed.

Handsome-wzw:
Confirmed, and it was worse than one early return - two guards hid it together. Fixed in `12301577`.

`optional_keyword` drops a `None` before it ever leaves the manager, and `_set_model` returned early on a falsy model. So the clear never became a frame at any layer: the remote session kept answering on B while the RPC, the chip and the reader all said the agent's own. The one surface that mattered was the wrong one.

**The way back is the session's own baseline** - the value it reported before this host first moved it. Both routes in already carry it (`session/new` and `session/load` each answer with the model option), so it is read where the modes beside it already are.

**Recorded once per session and never overwritten**, which is the subtle half: a resumed session reports what it is *currently* on, and after a switch that is the override. Adopting it as the baseline would restore the very thing the clear meant to undo. That rule was untestable until the stub started keeping a model per session and reporting it on resume - which is also what a real agent does - so the stub now does both.

Sent only when this host actually moved the session, so one it never touched still costs no frame on any dispatch. When the agent named no earlier model there is nothing to restore to; that is logged rather than passed over, because the reader has just asked for something this cannot deliver.

Two new cases, both mutation-checked:

- restore the old early return -> `set -> clear -> next turn` goes red with only the switch frame present
- let the baseline be overwritten on resume -> red, now that the stub reports per-session state
- and `an untouched session is never reset` guards the other direction, so the fix cannot be "always push".

`pytest` over acp, manager, instances and schema-match: 792 passed. `ruff check` and `ruff format --check` both pass. Web and TUI generation checks and type-checks pass.

For the record on your earlier round: the TUI generated client and the formatter were both real, and both were gates I had run only half of - `make lint-python` is check AND format, and `openrpc.json` feeds two generated clients, not one.

raven/acp_client/acp_agent.py
gloryfromca: Blocking: two remaining model-state paths still need correction.

The original set -> clear defect is fixed on the successful ACP path: this revision records the session's pre-switch model and sends it back before the next prompt. The new regression test covers that sequence, so this thread is settled. The blocking items I found in this revision are separate and will be posted with the review.

raven/agent/subagent/manager.py
```diff
@@ -320,6 +320,9 @@ def __init__(
         # ``set_instance_mode`` for why it is not persisted, and why the host
         # rather than the agent is what holds it.
         self._instance_modes: dict[tuple[str, str, str], str] = {}
+        # And which model, by the same key and on the same terms. Separate from
+        # the modes above because an agent can offer either without the other.
+        self._instance_models: dict[tuple[str, str, str], str] = {}
```
gloryfromca, line 325: Blocking: forgetting an instance must also discard its model override.

`instances_forget` clears `_instance_modes` through `set_instance_mode(..., None)`, but nothing clears this new map. Because the key includes the reusable `(session_key, agent, handle)` triple, forgetting an instance after selecting model B and later creating the same handle makes the replacement silently inherit B; its next dispatch pushes a choice the operator made for the deleted instance. Please clear both the named and paired handles' model overrides alongside their mode overrides and cover that lifecycle in the existing forget regression area.

raven/acp_client/acp_agent.py
```diff
@@ -1516,9 +1549,90 @@ async def request_or_busy(method: str, params: dict[str, Any]) -> Any:
...
+                self._model_pushed.pop(session_id, None)
+                return
+        try:
+            await client.request(
+                "session/set_config_option",
+                {"sessionId": session_id, "configId": "model", "value": target},
+                timeout=budget,
+            )
+        except AcpRemoteError as exc:
+            logger.warning(
```
gloryfromca, line 1613: Blocking: a rejected ACP model update must not be recorded as successful.

This `AcpRemoteError` branch logs but does not return, so execution reaches the success-only bookkeeping below. On a rejected clear, the remote session remains on model B while `_model_pushed` is erased; every later no-override turn then returns early and never retries, recreating the control/session mismatch this revision is meant to fix. A rejected set likewise records a model the agent never accepted. Keep the bookkeeping unchanged on refusal (as the `AcpError` branch already does) and add a clear-refusal regression.

gloryfromca (commented):
Blocking: forgetting and refused-reset paths must preserve truthful per-instance model state.

The successful set -> clear path now restores the ACP session baseline, so the previous thread is fixed and resolved. Two blocking state-lifecycle errors remain inline.

I reviewed the full `github/main...HEAD` diff, the manager/backend/RPC callers, relevant history, backward-compatible tuple placement, the ACP and RPC wire contracts, generated clients, UI behavior, and the repository rules in `AGENTS.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`. I found no weakened tests. Verification: `uv run pytest tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_rpc_instances.py -x` (446 passed); `uv run pytest tests/test_rpc_schema_match.py tests/test_acp_schema.py -x` (364 passed); changed-file Ruff check and format check passed; the web component test (10 passed), web type-check and RPC generation check passed; TUI RPC/i18n generation checks and type-check passed. (`make lint-python` could not run because `make` is unavailable, so I ran its relevant Ruff checks directly.)

Handsome-wzw:
Both confirmed in the code before I touched either, and both fixed in `8f682973`.

**The refused-reset one is the one that stung**, because the comment sitting directly above that bookkeeping said "recorded only once the agent has taken it" — and the typed refusal arm fell straight through to it. So the code did the opposite of what it claimed, on the likeliest refusal there is. Reproduced with a stub mode that takes a switch and refuses the way back: before the fix the restore was attempted once and never again, exactly as you described; after it, the third turn retries. I added the mirror too — a refused SET must not be recorded as a move, or a later clear sends a restore to undo something that never happened.

**Forgetting**: your reading is right and the handler had already written the argument for me. Its docstring says the key is reusable so anything left behind lands on whoever gets the name next — written for the mode, true word for word for the model. Both handles, named and paired, now clear both overrides.

That one surfaced a test-side problem worth naming: the existing `test_forgetting_an_instance_drops_its_mode_override` failed with `AttributeError` the moment the handler made a second call, because its fake manager had only `set_instance_mode`. I put both setters on the base fake rather than patching the one subclass — a fake missing a method the real object has is not a smaller manager, it is a blind spot, and here it failed against the fake while it would have passed against the real thing, which is the wrong way round.

Four new cases, each mutation-checked:

- let the typed refusal fall through again -> 2 red
- stop clearing the model on forget -> 2 red
- leave the paired handle behind -> 1 red

`pytest` over acp, manager, instances, schema-match and the error-declaration gate: 798 passed. `ruff check` and `ruff format --check` both pass. `ui-web` 1688 tests, type-check and `gen:check` pass; TUI rpc/i18n checks and type-check pass.

Still true and still worth repeating: none of this has been seen on a real agent. The gateway on this machine predates the server change, so confirming the chip end to end needs a restart I have not taken unprompted.

raven/acp_client/acp_agent.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in `8f682973`: both ACP error arms now return before the success-only bookkeeping. The regressions prove a refused restore remains eligible for retry and a refused switch is not recorded as a move, so this thread is settled.

raven/agent/subagent/manager.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in `8f682973`: `instances_forget` now clears the model override alongside the mode override for both the named handle and its paired graph handle. The two added regressions cover both shapes, so this thread is settled.

gloryfromca (commented):

`8f682973` fixes both outstanding state-lifecycle defects: refused ACP updates no longer alter success bookkeeping, and forgetting an instance clears model overrides for the named and paired handles. The regressions cover refused set, refused restore with retry, ordinary forget, and paired-handle forget. I replied to and resolved both prior blocking threads.

I reviewed the current full diff and the delta from `12301577`, including the backend and manager callers, history, backward compatibility, ACP/RPC wire boundaries, generated clients, UI behavior, and the repository rules in `AGENTS.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`; I found no weakened tests or new error worth raising. Verification: `uv run pytest tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_rpc_instances.py tests/test_rpc_schema_match.py tests/test_acp_schema.py -x` (814 passed); changed-file Ruff check and format check passed; the web component test (10 passed), web type-check and RPC generation check passed; TUI RPC/i18n generation checks and type-check passed.

gloryfromca (commented):

This revision rebases the already-reviewed feature onto the newer `main`. The range diff shows no behavioral change to the model-picker implementation or its two lifecycle fixes; the only feature-adjacent integration changes are updated generated catalogs/clients and upstream ACP test timing. The current full diff still preserves the manager/backend/RPC boundaries, backward-compatible tuple placement, ACP wire behavior, and resolved failure paths. All three prior finding threads remain resolved.

I covered the current full diff, rebase range diff, callers and history, backward compatibility, generated wire artifacts, UI integration, test changes, and the repository rules in `AGENTS.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`; I found no weakened tests or new error worth raising. Verification: `uv run pytest tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_rpc_instances.py tests/test_rpc_schema_match.py tests/test_acp_schema.py -x` (822 passed); changed-file Ruff check and format check passed; the web component test (10 passed), web type-check and RPC generation check passed; TUI RPC/i18n generation checks and type-check passed.

raven/agent/subagent/registry.py
0xKT, line 1: Non-blocking: `model_choices` can be dropped from either row/meta conversion and no test notices, while `modes` in the same two places is pinned.

This PR's own docstring on `AgentRow.meta()` states the rule:

> `model_choices` travels even though `format_agent_listing` does not render it, so a meta built from a row and one built from a config carry the same facts -- a field that is always empty on one of the two paths is one that rots.

Nothing enforces that. Measured on `8866f6af`, merged onto `2f631a6f`, over `tests/test_rpc_instances.py tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_subagent_registry.py`:

| withdrawn | result |
| --- | --- |
| `model_choices` on both conversions (`AgentRow.meta()` and `_row_for`) | 461 passed |
| `modes` on the same two conversions | 1 failed, 460 passed |

The control is the point: the failure for `modes` is `test_subagent_acp.py::test_the_relearned_menu_reaches_the_roster_the_clamp_and_picker_read`, so the round trip is a thing this suite already knows how to check. The new field just has no equivalent.

Why it is worth a test rather than nothing: the field is invisible by design. `format_agent_listing` deliberately does not render it, so the only reader is the RPC menu, and the pane drops the chip entirely when the menu is empty (`InstanceModel.tsx`, `if (!held || !held.menu.length) return null`). A later edit that loses the field on one path takes the whole control off the screen with every test still green -- which is exactly the rot the docstring names.

Not blocking, and not a request to restructure anything: the parallel with `modes` is sound and the rest of the wiring holds up. One test asserting a row built from a config and the meta built back off that row carry the same `model_choices` would close it.

Two things I checked that are NOT findings, recorded so they are not re-raised: removing the `session_model` push at `manager.py` direct-chat call site also leaves the suite green, but so does removing the sibling `mode=` argument at that same call site -- that site is untested for both, which predates this PR, so it is not yours. And the ui-web side does bite: withdrawing the empty-menu guard or the name-vs-value split each turns its tests red.

0xKT:
**Not a blocker** -- a note that does not block.

Body.

gloryfromca (commented):

This is another pure rebase of the five reviewed feature commits. The range diff is patch-equivalent except for the generated web RPC method count moving from 173 to 177 as `main` gained methods; the model picker and all lifecycle fixes are unchanged. The existing `model_choices` row/meta round-trip coverage gap raised by @0xKT remains a valid non-blocking follow-up. I did not resolve that thread because it belongs to its author. All three threads I opened remain resolved.

I covered the current full diff, rebase range diff, callers and history, backward compatibility, generated wire artifacts, UI integration, test changes, and the repository rules in `AGENTS.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`; I found no weakened tests or merge-blocking error. Verification: `uv run pytest tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_rpc_instances.py tests/test_rpc_schema_match.py tests/test_acp_schema.py -x` (830 passed); changed-file Ruff check and format check passed; the web component test (10 passed), web type-check and RPC generation check passed; TUI RPC/i18n generation checks and type-check passed.

gloryfromca (commented):

After fetching the current target branch, the range diff shows all five feature commits are exactly patch-equivalent to `92597947`; this is solely a rebase over newer dependency and UI changes on `main`. The existing non-blocking `model_choices` row/meta round-trip test gap remains the only follow-up, in @0xKT's thread. All threads I opened remain resolved.

I covered the current full diff, exact range diff, callers and history, backward compatibility, generated wire artifacts, UI integration, test changes, and the repository rules in `AGENTS.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`; I found no weakened tests or merge-blocking error. Verification: `uv run pytest tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_rpc_instances.py tests/test_rpc_schema_match.py tests/test_acp_schema.py -x` (830 passed); changed-file Ruff check and format check passed; the web component test (10 passed), web type-check and RPC generation check passed; TUI RPC/i18n generation checks and type-check passed.

gloryfromca (commented):

This revision rebases the reviewed feature over the instance turn-clock and dependency updates. The only manual feature-adjacent integration keeps the live pane controls ordered as clock, model, then mode; the generated web client advances to 178 methods. The current target has one later, unrelated web-search commit, and `git merge-tree --write-tree HEAD github/main` confirms a clean merge. The existing non-blocking `model_choices` row/meta round-trip test gap remains the only follow-up, in @0xKT's thread. All threads I opened remain resolved.

I covered the current full diff, rebase range diff, callers and history, backward compatibility, generated wire artifacts, pane integration, test changes, current-target compatibility, and the repository rules in `AGENTS.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`; I found no weakened tests or merge-blocking error. Verification: `uv run pytest tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_rpc_instances.py tests/test_rpc_schema_match.py tests/test_acp_schema.py -x` (834 passed); changed-file Ruff check and format check passed; the web component test (10 passed), web type-check and RPC generation check passed; TUI RPC/i18n generation checks and type-check passed.

raven/agent/subagent/registry.py
Handsome-wzw: Done in 52ac1393. You were right, and the control you built was the convincing part: `modes` failing while `model_choices` did not is the whole argument, and the docstring claiming the rule while nothing enforced it is worse than not claiming it.

`test_the_model_menu_survives_the_row_and_the_meta_built_back_off_it` records a snapshot carrying a two-entry menu, builds the row with `_row_for`, and asserts both halves of the trip: the config's menu reaches `row.caps.model_choices`, and the meta built back off that row carries it too.

Reproduced your measurement and then checked the new test against it, over the same four files:

| withdrawn | before | after |
| --- | --- | --- |
| `model_choices` from `AgentRow.meta()` | 461 passed | 1 failed, 463 passed |
| `model_choices` from `_row_for` | 461 passed | 1 failed, 463 passed |

Both mutations fail this test and only this test; the unmutated suite is 464 passed.

I left the two things you recorded as not-findings alone. The untested `session_model` push at the direct-chat call site takes its sibling `mode=` argument down with it, so it is one gap in that call site rather than something this PR introduced, and it is not mine to close here.

gloryfromca (commented):

`52ac1393` closes the sole non-blocking follow-up with a focused regression that records a non-empty two-model snapshot, builds the registry row, and asserts both conversions independently: config/snapshot to `row.caps.model_choices`, then row to `row.meta().model_choices`. That directly pins the failure mode @0xKT measured without broadening the change. I verified the test and current-target merge; the reviewer's thread remains for its author to resolve. All threads I opened remain resolved.

I covered the new delta, current full diff, callers and history, backward compatibility, generated wire artifacts, UI integration, test strength, current-target compatibility, and the repository rules in `AGENTS.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`; I found no weakened tests or error worth raising. Verification: `uv run pytest tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_rpc_instances.py tests/test_rpc_schema_match.py tests/test_acp_schema.py -x` (835 passed); changed-file Ruff check and format check passed; the web component test (10 passed), web type-check and RPC generation check passed; TUI RPC/i18n generation checks and type-check passed; `git merge-tree --write-tree HEAD github/main` completed cleanly.

0xKT (approved):

PR #404, opened 2026-09-14
fix(updates): the upgrade marker reads raven_home from its own module

gloryfromca:
## Summary

`install_guard.marker_path` imported `raven_home` from `raven.config.loader`, a
re-export. Importing any `raven.config` submodule runs that package's
`__init__`, which pulls in the settings stack and, with it, pydantic. The
function is defined in `raven.home`, whose only imports are `os`, `pathlib` and
a constants module with no imports of its own. One line, same function object,
no layering moved.

That cost lands at the worst moment. `write_marker` runs while uv is about to
delete the environment the process is executing from, so every module this path
needs is one more thing that has to still be importable inside the replacement
window.

This is the benefit `pyproject.toml` already describes. The import-linter
contract "the kernel stands alone" holds `raven.contracts`, `raven.home`,
`raven.spine` and `raven.tracing` away from the rest of the tree at import time,
and its comment records why `raven_home` sits where it does: "RAVEN_HOME
resolution moved in to raven/home.py". To be precise, nothing here was breaking
that contract, which constrains `raven.home` importing `raven.config` and not
`raven.updates` importing it, so this is not a lint fix. It is `install_guard`
taking the benefit that move was made for, at the call site with the least room
to spare.

It also broke `Windows self-upgrade` on every push since the history rewrite.
That test stands up a minimal fixture package declaring `httpx`, `rich` and
`typer`, installs it with a real uv, and has it import the handoff from the repo
source. The settings stack asks for more than the fixture declares, so the
subprocess died with `ModuleNotFoundError: No module named 'pydantic'`. The test
only runs in the Windows job, which is why it read as a platform problem.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

The failure reproduces on macOS, so it was never Windows-specific:

```
pytest tests/integration/test_cli_upgrade_real_uv.py -q -n 0
before: 1 failed, 1 passed   ModuleNotFoundError: No module named 'pydantic'
after:  2 passed
```

Import weight, measured:

```
from raven.config.loader import raven_home  ->  pydantic in sys.modules: True
from raven.home import raven_home           ->  pydantic in sys.modules: False
```

Same function object either way (`raven.home.raven_home is
raven.config.loader.raven_home` is True), and `marker_path()` returns the same
path. 178 tests matching `install_guard or upgrade` pass.

- [x] Relevant tests pass locally
- [ ] Relevant lint / type checks pass locally
[...]

PR #402, opened 2026-09-14
fix(ui-web): the page gate counts the asset digest script

gloryfromca:
## Summary

`ui-web/build.py` emits a third inline script, the asset-tree digest that hangs
`window.__ASSETV` on the page before any icon helper reads it, but
`ui-web/scripts/check-page.mjs` still asserted exactly two blocks. The `page
checks` job has failed on `main` ever since that feature landed. GitLab does not
run this gate, which is why the breakage stayed invisible on that side.

Raise the fixed count to three and name the blocks in load order. The count
stays fixed on purpose: a fourth block is still a failure.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Ran the exact CI step on a fresh build:

```
npm ci --prefix ui-web
npm run --prefix ui-web build
python3 ui-web/build.py
node ui-web/scripts/check-page.mjs            -> check-page: OK (1,392,248 bytes, 3 scripts parse)
node ui-web/scripts/count-shared-globals.mjs  -> exit 0
npm test --prefix ui-web                      -> 103 files, 1678 tests passed
```

The guard is still a guard, not a weakened check: inserting a fourth block into
the built page makes the gate fail with `expected exactly three <script>
blocks`, and removing it again returns the gate to OK.

- [x] Relevant tests pass locally
- [ ] Relevant lint / type checks pass locally
- [ ] User-facing docs or screenshots are updated when needed

## Risk

Gate-only change. No shipped code is touched and the served page is unchanged.
Reverting restores the red job and nothing else.

- [x] Security impact considered
- [x] Backward compatibility considered
- [x] Rollback path is clear for risky changes

## Related Issues

N/A

PR #409, opened 2026-09-15
test(*): guard the upgrade kernel boundary, de-flake mcp prewarm

0xKT:
## Summary

#402, #403 and #404 turned main green again after the 2026-09-11 history
rewrite. Reproducing those three failures locally surfaced four things
they leave open; this PR closes them.

- install_guard: #404 moved the import to raven.home but kept it inside
  marker_path, while its comment argues for importing as little as possible
  at the moment the marker is written. The two point in opposite
  directions: a module-level import is paid once at startup, before any
  replacement window opens, and raven.home has no cycle that would force
  the deferral. Hoisted, comment dropped.
- beta_channel._state_path carried the same raven.config.loader import.
  It runs on the `raven upgrade` path before the handoff, so nothing was
  broken there, but it now resolves the home from the module that defines
  it, like the rest of the upgrade path.
- A unit-level pin beside the marker tests: write_marker in an isolated
  interpreter (-I) must leave raven.config and pydantic unloaded. The
  integration test that caught the original bug runs only in the Windows
  job; this makes the one call site it tripped on visible on every
  platform. The integration test remains the proof for the whole handoff.
- test_a_turn_does_not_wait_for_a_server_that_is_waiting_on_a_person, the
  test that fails intermittently on the GitLab runner, polled the tool
  registry. The handshake registers the tool before the manager takes its
  lock, and the loop flips _mcp_connected only after the whole apply
  returns, so a poll could wake in between and read the flag as unset. It
  now awaits the prewarm task, as the rest of the file does.

Two things the diagnosis left for separate calls: doctor_commands.console
still follows the terminal width, so `raven doctor` on a narrow terminal
still breaks a long path mid-token (#403 pinned the test, not the
product); and check-page.mjs does not check that the __ASSETV__ marker
was substituted, while build.py has no count guard for it as it has for
the other marks.

The verification below was run on this tree with #402-#404 applied, so
the three repaired gates are included in the numbers.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other (tests, plus two import moves that serve them)

## Verification

```
npm run --prefix ui-web build; python3 ui-web/build.py
node ui-web/scripts/check-page.mjs               check-page: OK (1,392,248 bytes, 3 scripts parse)
node ui-web/scripts/count-shared-globals.mjs     exit 0
node ui-web/scripts/check-css.mjs                check-css: OK

uv run --frozen --all-extras pytest tests/test_cli_doctor_commands.py::test_doctor_answers_where_the_memories_are \
    -n 0 --basetemp=/private/tmp/aaaaaaaaaaaaaaaaaaaaaaaaaaa      1 passed   (81-char root; 1 failed on main)
uv run --frozen --all-extras pytest tests/test_cli_doctor_commands.py                 97 passed

uvx --python 3.12 --from "pytest>=8,<10" --with pytest-asyncio --with pytest-xdist \
    pytest --noconftest tests/integration/test_cli_upgrade_real_uv.py -q            2 passed   (1 failed on main)

uv run --frozen --all-extras pytest tests/test_updates_install_guard.py \
    tests/test_updates_beta_channel.py tests/test_updates_upgrade.py                191 passed
uv run --frozen --all-extras pytest tests/test_agent_loop_mcp_sync.py                51 passed

make lint-python        clean
make lint-imports       10 contracts kept, 0 broken
make lint-deps          no dependency issues
make test-python        22051 passed, 43 skipped in 443.67s
```

## Risk

- [x] Security impact considered -- no new input is parsed; two imports move
  between modules that return the same function object.
- [x] Backward compatibility considered -- the marker path and the beta state
  path are unchanged (raven.home.raven_home is raven.config.loader.raven_home).
- [x] Rollback path is clear for risky changes -- revert restores the lazy
  imports and the polling test; no shipped behaviour changes either way.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the full diff plus the affected upgrade handoff, beta-channel callers, MCP prewarm lifecycle, and commit history. I also checked backward compatibility, that the tests were strengthened rather than weakened, and the AGENTS.md/CLAUDE.md and CONTEXT-MAP.md architecture rules (including the canonical Kernel, Home, and Updates boundaries). The direct `raven.home` imports preserve runtime home resolution and keep the config/settings stack out of the upgrade path; awaiting the prewarm task observes the completed apply rather than the earlier registry mutation.

Verification:
- `uv run pytest tests/test_updates_install_guard.py tests/test_updates_beta_channel.py tests/test_agent_loop_mcp_sync.py -x` - 115 passed
- `uv run pytest --noconftest tests/integration/test_cli_upgrade_real_uv.py -q` - 2 passed
- `uv run lint-imports` - 10 contracts kept, 0 broken

gloryfromca (approved):

PR #410, opened 2026-09-15
feat(*): dead-end predicates, verbatim tool bodies and envelope failures

ZuyiZhou:
## Summary

Absorbs the first of two remaining commits from the research fork before it stopped
iterating. This one lands instruments only: nothing here changes what a shipped run does,
and the research flow label does not move because no model-visible text changed.

`raven/agent/loop/dead_end.py` (new) asks whether a finished turn ended on the model's own
prose. The test is structural, with a textual sub-label for the tail shape. Three decisions
are worth review:

- It lives under `raven/agent/loop/` rather than in the research plugin, because the loop
  is what would re-run a dead turn, and a host module cannot import a plugin without
  breaking `make lint-imports`.
- The ask labeller is injected (`ask_kind`) rather than imported, because only the product
  that writes the harness asks knows how they are worded. With no labeller the failure mode
  is still recognised structurally and reported as `harness_ask_unknown`, so the boolean
  never depends on wording.
- The fork marked its injected turns with a key of its own. This loop already had an
  equivalent in `_HOOK_INJECTED_KEY`, so the predicate reads that rather than adding a
  second marker for the same fact.

The conditional rerun that reads the predicate arrived on this branch afterwards, in
#418, along with the turn's wall clock. Both are supplied as data a hook writes onto the
turn's metadata rather than as config the loop reads, so an agent whose hooks leave no
budget runs exactly as it did before either existed: the rerun is unreachable for it and
the clock is never consulted. The rerun starts from the original question rather than
salvaging the failed attempt, because salvage was measured on 25 real duds to produce
content 23 times and the right answer none of them, turning a detectable zero into a
confident wrong answer.

Three defects in that work were found by review and fixed on this branch, each with a
reproduction that is red on the commit before it. The loop numbered episodes from an
iteration count that restarts, so a rerun emitted indices [0, 1, 2, 0] and the TUI, which
keys episode rows and their fold state by that index, had the rerun's first step collide
with the first attempt's. Each attempt took its own clock reading, so a first attempt
ending just under the limit handed the rerun a fresh full budget and a ten-second turn ran
to eighteen. And the research hook opens its per-turn ledger at iteration 1, which a rerun
reaches again, so the second open replaced the path the close reads and left the first
attempt's file unreachable on disk.

Reading that back afterwards turned up three more, none of them raised in review. The
turn's end record carried the attempt's iteration count beside the turn's elapsed time
with nothing saying which was which, so it gained ``attempt``. The hook contract described
that record as "status and iteration count" while it carried five fields, which hid
``stopped_by`` from anyone writing a terminal gate against the contract. And the rerun's
progress line named research in a loop that serves every agent, and called itself the
first attempt when the budget allows more than one.

A verbatim body sink (`RAVEN_VERBATIM_SINK`) records search and fetch results at three
chokepoints. It is a second file rather than more columns on the ledger because a 250 KB
row would leave no complete line in the 64 KB tail window `_resolve_session_seq` reads,
which would silently reset the re-run split. Off unless the environment names a file, and a
write failure is logged rather than raised.

`DigestOutput` lets a digest return a write-only ledger annotation beside its text. The
annotation is stripped from the tool result and merged into the fetch row, so a digest can
never label its own output where the model can read the label. Nothing emits one yet; this
is the pipe a later digest sidecar uses.

`is_hard_tool_failure` was blind to the whole reader path. `web_fetch` never returns a bare
string, it returns a JSON envelope, so a spent quota, a rejected URL and a failed
validation all read as success and reset the streak `web_search` had accumulated.
Transient markers are still tested first, so a 429 inside an envelope stays retryable.

Forced finalize now books its own salvage spend. The phase fires on every run that produced
no answer and burns up to `max_tokens` doing it, and it was counted nowhere.

The Chinese refusal opener lives in `raven/i18n/zh_lexicon.py`, which is the exemption zone
this repo keeps Chinese forms in, and the predicate imports it from there rather than
carrying a literal outside the zone.

Two defects found in the first review round are fixed on this head, both reproduced before
being changed.

The verbatim sink borrowed the ledger's `_resolve_session_seq`, which recovers the re-run
split by reading the last 64 KB of the file and parsing the last complete line. That works
for the ledger because its rows are small; this sink stores the large ones, so the first
record wider than the window leaves nothing parseable and the sequence resets to 1, making
two runs indistinguishable. Runs are now told apart by a tag minted in memory once per
process and never recovered from disk, which removes the read a large row defeats. The
argument for keeping this file separate from the ledger was already in its docstring; it
applies to the sequence too, and now says so.

`is_hard_tool_failure` started counting JSON envelopes while `failure_class` still read
them as undifferentiated text. The streak key is `(tool, failure_class)` and the break
threshold is two, so a blocked URL followed by a reader's HTTP refusal fired the
stop-repeating nudge at a model that had changed both its cause and its approach. Both
predicates now share one envelope parse, and an envelope is classified by its own `error`
string rather than by a textual ladder that cannot see into one. Coarseness is for model
prose, which varies without meaning anything; a tool writes its error from a small fixed
vocabulary, so equal strings are the same failure and different ones are different. The
parts that vary with the page travel in the envelope's other keys, so they cannot split a
streak.

The last commit removes an `isinstance` guard in the envelope reader that the diff-coverage
report flagged as the only uncovered changed line. JSON has one shape that opens with a
brace, so the branch was unreachable; removed rather than covered.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run at this head (95c975b3), on `origin/main` with nothing to rebase onto.

```
  22132 passed, 43 skipped, 109 warnings in 388.67s
```

Integration and e2e are deselected by default (`-m "not integration and not e2e"`), so the
run above does not include them. This change adds no integration surface.

```
make coverage
make coverage-ratchet
  line   current=88.66% baseline=87.30% delta=+1.36pp
  branch current=81.59% baseline=79.43% delta=+2.15pp
  Coverage ratchet passed with 0.05pp tolerance.
make coverage-baseline-check
  Coverage baseline update is monotonic.
COVERAGE_BASE_REF=origin/main make coverage-diff
  Diff coverage: 100.00% (126/126 executable changed lines)
```

Diff coverage reported 99.21% on the commit before last, with one uncovered changed line:
the sentence the rerun says to the reader, which no test drove because none passed an
`on_progress` into a retrying turn. The last commit pins it rather than waiving it.

```
make lint-python
  All checks passed! / 1946 files already formatted
make lint-imports
  Contracts: 10 kept, 0 broken.
make lint-deps
  Success! No dependency issues found.
make check-source-language
  clean
make check-large-files
  clean
make check-commits
  clean
```

The kernel budget gate caught one thing worth naming: documenting the turn-end record in
the hook contract put `raven/contracts` at 2808 lines against a ceiling of 2791. The
per-field detail moved to the write site in `turn_path`, which is not budgeted, and the
paper now names the fields and says where the rest is. `raven/contracts` is at 2788.

Each of the three review fixes was proved against the commit it fixes rather than
reasoned about: a detached worktree at that commit, the new tests copied in, red there and
green here. They reproduce the reviewer's own numbers -- the episode list `[0, 1, 2, 0]`,
the eighteen seconds, and the stranded ledger file. The strengthened comparison in
`test_the_retry_starts_from_the_question_not_from_the_wreckage` is a tightened guarantee
rather than a regression test and passes either way, which is why it is called out here.

`agents/raven-research/.env` was moved aside for every suite run: it is gitignored and
absent in CI, and two launcher tests that delete a key from the environment otherwise find
one in the file.

This machine has no search-provider key, so no research brief can actually be run here.
The retrieval-path changes are verified by unit tests only, which is why the three
chokepoints are pinned by tests that assert both the presence and the absence of a row
rather than by an observed run. The same limit applies to the rerun: the clock and the
dead-end trigger are driven by a deterministic clock in tests, never by a real turn.

No user-facing surface changed, so there is nothing to document. The one operator-visible
knob, `RAVEN_VERBATIM_SINK`, is an instrument that is off unless set and is documented in
the module that reads it.

## Risk

The `is_hard_tool_failure` change is the only unconditional one, and it reaches every agent
rather than just research: any tool returning a JSON envelope with a non-empty top-level
`error` now counts toward the tool-failure streak that triggers the change-approach nudge.
That is the intended correction, and the direction is conservative because transient
markers are still tested first. Reverting it is a single branch.

Everything else is inert by default. The body sink writes nothing unless the environment
names a file. `DigestOutput` has no emitter in this change, and a digest returning a plain
string produces the same envelope it always did, which
`test_a_bare_string_digest_still_works` pins.

The dead-end predicate now has a caller, and it reaches only an agent whose hooks write a
budget onto the turn's metadata. No agent on this branch does: `agents/raven-research`
ships `wallClockSeconds` and inherits the rerun from its plugin's class default, and every
other agent leaves the key unwritten and is bounded exactly as before. The turn's overrun
is one budget plus at most one iteration, because the clock is read between iterations and
never mid-generation: cancelling a call in flight would discard a finished generation and
leave no answer at all. Rolling the rerun back is the budget key; rolling the clock back is
the same.

Rollback is the whole commit; nothing here is depended on by anything already on main.

## Related Issues

N/A


agents/raven-research/plugins/research-flow/research_flow/support/ledger.py
```diff
@@ -198,6 +198,66 @@ def _resolve_session_seq(path: str) -> int:
...
+    append-only, read by nobody at run time, and a write failure is logged and swallowed,
+    because an instrument that can kill the run it measures is worse than no instrument.
+
+    ``session_seq`` is stamped here for the same reason :func:`ledger_append` stamps it:
+    several call sites, one definition of a line.
+    """
+    if not (path := verbatim_path()):
+        return
+    if path not in _session_seq:
+        _session_seq[path] = _resolve_session_seq(path)
```
gloryfromca: Blocking: rerun provenance in the verbatim sink must survive the large records it is designed to store.

`verbatim_append()` delegates to `_resolve_session_seq()`, which reads only the last 64 KiB and requires a complete JSON line. A normal source record can exceed that window (the module itself cites 250 KiB), so after a new process starts on an existing sink the last record cannot be parsed and the sequence resets to 1. I reproduced this by appending a 70,000-character record, clearing the per-process cache to simulate a rerun, and appending again: the rows were stamped `[1, 1]`. Those attempts can no longer be partitioned, so analysis includes stale tool bodies from the earlier run. The sink needs a rerun delimiter/sequence mechanism that does not depend on fitting a verbatim row in the ledger's 64 KiB tail window.

raven/agent/loop/failure_streak.py
```diff
@@ -75,6 +76,21 @@ def is_hard_tool_failure(result: object) -> bool:
...
+    # That made this predicate, the only tool-failure circuit breaker in the loop, blind
+    # to the entire reader path. Worse than not firing: a failed fetch read as a success
+    # and reset the streak ``web_search`` had already accumulated. Transient markers are
+    # tested above, so a 429 or a timeout inside the envelope still reads as retryable.
+    if s.lstrip().startswith("{"):
+        try:
+            payload = json.loads(s)
+        except (TypeError, ValueError):
+            payload = None
+        if isinstance(payload, dict) and payload.get("error"):
```
gloryfromca: Blocking: JSON-envelope failures must preserve the circuit breaker's distinction between different errors.

This adds every top-level `error` envelope to hard-failure accounting, but the streak key is still `failure_class(model_text)`. Common reader failures such as `URL validation failed: blocked host` and `Jina answered HTTP 403` both classify as `other`; with `_LOOP_BREAK_THRESHOLD == 2`, two consecutive but different `web_fetch` failures now trigger the stop-repeating nudge even though the model changed cause and approach. I confirmed both sample envelopes produce `(True, 'other')` for `(is_hard_tool_failure, failure_class)`. This violates `failure_class`'s stated contract and can steer ordinary research away from fetching while it is adapting. Detection and classification need a consistent JSON-envelope path.

gloryfromca (commented):
Blocking: rerun provenance and JSON-envelope streak classification need correction before merge.

I found two reproducible issues, marked inline. I reviewed the full `github/main...HEAD` diff, relevant loop and research-flow callers, the single-commit history, backward compatibility, test changes, and the repository's AGENTS/CLAUDE, context-map, source-language, dependency, test-naming, import-layer, commit, and asset constraints. The tests add coverage rather than weakening existing assertions.

Verification: the six directly affected modules passed (212 tests). The authoritative Python 3.12/all-extras suite produced 22,081 passed, 55 skipped, and one unrelated failure in untouched `test_cli_theme.py`; this environment exports `NO_COLOR=1`, and the isolated test fails with that variable present and passes when it is removed. The PR's Python 3.12 CI job is green. Import contracts, source language, large-file, and commit-message checks passed via their direct commands (`make` is unavailable in this environment).

gloryfromca (commented):
Blocking: verbatim rerun provenance and JSON-envelope failure classification remain incorrect.

This head only merges current `main`; every feature and test path in the PR is byte-for-byte unchanged from `865aca87abbe`. Both previously reported blockers still reproduce, neither thread has an author reply, and I found no new issue to add. The two existing threads therefore remain open.

I re-read the full `github/main...HEAD` diff and the relevant loop/research-flow callers, checked the merge and history, backward compatibility, test assertions, and the repository's source-language, naming, import-layer, dependency, commit, and asset constraints. The merge introduces no architecture conflict and the tests were not weakened.

Verification at this head: `uv run --frozen --python 3.12 --all-extras pytest -q` over the six PR-focused modules plus the two modules merged from `main` passed, 287 tests in 8.52s. The current CI unit job is still running; all completed checks are green.

agents/raven-research/plugins/research-flow/research_flow/support/ledger.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed. The sink now mints one opaque `run_id` per process instead of recovering a sequence from the file tail, so record size can no longer collapse two runs. The new regression test writes 70,000-character rows across simulated processes and verifies distinct run IDs; the focused suite passes. I am closing this thread.

raven/agent/loop/failure_streak.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed. Both predicates now share the same parsed top-level envelope error, and `failure_class` keys the streak by normalized error text rather than the catch-all. The tests cover distinct ordinary reader failures as separate classes and the same error across different payload fields as one class; the loop caller still keys on `(tool, failure_class)`. The focused suite passes. I am closing this thread.

gloryfromca (commented):

This revision fixes both previously reported defects. The verbatim sink now partitions process runs with a minted `run_id` that does not read the tail, and envelope failure classification now derives its key from the parsed top-level error while ignoring unrelated payload fields. I reproduced the relevant boundaries through the new regression tests and resolved both threads I opened.

I reviewed the current `github/main...HEAD` diff and the delta from `53f2170917c7`, plus the loop and research-flow callers and the two-commit history. I also checked backward compatibility, test changes for weakened assertions, and the project rules covering source language, domain naming, dependencies, test naming, import layers, commits, and assets. The new `run_id` schema has no released consumer because the sink is introduced in this PR, the opt-out path remains inert, and the fixes preserve the intended package boundary.

Verification: 225 focused tests passed in 8.57s. Import contracts passed (10 kept, 0 broken); source-language, large-file, commit-message, and diff-whitespace checks passed. All completed CI checks are green; the Python 3.12 unit job was still running when I posted this review.

agents/raven-research/plugins/research-flow/research_flow/support/ledger.py
ZuyiZhou: Fixed in a8872b81 (branch head; the two blockers landed in d2c23526).

Reproduced first, with your recipe: a 70,000-character record, the per-process cache
cleared, then a second append. Both rows came back stamped 1, so the two attempts could
not be partitioned.

The sequence is gone rather than repaired. `verbatim_append` no longer calls
`_resolve_session_seq`; runs are told apart by `_verbatim_run_id()`, a tag minted in
memory once per process and never recovered from disk. That removes the read a large row
defeats, so no record size can reach it. It is opaque and unordered on purpose: a reader
that wants the runs in order has `ts` on every row, and a sequence number here would imply
this process knows what came before it, which is the claim that was wrong.

The module docstring already argued that this file must be separate from the ledger
because its rows are the oversized ones. That argument applies to the sequence too, and
the docstring now says so, so the next person does not re-borrow the tail read.

Three tests pin it in `tests/test_agents_research_verbatim_sink.py`:
`test_two_runs_stay_apart_across_a_record_wider_than_a_tail_window` is your reproduction
and fails if the sequence comes back;
`test_one_process_stamps_one_run_however_many_rows_it_writes` holds the other direction,
since a tag that varied within a run would partition nothing; and
`test_a_caller_that_knows_better_keeps_its_own_tag` covers the `setdefault`.

raven/agent/loop/failure_streak.py
ZuyiZhou: Fixed in a8872b81 (branch head; the two blockers landed in d2c23526).

Confirmed your reproduction before changing anything: `URL validation failed: blocked
host`, `Jina answered HTTP 403`, `Jina API key not configured` and `Proxy error: ...` all
returned `(True, 'other')`, so with `_LOOP_BREAK_THRESHOLD == 2` two different causes in a
row fired the nudge.

The root cause was the split you name: detection learned about envelopes and
classification did not. Both now go through one `_envelope_error()` parse, so they cannot
diverge again, and `failure_class` classifies an envelope by its own `error` string
instead of running the textual ladder over a payload the ladder cannot see into.

On why the envelope class is the error text rather than a bucket: coarseness is for
model-facing prose, which varies without meaning anything. A tool writes its `error` from
a small fixed vocabulary, so equal strings really are the same failure and different ones
really are different, which is exactly the contract this function states. The parts that
vary with the page, the URL and the host, travel in the envelope's other keys, so they
cannot split a streak.

Both directions are pinned in `tests/test_agent_loop_tool_loop_break.py`.
`test_different_envelope_failures_are_different_classes` is your case, and it also asserts
all three still count as hard failures, so a fix that simply stopped counting envelopes
would not pass it. `test_the_same_envelope_failure_on_two_urls_is_one_class` keeps the
nudge reachable for a model walking one dead reader across different pages.
`test_an_envelope_class_is_read_off_the_error_not_the_payload` holds the class to the error
key alone.

Transient markers are still tested first, so a 429 inside an envelope stays retryable.

One follow-up from this change, in a8872b81: the diff-coverage report flagged the
`isinstance(payload, dict)` guard in `_envelope_error` as the only uncovered changed line.
JSON has one shape that opens with a brace, so that branch is unreachable. Removed rather
than covered.

gloryfromca (commented):

The only delta from the previously clean revision removes an unreachable non-dict guard after JSON input has both opened with `{` and parsed successfully. Standard JSON syntax guarantees an object in that case, so the remaining `.get()` is safe and behavior is unchanged. I found no new issue.

I rechecked the resulting `github/main...HEAD` diff, the envelope caller and history, backward compatibility, test assertions, and the project rules covering source language, domain naming, dependencies, test naming, import layers, commits, and assets. The change neither weakens tests nor alters the package boundary. Both earlier threads remain resolved; the later author replies only confirm those settled fixes.

Verification: the 53 focused failure-streak and loop-break tests passed in 6.75s. Import contracts passed (10 kept, 0 broken); source-language, large-file, commit-message, diff-whitespace, and all completed CI checks passed. The Python 3.12 CI unit job was still running when this review was posted.

raven/agent/loop/turn_path.py
```diff
@@ -2435,40 +2493,85 @@ async def _tap_reasoning(delta: str) -> None:
...
+                # The conditional rerun. A dead turn has no answer to damage -- "empty
+                # implies wrong" is a scoring rule, so the count of right answers among
+                # dead turns starts at zero and a second attempt can only raise it. It
+                # re-runs from the original question rather than salvaging the failed
+                # attempt, because salvage was measured to turn a detectable zero into a
+                # confident wrong answer and to empty this trigger at the same time.
+                # Unreachable for an agent whose hooks leave no budget, which is every
+                # agent but the one that asked for it.
+                while retries_left > 0:
+                    reasons = dead_reasons(
```
gloryfromca: No blockers; suggestions only, and they are marked inline.

Follow-up: dead-end retries currently pay for and discard terminal salvage before starting the fresh attempt. `_run_agent_loop()` has already called `terminal_answerless` before control reaches this predicate. In the shipped research chain, `ForcedFinalizeGate` may spend up to 240 seconds synthesizing an answer, but `LoopOutcome.status` remains `error`; this block therefore matches `status:error` and reruns from `retry_seed`, dropping the synthesized answer. I reproduced the ordering with a scripted terminal hook: it was called once, two attempts ran, and the returned value was the second attempt rather than the salvage. That conflicts with the stated rerun-instead-of-salvage policy and can consume enough of the shared clock to suppress the rerun. Consider evaluating the dead-end/retry decision before terminal salvage, or otherwise bypassing salvage while retry budget remains. This is nonblocking at this review stage because the shipped one-hour budget still normally permits the retry and cancellation/configuration remain escape hatches.

gloryfromca (commented):

Reviewed the new `a8872b81f971...6c60afda1b00` feature delta and rechecked the resulting `github/main...HEAD` change. The opt-in loop budgets preserve other agents' defaults, and the research flow's attempt reset, ledger lifecycle, persistence, and shared-clock paths are covered. One nonblocking follow-up is marked inline: the rerun is currently ordered after terminal salvage, so a dead attempt can pay for and discard a synthesized answer before starting fresh.

I covered the repository rules (`AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md`, and Runtime terminology), the diff, affected hook/tool/session callers, relevant history and stated intent, backward compatibility, test changes for weakening, and import-layer constraints. Verification: 148 focused tests passed; `git diff --check`, `lint-imports` (10 contracts), source-language, large-file, and commit-message checks passed. GitHub's unit job is still running; every other reported check is green.

gloryfromca (commented):

No new findings in the `6c60afda1b00...648eae4f1502` delta. Attempt numbering is monotonic across reruns, direct loop callers retain `attempt=1`, the final attempt's `turn_end` record reaches `after_send`, the progress notice is agent- and count-neutral, and the strengthened seed assertion does not weaken earlier coverage. The previously marked terminal-salvage ordering remains the only nonblocking follow-up.

I rechecked the repository rules (`AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md`, and Runtime terms), the full `github/main...HEAD` diff, affected loop/hook callers and history, backward compatibility, test changes, and import architecture. Verification: 248 focused loop, hook-contract, and research-flow tests passed; `git diff --check`, `lint-imports` (10 contracts), source-language, large-file, commit-message, Ruff lint, and Ruff format checks passed. All GitHub checks, including all four unit-test shards and coverage gates, are green.

gloryfromca (commented):

No new findings in the test-only `648eae4f1502...95c975b3ae40` delta. The added test drives a real dead attempt and rerun, observes exactly one progress notice, and pins the reviewed requirements without weakening existing assertions. The previously marked terminal-salvage ordering remains the only nonblocking follow-up.

I covered the applicable repository rules (`AGENTS.md`/`CLAUDE.md` and the existing Runtime terminology), the full change plus this delta, the progress-callback caller and relevant history, backward compatibility, test strength, and architecture boundaries. Verification: all 27 dead-end retry tests passed; `git diff --check`, source-language, large-file, commit-message, Ruff lint, and Ruff format checks passed. GitHub CI completed successfully, including all four unit-test shards and the coverage gates.

gloryfromca (commented):

No new findings. The requested `14b130bcabb2` delta makes the existing retry default explicit in the shipped config and pins that declaration against both the loader and vendored-fork parity. It resolves to the same behavior as before. While I was verifying, the branch advanced to `101dbfc3be88` by automatically merging current `main`; the remerge diff is empty, so there were no manual conflict resolutions, and I reviewed and tested that actual head. The previously marked terminal-salvage ordering remains the only nonblocking follow-up.

I covered the repository rules (`AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md`, and Runtime terms), the full `github/main...HEAD` feature diff and both new deltas, config/loader callers and history, backward compatibility, test strength, and architecture boundaries. Verification on the merge head: 78 focused launcher, parity, and retry tests passed; JSON parsing, `git diff --check`, source-language, large-file, commit-message, applicable Ruff, and all 10 import contracts passed. One initial Ruff command mistakenly included `config.json` and therefore reported Python `F821` errors for JSON booleans; rerunning the applicable Python and JSON checks passed. GitHub's non-unit checks are green; its four unit shards are still running.

raven/agent/loop/turn_path.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in `defdcf7ed`. The retry decision now happens once at the attempt boundary: an attempt with a rerun coming skips `terminal_answerless`, while the final attempt (or one whose shared clock denies another run) still reaches salvage. The regression tests cover the original discard/suppression paths, the shipped 240-second salvage cost, transient recovery scaffolding, and the unchanged no-rerun default. I reran the focused loop and research-gate suite: 158 passed. This settles the follow-up.

gloryfromca (commented):

The new feature delta fixes the only carried follow-up: the loop decides once whether an attempt has a rerun coming, skips terminal salvage only in that case, and still salvages the final or clock-denied attempt. The regression coverage reproduces the prior discard and trigger-suppression paths, the shipped salvage cost, transient recovery scaffolding, and the unchanged behavior for agents with no rerun budget. I replied on the settled follow-up; all threads I opened are resolved. The subsequent merge of current `main` has no manual conflict resolution.

I reviewed `101dbfc3be88...70211497f8a7` and the resulting full `github/main...HEAD` diff, relevant loop, hook, retry, and research-gate callers, and branch history. I also checked backward compatibility, test changes for weakening, and the repository rules in `AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md`/Runtime terms, source language, dependency/test conventions, import architecture, commits, and assets. The added tests strengthen coverage rather than removing assertions.

Verification: 158 focused loop, recovery, synthesis, broken-turn, and research-gate tests passed. Ruff lint and format checks passed on the changed Python files; import contracts passed (10 kept, 0 broken); source-language, large-file, commit-message, and diff-whitespace checks passed. Every GitHub check is green, including all four Python unit shards and coverage gates.

gloryfromca (commented):

This revision only merges updated `main`. The resulting `github/main...HEAD` feature patch is byte-for-byte identical to the previously clean revision, and the merge has no conflict-resolution delta, so it does not alter the reviewed behavior or tests. I found no new issue.

I compared the full feature diff and the `70211497f8a7...533a9619e866` delta, checked the merge and branch history, and rechecked the relevant loop, hook, retry, and research-gate paths. Because the feature patch is identical, backward compatibility, test strength, repository-rule compliance (`AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md` and Runtime terms), and import architecture are unchanged; no tests were weakened. All threads I opened remain resolved.

Verification on this head: 158 focused loop, recovery, synthesis, broken-turn, and research-gate tests passed. Ruff lint and format checks passed on the affected Python files; import contracts passed (10 kept, 0 broken); source-language, large-file, commit-message, and diff-whitespace checks passed. Every completed GitHub check is green; three of four Python unit shards were still pending when this review was posted.

gloryfromca (commented):

This revision only merges updated `main`. The merge has no conflict-resolution delta, and the resulting `github/main...HEAD` feature patch has the same byte-for-byte hash as the previously clean revision, so reviewed behavior and coverage are unchanged. I found no new issue.

I compared the full feature diff and `533a9619e866...4fc621155a21`, checked the merge and branch history, and rechecked the affected loop, hook, retry, and research-gate paths. Backward compatibility, test strength, repository-rule compliance (`AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md` and Runtime terms), and import architecture remain intact; no tests were weakened. All threads I opened remain resolved.

Verification on this head: 158 focused tests passed. Ruff lint and format checks passed on the affected Python files; import contracts passed (10 kept, 0 broken); source-language, large-file, commit-message, and diff-whitespace checks passed. Every completed GitHub check is green; three Python unit shards were still pending when this review was posted.

raven/agent/loop/turn_path.py
0xKT, line 1: **Blocking:** preserve live turn state across dead-end retries

The fresh list passed to `_attempt` at `raven/agent/loop/turn_path.py:2648` introduces two independently reproduced persistence failures.

1. **Consumed user corrections disappear on retry.** A correction sent with `turn.send` and `busy=inject` is drained while the first attempt is running and reaches its next model request. If that attempt ends without an answer, the retry starts from the seed captured before the injection. The consumed correction disappears from both the retry context and the saved session JSONL, while its handle resolves with the parent turn. In the probe, the first attempt sees a correction from Alice to Bob, then the retry sees only the original Alice question. The scheduler's fallback covers only undrained messages, so it cannot recover this one. Keep actual user steering when discarding the failed attempt's model/tool work.

2. **Cancelling a retry loses its completed tool transcript.** The retry operates on a fresh list, but the cancellation handler at `turn_path.py:2653-2660` still passes the first attempt's `initial_messages` to `_save_broken_turn`. A real `list_dir` completes successfully, emits `tool.complete`, and supplies its result to the next model request. Cancelling through `turn.cancel` then saves neither the completed assistant tool call nor its result. This violates the existing best-effort broken-turn persistence contract for work already produced by the active attempt. Keep the active attempt's buffer available to cancellation persistence. The adjacent generic-exception path was not executed and is not part of this finding.

Independent probes use the enabled shipped research-flow configuration, `make_hook`, its full hook chain, real scheduling, and public RPC handlers. The cancellation probe additionally uses `build_rpc_spine`, the real tool registry and subscription delivery. Model replies and latency are controlled; executor/MCP startup are bypassed. Each public-RPC probe reproduces the failure twice after asserting its live-state preconditions, with a passing no-retry control. The same injection probe passes all three cases on exact main `7beffb9d0f85488b2b5442caf269e398b2ffba8b`. Separate targeted alternatives preserving consumed user messages and retaining the active retry buffer make their corresponding probes pass. These alternatives demonstrate causality, not complete proposed fixes. All temporary source edits were restored. The coordinator independently replayed both public-RPC probes in another pristine merged tree, with the same results.

The existing retry tests pass and catch a control that disables retry. The findings concern state discarded by the new retry path; preserving failed research output is not required. Five installation-guard failures also occur on exact main in the shared uninstalled environment and are not PR findings.

Independently re-verified before posting, in a separate tree, by the reviewer signing this:

- The mechanism, read off the shipped source rather than a summary: `retry_seed` is snapshotted at `turn_path.py:2582`, the first attempt runs on `initial_messages` itself at `:2622`, and the retry runs on a copy of that snapshot at `:2648` -- so anything appended to `initial_messages` while the first attempt is running is absent from the retry. The `except asyncio.CancelledError` handler at `:2649-2660` still hands `initial_messages` to `_save_broken_turn`, which is the list the retry is not writing to.
- Both probes, re-run: 2 failed and 1 passing no-retry control each, matching the report.
- Reachability, which is the half I tried hardest to knock down: `dead_end_retries` defaults to 0 in `raven/agent/loop/_shared.py:660`, so I checked whether anything ships it on. `agents/raven-research/config.json` carries `"deadEndRetry": {"enabled": true, "maxRetries": 1}`, and `research_flow/flow.py:229` turns that into the budget the loop reads. The retry path is live in the shipped research agent, not dead code.
- Not pre-existing: on exact `7beffb9d` the first probe passes 3/3. The second probe cannot run there at all -- `FlowConfig` has no `dead_end_retry` attribute and the config key raises `KeyError: 'deadEndRetry'` -- which says the same thing more strongly: the whole dead-end-retry feature arrives with this PR, so neither failure mode can exist before it.

Every attempt I made to refute these failed. The probe sources are available if you want them.

gloryfromca (commented):
Blocking: live injected corrections and completed retry tool transcripts must be preserved.

The new open review comment changes my stance on this unchanged revision from clean to blocker. I independently reproduced both reported failures. In the injection probe, the first attempt consumed and used `Correction: Bob, not Alice`, but the fresh retry saw only the original question and the saved session JSONL contained no correction. In the cancellation probe, the retry completed a real `list_dir` and reached its next model call; cancellation then saved the `cancelled` marker but zero tool calls and zero tool results because `_save_broken_turn` received the first attempt's list.

The source paths agree with those observations: drained messages append only to the active attempt's list, the rerun starts from the pre-drain `retry_seed`, and both exception handlers persist `initial_messages` rather than the active rerun buffer. I am not resolving or re-grading the other reviewer's thread; it remains their open blocker. All threads I opened remain resolved.

Verification: both independent diagnostic probes reproduced their respective losses. The adjacent existing suites still pass (84 passed in 8.40s), with four non-strict wall-clock-ceiling warnings, confirming the current tests do not cover these retry-state persistence paths.

raven/agent/loop/turn_path.py
ZuyiZhou: Fixed in the branch this stacks under, as #443.

Reproduced both before changing anything, with your recipes.

**The consumed correction.** The seed was snapshotted before the attempt, and the
attempt is where the drain lands, so the correction existed only on the list the
rerun throws away -- and the inject queue had already given it up. Drained messages
now carry a mark, and the rerun's seed is the question plus every marked message on
the finished list. Two things about that seed you may want to push on: it reads the
whole list rather than its tail, because compaction can move the tail, and it
rebuilds from the untouched pre-attempt snapshot each time rather than appending to
a growing seed, so a second rerun carries the correction exactly once.

The mark is a message key in the family this loop already keeps -- it persists in
the transcript, where it says the message arrived mid-turn, and `_ALLOWED_MSG_KEYS`
drops it on the way to the provider.

**The cancelled rerun.** The attempt now records the list it runs on and both rescue
paths file that one, which is the parity the first attempt always had. Your finding
scoped itself to the cancellation path; the generic-exception handler is the same
line and changed with it.

Pinned in `tests/test_agent_loop_dead_end_retry.py`:
`test_a_correction_typed_while_the_attempt_ran_survives_the_rerun` is your first
probe read off the rerun's seed, `test_the_correction_the_rerun_carried_is_in_what_the_turn_files`
is the persistence half you also reported, and
`test_a_cancel_during_the_rerun_files_the_rerun_s_own_work` is your second probe --
a real `list_dir` completes in the rerun, then the next call is cancelled, and it
asserts the call id and its result are in what was filed.

Reverting each half on its own reddens only its own tests (2 and 1), so neither test
is covering the other's fix. Three mutants on the new lines are all killed, with a
behavioural control killed and a no-op control surviving.

Current `main` is merged in.

gloryfromca (commented):

The author has explicitly assigned the live-turn-state repair to stacked #443, whose base is this exact head. That scope decision lifts my blocker on #410. The defect remains real on this revision and is the named follow-up: #443 preserves consumed injected corrections in rerun context and session persistence, and files the active retry transcript on cancellation. Its focused suite passes locally: 36 passed in 10.46s.

I covered the new discussion, the exact branch relationship, and the focused fix tests; I did not rereview the unchanged #410 diff. @0xKT's thread remains open for its opener to assess and resolve.

gloryfromca (commented):

This revision only merges current main. The eight PR commits are unchanged under range-diff, the two merge parents changed no common paths, and the merge has no manual-resolution diff. I found no new integration issue.

The focused feature suite passes locally (306 passed in 9.41s), as does the adjacent loop/RPC-turn suite (101 passed in 7.76s). I checked the repository rules and runtime terminology, the merge delta and parent history, affected callers, backward compatibility, test integrity, and architecture boundaries; no tests were weakened. The previously named live-turn-state defect remains assigned to follow-up #443, so the stance remains nonblocking rather than clean. All threads I opened are resolved; @0xKT's thread remains theirs.

gloryfromca (commented):

This revision absorbs #443 as `3f3778e73`, fixing both live-turn-state losses: reruns retain consumed mid-turn user corrections, and cancellation or failure persists the active attempt's transcript. The subsequent main merge does not touch those files or require manual resolution.

I reviewed the new diff, surrounding loop and context-history paths, provider filtering, repeated-rerun and compaction behavior, callers and history, backward compatibility, test integrity, repository rules, and architecture boundaries. The focused feature suite passes locally (309 passed in 9.64s), and the retry plus adjacent broken-turn/RPC suite passes (137 passed in 9.49s). No tests were weakened. All threads I opened remain resolved; @0xKT's thread remains for its opener to close.

gloryfromca (commented):

This revision is only a conflict-free main merge. The incoming changes and the PR touch no common paths, all nine PR commits are identical under range-diff, and there is no manual-resolution diff. The focused feature suite passes on the merged tree: 309 passed in 9.44s.

For this revision I checked the repository rules and runtime terminology, the merge and full PR diffs, parent history, callers, backward compatibility, test integrity, and architecture boundaries. No tests were weakened and I found no new issue. All threads I opened remain resolved; the remaining open thread belongs to its original reviewer.

gloryfromca (commented):

Reviewed the merge update from `4e0a7bae7057` to `32fc1d96416b`. The nine feature commits are unchanged. The only merge conflict correctly follows main's relocation of `LoopOutcome` into `_shared` while preserving the package exports for `AgentLoop`, `LoopOutcome`, and the turn-budget symbols; the other five overlapping paths preserve both sides' behavior.

I covered the repository rules and Runtime context, the final comparison, overlapping callers and history, backward compatibility at the package export, and whether tests were weakened. `git diff --check` passed, and the exact-head isolated focused suite passed all 312 tests. The full CI run is not green, but its failures are in incoming-main paths outside this PR comparison (the new type gate, image-search/config expectations, and RPC idle ceilings), so they do not establish a defect in this revision. The patch remains human-review-required rather than an automatic-merge candidate because it changes shared retry contracts and persisted records.

gloryfromca (commented):

Reviewed the merge update from `32fc1d96416b` to `adc9f47467f1`. The two incoming main commits change eight files, none of which overlap this PR's files; there is no conflict resolution or combined diff, and `git range-diff` shows all nine feature commits unchanged.

I checked the repository rules and Runtime context, the final comparison, callers and history, backward compatibility, test changes, and the relevant architecture boundaries. No tests were weakened. `git diff --check` passed, and the exact-head focused suite passed all 312 tests. The pre-existing idle-ceiling warning came from an untouched test and does not establish a defect in this revision.

gloryfromca (commented):

Reviewed the main merge from `adc9f47467f1` to `e9bb3037939e`. Its seven incoming commits change 37 files, none overlapping this PR's files; there is no conflict resolution, and `git range-diff` shows all nine feature commits unchanged.

I checked the current AGENTS/CLAUDE rules and Runtime context, the final comparison, callers and history, backward compatibility, architecture boundaries, and test changes; no tests were weakened. `git diff --check` passed and the exact-head focused suite passed all 312 tests. All completed CI jobs pass; two unit shards are still running.

gloryfromca (commented):

Reviewed the main merge from `e9bb3037939e` to `d43384c96b3c`. Main's loop/window refactor overlaps this PR in `_shared.py` and `turn_path.py`; the remerge diff shows the manual resolution retains the PR's shared transient-key filter and correctly binds the attached-image marker to its new `raven.agent.window.images` home. The retry budgets, mid-turn user marker, rerun decision, cancellation rescue, and persistence paths remain wired through the refactored Memory role. `git range-diff` shows all nine feature commits unchanged.

I checked the current AGENTS/CLAUDE rules and Runtime context, the final diff, callers and history, backward compatibility, architecture boundaries, and whether tests were weakened. `git diff --check` passed; the exact-head feature and overlapping loop/window suite passed all 523 tests. All completed CI jobs pass, with one unit shard still running.

raven/agent/loop/turn_path.py
```diff
@@ -1968,47 +2073,141 @@ async def _tap_reasoning(delta: str) -> None:
...
+            """
+            return [dict(m) for m in retry_seed] + [dict(m) for m in msgs if m.get(_MID_TURN_USER_KEY)]
+
+        def _dead_reasons(final_content: str | None, msgs: list[dict], status: str) -> list[str]:
+            # The scaffolding goes before the reading. This question used to be asked
+            # after the attempt returned, on the list it had already been dropped from,
+            # and it is asked inside the attempt now; a recovery nudge left sitting at
+            # the end would otherwise become what the turn is judged to have ended on.
+            turn = [m for m in msgs[turn_start_idx:] if not any(m.get(k) for k in _TURN_TRANSIENT_KEYS)]
+            reasons = dead_reasons(messages=turn, final_content=final_content, status=status)
```
LivXue: **Nonblocking - wire the product's ask labeller into `dead_reasons`, or say the seam has no caller yet.** `dead_end.AskKind` and `stranded_of(..., ask_kind=...)` exist so a product can name the harness question an injected turn carries, and `research_flow.support.harness_text.harness_ask_kind` is written, exported and tested to be exactly that labeller. Nothing connects the two. This line is the only `dead_reasons` call in `raven/` and it passes no `ask_kind`, and `harness_ask_kind` has no caller outside its own `__all__` entry and `tests/test_agents_research_harness_text.py`.

Verified by grep at this head rather than by reading the design: `git grep -n harness_ask_kind d43384c96 -- ':!tests'` returns only the definition and the export, and `git grep -n dead_reasons d43384c96 -- raven/` returns only the definition, the import, this call, and two docstring mentions.

The boolean is unaffected, which is the seam's own stated guarantee, and it holds. What is affected is the sub-label: every harness-injected dead end reports `stranded:harness_ask_unknown`, so the three labels `harness_ask_kind` can produce (`finalize`, `checkpoint`, `verify_reject`) are unreachable. `TurnBudgets.dead_end_reasons` matches these labels by prefix, so an operator who narrows `deadEndRetry.reasons` to `stranded:finalize` would match nothing and disable the rerun they meant to narrow. The shipped config sends `reasons: []`, so nothing misbehaves today.

Either pass the labeller through from the flow that owns the asks, or note in `dead_end.py` that the parameter has no caller yet, so the next reader does not take the sub-labels as reachable.

raven/agent/loop/failure_streak.py
```diff
@@ -28,14 +29,50 @@
...
+
+    The envelope class is the error text itself, not a bucket, and that is the one
+    place coarseness buys nothing. Coarseness is for model-facing prose, which varies
+    without meaning anything; an envelope's ``error`` is written by the tool from a
+    small fixed vocabulary, so equal strings really are the same failure and different
+    ones really are different. The variable parts a page brings with it -- the URL, the
+    host -- travel in the envelope's own keys, not in this one.
     """
+    if (error := _envelope_error(model_text)) is not None:
+        return f"envelope:{' '.join(error.lower().split())[:120]}"
```
LivXue, line 83: **Nonblocking - the envelope class rests on an invariant its producers do not all keep.** This keys the streak on the envelope's own `error` text, and the docstring above states why that is safe: "The variable parts a page brings with it -- the URL, the host -- travel in the envelope's own keys, not in this one." That holds where the error is written from a fixed vocabulary, but `research_flow/tools/web.py` has two writers that interpolate free exception text instead: `f"Proxy error: {e}"` at :1844 and `str(e)` at :1847.

Reproduced against this head's `failure_streak.py`, not reasoned about: two envelopes reaching the `str(e)` writer with an SSL hostname mismatch, one for `alpha.example.com` and one for `beta.example.org`, both return `is_hard_tool_failure == True` and two different `failure_class` values. The hostname sits inside the first 120 characters, so the truncation does not collapse them.

With `_LOOP_BREAK_THRESHOLD == 2` (`loop/main.py:94`), a model walking one broken-certificate reader across different hosts therefore never reaches the stop-repeating nudge. That is this function's own failure mode in the other direction: it keeps distinct causes apart so the nudge does not fire at a model that is adapting, and here one repeated cause is split so the nudge never fires at all.

Not a regression, and I am not asking for the branch to be held on it: before this PR no envelope counted as a hard failure at all, so the direction is strictly better. Worth either narrowing those two writers to a fixed vocabulary, or stating in this docstring that the invariant is a convention producers are expected to keep rather than one this module can enforce.

LivXue:

Scope: I reviewed only the `raven/agent/**` portion of this PR -- 5 of 24 changed files (`loop/__init__.py`, `loop/_shared.py`, `loop/dead_end.py`, `loop/failure_streak.py`, `loop/turn_path.py`). The remaining 19 files were read as context but are not covered by this verdict.

I examined the in-scope diff against merge base `08596330cb30`, the rerun and budget paths in `turn_path.py`, the two envelope predicates in `failure_streak.py`, their callers under `agents/raven-research/plugins/research-flow/`, the `turn_end` text in the `loop_hooks` contract, and the conflict resolution carried by this head.

The conflict with `main` is resolved and the resolution is clean. `869ad78ab` (#468) is an ancestor of this head, `raven/agent/loop/compaction.py` is correctly gone, `raven/agent/window/` is present, and `git diff --stat origin/main..d43384c96` lists only this PR's own 24 files, so nothing from `main` was reverted while resolving. The in-scope feature patch is byte-identical to the previously reviewed revision except `_ATTACHED_IMAGE_KEY` -> `ATTACHED_IMAGE_KEY`, which is the correct adaptation to #468 making that constant public. gloryfromca reached the same conclusion on this head while I was drafting, and backed it with a 523-test run over the feature and the overlapping loop/window suites, which mine does not include; I ran my own check rather than relying on theirs, and the two agree. Neither of the findings below appears in that review.

Confirmed findings, both nonblocking, marked inline:

- `turn_path.py:2147` calls `dead_reasons` without `ask_kind`, and `harness_ask_kind` has no caller, so the injected-labeller seam is unreachable and every harness-injected dead end reports `stranded:harness_ask_unknown`. A narrowed `deadEndRetry.reasons` would match nothing.
- `failure_streak.py:75` keys the streak on envelope `error` text, but two of `web.py`'s writers interpolate free exception text, so one repeated reader failure across two hosts yields two classes and never reaches the two-strike nudge.

One item has no line to anchor to, so it belongs here. The Verification section still reads "Run at this head (95c975b3), on `origin/main` with nothing to rebase onto." This head is `d43384c96b3c`, and both halves of that sentence are now false: the reported `22132 passed` run predates `14b130bca`, `defdcf7ed` and `3f3778e73`, and the last of those is the fix for the live-turn-state defect, the most consequential commit on the branch. Because this repository squash-merges with the PR body as the commit body, that claim lands on `main` permanently. CI is green on this head, so this is a matter of updating the sha and the numbers rather than re-running anything.

Verification:

- Merge resolution: `git merge-base --is-ancestor 869ad78ab d43384c96` (ancestor), `git cat-file -e d43384c96:raven/agent/loop/compaction.py` (absent), `git diff --stat origin/main..d43384c96` (24 files, all this PR's).
- Finding 1: `git grep -n harness_ask_kind d43384c96 -- ':!tests'` and `git grep -n dead_reasons d43384c96 -- raven/` at this head.
- Finding 2: ran this head's `failure_class` and `is_hard_tool_failure` over two SSL-hostname-mismatch envelopes on different hosts; both hard failures, classes differ.
- I did not run the test suite. All 17 CI checks are green on this head, including the four Python shards and the coverage gates; I read that rather than re-running it.

- The live-turn-state fix from #443 holds up when tested rather than read off the commit subject: `_seed_for_the_rerun` carries `_MID_TURN_USER_KEY` messages onto the rerun seed, and the cancellation path now persists `live["messages"]`, the list the active attempt appends to. Rebuilding from the untouched snapshot each time also means a correction cannot be carried twice when more than one retry is budgeted.
- `live["messages"]` goes stale if `turn_path.py:996` rebinds `messages` to a shrunk list mid-attempt. Before this PR the same handler received `initial_messages`, the same object with the same staleness, so this is pre-existing rather than an escalation.
- `_dead_reasons` slices `msgs[turn_start_idx:]` where `turn_start_idx` is `len(initial_messages) - 1`; after an emergency shrink that slice can be empty, and `stranded_of([])` is False, so the structural signal is lost while the textual ones survive. Not reported because `_save_turn` and `_stamp_turn_observers` already index the same way on `main`, so the assumption predates this branch. Code-reading only; I did not drive a shrink.
- `_MID_TURN_USER_KEY` joins the existing `_HOOK_INJECTED_KEY` family of underscore keys and inherits its handling, so it raises nothing new here.
- `turn_budgets` rejects `bool` where an `int` or `float` is expected, which is the right way round for a key a plugin writes.
- Nothing under `raven/agent/subagent/` is touched.
- The branch carries four `Merge branch 'main'` commits rather than a rebase onto the target. Since this repository squash-merges, what lands on `main` is identical, so I am not raising it.

gloryfromca (commented):

The new review changes my standing stance on this unchanged revision from clean to nonblocking. I confirmed both code observations: `harness_ask_kind` has no production caller and the sole `dead_reasons` call supplies no `ask_kind`; the two proxy-error envelope writers interpolate exception text, so equivalent failures can split by host. I also confirmed that the PR description still attributes its full-suite verification to `95c975b3`, before the later fixes now present on this head.

These are concrete follow-ups, but none makes the revision unmergeable: the ask labeller currently leaves a narrowed optional retry filter ineffective, variable envelope classes still improve on the prior behavior where envelopes were not counted, and the stale verification text can be corrected without changing code. I did not re-review the unchanged diff or duplicate the existing inline threads; this comment only updates the whole-record stance.

LivXue (dismissed):

gloryfromca (commented):

This revision is a conflict-free merge of current `main`. The nine PR commits are unchanged under `git range-diff`, and none of the 17 incoming-main files overlaps the PR diff, so the reviewed feature behavior, callers, backward-compatibility boundary, architecture constraints, and tests are unchanged. I also checked the repository rules and updated runtime context, the full diff and history, and test changes for weakening; I found no new error.

The two existing inline suggestions and the stale PR-body verification provenance remain as nonblocking follow-ups. Focused verification at `eb986cf60abc`: `uv run pytest -q` over the dead-end, retry, failure-streak, research-flow, harness, ledger, tool, and verbatim-sink suites passed, 340 tests. GitHub CI has passed every completed job; unit shard 1/4 was still pending when reviewed.

raven/agent/loop/turn_path.py
ZuyiZhou: Fixed in 712ed3d6, and the seam is wired rather than documented away.

Your grep was right at that head and still was at eb986cf6: `harness_ask_kind` had no caller outside its own `__all__`, and the one `dead_reasons` call in `raven/` passed no `ask_kind`. So the three labels were unreachable and a `deadEndRetry.reasons` narrowed to `stranded:finalize` matched nothing, which switches off the rerun it meant to narrow rather than narrowing it.

The labeller now travels the seam the budgets already use: `TURN_ASK_KIND_KEY` on the turn's hook metadata, written by `TurnFrame.before_user_inbound` and read by `turn_ask_kind` beside `turn_budgets`. Two things about that choice you may want to push on. It carries a callable rather than a copy of the prefixes, because the classification is meant to be one implementation in the loop and the wording is the product's, which is what `AskKind`'s own docstring asks for; the entry is never filed onto a message or sent to a client -- only `metadata["observers"]` is, per `turn_path._file_observers` -- so it crosses no serialization boundary. And the reader is tolerant the way `turn_budgets` is: a non-callable leaves the ask unnamed rather than raising inside the loop, since an instrument may not end the turn it measures.

The boolean still comes from the structural marker, so an agent that writes nothing is labelled exactly as before.

Pinned in `tests/test_agent_loop_dead_end_retry.py`: `test_a_product_that_names_its_ask_can_narrow_the_rerun_to_it` is your finding -- a gate bounces one iteration with a `[finalize]` ask, the attempt then dies on it, and the turn re-runs only because the filter matched the name. `test_a_product_that_names_nothing_still_re_runs_on_the_structure` is the control for the guarantee above, and `test_a_labeller_that_is_not_callable_leaves_the_ask_unnamed` covers the tolerant reader.

Reverting `turn_path.py` alone -- keeping the key, the reader and the product's write -- reddens the first and leaves the control green, so neither test is covering the other's fix. Full suite: 23413 passed, 109 skipped.

raven/agent/loop/failure_streak.py
ZuyiZhou: Fixed in 712ed3d6, at the producers rather than in the docstring, and in both copies.

Reproduced your case first: two envelopes through the `str(e)` writer with an SSL hostname mismatch, `alpha.example.com` and `beta.example.org`, both hard, two different `failure_class` values, the host well inside the first 120 characters. With `_LOOP_BREAK_THRESHOLD == 2` one broken reader walked across hosts therefore never reaches the nudge.

Three writers now instead of two. The proxy case names itself (`"Proxy error"`). The generic case names the exception type. The exception's own text moves to a `detail` key, which is deliberate: dropping it would have taken the cause away from the model and, worse, out of `is_hard_tool_failure`'s transient-marker scan, so a timeout inside the envelope would have turned from retryable into hard -- the fix trading one bug for another. And `_ProviderPageError` is split out and keeps its composed message, because that type exists to say the vendor answered without a page and its text is assembled here from a fixed set of phrases; folding it onto the type would make a drained key and a blocked page one streak, which is the failure the function was written to avoid.

You cited `research_flow/tools/web.py`. The host tool `raven/agent/tools/web.py` carried the same two writers at `:831` and `:834`, and it serves every agent rather than one, so both are fixed the same way.

The docstring changed too, but to a weaker claim than it made: the vocabulary is a convention its producers keep, not something that module can enforce, and it now names the failure mode a writer that breaks it causes.

Pinned on both sides, each with its own control: `test_one_transport_fault_on_two_hosts_is_one_streak_class` and `test_a_vendor_that_answered_without_a_page_keeps_its_own_class` in `tests/test_agents_research_tools.py`, and the same pair in `tests/test_agent_tools_web_providers.py`. Reverting either web tool reddens its own streak test and leaves its vendor-refusal control green. Full suite: 23413 passed, 109 skipped.

gloryfromca (commented):

The new commit resolves both prior inline suggestions. I traced the ask labeller from `TurnFrame.before_user_inbound` through the per-turn metadata snapshot into `dead_reasons`; agents that do not provide it retain the structural `harness_ask_unknown` behavior. I also checked both web-tool copies and their exception branches: transport details remain model-visible but no longer fragment the streak key, while distinct provider refusals remain distinct. The tests add controls for both compatibility directions rather than weakening existing assertions.

I covered the repository rules and relevant runtime context, the full diff against current `main`, the new commit and merge history, callers, backward compatibility, test changes, and the host/plugin import boundary. Focused verification passed: 427 tests; the import contracts passed with 10 kept and 0 broken; `git diff --check` passed. Every completed GitHub job was green when reviewed, with unit shards 1/4 and 4/4 still pending.

The remaining nonblocking record follow-up is the PR description: its Verification section still says the full suite was run "at this head (95c975b3)", although the current head is `712ed3d60d16` and this commit reports a newer full-suite result. Refreshing that provenance would make the merge record accurate; it does not hold up the code.

raven/agent/loop/turn_path.py
LivXue: Verified at `712ed3d60`, by re-running the reproduction rather than by reading the fix.

The three labels are reachable now. Driving `dead_reasons` with `harness_ask_kind` over a harness-injected tail returns `stranded:finalize`, `stranded:checkpoint` and `stranded:verify_reject`; the same three inputs with no labeller still return `stranded:harness_ask_unknown`, so the difference is the wiring rather than the probe. I traced the seam end to end as well: `TurnFrame` writes `TURN_ASK_KIND_KEY`, `_shared` exports the tolerant reader beside `turn_budgets`, and `turn_path` reads it at :2123 and passes it at the one call site. The three tests you name exist.

Carrying a callable rather than a copy of the prefixes is the right call as far as I can see, and your reason for it holds: `metadata["observers"]` is the entry that gets filed, so this one crosses no serialization boundary.

I did not re-run your revert check, so the independence of those tests is your measurement and not mine.

You resolved this one already; I agree with the closure.

raven/agent/loop/failure_streak.py
LivXue: Verified at `712ed3d60`, by re-running the reproduction rather than by reading the fix.

Both writers moved and the class collapses: the two SSL hostname-mismatch envelopes that gave two classes now give `envelope:sslcertverificationerror` for both, and the proxy pair gives `envelope:proxy error` for both. I also checked the direction a careless fix would have broken, since `detail` keeps the exception text inside the same JSON string that `is_hard_tool_failure` scans: a `timed out` envelope and a `429` envelope both still read as transient rather than hard, so nothing turned retryable into a streak. Thank you for finding the host copy -- I cited only the plugin's, and the host tool is the one that serves every agent.

One disclosure, not a request. I tested the reason given for keeping `_ProviderPageError` composed, because it is a claim about the input space. Five of its eight raise sites are fixed phrases. The other three -- `f"Tavily: {reason or 'extract failed'}"`, `f"Firecrawl: {message}"` and `f"AnySearch: {data.get('message') or 'extract failed'}"` -- interpolate a string taken from the vendor's response body. "Not taken from an exception" holds; "assembled here from a fixed set of phrases" does not hold for those three, and whether a vendor varies that message per URL is not observable from a machine with no provider key. The carve-out itself is right, since two vendors refusing are two causes, so this is only about the variable part of three messages and it is much smaller than what you fixed.

Separately, and outside this PR: `raven/agent/tools/media_gen.py:251` writes `{"error": f"HTTP {status}: {body}{hint}"}` into the same envelope shape on the same `failure_class` path, with the raw response body in the key. Untouched here, so not a finding against this branch. I mention it because the new docstring is right that this is a convention producers keep rather than something the module can enforce, and that is what an unfixed site looks like.

You resolved this one already; I agree with the closure.

LivXue (approved):

PR #411, opened 2026-09-15
feat(rpc): let a client change the playbook library, not only read it

Handsome-wzw:
## Summary

The playbook library has been reachable from `raven playbook` since it shipped
and reachable from nothing else. A client could list playbooks, open one, and
hold their credentials -- and could not enable one, could not check one, and
could not remove one.

The clearest symptom is on the page already: `PlaybooksPage` renders a playbook
as disabled in two places and has never had a way to undisable it. A state drawn
and not actionable.

Three handlers close the gap, each wrapping the call the CLI already makes:

**`playbooks.set_enabled`** takes the state wanted rather than a toggle, so two
clients asking at once agree on where the name lands instead of on who arrived
last. It answers whether this call is what moved it -- "you did that" and "it
was already so" are different answers, and a caller should not need a second
read to tell them apart. Disabling removes a playbook from what the model is
offered and nothing else; it stays runnable by name from the CLI, which is the
user's own hand rather than the model's.

**`playbooks.validate`** answers its findings instead of raising. A playbook that
does not validate is the ordinary reason to call it, so an error code would
dress the normal answer as a broken call -- and would carry one string where the
validator produces a list. Validated against this machine's agent table, because
a playbook is a distribution unit: a step naming an agent this host does not
have is a real finding, and the CLI reports it the same way.

**`playbooks.delete`** refuses a builtin and names the operation that does exist
for it: there is no file of the host's to remove and the next install would put
it back. The name comes off the deny list as the directory goes, for the reason
the CLI records -- a gone name has no business there, and a later playbook
reusing it should start enabled like any other. When a user playbook was
shadowing a builtin, the answer says the builtin is now uncovered, rather than
leaving a caller to re-read the list to discover the name survived.

**One name guard sits in front of all of them.** `PlaybookStore.origin_of`
joins a name to the library root and answers `user` for `../victim`, so it is
the escape rather than the guard -- a check placed after the lookup has already
been told the wrong answer, and `delete` would then remove whatever came back.
`_known_name()` matches `NAME_RE` before the first store call, and every
name-taking handler goes through it, including the seven that existed:
`playbooks.get` and the five credential and oauth handlers. The contract carries
the same `pattern` on each `name` param, so a client is told the shape rather
than only refused at the door.

**`run` and `create` are deliberately absent**, for a reason worth stating
accurately rather than the one that first suggested itself. It is not that they
would hold the socket: this transport dispatches every frame as its own task, so
a slow handler blocks nothing else on that connection, and a `dag` playbook
reached through the loop's runtime already dispatches in the background and
answers with a receipt carrying the run id.

What they actually need is an address. Progress and the completion announce are
addressed to a conversation, which the loop sets per turn; an RPC entry is a
third origin nothing sets one for, and every consumer falls back to a default
`cli:direct`. Answering that -- which conversation a page-initiated run reports
to, and what the graph-level `confirm` gate does when the caller is not a model
-- is the change, and it is a different one from widening the library's reads.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/test_rpc_playbooks.py                          48 pass
pytest -k "playbook or rpc_schema or declare_their_errors" 748 pass
cd ui-web && npx tsc --noEmit -p .                          clean
cd ui-web && npm run gen:check                              matches the contract (171 methods)
cd ui-web && npx vitest run                                 1678 pass
cd ui-tui && npm run lint:rpc / lint:i18n / tsc             pass
scripts/check_source_language.py <base>..HEAD               exit 0
git ls-files | grep -c node_modules                         0
```

`test_rpc_schema_match.py`), and **both** generated clients were regenerated --
`ui-web` and `ui-tui` each hold their own, and each has its own sync gate.

Fifteen new cases (48 in the file, from 33), every one mutation-checked by
reverting the line it covers:

- flip the enabled/disabled polarity -> red
- accept a non-boolean as a state -> red
- let a builtin be deleted -> red
- leave the deny entry behind on delete -> red
- drop the `NAME_RE` check -> 13 red
- keep the check but move it after the lookup -> 5 red
- give one handler its own guard instead of the shared one -> 8 red
- put a default back on one result field -> `test_rpc_schema_match.py` red

The traversal pair had to be rewritten after it passed for the wrong reason: the
library was empty, so `origin_of` returned `None` and the unknown-name branch
raised while the shape guard was never reached. The cases now save a playbook
first and assert `library.origin_of("../victim") == "user"`, pinning the escape
as reachable, and one case spies on `origin_of` to assert nothing resolved at
all.

The validate pair is deliberately two cases: one asserting findings come back
and one asserting a sound playbook answers `ok` with an empty list, so neither is
satisfied by a handler that always returns the same shape.

**Driven on a real machine, not only against fixtures.** A throwaway playbook was
written into this host's user layer, then:

The negative half is the one worth stating precisely, because a refusal can come
from the right place or the wrong one. A `victim` directory holding an
`IMPORTANT.txt` was planted where `../victim` actually resolves, then all nine
name-taking methods were driven with five bad shapes each -- 45 calls over the
socket:

- 45 refused by the shape guard (`playbook names are kebab-case`), 0 refused only
  after a lookup, 0 got through. The distinction matters: a `no playbook named`
  answer would mean the name had already been joined to a path and the guard was
  merely lucky that nothing was there. Before the fix that is exactly what the
  gateway answered.
- `victim` and its file survive.

Then the whole positive round trip on a throwaway playbook written through the
store's own `save`: validate `ok: true` with an empty `errors` list -> disable
(`changed: true`) -> disable again (`changed: false`) -> list shows it disabled
-> enable -> delete (`deleted: true`) -> second delete refused -> gone from the
list.

The host was left as found: the throwaway playbook and the planted directory are
removed, and `~/.raven/config.json` hashes to the backup taken before the check
(`44b439bd`) -- the empty `playbooks: {"disabled": []}` block that
`set_playbook_disabled` creates was removed again.

No UI change here -- this is the interface only, by request. The page still has
no control bound to `set_enabled`; wiring one is a follow-up.

## Risk

Security: `delete` is the one destructive addition, and the first version of it
was exploitable -- a caller-supplied `../victim` resolved to a sibling directory,
which `rmtree` removed. That is fixed and the fix is the wider half of this PR:
a name is matched against `NAME_RE` before it reaches any store call, so it
cannot carry a separator or a `..` segment, and the same guard was extended to
the seven handlers that already existed and had the same hole. `delete` is then
bounded on three further sides -- it refuses a builtin, it refuses an unknown
name, and it removes only the directory the store resolves for a name of that
shape. Nothing here returns a secret; the credentials handlers beside it gained
the guard and are otherwise untouched.

Backward compatibility: three new methods and no change to the seven that
existed. Every new result field is additive. A client that knows none of this is
unaffected.

Rollback: one commit, reverts cleanly. Reverting also removes the contract
entries, which nothing else reads.

User-visible behaviour change: a client can now disable, validate and delete
playbooks over RPC. `delete` is irreversible for a user playbook and is not
confirmed at this layer -- the CLI asks before deleting, and a UI binding this
should ask too. Said plainly rather than left implicit, because the RPC is the
wrong place to put that prompt and the right place is whatever draws the button.

## Related Issues

N/A


gloryfromca (commented):
Blocking: validate playbook names before any store path lookup, and make the new success payloads required.

I found two introduced defects: `playbooks.delete` can escape the playbook root through a traversal name, and all fields in the three new success results are optional in the wire contract. The rest of the change is coherent with the existing CLI behavior and layered library semantics.

I covered the full diff; AGENTS.md/CLAUDE.md and the canonical Playbook, RPC Protocol, Live preference, and Wire Schema constraints in CONTEXT-MAP.md/CONTEXT.md; the CLI counterparts; PlaybookStore and config-update callers; relevant history; backward compatibility; generated clients; and test integrity. The tests only add coverage and do not weaken existing assertions. The direct RPC confirmation choice is consistent with other client-owned destructive confirmations, so I did not raise it.

Verification: `uv run pytest tests/test_rpc_playbooks.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py tests/test_cli_playbook_commands.py -x` passed 421 tests. I also reproduced the traversal entirely under a temporary directory: `../victim` was classified as a user playbook and `playbooks_delete` removed that sibling directory.

raven/rpc/methods/playbooks.py
```diff
@@ -352,6 +352,131 @@ async def playbooks_oauth_clear(params: dict) -> dict:
...
+    if not name:
+        raise ConfigValidationError("name is required")
+    store = _store()
+    origin = store.origin_of(name)
+    if origin is None:
+        raise ConfigValidationError(f"no playbook named {name}")
+    if origin == "builtin":
+        raise ConfigValidationError(f"{name} is a builtin and cannot be deleted; disable it instead")
+
+    directory = store.path_for(name).parent
```
gloryfromca: Blocking: validate playbook names before using them to resolve and delete a directory.

`name` is only stripped and checked for emptiness, while `PlaybookStore.origin_of()` and `path_for()` join it directly to the user root. With an existing user root and a sibling directory containing `playbook.md`, a request with `name: "../victim"` is classified as origin `user`; this line resolves the sibling and the next line recursively deletes it. I reproduced exactly that in a temporary directory and got `deleted: true` with the sibling removed. Enforce the canonical `NAME_RE` boundary before the first store lookup (and represent it in the request contract) so this operation cannot escape the playbook library.

raven/rpc/models.py
```diff
@@ -3876,6 +3876,56 @@ class PlaybooksOauthClearParams(_Strict):
...
 
+class PlaybooksSetEnabledParams(_Strict):
+    name: str
+    enabled: bool = Field(
+        ...,
+        description="The state wanted. true takes the name off the deny list, false puts it on.",
+    )
+
+
+class PlaybooksSetEnabledResult(_Strict):
```
gloryfromca, line 3887: Blocking: declare the fields that every successful handler returns as required.

All fields in this result and the two new result classes below have defaults. Consequently, each model accepts `{}`, OpenRPC emits `required: []`, and both generated TypeScript clients expose every field as optional (`name?:`, `enabled?:`, etc.). That contradicts the handlers, which always return the complete payload, and prevents clients from relying on the advertised success shape. Remove the defaults from fields that are unconditionally present, then regenerate both clients.

raven/rpc/methods/playbooks.py
Handsome-wzw: Confirmed, and worse than the report: `origin_of` is not merely unhelpful here, it is the escape. It answers `user` for `../victim`, so a guard placed after the lookup has already been told the wrong answer. Reproduced against a real library -- the sibling directory and its `IMPORTANT.txt` were removed.

Fixed in 64cf3e4c. `_known_name()` matches `NAME_RE` (`^[a-z0-9][a-z0-9-]*$`) **before** the first store call, and every name-taking handler goes through it -- not only `delete`: `playbooks.get`, the five credential/oauth handlers via `_spec_or_raise`, and the three added in this PR. The same `pattern` is now on each `playbooks.*` `name` param in `rpc-schema/openrpc.json`, so a client is told the shape rather than only refused at the door.

On the tests: the first pair I wrote passed for the wrong reason. The library was empty, so `origin_of` returned `None` and the `no playbook named` branch raised -- the shape guard was never exercised. They now `library.save(...)` first and assert `library.origin_of("../victim") == "user"`, which pins the escape as reachable, plus one case that spies on `origin_of` and asserts nothing resolved at all (`asked == []`). Mutation-checked: removing the `NAME_RE` check turns 13 red, removing the pre-lookup ordering 5, removing the shared helper 8.

Verified live on this host afterwards: all five name-taking handlers refuse `../victim`, and the sibling survives.

The PR description carried the claim "It reaches no path a caller supplies", which was false when written. It has been rewritten to describe the boundary that now makes it true.

raven/rpc/models.py
Handsome-wzw: Agreed. Fixed in 64cf3e4c.

All nine fields across the three new result classes moved from defaults to `Field(...)`: `PlaybooksSetEnabledResult` (`name`, `enabled`, `changed`), `PlaybooksValidateResult` (`name`, `ok`, `errors`, `path`) and `PlaybooksDeleteResult` (`name`, `deleted`, `uncovered_builtin`). `rpc-schema/openrpc.json` carries the matching `required` lists -- it is the source of truth, so the list there is the change and the model mirrors it, held equal by `test_rpc_schema_match.py`.

Both generated clients were regenerated, since the contract feeds two: `ui-web` (`npm run gen:check` -> matches, 171 methods) and `ui-tui` (`npm run lint:rpc` -> in sync, sha256 `bcdedd558cca`). The delete result now reads `name: string; deleted: boolean; uncovered_builtin: boolean;` with no `?`.

Mutation-checked by putting one default back: `test_rpc_schema_match.py` goes red.

raven/rpc/methods/playbooks.py
Handsome-wzw: Re-ran the machine check over the real gateway's WebSocket after restarting it onto this head, because the earlier run in the description predates the guard commit and would otherwise be describing code that is no longer here. (The running gateway was indeed still on the old build -- it answered `no playbook named ../victim`, which is the pre-fix path.)

A `victim` directory holding an `IMPORTANT.txt` was planted where `../victim` actually resolves (`~/.raven/workspace/victim`), then all nine name-taking methods were driven with five bad shapes each -- 45 calls:

```
playbooks.get                    shape shape shape shape shape
playbooks.set_enabled            shape shape shape shape shape
playbooks.validate               shape shape shape shape shape
playbooks.delete                 shape shape shape shape shape
playbooks.credentials.get        shape shape shape shape shape
playbooks.credentials.set        shape shape shape shape shape
playbooks.credentials.clear      shape shape shape shape shape
playbooks.oauth.authorize        shape shape shape shape shape
playbooks.oauth.clear            shape shape shape shape shape

refused by the SHAPE guard: 45   refused only after a lookup: 0   leaked: 0
victim survives: True
```

The two refusal kinds are counted apart on purpose. A `no playbook named` answer is not a pass here -- it means the name was already joined to a path and the call was merely lucky that nothing was there, which is the state this PR started in. Only `playbook names are kebab-case` says the name never resolved.

Positive round trip on a throwaway playbook written through the store's own `save`: validate `ok: true` / empty `errors` -> disable (`changed: true`) -> disable again (`changed: false`) -> list shows it disabled -> enable -> delete (`deleted: true`) -> second delete refused -> gone from the list. Host left as found: playbook and planted directory removed, `~/.raven/config.json` hashes back to the pre-check backup with the empty `playbooks: {"disabled": []}` block gone.

Handsome-wzw:
Both blocking findings are fixed in 64cf3e4c and answered on their threads; re-review requested.

- **Path traversal** — confirmed, and the mechanism was worse than reported: `origin_of` is not a weak check but the escape itself, answering `user` for `../victim`. `_known_name()` now matches `NAME_RE` before the first store call, and the guard was extended to the seven handlers that already existed with the same hole, not only to the three added here. Driven over the real gateway socket: all nine name-taking methods × five bad shapes, 45/45 refused by the shape guard with 0 refused only after a lookup, and a planted `victim` directory survives.
- **Optional result fields** — confirmed. Nine fields moved to `Field(...)`, `required` set in `openrpc.json`, both generated clients regenerated (`ui-web` 171 methods; `ui-tui` sha256 `bcdedd558cca`).

CI is green on this head: 12/12 checks pass, including `unit (py3.12)` with the coverage ratchets and PR diff coverage. Diff coverage was pre-checked locally — every uncovered line in `playbooks.py` is pre-existing (`_store`/`_disabled`, the mcp-server lookup, one oauth error arm); no added line is uncovered.

I have deliberately not resolved either thread. The findings were yours and the evidence that they are closed should be yours to accept -- resolving my own would hide them.

The PR description was rewritten where it was wrong: the Risk section previously claimed "It reaches no path a caller supplies", which was false when written, and the Verification section described a wire run made before the guard existed. Both now describe this head.

raven/rpc/models.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified on 64cf3e4c: all nine fields are required in the Python result models and OpenRPC, both generated clients expose them as non-optional, and each model now rejects an empty success payload. This thread is settled.

raven/rpc/methods/playbooks.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified on 64cf3e4c: the canonical-name check runs before the first store lookup, the temporary traversal reproduction now returns `config_validation_error`, and the sibling directory remains intact. The shared helper also closes the pre-existing read-side variants. This thread is settled.

gloryfromca (commented):

Both prior blockers are fixed on this revision. The name boundary now rejects noncanonical values before the first store lookup, including the traversal that previously reached `shutil.rmtree`; the guard is shared by all name-taking handlers. The three new success models, OpenRPC results, and both generated clients now agree that all nine returned fields are required. I found no new defect in the fix delta or full change.

I covered the complete `github/main...HEAD` diff and the fix delta; AGENTS.md/CLAUDE.md and the Playbook, RPC Protocol, Live preference, and Wire Schema constraints; the RPC callers, dispatcher, PlaybookStore, config updater, CLI counterparts, and relevant history; backward compatibility; generated-client drift; and whether tests were weakened. The tests add negative traversal and positive valid-name cases without removing prior coverage. A bounded security diff pass found no surviving candidate after the new pre-store guard.

Verification: the temporary `../victim` reproduction now returns `config_validation_error` and leaves the sibling directory intact; all three result models reject `{}`. `uv run pytest tests/test_rpc_playbooks.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py tests/test_cli_playbook_commands.py -x` passed 436 tests. The local `ui-web` generator check could not start because `json-schema-to-typescript` is absent from this worktree, so I did not count it as green; the current GitHub `ui rpc contract` and `TUI checks` jobs both pass on this head.

raven/rpc/methods/playbooks.py
0xKT, line 1: The new guard has three clauses and the third one has no gate. Mutating only
it leaves all 48 tests green.

`_known_name` normalises, checks, then dispenses -- and the dispense is
written down as a promise in its own docstring (playbooks.py:221):

    Answers the clean name so a caller cannot go on using the raw one.

That clause is what stops a caller resolving a path from the unnormalised
value. It is the only one of the three the suite does not hold.

WHAT I RAN, in a worktree at head 64cf3e4c merged onto 09a5f7f8, restoring
the file from a saved copy between runs (`git status --porcelain` empty
afterwards):

  0. BASELINE
     bash board/pyt.sh tests/test_rpc_playbooks.py   -> 48 passed

  1. CONTROL -- does the suite read the RETURN VALUE at all?
     `return name`  ->  `return name[:-1]`
     -> 25 failed, 23 passed
        (test_deleting_a_builtin_is_refused_and_says_what_to_do,
         test_a_library_shaped_name_still_works, and 23 more)
     So the value is live under the suite. A mutant surviving below is not a
     selection that never runs.

  2. MUTANT -- withdraw the dispense clause only
     `return name`  ->  `return raw`
     -> 48 passed        <-- SURVIVES

It is not an equivalent mutant. With `return raw` the guard checks one
string and hands the store a different one, so a name that is clean after
`.strip()` but dirty before it -- `"  competitor-scan  "`, or one with a
trailing newline -- passes the shape check and then fails as an uncaught
FileNotFoundError inside the store, which the dispatcher turns into -32603.

WHY NOTHING CATCHES IT. The suite's own traversal table
(tests/test_rpc_playbooks.py, the parametrised names) is
["../victim", "..", "a/b", "/etc", "Up", "a b", "-lead", ""] -- every one of
them is rejected by the shape check, and every other name in the file is
already equal to its own `.strip()`. The one set the dispense clause exists
for, clean-after-strip and dirty-before, is not in the suite.

One name of that shape in the existing table closes it.

For completeness, the other three mutants on this guard are all killed, so
this is the single gap rather than a pattern:
  drop the NAME_RE check entirely            -> 13 failed
  move the check after origin_of             -> 8 failed
     (test_the_shape_is_checked_before_anything_resolves_a_path is the one
      that pins the order, and it is the one that fails -- the docstring's
      "a guard placed after it has already been told the wrong answer" is
      verified, not just asserted)

INTRODUCED BY THIS PR: `git show 09a5f7f8:raven/rpc/methods/playbooks.py |
grep -c _known_name` -> 0. The guard, the promise and the tests all arrive
together here; nothing regressed, the new guard just ships with one of its
three clauses ungated.

Not blocking: the shipped code is correct and the gap is in the guard's
guard. Filed because this is a security boundary, and an untested clause on
one of those is the one that quietly stops holding.

0xKT, line 1: Two error surfaces in the new write methods. Both are narrow, both are
measured, and one of them contradicts a promise this PR publishes to the
generated clients.

(1) playbooks.validate says it answers rather than raises, and a YAML syntax
    error is the one shape that escapes.

    The docstring (playbooks.py:414-417) makes this the method's whole point:

        Answers rather than raises. A playbook that does not validate is the
        normal reason to call this, so the findings are the result -- an
        error code would make the ordinary answer look like a broken call.

    The handler catches `(ValidationError, ValueError)` around `store.load`.
    `yaml.parser.ParserError` is neither:

        ParserError -> MarkedYAMLError -> YAMLError -> Exception

    so it propagates and the dispatcher turns it into -32603 with a raw
    traceback -- exactly the "ordinary answer looks like a broken call" the
    docstring set out to avoid.

    Measured, four broken shapes through the real handler, one temp library:

        yaml-syntax      -> raises ParserError
        no-frontmatter   -> answers ok=False, 1 error
        empty-file       -> answers ok=False, 1 error
        wrong-types      -> answers ok=False, 1 error

    Three of the four are already right, which is what makes the fourth worth
    fixing rather than rethinking: a hand-edited playbook with an unclosed
    bracket is the commonest way this breaks, and it is the one case that
    does not come back as a finding. Widening the except to include
    `yaml.YAMLError` is the whole change.

(2) playbooks.delete decides "is this a builtin" and "which directory" from
    two separate filesystem reads, and acts on the second.

        origin = store.origin_of(name)        # :470  -- reads the layer
        if origin == "builtin": raise         # :471
        directory = store.path_for(name).parent   # :474  -- reads it AGAIN
        shutil.rmtree(directory)              # :475

    `path_for` prefers the user layer, falls back to the builtin layer, and
    falls through to the user path. So if the user directory goes away
    between :470 and :474, the answer to "which directory" is taken on a
    different world than the answer to "is this a builtin".

    There is no `await` between those lines, so two RPC calls on one event
    loop cannot interleave here. It needs a second PROCESS, and the repo has
    one: `raven playbook delete` (raven/cli/playbook_commands.py) removes the
    same directories, with no lock between the two doors.

    WHAT HAPPENS TODAY, measured on a library shaped like the shipped one:

        builtin layer contents: []   (empty)
        [second actor removes the user directory after origin_of answers user]
        handler raises FileNotFoundError -> -32603 with a traceback
        library root intact, builtin layer intact

    So today this is an ugly error, not a destructive act, and I am filing it
    as non-blocking on that basis. I checked before assuming: `git ls-files |
    grep -c playbook.md` is 0 -- no builtin playbook ships, so `path_for`'s
    builtin branch cannot be reached, and commit 2dff96ae removed the
    packaged ones on purpose.

    What makes it worth writing down anyway is that the destructive form is
    one commit away. The code keeps the builtin layer as a first-class
    concept -- `origin_of` answers "builtin", delete refuses it by name, the
    answer carries `uncovered_builtin` -- so the day a builtin playbook ships
    again, the same race points `rmtree` at the packaged directory instead of
    an absent one. I reproduced that form too, by giving the temp library a
    builtin layer: the packaged directory and a file beside it were removed,
    and the call still answered `{"deleted": true, "uncovered_builtin": false}`.

    Resolving the directory once and acting on that one answer closes both
    forms.

INTRODUCED BY THIS PR for both: at 09a5f7f8 neither `playbooks_validate` nor
`playbooks_delete` exists (`git show 09a5f7f8:raven/rpc/methods/playbooks.py
| grep -c 'async def playbooks_delete'` -> 0). The CLI's own delete has the
same two-reads shape and is untouched here, so item 2 is a second door
inheriting the first door's flaw rather than a regression -- worth saying
because the fix belongs in one place, not two.

Neither is blocking.

gloryfromca (commented):

The new revision closes the three measured gaps: padded names are proven to reach the store in normalized form, YAML parser failures return validation findings, and deletion can resolve only a user-layer directory rather than falling through to a builtin. I found no new defect in the delta or the complete change.

I covered the full `github/main...HEAD` diff and `64cf3e4c...HEAD` delta; AGENTS.md/CLAUDE.md and the Playbook, RPC Protocol, Live preference, and Wire Schema constraints; the handler/store/config/CLI callers and relevant history; backward compatibility; test integrity; and the generated-client contract. The tests add focused regressions without weakening existing coverage. I also tried to refute the deletion fix through the remaining post-resolution filesystem race; it requires a second process to land inside a synchronous window, cannot select the builtin layer after this change, and has a working retry path, so it does not meet the late-round blocking bar.

Verification: `uv run pytest tests/test_rpc_playbooks.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py tests/test_cli_playbook_commands.py -x` passed 439 tests. `git merge-tree --write-tree HEAD github/main` exited 0. Current CI has the contract, TUI, page, lint, repository, Windows, kernel, and three of four unit shards passing; the fourth unit shard is still running. My two earlier threads remain resolved, and I did not alter the other reviewer's threads.

gloryfromca (commented):

This head is a content-preserving rebase onto current `main`. `git range-diff` reports all three commit pairs as identical (`81377c8 -> 5a72ede`, `64cf3e4 -> bc95b54`, and `3a137de -> 3434c72`), so the previously reviewed code, callers, history, compatibility, tests, project rules, and RPC/playbook architecture have no new PR delta. The complete `github/main...HEAD` diff remains the same seven-file change.

Verification on the rebased tree: `uv run pytest tests/test_rpc_playbooks.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py tests/test_cli_playbook_commands.py -x` passed 439 tests. Current lint, contract, TUI, page, repository, Windows, kernel, and one unit-shard checks pass; three unit shards are still running. My two threads remain resolved.

raven/rpc/methods/playbooks.py
Handsome-wzw: Reproduced exactly as you wrote it, and fixed in `3434c722`.

Your control is what made this worth acting on rather than arguing about: `return name` → `return name[:-1]` takes 25 tests with it, so the value is live under the suite, and `return raw` surviving is a real hole rather than a selection nothing reaches. I ran both before changing anything and got the same two numbers.

The clause now has a test of its own — `"  competitor-scan  "` through `playbooks.get`, asserting the playbook that comes back is `competitor-scan`. Your mutant dies on it: `return name` → `return raw` is now 1 failed, 50 passed.

I put it as a positive case rather than a new row in the traversal table, because the table asserts refusal and this name must be *accepted*. It is the one shape you identified — clean after a strip, dirty before — and the test says so in as many words, since the reason it is absent from the rest of the file is that every other name there already equals its own strip.

Taking your framing seriously about where this matters: an untested clause on a security boundary is the one that quietly stops holding. The other three mutants you list are still killed, so the guard now has all four clauses held.

Handsome-wzw: Both reproduced, both fixed in `3434c722`.

**(1) The YAML shape.** Confirmed through the real handler on a temp library: a file whose spec block reads `nodes: [ unclosed` raises `ParserError`, and your chain is the reason — `ParserError → MarkedYAMLError → YAMLError → Exception`, so neither arm of `(ValidationError, ValueError)` catches it. `yaml.YAMLError` added to the catch; the same file now answers `ok=False` with the parse failure as its finding. A regression test covers it, and removing `yaml.YAMLError` again turns that test red.

The reason your framing landed: three of the four broken shapes were already right. That is what makes this a hole in one arm rather than a sign the method wanted rethinking.

**(2) The two reads.** Fixed by resolving once rather than by ordering the two reads more carefully, since ordering alone still leaves two answers about two views of the disk.

`PlaybookStore.user_directory(name)` answers both questions a caller intending to remove has — which directory, and whether it is the user's to remove — from a single lookup. `delete` acts on that, and `None` covers "packaged" and "already gone" alike, which are exactly the two cases where there is nothing here to remove. Your race now produces a typed refusal (`x-scan is no longer in the library; nothing was removed`) instead of a `FileNotFoundError` reaching the dispatcher as -32603.

That also closes the destructive form you reproduced, and closes it structurally rather than by luck: the path handed to `rmtree` was proved to be in the user layer at the moment it was resolved, so it cannot be a packaged directory no matter what the second process did in between, and no matter whether a builtin playbook ships again.

Your note that the CLI carries the same two-reads shape is right, and I have left it alone here — the fix belongs in one place, and `user_directory` is now that place for whoever takes the CLI door next. I would rather that be a deliberate second change than a drive-by in a PR about the RPC surface.

Both mutations are held: putting the `except` back to two arms, and putting `delete` back on `path_for`, each turn one test red.

0xKT: Verified at 3434c722. I re-ran the mutation rather than reading the diff:

  `return name` -> `return raw`
     before your fix   48 passed          <- the clause had no gate
     now               1 failed, 50 passed
        test_the_name_that_reaches_the_store_is_the_cleaned_one

So the dispense clause is held now, and by its own test rather than as a side
effect of another one.

I also re-checked the control I used when filing this, to be sure the gate is
not passing for a different reason: `return name[:-1]` still reddens a wide
set, so the suite reads the value, and the new test is what distinguishes the
raw string from the cleaned one.

Closing.

0xKT: Verified at 3434c722, both halves, and the second one is a better fix than the
one I suggested.

(1) The YAML escape. Mutation:

  drop `yaml.YAMLError` from the except tuple
     -> 1 failed, 50 passed
        test_a_file_yaml_cannot_parse_is_a_finding_like_any_other

Held, and by a test that names the case rather than a general "validate does
not raise".

(2) The delete race. I re-ran the reproduction from my note -- the same
injected interleave, a builtin layer present, the user directory removed after
`origin_of` has already answered `user`:

     before   the packaged builtin directory and a file beside it were removed,
              and the call still answered {"deleted": true}
     now      ConfigValidationError: "demo is no longer in the library; nothing
              was removed", builtin layer intact

I asked for one resolution acted on once. `store.user_directory` is more than
that: it answers "which directory" and "is it the user's to remove" from a
single read, so there is no second view of the disk for the first answer to go
stale against. The window is gone rather than narrowed, and the docstring names
the second writer -- `raven playbook delete` -- which is the part that makes it
checkable by the next reader.

One thing I did NOT verify, so it is yours to weigh: I drove the race through
an injected interleave, not through two real processes. The structural argument
is what convinces me here, and it does not depend on my harness.

Closing.

0xKT (approved):

PR #413, opened 2026-09-15
ci(*): run unit tests as four shards and speed up the suite

0xKT:
## Summary

The unit job takes twelve minutes on one runner. The first cut of this PR took the production waits out of the tests, which made the suite a quarter faster on a laptop and moved CI by 14 s (pytest 720.95 s on main, 706.57 s here): the runner is three to four times slower than a laptop on CPU-bound tests, and two of the twelve minutes are the four xdist workers importing every test module. So the PR now attacks the wall clock from every side that CI's own numbers pointed at.

- `ci(*)`: the unit job runs as four shards. pytest gains `--shard K/N` in `tests/conftest.py`, decided in `pytest_ignore_collect` so a shard never imports the other shards' files; test files are dealt round-robin over their sorted list so the slow families that sort together (the ppt engine's) are spread out. Each shard writes coverage data and no report; a `coverage` job downloads the four data files, combines them, writes `coverage.json` and `coverage.xml`, and runs the gates that used to follow the tests.
- `perf(ppt-engine)`: `menu()` caches its reading of a template, capped at eight entries. One bind reads the bound template and then every other bundled template for the borrow offer; with ten templates bundled, every bind evicted what the next one needed and re-measured all of them with freetype. The cap is 32 and a hit moves its entry to the back. On this PR's first CI run the template-tool file alone was 257 s of the over-ceiling list, its tests 10 to 26 s each.
- `test(*)`: an autouse fixture collapses the provider chat retry ladder, the loop's LLM-error ladder and the memory store backoff to a millisecond per step, and the steer hook grace to one poll. It keeps each tuple's length because both ladders count attempts with `len()`; the steps are a millisecond rather than zero because the store pipeline waits with `wait_for(stopping.wait(), timeout=delay)`, and a zero timeout cancels that wait before it runs. A test that proves a timing property opts out with the new `production_timing` marker. Ten tests spent 257 s of the suite's worker time in those waits; none asserted on the time.
- `test(*)`: a wall-clock ceiling. Setup plus call over 2 s on a test marked neither `slow` nor `production_timing` is listed at the end of the run; `--wall-clock-ceiling-strict` fails an otherwise green session on it. Warning only here; CI flips to strict once the remaining tail is marked.
- `test(*)`: every test created two numbered temp directories, and pytest picks each number by scanning the base temp dir; the fake home is now a sibling of `tmp_path`. `MPLCONFIGDIR` points at a git-ignored dir in the checkout, because the fake home changed per test and matplotlib rebuilt its font list from the system fonts on every run.
- `refactor(agent)`: the outer LLM-error retry ladder default `(15, 30, 60)` lived in three places; it is one constant in `raven/config/schema.py` now, read by both `RecoveryLimits` and `_ladder`. Behaviour unchanged.
- `test(updates)`: the first sharded run turned up an order dependency. `beta_channel` binds `ReleaseInfo` from `raven.updates.upgrade` at import, the update-notice tests stand a fake in for that module through `sys.modules`, and `_refresh` imports `beta_channel` lazily; a worker that imported `beta_channel` for the first time inside that test got an `ImportError` swallowed as a failed refresh. The test module now imports `beta_channel` at the top.
- `ci(*)`: the PR diff gate named every `.py` under `plugins-dist` as production code while coverage measures only `raven` and `raven_everos`, so a change to the ppt or design engine could never appear in the report and the gate refused every such PR on sight; the first sharded run hit it on `menu.py`. The pathspec now names `plugins-dist/everos-memory`, and a test holds every production pathspec to a package some `--cov` flag in the Makefile measures.
- Two slow tests were not waits: the sandbox second-cancel test cancelled one tick too early (two pending cancels coalesce into one `CancelledError`) and now waits, under a one second timeout, for the kill to land; the two thousand-session dispatch tests wrote a span per refused enqueue through a store that opens its file per span, so tracing is off for them. Follow-up on the product side: `raven/tracing/store.py` should reuse its handle.

The suite has 30 upper-bound assertions of the form `assert elapsed < X`; shortened delays can only make those looser.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [x] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Whole suite, `uv run --frozen --all-extras pytest -q` (macOS, 4 workers), 22051 passed / 43 skipped each time:

| tree | wall clock | unmarked tests over 2 s |
|---|---|---|
| main before this branch | 372.82 s | (no ceiling yet) |
| after the waits were taken out | 279.65 s | 63 |
| this branch | 224.25 s | 39 |

The shards partition the suite exactly (`uv run --frozen --all-extras pytest --collect-only -q -n 0 --shard K/4`, K = 1..4):

| whole | shard 1 | shard 2 | shard 3 | shard 4 | union | ids in two shards |
|---|---|---|---|---|---|---|
| 22095 | 5058 | 5555 | 5955 | 5527 | 22095 | 0 |

196 / 195 / 194 / 194 test files per shard. A shard collects in 7 s where the whole suite takes 23 s on the same machine. `--ignore` still works under `--shard`: with four `--ignore` flags on shard 1, the ignored file that shard owns is gone from its collection (5034 ids against 5058).

Combining loses nothing. Two small runs with `--cov=raven --cov=raven_everos --cov-branch --cov-report=` and `COVERAGE_FILE` set, then `make coverage-combine COVERAGE_DATA_DIR=<dir>`, against one `make coverage` of the whole suite on main:

| report | files | statements | branches |
|---|---|---|---|
| one run of the whole suite | 704 | 82634 | 25548 |
| two shards combined | 704 | 82636 | 25548 |

Same file set; the two extra statements are the new constant and its import. A source package's never-imported files reach the combined report through the shard that imported it, as they reached the single run.

The menu cache, the three files that bind templates (`uv run --frozen --all-extras pytest -q tests/test_ppt_engine_template_tool.py tests/test_ppt_engine_outline_checks.py tests/test_ppt_engine_default_templates.py`): 101 s of worker time in the last profile, now `196 passed in 11.84s` with no test over 2.9 s. The files that call `menu()` directly: `219 passed in 8.06s`.

The font cache, the formula render test at `-n 0`: 8.91 s cold, 0.69 s on the next run. Seven home-sensitive files: `484 passed in 5.30s`.

Ceiling guard, both modes, with the ceiling forced low so it has something to list:

```
uv run --frozen --all-extras pytest -q -n 0 --wall-clock-ceiling=0.01 tests/test_import_cycle_budget.py tests/test_updates_install_guard.py
uv run --frozen --all-extras pytest -q -n 2 --wall-clock-ceiling=0.01 --wall-clock-ceiling-strict tests/test_updates_install_guard.py
```

First: 28 passed, five listed, exit 0. Second: 24 passed, four listed, exit 1 (the controller flips the status; workers do not).

`make lint-python` clean, `make lint-imports` passes. The CI run on this PR is the measurement that matters: four `unit` shards and one `coverage gates` job.

- [ ] User-facing docs or screenshots are updated when needed (none: CI plumbing, a cache cap, test setup)

## Risk

- The coverage gates now run in a job after the four shards. A red shard skips the gates, a deliberate trade: combining three of four data files would report a coverage drop that is not one, and the PR is red either way. The gate scripts and their thresholds are unchanged.
- The diff gate no longer sees `plugins-dist/ppt-engine` or `plugins-dist/design-engine` at all, as it could never judge them; measuring them (`--cov=raven_ppt --cov=raven_design`, a new ratchet baseline) is the follow-up that would bring them under the gate.
- `--shard` is opt-in; a plain `pytest` run collects everything as before. Shard data artifacts are kept one day, the combined reports 30 days as before.
- No product behaviour change: the ladder default is the same tuple read from one place; the menu cap only changes how long the cache keeps entries.
- Test-only risk: a test that implicitly relied on a production-length retry delay would see different timing. The full suite passed with the fixture on, and a single test can opt out with `production_timing`.
- Rollback: revert the PR. The unit job goes back to one runner with the gates inline; both markers are additive.

- [x] Security impact considered (no security surface touched)

## Related Issues

N/A


gloryfromca (commented):

I reviewed `github/main...HEAD` and the affected callers/history. The shared retry default keeps the production tuple unchanged; the per-test fixture preserves retry counts and uses positive shortened steps where zero would change cancellation semantics; and the `slow`/`production_timing` markers exempt tests from reporting without deselecting them. I also exercised the wall-clock hook under serial and xdist execution: normal mode reports over-ceiling tests, while strict mode changes an otherwise green session to a failure on the controller. I found no weakened assertion or bypassed behavior.

Coverage included AGENTS.md/CLAUDE.md and the Runtime context terms, every changed file, affected callers and commit history, backward compatibility, test-integrity concerns, and the layer/import contracts. `lint-imports` kept all 10 architecture contracts; Ruff check/format, source-language validation, commit-message validation, and `git diff --check` passed.

Verification: the four directly changed test modules passed, 222/222, under the repository's normal four-worker run. A serial strict run executed the same 222 assertions successfully and correctly failed the session for one setup-plus-call ceiling hit. After syncing the repository workspace packages, the full default suite completed with 21,784 passed, 65 skipped, and one failure in the untouched July-era CLI theme test; that same theme test fails alone and has no path through this diff. The new ceiling reported 75 warning-only outliers, consistent with this PR's documented staged rollout.

Makefile
```diff
@@ -11,6 +11,11 @@ COVERAGE_DIFF_THRESHOLD ?= 90
 COVERAGE_RATCHET_TOLERANCE ?= 0.05
 DOCKER_IMAGE ?= raven:local
 COVERAGE_REPORT_ARGS = --cov=raven --cov=raven_everos --cov-branch --cov-report=term-missing:skip-covered --cov-report=xml --cov-report=json --cov-report=html
+# One slice of the suite, K/N; the shard writes its .coverage data and no report.
+COVERAGE_SHARD ?= 1/1
+COVERAGE_DATA_ARGS = --cov=raven --cov=raven_everos --cov-branch --cov-report=
```
gloryfromca, line 16: Blocking: the sharded coverage command omits raven_ppt, so this revision's menu.py change is absent from coverage and the required diff gate fails

`COVERAGE_DATA_ARGS` traces only `raven` and `raven_everos`. The four shard jobs therefore pass and combine, but `make coverage-diff` reports `plugins-dist/ppt-engine/raven_ppt/services/template/menu.py` as a changed production file missing from the report and exits nonzero. Please make every production file changed by this revision measurable by the required gate.

gloryfromca (commented):
Blocking: restore diff coverage for the changed ppt-engine file and update the PR body for the merged CI/cache delta.

I reviewed the three commits added since the previously reviewed head, then checked the cumulative diff and affected callers/history. The code blocker is inline: all four unit shards pass and combine successfully, but the required [coverage gate](https://github.com/EverMind-AI/Raven/actions/runs/34938394252/job/104281969418) rejects `menu.py` because that changed production package is absent from the coverage report.

The PR description also remains at the prior revision: its Summary, Verification, and Risk sections do not mention the CI sharding, persistent matplotlib cache, per-test home allocation, or ppt-engine menu-cache change merged from #417. Please update those sections before squash merge so the permanent merge record describes and verifies the revision that actually lands.

Verification: `tests/test_ppt_engine_template.py` passed 142 tests with one LibreOffice-dependent skip. A collection comparison found the four shards contain 21,843 unique default tests, exactly matching unsharded collection, with no overlap, missing test, or extra test. I also investigated an isolated `update_notice` failure: the test passes with its companion module and its actual CI shard passes, so I dropped it as a pre-existing collection-order weakness on untouched lines rather than treating it as a regression. `git diff --check` passed, and current CI passes the Python lint and import-contract jobs.

Coverage of this review included AGENTS.md/CLAUDE.md and the relevant context constraints, the diff, callers, commit history, backward compatibility, test-integrity/weakening concerns, and architecture/import boundaries. I found no additional surviving issue in the new cache or fixture behavior.

gloryfromca (commented):
Blocking: the existing coverage configuration must measure the changed ppt-engine file before this revision can merge.

The new commit correctly removes the update-notice collection-order dependency, and the updated title and PR description now cover the cumulative CI/cache scope. Local verification passed `tests/test_updates_update_notice.py` both alone and under `--shard 3/4`, 38/38 each. On the current CI run, all four unit shards passed; coverage combination, the overall ratchet, and the baseline check also passed.

The existing inline finding remains open because the current head reproduces it unchanged: `PR diff coverage` reports `plugins-dist/ppt-engine/raven_ppt/services/template/menu.py` missing from the coverage report and exits nonzero. At this fourth review round, this meets all three blocking criteria: this revision introduces the changed unmeasured production file, every ordinary PR run reaches the required diff gate, and an operator has no working green-check path without a branch change.

I rechecked the new delta and cumulative diff against AGENTS.md/CLAUDE.md and the relevant context constraints, including affected callers and history, backward compatibility, test-integrity/weakening risk, and architecture/import boundaries. The current lint and import-contract jobs pass, `git diff --check` passes, and I found no additional issue worth raising.

Makefile
gloryfromca: No blockers; this can merge as far as I am concerned.

Resolved at `78e4bd8ea86b`: history confirms the broad `plugins-dist` wildcard exceeded the gate’s intended EverOS scope, while coverage has only ever measured `raven` and `raven_everos`. The narrowed pathspec restores that boundary, the new contract test keeps pathspec packages tied to real `--cov` sources, and the exact-head diff-coverage job passes. I accept this as the fix and am closing the thread.

gloryfromca (commented):

I reviewed the new coverage-scope commit and rechecked the cumulative `github/main...HEAD` patch, affected callers, and history. The gate now evaluates exactly the distributions its report measures (`raven` and `raven_everos`); the former all-plugins wildcard contradicted its original EverOS-specific comment and made any PPT/design change unconditionally fail. The new contract test prevents that pathspec/source mismatch from returning. I accepted the fix, replied to my prior finding, and resolved the thread.

Verification at this head: 53 focused coverage-gate and update-notice tests pass; the four unit shards, combined coverage, diff gate, ratchet, baseline, lint, import contracts, platform checks, and other required CI jobs all pass. `git diff --check` is clean. I also confirmed the shard union still matches the unsharded default collection without overlap, so the speedup does not silently drop tests.

Review coverage included AGENTS.md/CLAUDE.md and relevant context constraints, the full diff, callers and commit history, backward compatibility, test-integrity concerns, and architecture/import boundaries. Patch-risk assessment: moderate impact if wrong, low regression likelihood, strong protection, easy recovery, and high confidence; merge is supported, with human review still appropriate because this changes broad shared test/CI defaults. No additional finding survived refutation.

gloryfromca (dismissed):

gloryfromca (approved):

gloryfromca (commented):

I reviewed the changes since the prior clean head and rechecked the cumulative diff. The shard hook now returns `None` for files this shard owns, preserving pytest's first-result `--ignore` and `collect_ignore` handling; its index follows `python_files` and `norecursedirs`; and the CI matrix carries explicit `K/N` values instead of deriving the denominator from total matrix jobs. The related state resets, conditional matplotlib-cache creation, documentation correction, and coverage-contract failure message are consistent with their callers and intended behavior.

Verification: `tests/test_coverage_gate.py` passes 15/15. An independent collection comparison found all 21,844 local default nodes exactly once across the four shards, with zero missing, extra, or overlapping nodes; ignoring a shard-owned file removed exactly its 26 nodes and changed nothing else. Exact-head CI passes all four unit shards, combined coverage and diff gates, lint/import contracts, platform, UI, bridge, and wheel checks. `git diff --check` passes.

I covered AGENTS.md/CLAUDE.md and relevant context constraints, the incremental and cumulative diffs, affected pytest/CI callers and history, backward compatibility, test-weakening risk, and architecture/import boundaries. The only thread I opened remains correctly resolved. No candidate finding survived refutation.

PR #415, opened 2026-09-15
fix(everos-memory): one refused write must not take the rest of the run with it

gloryfromca:
## Summary

An import of five memory-file sources lost four of them to a single bad record. Diagnosing it turned up three separate faults; two are ours and are fixed here.

**One answered-but-failed request must not kill the backend.** The demotion classifier read any exception that was not a timeout or a connection refusal as a dead service. EverOS returned a 500 for one unprocessable payload -- a status line, which is proof the service is up and answering -- and the backend went UNRESPONSIVE. Every later write then short-circuited on the state guard and returned "not landed" without reaching the wire, so the two sources queued behind the bad one failed in milliseconds with no network activity at all. An `httpx.HTTPStatusError` now leaves the state exactly as it was. A service that does go away still demotes, through the timeout and connection-refused paths that already existed.

**The short write budget was sized for the wrong caller.** A non-final append got a flat ten seconds, which is a turn's patience for a handful of messages. An append is not free work -- EverOS may carve a boundary out of one, which runs a model -- and the importer hands over up to a hundred messages at a time. That call overran a server that was still working on it, which the classifier above then filed as a hung service. The budget now follows the slice's size, floored at the turn-sized value, so a turn is unaffected and a bulk batch gets proportionate patience.

**A config key nobody declared.** The onboarding wizard records a `port` in the plugin's config slice, but the manifest's `config_schema` never listed it, so admission logged a passthrough warning on every backend build. Declared.

The third fault is upstream and not fixed here: EverOS's episode extraction retries malformed model output twice, then makes the final attempt outside the try block, so a third bad response escapes as an unhandled `JSONDecodeError` and the whole `/memory/add` request becomes a 500. Reporting the failure is right; a 500 with a traceback is not the way to report "this content could not be processed", and EverOS already has a shape for "content accepted, nothing extracted". With the first fix above, that upstream behaviour costs one source instead of the rest of the run.

## Type

- [x] Fix

## Verification

- Reproduced on a real isolated install (temporary RAVEN_HOME, isolated EverOS root and port) with the everos server log captured: `EverosBackend.store timed out after 10.0s` on the 100-message batch, and `unhandled_exception exception_type=JSONDecodeError path=/api/v2/memory/add` for the 500, followed by two sources failing with no request logged at all.
- A billing hypothesis was ruled out by rerunning against a key with headroom: the server log holds only `200` x6 and `500` x1, no 401/403/429.
- Each fix was mutated away and the new test confirmed red, then restored and confirmed green.
- `uv run pytest tests/test_everos_backend.py tests/test_plugin_manifest.py tests/test_plugin_config_admission.py tests/test_core_plugin_stack.py tests/test_plugin_registry.py tests/test_everos_discover.py tests/test_everos_server.py`: 294 passed, 1 skipped (inotify is Linux-only).
- `uv run ruff check plugins-dist/ tests/` and `ruff format --check` on the changed files: clean.
- `scripts/check_commit_messages.py`, `scripts/check_large_files.py`, `scripts/check_source_language.py` over the branch range: clean.

## Risk

User-visible changes:

1. A backend that answers with an error status stays usable. Previously one such response silenced memory for the rest of the process, recoverable only through an out-of-band probe; now each request stands on its own.
2. A bulk append waits longer before giving up. A turn's append is unchanged in practice -- the floor is the old value -- and the ceiling only grows with the number of messages handed over in one call.

Both are local to the EverOS plugin's write path. Rollback is a revert; neither touches stored data or the wire protocol.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the complete `github/main...HEAD` diff and the surrounding EverOS state machine, HTTP adapter, importer batching, StorePipeline caller, plugin admission path, and relevant history. I also checked backward compatibility, the architecture/domain rules in `AGENTS.md`, `CLAUDE.md`, and `CONTEXT-MAP.md`, and whether the changed tests weakened the old guarantees.

The failure-isolation behavior holds: an `HTTPStatusError` still reports the individual write as not landed but no longer changes `READY`, so the next source reaches the wire; timeout and connection-refusal paths still demote. A direct two-write reproduction produced `False`, then `True`, with two adapter calls and state still `ready`. The scaled non-final budget preserves the existing ten-second floor and gives the importer's maximum 100-message batch 60 seconds; final flushes keep the existing six-minute budget. The `port` declaration matches the integer value persisted by onboarding and only removes the admission warning.

Local verification:
- `uv run --extra dev pytest tests/test_everos_backend.py tests/test_everos_http_adapter.py tests/test_importer_orchestrator.py tests/test_everos_plugin_discovery.py tests/test_plugin_config_admission.py -x`: 213 passed.
- `uv run pytest tests/test_everos_backend.py tests/test_plugin_manifest.py tests/test_plugin_config_admission.py tests/test_core_plugin_stack.py tests/test_plugin_registry.py tests/test_everos_discover.py tests/test_everos_server.py`: 295 passed.

The initial default-environment attempt collected no tests because the optional `raven_everos` workspace package was not installed; installing the repository's `dev` extra resolved setup, after which both suites above passed. No worktree files were changed.

gloryfromca (commented):

The only revision change is a clean merge of current `main`. The three PR-owned files are byte-for-byte unchanged from the previously reviewed revision, there are no merge conflict resolutions, and the full `github/main...HEAD` diff remains limited to those same files.

For this revision I rechecked the full diff, affected callers and history, repository rules (`AGENTS.md`, `CLAUDE.md`, and `CONTEXT-MAP.md`), architecture boundaries, backward compatibility, and the changed-test assertions; the merged base does not alter those conclusions or weaken the coverage. Under the newly merged test harness, `uv run --extra dev pytest tests/test_everos_backend.py tests/test_everos_http_adapter.py tests/test_importer_orchestrator.py tests/test_everos_plugin_discovery.py tests/test_plugin_config_admission.py -x` passed all 213 tests. All completed GitHub checks are green; the combined coverage job was still running when this review was posted. No review threads exist to sweep, and no worktree files were changed.

0xKT (approved):

PR #416, opened 2026-09-15
chore(*): gate call correctness with a pinned type checker

gloryfromca:
## Summary

Nothing in this repo could see a signature change reach its callers. Ruff reads one file at a time, the import contracts read edges rather than calls, and the suite only covers lines it executes. The consequence was concrete: a required `sandbox_home` argument added to `BoxliteExecutor` reached `plugins-dist/design-engine`'s render path with CI green, and a reviewer -- not a gate -- found it.

`ty` answers that question. This adds it as a pinned dev dependency, a `make lint-types` target and a CI step, with **110 of its 133 rules live**: nine switched off with their measured backlog recorded beside each, fourteen shipped disabled upstream and untouched.

Getting there meant clearing 81 diagnostics. This PR folds what were three stacked PRs (#422 and #423, now closed); each rule's story is a commit of its own.

### 1. The gate, at 108 rules (`chore(*): gate call correctness...`)

21 diagnostics to clear, five of them real:

| Where | What |
|---|---|
| `evolver/orchestrator/production.py` | `design_of` annotated as taking two parameters while its docstring, its only implementation and its one call site all use three |
| `raven/cli/provider_commands.py` | `dict[str, callable]` -- the builtin function, not a type, so the annotation asserted nothing |
| `raven/config/schema.py` | `Config` is a `BaseSettings` but declared `env_prefix` / `env_nested_delimiter` through pydantic's `ConfigDict`, which does not carry those keys |
| `raven/spine/delivery.py` | `StreamDelta.conversation_id` is optional and was used as a `dict[str, str]` key; a delta without one stored a `None` entry the close path then read as "no open stream", losing that stream's done marker |
| `research_flow/support/harness_text.py` | a predicate table typed its matchers as `object`, so nothing checked they are callable |

The rest were unproven invariants and annotations wider than the code: three late-bound callables asserted where a comment used to carry the claim, one closure binding the callable it guarded instead of re-reading it, four `hasattr` probes that now check what the call actually needs, and a contextmanager yielding the value it just set.

### 2. `unresolved-import` (`fix(*): resolve the imports the type gate could not`)

21 diagnostics; three were stale module paths nothing had run into:

- `HarnessNode.current_core_version()` read `raven.__core_version__`, a module that exists neither here nor in the wheel `scripts/build_core_wheel.py` produces. The read sat inside `try/except ImportError`, so the except branch was the only branch and **every node `store.py` saved recorded `core_version="unknown"`**. It reads `raven.__version__` now and returns `0.1.13`.
- `raven/agent/loop/wiring.py` imported `SkillForgeRouterConfig` from `raven.config.schema`; it lives in `raven.config.raven`. Under `TYPE_CHECKING`, so the annotation simply meant nothing.
- three scripts imported `raven.context_engine.base` and `raven.memory_engine.base`, neither of which exists. **Two failed at module import** -- running them was not possible.

The other eighteen were the environment: the three agent products keep their plugin package beside them, so their roots are declared in `[tool.ty.environment]`; `lxml` ships a compiled `etree` with no type information, so `lxml-stubs` is a dev dependency (real types rather than five silenced lines); `uno` and `com.sun.star.*` are suppressed per line, since that worker runs inside LibreOffice's own interpreter.

### 3. `invalid-return-type` (`fix(*): make the declared return types true`)

39 diagnostics, and **one stale annotation held eight of them**: `run_turn` and the `_run_turn` behind it were declared `-> LoopOutcome` while all three of their returns build a `TurnOutcome` -- a different dataclass with no field in common. Nothing broke, because the spine worker reads what it actually gets and `LoopOutcome` never leaves `_run_agent_loop`. The annotation was the only thing wrong.

The other 31: nine annotations narrower than the code (five `return tuple(<genexpr>)` typed as fixed-arity tuples, an inner `one()` still declaring the two-tuple it returned before a third element was added, `image_block`'s TypedDict), five declarations refusing a value the function returns (four tool `execute` overrides saying `-> str` while returning `ToolResult`), seven invariants that lived in a `# type: ignore` comment, and ten honest widenings.

### 4. What the gates then caught in this PR's own work

- The qq factory guard passed locally only because an earlier test had left an event loop behind. Reproduced the CI state (`asyncio.set_event_loop(None)`), then made the test async so the loop is its own.
- `invalid-return-type`, while still disabled, hid a regression introduced here: the qq factory returning the class instead of an instance. Found by diffing the rule histogram, which is why that rule is on now.
- The diff-coverage gate named four lines in `raven/routing/cache.py` -- the asserts this work had just added, sitting inside a `try` whose `except Exception` would have swallowed the `AssertionError` and served the stale cache. `_do_refresh` returns its data now; the asserts, their noqa and the swallowing all go with it.

`child_scope`, `QQChannel.start()` and `BenchmarkCache` had no tests at all and have them now. Every one was mutation-checked.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [x] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `make lint-types` -> All checks passed; 81 diagnostics cleared across the three rule sets
- Gate proven live three times, once per rule: a missing argument, an import of a module that does not exist, and a function returning a str under `-> int` each turn it red, and removing the probe turns it green
- `make coverage` locally: **22050 passed, 45 skipped, 4 failed**; diff coverage 92.59%
- All four failures are **pre-existing on `main`**, checked by a real checkout in the same environment where they fail identically: two `test_the_products_tool_face_*` launcher tests, `test_a_ceiling_as_large_as_the_window_still_leaves_room_for_history` (`reserved_tools = 6044 > 6_000`), and `test_a_missing_key_is_said_once_rather_than_per_query` (environment-dependent: it asserts `SERPER_API_KEY` is absent)
- Rule counts measured with both `--extra dev` and `--all-extras` -- identical, so this gate's verdict does not depend on which optional packages are installed

## Risk

Behaviour changes are confined to the fixes listed above, and each replaces an error path with a correct one: nodes are stamped with the real raven version instead of `"unknown"`, two scripts that could not start now start, the stream bookkeeping no longer stores a `None` key, and a cache refresh returns what it fetched. Everything else is a declaration moved to match what the function already returned.

`tests/` is out of the gate's scope (113 diagnostics under the same rules). Tests are executed, so a broken call there is already loud; this gate is for code CI does not run. Bringing them in is sized, separate work.

Rollback: revert the whole PR to drop the gate and its fixes together, or delete the `[tool.ty]` block, the `lint-types` target and its CI step to keep the fixes and drop only the gate.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the full `github/main...HEAD` diff and the relevant callers and history for the CI/Makefile gate, every behavior-changing type cleanup, QQ construction, stream bookkeeping, provider resolution, config loading, detached upgrade launch, and the bundled agent/plugin paths. I found no concrete runtime, backward-compatibility, test-integrity, or architecture regression. No tests were weakened; the QQ factory change adds a focused regression test.

Nonblocking follow-up, kept here rather than as an open thread: the explanatory rule counts in `[tool.ty]` should be reconciled. The block explicitly disables eleven rules, not nine, and an all-rule scan with pinned ty 0.0.80 reports diagnostics in 21 of 133 rule categories, so the stated `124 of 133` baseline is not reproducible from the checked-in configuration. This does not change the gate that CI actually executes.

Verification: the exact frozen command behind `lint-types` passed; 934 focused tests across all touched behavior areas passed; Ruff check and format, all 10 import contracts, deptry, `uv lock --check`, source-language validation, and `git diff --check` passed. The local `make lint-types` wrapper itself could not be invoked because this environment has no `make`, so its exact `uv` recipe was run directly. GitHub's Python lint job is also green; the unit job was still running when this review was posted.

Coverage included `AGENTS.md`/the linked `CLAUDE.md`, `CONTEXT-MAP.md` and the Runtime architecture terms, the diff, callers, relevant history, backward compatibility, test weakening, and architecture boundaries.

gloryfromca (commented):

No new findings on this revision. The comment-only delta makes the live-rule arithmetic reproducible: 118 upstream-live rules, minus eleven local disables, plus `blanket-ignore-comment`, equals 108. It does not alter the configured gate or any runtime behavior. The previously recorded wording follow-up remains nonblocking and is not repeated here.

Verification on this head: the exact frozen `ty` command passed, all 934 focused tests across the behavior-changing areas passed, and `git diff --check` passed. GitHub's Python lint, pre-commit, commit-message, repository-file, Windows self-upgrade, kernel-wheel, TUI, UI RPC, page, and bridge checks are green; the unit job was still running when this review was posted.

This round covered the new delta, the full PR diff, the applicable `AGENTS.md`/linked `CLAUDE.md` and Runtime context constraints, callers and relevant history from the prior reviewed revision, backward compatibility, test integrity, and architecture boundaries.

gloryfromca (commented):

No new findings on this revision. The comment-only delta was checked against pinned ty 0.0.80: its diagnostic totals, mixin/optional breakdown, and cited override/import examples are reproducible, and no rule severity or runtime behavior changed. The previously recorded wording follow-up remains nonblocking and is not repeated here.

Verification on this head: the exact frozen `ty` gate passed, all 934 focused tests across the behavior-changing areas passed, and `git diff --check` passed.

This round covered the new delta, the full PR diff, applicable repository and Runtime context rules, the relevant checker output and history, backward compatibility, test integrity, and architecture boundaries. No review threads from this reviewer are open.

gloryfromca (commented):

No new findings on this revision. The delta only makes the QQ factory regression test run inside pytest's configured async loop; this removes dependence on ambient thread-loop state while preserving the behavior assertion, and it changes no production path. The existing nonblocking item is unchanged and is not repeated here.

Verification on this head: the exact frozen `ty` gate passed, all 35 tests in `tests/test_channels_qq.py` passed, source-language validation passed, and `git diff --check` passed.

This round covered the new delta, the full PR diff and record, applicable repository and Runtime context rules, the test's caller and history, backward compatibility, test integrity, and architecture boundaries. No review threads from this reviewer are open.

gloryfromca (commented):

No new findings on this revision. The rebase onto the sharded-test `main` is clean, the added QQ and tracing tests cover the new behavior, and restoring the broker call expressions to their pre-existing form leaves only the stricter callable guards as the production delta. The existing nonblocking item is unchanged and is not repeated here.

Verification on this head: the exact frozen `ty` gate passed; 189 targeted QQ, tracing, deep-research, and gateway tests passed; Ruff on the touched delta, source-language validation, and `git diff --check` passed. GitHub's four unit shards, combined coverage gate, Python lint, pre-commit, commit checks, Windows upgrade, kernel wheel, TUI, UI RPC, page, and bridge checks are all green.

This round covered the rebase range-diff, new commit and full PR diff, applicable repository and Runtime context rules, callers and history, backward compatibility, test integrity, coverage accounting, and architecture boundaries. No review threads from this reviewer are open.

gloryfromca (commented):

Reviewed the new `d5f0588f82eb...f64bff7be1db` delta and rechecked the full `github/main...HEAD` change against AGENTS.md/CLAUDE.md and the CONTEXT architecture contracts, including callers and history, backward compatibility, and whether tests were weakened. The two newly enabled rules (`unresolved-import` and `invalid-return-type`) are cleared with corrections that preserve the runtime contracts; in particular, `_do_refresh()` now returns its fetched value directly to both foreground callers while its background caller safely ignores that value. The gate accounting now matches the configuration at 110 live rules and nine locally disabled rules. The delta only adds tests rather than weakening existing coverage.

Verification: the exact pinned `ty` gate passed locally; 519 focused routing, evolver, loop, CLI, QQ, and tracing tests passed; and Ruff passed on the newly changed implementation/tests. GitHub's four unit shards, coverage gates, Python lint/type checks, kernel-wheel smoke, repository checks, TUI/page/RPC checks, and Windows upgrade check all pass. The red Bridge job is unrelated to this diff: its TypeScript build passed, then npm's retiring audit endpoint returned `400 Invalid package tree`; bridge sources and lockfile are unchanged. A local bridge build could not start because this checkout has no bridge dependencies installed (`tsc` was absent), so the successful CI compile is the meaningful build result. No review threads are open.

gloryfromca (commented):

This revision only merges the current target branch. I checked the refreshed `github/main...HEAD` diff, the merge against both parents, commit history/range-diff, and the stable patch ID: all nine PR commits and the complete PR patch are unchanged, with no conflict-resolution delta. That also preserves the prior conclusions on callers, backward compatibility, test strength, AGENTS.md/CLAUDE.md rules, and the CONTEXT architecture contracts.

On the merged tree, the exact pinned `ty` gate passed and the focused routing, evolver, loop, CLI, QQ, and tracing run passed all 519 tests. CI's lint, repository, commit, bridge, kernel-wheel, page, TUI, RPC, Windows, and three unit-shard checks pass. The remaining unit shard ran all 5,429 tests successfully but failed the new idle ceiling because an upstream `tests/test_rpc_model.py` case recorded 3.44 seconds idle against a 3-second limit. This PR changes neither that test nor the idle-ceiling machinery, the test passes locally with `--idle-ceiling-strict`, and the target commit itself already has a failing unit shard; this is not a regression introduced by this revision. No review threads are open.

0xKT (dismissed):

gloryfromca (commented):

This revision is another merge of `main`. I checked both parents, the refreshed `github/main...HEAD` diff, history/range-diff, and stable patch IDs: all nine PR commits and the full PR patch remain identical, with no conflict-resolution change. The prior review of repository rules, callers, compatibility, architecture contracts, and test strength therefore still applies; the new work here was checking interaction with the updated target.

Named nonblocking follow-up (summary-only because the affected line is untouched target code): the pinned gate now reports one `unresolved-import` in `raven/rpc/knowledge_preview.py:34`. `KnowledgeDocumentRecord` is imported from `raven.knowledge._types`, but it is defined in `_records` and exported from `raven.knowledge`. The line came from target commit `7beffb9d0f`, not this PR, so this is a pre-existing issue newly exposed by the gate and does not meet the late-round blocker requirement that this change introduce the defect. Correcting that import on `main` or carrying the same correction here will clear the currently red Python-lint job.

Verification: the exact `ty` command fails only on that import; the focused merged-tree run passed all 519 tests; all four CI unit shards, coverage, and every other reported job pass. No tests were weakened and no review threads are open.

gloryfromca (commented):

The new one-line delta resolves the target-branch interaction noted on the previous revision: `KnowledgeDocumentRecord` is defined in `raven.knowledge._records`, and that is the class returned by `KnowledgeManager.get_document()`. Because the import is under `TYPE_CHECKING`, the correction changes annotation resolution only and preserves runtime/backward compatibility. I checked the refreshed full diff, the symbol's definition and caller, commit history, AGENTS.md/CLAUDE.md and CONTEXT constraints, and test changes; no tests were weakened.

Verification: the exact pinned `ty` gate now passes, 49 knowledge-preview/routing/evolver tests pass locally, and the revision's Python lint/type-check CI job plus all completed non-unit jobs are green. The remaining unit shards were still running when reviewed. No review threads are open.

gloryfromca (commented):

This revision is another merge of current `main` into the branch. The range-diff preserves all 10 PR commits one-for-one, and the stable patch ID is unchanged from the clean `891316dd7c58` revision. I checked the repository rules, the full PR diff and new-main overlap, affected callers and history, backward compatibility, test-strength changes, and the existing architecture constraints; the merge introduces no new PR-side behavior or weakened tests.

Verification:
- `uv run --frozen --python 3.12 --all-extras ty check raven evolver agents plugins-dist scripts` - passed
- `uv run pytest tests/test_routing_benchmark_cache.py tests/test_evolver_node_core_version.py tests/test_agent_loop_run_emit.py tests/test_cli_session_commands.py tests/test_cli_onboard_commands.py tests/test_channels_qq.py tests/test_tracing_api.py -q` - 529 passed

All completed GitHub checks were green when inspected; the four unit shards were still pending.

0xKT (approved):

PR #417, opened 2026-09-15
ci(*): run the unit suite as four shards and fix two slow caches

0xKT:
## Summary

The unit job takes twelve minutes on one runner, and #413 did not move it (pytest 720.95 s on main, 706.57 s there): the runner is three to four times slower than a laptop on CPU-bound tests, and two of the twelve minutes are the four xdist workers importing every test module before the first test runs. This PR attacks the wall clock from three sides.

- `ci(*)`: the unit job runs as four shards. pytest gains `--shard K/N` in `tests/conftest.py`, decided in `pytest_ignore_collect` so a shard never imports the other shards' files; test files are dealt round-robin over their sorted list so the slow families that sort together (the ppt engine's) are spread out. Each shard writes coverage data and no report; a `coverage` job downloads the four data files, combines them, writes `coverage.json` and `coverage.xml`, and runs the gates that used to follow the tests.
- `perf(ppt-engine)`: `menu()` caches its reading of a template, capped at eight entries. One bind reads the bound template and then every other bundled template for the borrow offer; with ten templates bundled, every bind evicted what the next one needed and re-measured all of them with freetype. The cap is 32 and a hit moves its entry to the back. On the CI run of #413 the template-tool file alone was 257 s of the over-ceiling list, its tests 10 to 26 s each.
- `test(*)`: every test created two numbered temp directories, and pytest picks each number by scanning the base temp dir; the fake home is now a sibling of `tmp_path`. `MPLCONFIGDIR` points at a git-ignored dir in the checkout, because the fake home changed per test and matplotlib rebuilt its font list from the system fonts on every run.

The branch is stacked on #413 and targets its branch; GitHub retargets it to `main` when that PR merges.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [x] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

The shards partition the suite exactly. Collected ids of the whole suite against the union of the four shards (`uv run --frozen --all-extras pytest --collect-only -q -n 0 --shard K/4`, K = 1..4):

| whole | shard 1 | shard 2 | shard 3 | shard 4 | union | ids in two shards |
|---|---|---|---|---|---|---|
| 22094 | 5527 | 5058 | 5555 | 5954 | 22094 | 0 |

A shard collects in 7 s where the whole suite takes 23 s on the same machine.

Combining works and loses nothing: two small runs with `--cov=raven --cov=raven_everos --cov-branch --cov-report=` and `COVERAGE_FILE` set, then `make coverage-combine COVERAGE_DATA_DIR=<dir>`:

| report | files | statements | branches |
|---|---|---|---|
| one `make coverage` run of the whole suite (main before #413) | 704 | 82634 | 25548 |
| two shards combined (one of them importing `raven_everos`) | 704 | 82636 | 25548 |

Same file set; the two extra statements are the constant and import #413 added.

The menu cache, three files that bind templates (`uv run --frozen --all-extras pytest -q tests/test_ppt_engine_template_tool.py tests/test_ppt_engine_outline_checks.py tests/test_ppt_engine_default_templates.py`): 101 s of worker time in the last profile, now `196 passed in 11.84s` with no test over 2.9 s. The files that call `menu()` directly: `219 passed in 8.06s`.

The font cache, `tests/test_ppt_engine_assets_script_helpers.py::test_a_formula_that_stacks_is_typeset_as_a_picture_in_the_decks_ink` at `-n 0`: 8.91 s with the cache cold, 0.69 s on the next run. Seven home-sensitive files (`test_config_loader`, `test_updates_upgrade`, `test_cli_tui_commands`, `test_agents_code_launcher`, `test_deep_research_tool`, `test_media_gen_tool`, `test_subagent_mcp_grant`): `484 passed in 5.30s`.

Whole suite on this branch, `uv run --frozen --all-extras pytest -q` (macOS, 4 workers): `22051 passed, 43 skipped in 224.25s`, against 279.65 s on #413 and 372.82 s on main before it. The wall-clock ceiling lists 39 unmarked tests over 2 s, down from 63 on #413.

`make lint-python` clean. The CI run on this PR is the measurement that matters: four `unit` shards and one `coverage gates` job.

- [ ] User-facing docs or screenshots are updated when needed (none: CI plumbing, a cache cap, test setup)

## Risk

- The coverage gates now run in a job after the four shards. A red shard skips the gates; the PR is red either way. The gate scripts and their thresholds are unchanged.
- `--shard` is opt-in; a plain `pytest` run collects everything as before. Shard data artifacts are kept one day, the combined reports 30 days as before.
- A source package's never-imported files reach the combined report through whichever shard imported the package (coverage adds them when that process saves its data), the same way they reached the single run; the check below shows the file set is identical.
- The menu cap only changes how long the cache keeps entries; nothing reads the cache's size. No security surface is touched.
- Rollback: revert the PR. The unit job goes back to one runner with the gates inline.

## Related Issues

N/A


Makefile
```diff
@@ -11,6 +11,11 @@ COVERAGE_DIFF_THRESHOLD ?= 90
 COVERAGE_RATCHET_TOLERANCE ?= 0.05
 DOCKER_IMAGE ?= raven:local
 COVERAGE_REPORT_ARGS = --cov=raven --cov=raven_everos --cov-branch --cov-report=term-missing:skip-covered --cov-report=xml --cov-report=json --cov-report=html
+# One slice of the suite, K/N; the shard writes its .coverage data and no report.
+COVERAGE_SHARD ?= 1/1
+COVERAGE_DATA_ARGS = --cov=raven --cov=raven_everos --cov-branch --cov-report=
```
gloryfromca, line 16: Blocking: include the changed PPT engine in the shard coverage data.

`COVERAGE_DATA_ARGS` measures only `raven` and `raven_everos`, while `coverage_gate.py` treats every `plugins-dist/**/*.py` file as production. This revision changes `plugins-dist/ppt-engine/raven_ppt/services/template/menu.py`, so after all four shards pass, `make coverage-diff` reports that file as missing and fails. I reproduced this against the stated target, and GitHub's `coverage gates` job fails the same way. The combined report needs to measure the changed production package before this PR can go green.

gloryfromca (commented):
Blocking: the combined coverage report omits the changed PPT engine package.

I found one inline blocker. The four GitHub unit shards all pass on the rerun, but the downstream coverage gate fails because the combined report does not contain the changed `raven_ppt` file.

Verification performed:
- Python 3.12 collection: the unsharded suite and the four-shard union each contain 21,843 node IDs, with no missing or duplicate ID; invalid shard specifications exit with pytest usage errors.
- `tests/test_ppt_engine_default_templates.py`: 83 passed.
- `tests/test_ppt_engine_template_capacity.py`: 22 passed.
- `tests/test_updates_update_notice.py`: 38 passed in each of three runs; the earlier GitHub failure in that file did not recur, so I did not raise it.
- Focused coverage reproduction: the tests pass, report generation succeeds, and the diff gate exits 1 because `plugins-dist/ppt-engine/raven_ppt/services/template/menu.py` is unmeasured. GitHub's `coverage gates` job reports the same failure.

I covered the full diff and its three commits, surrounding menu callers and cache history, pytest collection/xdist and coverage-combine behavior, workflow behavior, backward compatibility of the unchanged `coverage` target, whether sharding weakened the suite, and the applicable AGENTS.md rules. No domain-term or architectural-boundary change is implicated; the repository lint, import-contract, commit, source-language, and asset checks pass on GitHub.

PR #418, opened 2026-09-15
feat(agent): a wall clock and a dead-end rerun, bounded by the product

ZuyiZhou:
## Summary

Stacked on #410, which supplies the dead-end predicate this builds on. Base is that
branch; when #410 squash-merges GitHub moves this to `main` and the branch will be
replayed onto it.

Two mechanisms the research product could not reach. It had no clock of any kind: only
the iteration cap could end a turn, so a turn that kept finding leads ran until it hit
one. And the conditional rerun, the highest measured pp-per-compute lever this product
has, existed only as an offline pass over persisted rows, where no user could reach it
however the config was written.

Both land as mechanism in the loop and switch in the product, because the loop serves
every agent and must not know one of them. A hook writes the bounds it wants under
`TURN_BUDGETS_KEY` at `before_user_inbound`; the loop reads that and applies whatever is
there. An agent that writes nothing is bounded exactly as it was before the key existed,
and `turn_budgets` treats a malformed value as no budget rather than raising inside the
loop, because it is parsing a dict a plugin wrote.

The clock is checked between iterations and never mid-generation. Cancelling a call in
flight discards a finished generation and leaves no answer, and a deadline landing
between a tool result and the model reading it produces a trajectory nothing can
interpret, so the deliberate cost is an overrun of at most one iteration. It ends the
turn through the same wrap-up path as the iteration cap, so a spent deadline never
produces a silent turn.

The rerun re-runs from the original question rather than salvaging the failed attempt.
Salvage was measured on 25 real duds: 23 produced content, none hit the right answer,
all confidently wrong, and it empties this trigger while doing it. The seed is copied
before the first attempt, because the loop appends to the list it is handed and a seed
rebuilt afterwards would carry the wreckage it is meant to escape. A turn whose first
attempt spent the clock is not re-run: each attempt gets its own clock, which is safe
only because a failing run is a shortcut and duds end early.

`turn_end` now records `stopped_by`, the budget and the elapsed time, and is written for
every turn rather than only an answerless one. "This turn was cut short" is not
recoverable from `status`: the iteration cap, a stalled tool and a spent clock all land
on `interrupted`, and the wrap-up they produce reads like an ordinary answer to
everything downstream. Ints only, because a reader that keeps scalars by type drops a
float without a word.

Three decisions worth review:

- The shipped config takes a one-hour clock (`wallClockSeconds: 3600`). The measured
  arms this product descends from were bounded from outside, by the harness that
  launched them, so nothing here was measured with a clock; an unbounded research turn
  is the status quo rather than a validated setting. One line to change.
- `deadEndRetry` is on by class default and is not written into the shipped slice. A
  dead turn has no answer to damage, so the count of right answers among the turns it
  fires on starts at zero and a second attempt can only raise it: the usual
  benefit-to-risk question has a zero in the denominator.
- `TURN_BUDGETS_KEY` lives beside its parser in the loop package rather than in
  `raven/contracts`. It reads as a contract and that was the first placement, but
  `raven/contracts` carries a reviewed line ceiling and adding it there would have meant
  raising the ceiling for one constant. One module owning both the key and the only
  reader of it is the better shape anyway.

`monotonic()` is a seam, and the only production line here that exists for a test to
reach. The budgets are checked against elapsed time and a test cannot wait for a
deadline, while replacing `time.monotonic` itself would replace the clock asyncio
schedules on. The datetime equivalent already goes through `_now_fn` for the same
reason.

One thing the fork fixed that needs no fix here: it had to fill an answerless turn with
the no-answer sentence before appending the research appendix, or the appendix rode on
an empty string and emptied the dead-end trigger. This product appends its appendix in
`after_send`, which runs after the loop has already filled that sentence, so the
ordering is correct as it stands. Verified by reading the call order rather than assumed.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
  22124 passed, 43 skipped, 109 warnings in 366.33s
```

Integration and e2e are deselected by default (`-m "not integration and not e2e"`), so
the run above does not include them. This change adds no integration surface.

```
make lint-python
  All checks passed! / 1945 files already formatted
make lint-imports
  Contracts: 10 kept, 0 broken.
make lint-deps
  Success! No dependency issues found.
PYTHONPATH=. uv run --frozen --python 3.12 --extra dev python scripts/check_source_language.py origin/main...HEAD
  clean
make check-large-files
  clean
make check-commits
  clean
PR_TITLE=... make check-pr-title
  clean
```

23 new cases in `tests/test_agent_loop_dead_end_retry.py`, covering the rerun (a dead
turn runs again and the second answer wins, a live turn never does, an agent that asks
for nothing never does, the budget is a ceiling and not a loop, the rerun starts from
the question and not the wreckage, the reason filter narrows the trigger, a spent clock
stops it and an unset one does not), the clock (a deadline wraps up rather than going
silent, `stopped_by` names the cap and the clock apart, the record carries only scalars
a reader keeps) and the budget reader (nothing readable means unbounded, and a value
that is not a budget -- a bool in a number's place, a string, a zero, a negative --
leaves the turn unbounded rather than ending it early).

The spent-clock test drives `monotonic()` as a counter rather than waiting, and asserts
the first attempt ran to empty-response exhaustion, so it cannot pass by the loop's own
clock cutting the attempt short and leaving no dead end to decline.

`agents/raven-research/.env` was moved aside for every suite run: it is gitignored and
absent in CI, and two launcher tests that delete a key from the environment otherwise
find one in the file.

This machine has no search-provider key, so no research brief can be run here. The
budgets are verified by unit tests; a real run on a machine that holds the key is what
would exercise them end to end.

The new config fields carry their reasoning in their own docstrings, which is where this
repo documents a knob. No user-facing surface changed.

## Risk

This is the only change in the series that touches the loop's hot path, and that is why
it is its own PR rather than part of #410: it can be reverted on its own.

Every behaviour change is gated on a hook writing the budget key. No agent in this repo
writes it except the research flow, so for every other agent the loop runs the same
sequence it ran before, and the tests assert that directly rather than by inspection.
The one unconditional change is that `turn_end` is now written for every turn rather
than only an answerless one, which adds a metadata key for hooks and changes nothing a
user sees.

For the research product itself the behaviour does change: a turn now ends after an hour
where it previously had no limit, and a turn that produced no answer runs a second time,
which costs a second turn's compute on the turns it fires on. Rolling either back is one
config line; rolling the whole thing back is one commit.

## Related Issues

N/A


raven/agent/loop/turn_path.py
```diff
@@ -2446,29 +2488,70 @@ async def _tap_episode(index: int) -> None:
...
+                seed,
+                on_progress=on_progress,
+                session_key=key,
+                model=routed_model,
+                fallback_models=fallback_models,
+                injected_skill_ids=self._collect_injected_skill_ids(selected_skills),
+                on_token_delta=_tap_token if on_token_delta is not None else None,
+                on_reasoning_delta=_tap_reasoning if on_reasoning_delta is not None else None,
+                on_tool_event=on_tool_event,
+                on_episode_start=_tap_episode,
```
gloryfromca, line 2520: Blocking: keep episode indices unique across the retried turn.

`_run_agent_loop` numbers episodes from its local iteration counter, so invoking it again here restarts the wire indices at zero. With the existing dead-then-answer scripted case plus an `on_episode_start` collector, this emits `[0, 1, 2, 0]`. `EpisodeStart` defines the index as the 0-based step within the turn, and the TUI uses `ep.index` as both React keys and fold-state identifiers; distinct calls from the two attempts therefore collide and can share rendering/fold state. The retry needs to continue the outer turn's episode sequence.

raven/agent/loop/turn_path.py
```diff
@@ -2446,29 +2488,70 @@ async def _tap_episode(index: int) -> None:
...
+                    # the budget, a retry would hand back a fresh copy of the clock that
+                    # just fired, so the turn stops here with whatever it has.
+                    if budgets.wall_clock_seconds and (monotonic() - turn_t0) >= budgets.wall_clock_seconds:
+                        logger.info("Dead end ({}); not re-running: the turn's clock is spent", ", ".join(reasons))
+                        break
+                    retries_left -= 1
+                    logger.info("Dead end ({}); re-running the turn", ", ".join(reasons))
+                    if on_progress is not None:
+                        await on_progress("The first attempt produced no answer; researching again.")
+                    final_content, _, all_msgs, outcome = await _attempt([dict(m) for m in retry_seed])
```
gloryfromca, line 2574: Blocking: keep the configured wall clock scoped to the whole turn.

Each call to `_run_agent_loop` starts a new `turn_t0`, while the outer check only declines a retry once the first attempt has already consumed the entire budget. A first attempt ending just below the limit therefore gives the retry a fresh full limit. In a deterministic reproduction with three-second model calls, a 10-second budget, a dead three-call first attempt, and a successful three-call retry, the user turn returned at 18 seconds. That is more than the documented one-budget-plus-at-most-one-iteration overrun; with the shipped one-hour value, a late dead end can similarly make the same turn run for nearly two budgets.

gloryfromca, line 2574: Blocking: close or reuse the first attempt's product ledger before retrying.

The research hook treats `ctx.iteration == 1` as a turn boundary and calls `_start_turn`, which opens a fresh ledger token. Because the retry restarts iteration at 1, it opens a second ledger and overwrites the ContextVar path without closing the first; `after_send` later closes only the second path. Whenever the dead attempt wrote a search/gate row, its first per-turn file is consequently unreachable and accumulates on disk, contradicting the ledger's one-file-per-turn cleanup contract.

gloryfromca (commented):
Blocking: keep retries within the turn clock and preserve per-turn episode and ledger invariants.

I reviewed the full target diff, the relevant loop, hook, checkpoint, research-flow and UI callers, the base/history, and the repository rules in AGENTS.md, CONTEXT-MAP.md, CONTEXT.md and ui-tui/CONTEXT.md. I also checked backward compatibility for agents that provide no budgets and looked for weakened tests; the existing tests were not weakened, but the new coverage does not exercise these cross-attempt runner/product lifecycle invariants.

Verification: `uv run pytest tests/test_agent_loop_*.py tests/test_agents_research_flow_*.py tests/test_agents_research_launcher.py tests/test_agents_research_mode_tiers.py -x` passed: 1,138 tests. `git diff --check github/feat/research_dead_end_verbatim...HEAD` also passed.

The two direct reproductions emitted episode indices `[0, 1, 2, 0]` for a dead-then-live retry, and allowed a 10-second configured turn budget to finish at 18 simulated seconds. The inline comments contain the concrete failures.

raven/agent/loop/turn_path.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed at 703b1844ac50: the outer turn now owns the episode counter, and the regression test verifies a continuous sequence across the dead attempt and rerun. I reproduced the new behavior through the relevant suite.

raven/agent/loop/turn_path.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed at 703b1844ac50: one monotonic start is passed into every attempt, so the rerun spends the remainder of the turn budget. The new deterministic regression reaches five calls, including the permitted overrun and wrap-up, rather than the former six calls and 18 seconds.

gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed at 703b1844ac50: reopening a product ledger first closes the path currently held by the turn context. The lifecycle tests cover replacement after rows exist, final cleanup, writes moving to the new file, and disabling the ledger between opens.

gloryfromca (commented):

The new revision fixes all three prior findings: episode indices now remain turn-scoped across attempts, every attempt shares the turn's monotonic start, and reopening the research ledger releases the previous file. I closed the three threads I opened after verifying the fixes.

I reviewed the fix delta and the complete target diff, relevant loop/research-flow/ledger/UI callers and history, AGENTS.md and the Runtime/TUI context contracts, backward compatibility for direct loop callers and agents without budgets, and the test changes for weakening or false-green behavior. I found no new issue worth raising.

Verification: `uv run pytest tests/test_agent_loop_*.py tests/test_agents_research_flow_*.py tests/test_agents_research_launcher.py tests/test_agents_research_mode_tiers.py -x` passed: 1,144 tests. `git diff --check github/feat/research_dead_end_verbatim...HEAD` passed.

PR #419, opened 2026-09-15
feat(rpc): a client can run a playbook, and says where the run reports

Handsome-wzw:
## Summary

The library became readable and manageable over RPC and stayed unrunnable, so
the one thing a playbook is for remained behind the CLI and the model's own
tool. `playbooks.run` goes through `PlaybookRuntime.load` -- the same door both
of those use -- and answers the executor's plan verbatim: `kind` and `reply`,
the receipt's run id included, rather than a shape invented at this layer.

**`session_key` is required, and is the whole reason this is not a thin
wrapper.** A run's progress and its completion announce are addressed to a
conversation, which the loop sets per turn for every origin it knows about. An
RPC call is one it does not: the address is a `ContextVar` whose every consumer
falls back to a `cli:direct` default, so a page-initiated run would have
dispatched correctly and reported where nobody is looking. No RPC method has had
to set that address before; this is the first that dispatches a graph. The key is
the conversation's own `channel:chat_id`, the pair every origin dict already
carries, and a half-formed one is refused rather than defaulted -- `set_context`
forwards to the graph tool only when both halves are present, so the failure it
would otherwise cause is a silent one.

**`confirmed` is the caller's own statement**, not this layer's assumption.
`PlaybookExecutor.execute` has carried the parameter with no caller at all since
the funnel was deleted, kept, in its own words, "because an entry point that
*does* ask must be able to say so". This is that entry point. It is relayed
through `load` rather than decided here because a page that shows a confirmation
and one that fires on a single click are both legitimate, and only the page knows
which it is; asserting it on their behalf would skip the graph-level gate with
nobody having seen the graph.

**A run is placed in its session, not only addressed to it.** `set_context`
gives the graph its reporting and history origin; the workspace a node reads and
writes, and the pair a raven-backed node resolves, come from two context
managers only the turn path enters (`workdir.bind` at `turn_path.py:3127`,
`use_binding` at `main.py:950`). Without them a run aimed at one conversation
worked in the loop-wide workspace and on the default model while reporting under
that conversation's name -- the wrong project and the wrong model, in the right
session's name. Both are entered around `load` rather than before it: the
dispatch backgrounds itself with `create_task`, which copies the context it is
created in, so a binding already reset by then is one the detached run never
sees.

**A caller on this socket is read as the user's own hand.** Disabling takes a
playbook out of what the *model* is offered -- this caller holds the session
cookie, which is the user's credential and not something a model is handed -- so
`allow_disabled` is true, the answer the CLI gives for the reason it gives it.

No new field carries the run id. `_run_id_of` states why it is read out of the
receipt text rather than threaded as a field: "a second channel for the same fact
is a second thing to keep in step", and the existing clients already parse it
there.

`PlaybookRuntime.load` gained the `confirmed` relay it was missing. The parameter
was reachable only by a caller holding the executor directly, which no entry
point does.

This is stacked on the branch of the playbook-library PR, because the name guard
it adds (`_known_name`, the kebab-case check that runs before any store lookup)
is what keeps a traversing name out of this handler too. Until that merges, the
diff here shows its two commits as well.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/test_rpc_playbooks.py                          65 pass
pytest tests/test_playbook_executor.py                      58 pass
pytest -k "playbook or rpc or subagent_dag or agent_loop"  3509 pass
cd ui-web && npm run gen:check                              matches the contract (172 methods)
cd ui-web && npx tsc --noEmit -p .                          clean
cd ui-web && npx vitest run                                 1678 pass
cd ui-tui && npm run lint:rpc                               in sync (sha256 010998e35f8b)
cd ui-tui && npm run lint:i18n / npx tsc --noEmit -p .      pass
scripts/check_source_language.py upstream/main..HEAD        exit 0
```

The contract moved through all three layers: `rpc-schema/openrpc.json` is the
source of truth, `raven/rpc/models.py` mirrors it (held equal by
`test_rpc_schema_match.py`), and **both** generated clients were regenerated --
`ui-web` and `ui-tui` each hold their own and each has its own sync gate.

Eighteen new cases, every one mutation-checked by reverting the line it covers.
Thirteen mutations, thirteen caught:

- never set the address -> red
- set the address *after* the dispatch instead of before -> red (the order is
  the point: the origin is read inside `execute`, so one set afterwards arrives
  for the next call)
- accept a half-formed `session_key` -> red
- assert `confirmed: true` on the caller's behalf -> red
- flip `allow_disabled` to false -> red
- drop the name guard -> red
- degrade "no runtime" to an empty answer instead of refusing -> red
- answer success when the runtime cannot load the playbook -> red
- stop relaying `confirmed` through `load` -> red
- enter neither the workspace nor the binding -> red
- enter the workspace but not the binding -> red
- enter both but exit before `load`, so the detached task captures neither -> red
- dispatch anyway when the session's workspace will not resolve -> red

**Driven over the real gateway's WebSocket**, on a build restarted onto this
head. A throwaway playbook with one required param was written through the
store's own `save`, then:

- an unknown name answers a typed `config_validation_error` rather than
  method-not-found, which is what says the method is registered;
- `../victim`, `a/b` and `Upper` are refused by the kebab-case rule;
- `''`, `'web'`, `':c1'` and `'web:'` are all refused as session keys;
- the positive path answers `kind: "gaps"` with `params.target` named, which
  exercises the whole route -- address set, runtime reached, plan relayed -- at
  the cost of **zero dispatches and zero model calls**, because a missing
  required param is reported with nothing dispatched by design.

The host was left as found: the throwaway playbook is removed and
`~/.raven/config.json` hashes to the backup taken before the check.

No UI change here -- this is the interface only. No page control is bound to it
yet.

## Risk

Security: this dispatches work, which is the sharpest thing in the `playbooks.*`
surface so far, and it is bounded on four sides. The name is matched against
`NAME_RE` before any store lookup. Secret params are never accepted here and
never were: a conversation is the one place a credential may not pass, and the
runtime fills them from this machine's own credential store. The graph-level
`confirm` gate still runs unless the caller states it already asked, and that
statement is the caller's to make. And a run cannot be aimed at a conversation
by accident -- a malformed address is refused rather than silently defaulted.

Known gap, checked and deliberately not closed here: the turn path also enters
`usage_context.bind(cid, ...)`, which `usage_tracker` reads to attribute token
spend to a session, so a run entered here has its nodes' usage attributed to no
session. Closing it means calling `sessions.get_or_create(cid)` as the turn path
does, which would have an RPC call create a session record as a side effect of
running a playbook -- a separate decision. Misattributed usage is an accounting
gap, where the workspace and the binding were the run doing the wrong work.

Worth a reviewer's judgement, and not decided here: the `gaps` reply ends "Call
`load_playbook` again with those filled in." That sentence is written for the
model -- `_gap_reply` says as much -- and `load_playbook` is a tool name this
caller does not have. Relaying the string verbatim is what keeps this method from
inventing its own answer shape, and rewording it would change the model's path
too, so it is left as it is and flagged. A client has `kind` and the named
arguments regardless.

Backward compatibility: one new method, and the only change to existing code is
a keyword with a default on `PlaybookRuntime.load` -- callers that do not pass it
behave exactly as before, which the existing suites cover. Nothing in the answer
shape of the ten existing methods moves.

Rollback: one commit, reverts cleanly.

Known behaviour a client must be built around, stated rather than left implicit:
this call can block. A playbook whose `confirm` is true (the default) puts the
run to the conversation named by `session_key` and waits up to the ask broker's
600-second deadline, and a conversation still busy at that deadline is a "not
now" -- the graph does not run. That does not hold the connection hostage: this
transport dispatches every frame as its own task, so other calls on the same
socket are unaffected.

## Related Issues

N/A


raven/rpc/methods/playbooks.py
```diff
@@ -352,14 +371,220 @@ async def playbooks_oauth_clear(params: dict) -> dict:
...
+    if runtime is None:
+        # A write, so it refuses rather than degrading to empty: answering a run
+        # with "nothing happened" would leave the caller drawing a dispatch that
+        # was never made. Both causes read the same to a caller and are named
+        # together -- playbooks switched off builds no runtime at all.
+        raise ConfigValidationError("playbooks are not running on this host, so nothing can be dispatched")
+
+    # Addressed before the dispatch, not after: the origin is read inside
+    # ``execute``, so setting it afterwards would arrive for the next call.
+    runtime.set_context(channel=channel, chat_id=chat_id, session_key=session_key)
```
gloryfromca, line 590: Blocking: bind the run to the named session's execution context before dispatch.

`set_context` establishes only the DAG's reporting/history origin. Unlike the turn path, this RPC task never enters the named session's `workdir.bind(...)` or `use_binding(...)` contexts. Consequently `SubAgentDagTool` sees `workdir.current() is None` and falls back to the loop-wide workspace, while its late `_binding_for()` read falls back to the default model. A session pinned to another project or switched to another model can therefore receive this run's progress and history while its nodes read/write the wrong workspace and Raven-backed nodes use the wrong model. I reproduced this call path with a runtime probe and observed both contexts as `None`; please establish the session execution contexts around `runtime.load` so the detached DAG task captures them.

gloryfromca (commented):
Blocking: the RPC run must inherit the named session's working directory and model binding.

I reviewed the full `github/main...HEAD` diff and the relevant playbook runtime/executor, DAG dispatch, RPC transports, session context, generated contracts, tests, and commit history. I also checked the repository rules and canonical Runtime/TUI terms, backward compatibility of the added `confirmed` keyword, that the existing tests were not weakened, and the architectural RPC boundary.

Verification: `uv run pytest tests/test_rpc_playbooks.py tests/test_playbook_executor.py -q` passed (120 tests), and `uv run pytest tests/test_rpc_schema_match.py tests/test_rpc_registration.py tests/test_rpc_bootstrap.py tests/test_agent_loop_playbook_entry.py tests/test_agent_playbook_delegate.py tests/test_playbook_tool.py -q` passed (455 tests). A focused handler probe printed `workdir=None` and `binding=None`, confirming the missing session execution context described inline.

raven/rpc/methods/playbooks.py
Handsome-wzw: Confirmed, reproduced, and fixed in 4f961b5d.

I probed the call path independently before agreeing, and read both as you did — at the moment the dispatch is made:

```
workdir.current() = None
active_binding()  = None
```

Structurally it is the two entries: `workdir.bind(...)` appears only at `turn_path.py:3127`, and `use_binding(...)` only at `main.py:950`, alongside the other turn scopes. An RPC handler enters neither, so `SubAgentDagTool` falls to `workdir.current() or self._workspace` (`dag_tool.py:1102` and `:1258`) and `_binding_for()` to the default pair. The consequence is worth stating as sharply as you put it: the run was addressed correctly and *placed* wrongly — right conversation's name, wrong project and wrong model.

The fix enters both around `runtime.load`:

```python
with workdir.bind(session_workdir), use_binding(loop.binding_for_session(session_key)):
    plan = await runtime.load(...)
```

Around, not before, for the reason your last sentence names: `_execute` backgrounds itself with `create_task`, which copies the context it is created in — `dag_tool.py:1063` says so in as many words — so a binding already reset by then is one the detached run never sees. The workspace comes from `loop.session_workdir(session_key)` and the pair from `loop.binding_for_session(session_key)`, the same two the turn path resolves; a directory override that will not resolve is now refused by name rather than surfacing as an internal error.

Four new mutations, all caught:

- enter neither context -> red (the state you reported)
- enter the workspace but not the binding -> red
- enter both but exit *before* `load` -> red (the subtle half: bound, and not captured by the detached task)
- dispatch anyway when the workspace will not resolve -> red

One test reads both values inside `load`, which is the only moment that proves anything; another asserts they are gone once the call returns, since a handler that set them and never reset would hand the next frame on that task whatever the last one chose.

**One thing I found alongside this and deliberately did not fold in.** The turn path also enters `usage_context.bind(cid, ...)`, and `usage_tracker` reads it to attribute token spend to a session — so a run entered here has its nodes' usage attributed to no session. It is the same shape of gap, but closing it means calling `sessions.get_or_create(cid)` the way the turn path does, which would have an RPC call create a session record as a side effect of running a playbook; that is a separate decision and not one to make inside a fix. Misattributed usage is an accounting gap, where the two above were the run doing the wrong work, so I have left it and am naming it rather than quietly widening the change. Happy to take it in this PR if you would rather it not wait.

gloryfromca: No blockers; suggestions only, and they are marked inline.

Verified on `4f961b5d`: the named session's workspace and model binding now surround `runtime.load`, so the detached DAG task captures both contexts; the handler resets them afterward and refuses an unusable workspace before dispatch. The focused playbook/RPC suites passed (123 tests), as did the broader RPC schema, registration, bootstrap, and playbook-call-path suites (455 tests).

The original execution-placement blocker is settled. I accept session usage attribution as the separate, nonblocking follow-up you scoped: it affects accounting, not which workspace/model executes or where the run reports.

gloryfromca (commented):

The session execution-context fix is correct: workspace and model selection now match the named session at dispatch, the detached DAG inherits both, cleanup is scoped, and an invalid workspace cannot dispatch. I resolved the original blocker.

Named follow-up: RPC-started playbook nodes still lack the turn path's session usage binding, so token usage can be attributed to no session. As the author scoped, this is accounting-only and nonblocking for this change.

I covered the repository rules (AGENTS.md, CLAUDE.md, CONTEXT-MAP.md and routed context), the full diff, affected callers and history, backward compatibility and generated RPC contracts, test-strength changes, and the RPC/session architecture. A security-focused pass found no additional reportable issue. `git diff --check github/main...HEAD` passed. Tests: `uv run pytest tests/test_rpc_playbooks.py tests/test_playbook_executor.py -q` (123 passed); `uv run pytest tests/test_rpc_schema_match.py tests/test_rpc_registration.py tests/test_rpc_bootstrap.py tests/test_agent_loop_playbook_entry.py tests/test_agent_playbook_delegate.py tests/test_playbook_tool.py -q` (455 passed).

raven/rpc/methods/playbooks.py
0xKT, line 1: This change discharges a promise the file still states as pending, and the
reason that promise gave is contradicted by the server's own comment.

WHAT THE FILE SAYS, playbooks.py:382-385, unchanged by this PR:

    Deliberately not here: `run` and `create`. Both take model time -- a
    graph to completion, a generation -- and a JSON-RPC call that occupies
    the socket for minutes blocks every other call on it. They need a
    dispatch that answers with a handle and reports progress, which is a
    different change.

WHAT THE FILE DOES, 113 lines below it:

    playbooks.py:495   async def playbooks_run(
    playbooks.py:608   dispatcher.register("playbooks.run", _run)

Same module, same section header. A reader at :382 is told `run` and
`create` are both deliberately absent for a stated architectural reason.
`create` IS still absent, which is what makes the half that is now false
actively misleading rather than obviously dead: the sentence reads as a
current design decision, and half of it is.

THE REASON IT GIVES IS NOT TRUE EITHER, and the contradiction is already in
the tree. raven/rpc/server.py:192-194:

    # Spawn the dispatch as an independent task so streaming /
    # slow handlers don't block subsequent reads.
    task = asyncio.create_task(self._handle_frame(line))

and raven/rpc/transports/ws.py:357 has the same shape. So a slow handler
does not occupy the socket, and has not for as long as those lines have been
there -- both predate the sentence. Two comments in this repo say opposite
things about the same mechanism.

Measured during this acceptance on a real RpcServer over a socketpair, with
the PR's own registration: a `playbooks.run` made to take 4 seconds answered
in 4.003s, and a `playbooks.list` sent while it was still in flight answered
in 0.0029s, with the run asserted still pending at that moment. I read the
two dispatch sites above myself; the timing run is the acceptance's rather
than mine, and I am saying which is which.

None of this is an argument against the change. The opposite: `playbooks.run`
does exactly what the sentence asks for, and better than the sentence expects.
Every plan kind answered in about 6 ms over that socket, `dag` returning its
receipt with the node backend provably still running, and the production
progress sink emitting run_started, node_updated and run_completed afterwards.
`runtime.load` reaches no model on any of those paths. The promise is kept.

So the only thing left wrong is the writing. Whoever reads :382 next is told
this feature is deliberately absent, by a comment sitting above the feature.

A SECOND STATEMENT OF THE SAME CLASS, in a docstring this PR edits.
raven/playbook/runtime.py:418 still reads:

    ``allow_disabled`` is for the CLI, where the user named the playbook
    themselves.

This PR adds a second, non-CLI caller that passes it True
(playbooks.py:492 `_CALLER_NAMED_IT = True`, used at :576), and the new
comment above that constant quotes this very sentence as its authority. The
PR is editing that docstring -- it adds a `confirmed` paragraph immediately
below this one and the sentence itself is byte-identical at base and head, so
this is not an unvisited file.

I checked the gate is the whole gate, at this head:

    runtime.py:428   if spec is None or (name in self.disabled() and not allow_disabled):
    one disabled playbook, same runtime:
        allow_disabled=False -> plan=None, executor called: []
        allow_disabled=True  -> plan=dag,  executor called: ['demo']

Three more places still name the CLI as the only entrance that resolves a
disabled playbook, none updated here: runtime.py:99, config/schema.py:2096,
CONTEXT.md:1076.

I am not arguing the behaviour is wrong -- disabling takes a playbook away
from the MODEL, and a client on this socket is not the model. If that is the
intent, the change that widens the rule is the one that should widen the
writing.

Smaller, same shape: the new constant's justification is "a client here holds
the session cookie". The handler is registered unconditionally
(rpc/methods/__init__.py:228) and reaches two transports; the TUI one has no
cookie in it at all (`grep -c -i cookie` is 0 in both cli/tui_commands.py and
rpc/server.py), its credential being a per-boot token checked at
server.py:150-160. The conclusion holds there, arguably more strongly; the
premise as written does not.

Not blocking. Nothing shipped behaves wrongly. Filed because a comment that
says a feature is deliberately absent, 113 lines above that feature, costs the
next reader more than it cost to write.

0xKT, line 1: `playbooks.run` and `playbooks.validate` judge agent names against two
different tables, and `run`'s refusal sends the user to `validate` -- which
can answer "it is fine".

Both endpoints are new in this PR, so this never worked rather than broke.

THE TWO TABLES.

  run     raven/agent/loop/wiring.py:1340
          known_agents=self.subagents.registry.all_names
          a callable onto the LIVE in-process registry

  validate raven/rpc/methods/playbooks.py:442-443
          registry = AgentRegistry()
          registry.apply(load_config().subagents.agents)
          a fresh registry built from the config FILE

They agree until something writes config.json without going through this
process. Nothing keeps them in step: raven/config/live.py:27 says the agents
table is deliberately kept on the explicit apply path rather than a watcher,
and `apply_agents(` has exactly three call sites, all in-process
(agent/subagent/manager.py:356, agent/loop/wiring.py:1662, and the
subagents.* RPC's own hot-apply at rpc/methods/subagents.py:378). An external
writer reaches none of them.

THE WINDOW IS A DOCUMENTED NORMAL STATE, NOT A CONTRIVANCE. `raven agents new
--register` writes the row from another process through
raven/config/update_subagents.py, and that command's own closing advice says
so in as many words (raven/cli/agents_commands.py, the 4/4 next block):

    restart the gateway to make a live raven re-read the tree

So the repo already tells users that after registering an agent, a running
gateway has not seen it.

WHAT THE USER THEN HITS. A playbook naming the newly registered agent:

  playbooks.run      -> refused: "playbook 'x' does not load; validate it to
                        see why"                     (playbooks.py:585)
  playbooks.validate -> ok=true, errors=[]

The refusal names the remedy, and the remedy reports no problem. There is
nothing else in the answer to act on.

WHAT I VERIFIED MYSELF, at head 4f961b5d merged onto 56c79edf: the two
sources above read as quoted; live.py's agents note reads as quoted; the
three `apply_agents(` call sites are the only ones; and the CLI's restart
advice is in the tree. The end-to-end reproduction -- driving both handlers
after an external write and getting the two answers above, with a control
naming an agent NEITHER table has so that validate reports it correctly --
was produced during this acceptance run rather than re-run by me, and I am
saying so rather than presenting it as my own execution.

THE FIX IS ALREADY IN REACH. register_playbooks_methods receives
`agent_loop_factory` and currently passes it only to the run handler. validate
could ask the same live table the dispatch is judged against, and the two
surfaces would stop disagreeing about one file.

Rule this misses, from the code that owns the live table
(raven/agent/loop/wiring.py:1335-1340):

    The live table, asked rather than copied ... a copy taken here would go
    stale against the very dispatch it is meant to agree with.

and from raven/playbook/runtime.py:

    two surfaces disagreeing about one file is worse than either answer alone.

Not blocking: the wrong answer is a misleading "ok", not data loss or a wrong
dispatch. Filed because the misleading answer is the one the error message
sends the user to.

gloryfromca: No blockers; suggestions only, and they are marked inline.

Agreed on the concrete mismatch. `playbooks.run` refreshes through the loop runtime's callable over the live in-process registry, while `playbooks.validate` reconstructs a registry from the current config file. After an external `raven agents new --register`, before the documented gateway restart, validate can therefore report `ok=true` for an agent the running gateway still refuses, even though the run error directs the user to validate.

The documented gateway restart reconciles the tables, so this has a working operator escape hatch and remains a nonblocking follow-up rather than a late-round merge blocker. The relevant playbook validation/runtime suites passed (130 tests).

gloryfromca: No blockers; suggestions only, and they are marked inline.

Agreed. The transport dispatches each frame in its own task, and `playbooks.run` returns the executor's immediate plan/receipt rather than waiting for DAG completion, so the section-level claim that `run` is deliberately absent because it would block the socket is stale. The CLI-only wording around `allow_disabled` and the session-cookie premise are stale for the same reason: this PR adds RPC callers, including the token-authenticated TUI transport. These are concrete documentation corrections, but they do not make the shipped behavior wrong, so I retain a nonblocking stance.

Verification here: `tests/test_rpc_server_socket.py` passed (3 tests), and the playbook/runtime suites passed (130 tests).

gloryfromca (commented):

The new revision correctly closes the run/validate table mismatch: production registration supplies the loop, validation reads the same live registry dispatch uses, and direct/offline callers retain the config-backed fallback. The revised section comment no longer claims `playbooks.run` is absent or socket-blocking. I also checked the rewritten branch's name normalization, malformed-YAML result handling, user-layer-only deletion path, prior session execution-context fix, RPC contracts, and generated clients; I found no new plain error.

Previously named nonblocking follow-ups remain: RPC-started node usage is not attributed through the turn path's session usage binding, and several CLI-only/session-cookie descriptions of explicit disabled-playbook runs still need widening for both RPC transports.

Covered this turn: repository rules and domain context, the current full diff and revision delta, affected callers and history, backward compatibility, test-strength changes, and RPC/session architecture. `git diff --check github/main...HEAD` passed. Tests: the RPC/playbook/schema/transport set passed 558 tests; the store, agent-loop playbook-entry/delegation, and RPC bootstrap set passed 99 tests.

gloryfromca (commented):

This revision is a patch-equivalent rebase onto the newer `main`: `git range-diff` pairs all six PR commits exactly, and the reviewed fixes retain identical stable patch IDs. The rebase changes none of the PR's affected files relative to the prior head, and integration with the new base produced no regression.

The previously named nonblocking follow-ups remain unchanged: session usage attribution for RPC-started nodes, and widening the remaining CLI-only/session-cookie descriptions of explicit disabled-playbook runs.

Covered this turn: unchanged repository rules/domain context, revision and target diffs, callers/history via exact range comparison, backward compatibility, test strength, and RPC/session architecture. `git diff --check github/main...HEAD` passed. The affected RPC/playbook/schema/transport suite passed 558 tests; store, agent-loop playbook-entry/delegation, and RPC bootstrap passed 99 tests.

raven/rpc/methods/playbooks.py
Handsome-wzw: Fixed in `de01592c`, and you are right about which half is the worse one.

The stale half was easy to see once named: the sentence says `run` is deliberately absent and `playbooks.run` is defined below it in the same section. What makes it actively misleading rather than obviously dead is exactly what you say — `create` *was* still absent, so the sentence reads as a live design decision and half of it was.

The false half is the one I want to acknowledge properly. I found that same contradiction while designing `run` — `asyncio.create_task` per frame in `ws.py`, and `server.py` has the shape you quote — and I corrected the claim in the pull-request description at the time and then left the copy in the source. So the sentence outlived the belief that produced it by my own hand, in the very change that disproved it. That is worse than never having checked, and it is the part of your comment I have taken most seriously.

The replacement states the rule the handlers keep rather than a snapshot of which ones exist: each reaches the library through the door that already exists for it rather than around it. It does not expire the next time this surface grows, which is the property the old one lacked.

I also swept the rest of the module for claims of the same shape and found none.

Handsome-wzw: Confirmed and fixed in `de01592c`. `validate` now reads the live registry when there is a loop to ask, and the config file only when there is not.

What settled it for me was the comment already sitting on the `run` side at `wiring.py:1336-1340`: `known_agents` is passed as a *callable* onto the live registry precisely because "a copy taken here would go stale against the very dispatch it is meant to agree with". `validate` was taking exactly the copy that comment warns about — and then the refusal at dispatch points the user at `validate`, so the two disagree at the one moment a user is following the advice.

Your point that the window is a documented normal state, not a contrivance, is what makes this worth fixing rather than noting: `raven agents new --register` writes from another process and its own closing advice tells the user to restart the gateway afterwards. Between those two, a check that reads the file says the step is fine and the run refuses it.

Two tests, and the second exists because of a mutant that survived the first. The handler falls back to the config file when it has no loop — and a registration that simply forgot to pass the factory looks identical to that fallback, silently, and only on a machine where the two tables have drifted. So one test drives the handler with a live table directly, and one goes through `register_playbooks_methods` and the real dispatcher. Dropping the factory at the registration site is now 1 failed; before the second test it was 440 passed.

Left deliberately: the fallback itself. A check with no loop to ask should still answer rather than refuse, and the config file is the only table there is in that case.

0xKT: Verified at de01592c.

  `grep -c "Deliberately not here"` -> 0
  `playbooks_run` still at :534, registered at :651

So the sentence went and the feature stayed, which is the pair I wanted to
check: a comment removed by deleting what it described would have been the
other way of making it true.

The replacement is stronger than the removal. Saying these handlers reach the
library through the same door rather than around it -- enabling writes the deny
list the CLI writes, running goes through the runtime the model's tool goes
through -- states a rule a reader can check a future handler against. The
sentence it replaces only recorded a decision that had since been reversed.

You were right about which half was the worse one. `create` still being absent
is what made the old sentence read as current rather than obviously dead, and
that is exactly the half a reader would have trusted.

Closing.

0xKT: Verified at de01592c, by mutation rather than by reading.

  `_known_agent_names` reverted to the config file alone
     -> 2 failed, 68 passed
        test_validate_judges_agents_against_the_table_a_run_would_use
        test_the_registered_validate_is_given_the_loop_to_ask

The first of those names the defect I filed, and the second pins the wiring
that makes it reachable -- the factory actually being handed to validate. Both
were needed: the fix could have been correct in the function and never reach it
through registration, and one test alone would not have told those apart.

The fallback to the file when there is no loop is the right shape and the
docstring says why it is a fallback rather than a peer. What convinced me was
the ordering: live registry first, file only when there is nothing to ask. A
check that answers "would this run" has to read the table the run would use,
which is the sentence I would have wanted and did not write as well.

Closing.

0xKT (dismissed):

gloryfromca (commented):

Reviewed `bdbc3dd4fc8c` as a new revision. The three PR commits are patch-equivalent to the previously reviewed run, execution-context, and live-agent-table commits; the earlier library-management commits are now in `main`. The newer base changes do not alter playbook dispatch: the only inherited RPC overlap adds unrelated model-picker metadata, while the new session-isolation change is confined to the design plugin.

The two previously named nonblocking follow-ups remain: RPC-started node usage is not yet attributed through `usage_context.bind`, and the remaining disabled-playbook/session-auth prose still describes the caller as CLI/cookie-specific. Neither holds this change at this review stage.

Covered the repository rules and context documents, the diff, affected callers and registration/schema surfaces, history/range-diff, backward compatibility, test changes (including checking that the rebase did not weaken them), and the session/workdir/model-binding architecture. `git diff --check github/main...HEAD` passed. Focused playbook/RPC suites passed: 558 tests; the store, loop-entry, delegation, and RPC-bootstrap group passed: 99 tests. All threads I opened are resolved.

0xKT (approved):

PR #420, opened 2026-09-15
test(cli): clear rich's per-style ansi cache between theme tests

### LINKED ISSUE #339: test(cli): deflake order-dependent bold accent assertion in test_cli_theme
## Problem

`tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare` (tests/test_cli_theme.py:276) is order-dependent: it passes when run alone and fails when the whole file (or the full suite) runs. It shows up intermittently as the lone unexplained failure in otherwise green full-suite runs, which weakens the whole-suite signal.

## Evidence

- Run alone: passes. Run the whole file: fails. The failure mode is a style degradation: the accent renders as 16-color bright_yellow (SGR `1;93`) instead of the truecolor accent `#fbe23f`, so the assertion that the output is styled-not-bare sees the degraded form.
- Reproduced byte-identically in two pristine checkouts of recent main-history commits containing no local changes, so it is in-file state pollution (or environment-sensitive terminal capability detection), not a regression from any one change.
- The degradation direction (truecolor down to 16-color) points at cached console/terminal capability state: some earlier test in the file plausibly constructs a Console (or mutates COLORTERM/TERM/FORCE_COLOR detection) whose downgraded color system leaks into this test.
- Several independent full-suite runs on the same machine reproduced it while others did not, consistent with ordering/parallelism sensitivity rather than a deterministic bug.

## Suggested direction

Find the in-file pollution: run progressively larger prefixes of tests/test_cli_theme.py to identify the first test whose presence flips the result, then audit it for module-level or cached Console/theme state and env mutations that survive the test. Isolate with a fixture that resets the cached theme/console and pins a truecolor-capable fake terminal for this assertion.

AmirF194:
## Summary

rich caches a Style's rendered ANSI on the object `Style.parse`/`Style._add` return (both `lru_cache`'d), and never re-renders it for a different `color_system` once that happens once. The dark accent+bold combo `test_bold_accent_renders_styled_not_bare` checks is the same style `_ThemedConsole` and a couple of other tests in this file print through consoles that don't pin `color_system="truecolor"`, so whichever one hits it first under xdist's scheduling decides the ANSI bytes for every truecolor render after it, including this test's.

`_reset_theme_cache` now clears `Style.parse` and `Style._add`'s caches along with its existing reset, before and after each test in the file. Added `test_bold_accent_survives_a_downgraded_render_elsewhere` right before the assertion test: it renders the same markup through a 256-color console first, which reproduces the failure on main and passes once the caches are cleared between tests.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other (test-only deflake, no production code touched)

## Verification

```
uv run --frozen --all-extras pytest -n 0 \
    tests/test_cli_theme.py::test_bold_accent_survives_a_downgraded_render_elsewhere \
    tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare -v
2 passed   (the second one fails on main with only the fixture hunk reverted, same command)

uv run --frozen --all-extras pytest tests/test_cli_theme.py
54 passed, both under -n 0 and the default -n 4

make lint-python
clean
```

## Risk

- [x] Security impact considered: test-only change, no production code touched.
- [x] Backward compatibility considered: no shipped behavior changes either way.
- [x] Rollback path is clear for risky changes: revert restores the status quo, including the flake.

## Related Issues

Fixes #339


AmirF194:
No review yet on this one, a week in. It's test-only (clears rich's per-style ANSI cache between the two theme tests so they don't order-depend), so it should be a quick look whenever you get to it.

PR #422, opened 2026-09-15
fix(*): resolve the imports the type gate could not, and turn that rule on

gloryfromca:
## Summary

Stacked on #416 -- it adds the gate, this turns one of its disabled rules on.

`unresolved-import` was switched off with 21 diagnostics. Three of them were stale module paths that nothing had ever run into:

| Where | What was wrong |
|---|---|
| `evolver/tree/node.py` | `HarnessNode.current_core_version()` read `raven.__core_version__`, a module that exists neither in this tree nor in the wheel `scripts/build_core_wheel.py` produces. The read sits inside `try/except ImportError`, so the except branch was the only branch -- **every node `store.py` saved recorded `core_version="unknown"`**. The docstring already named the intended source (`raven` version); it now reads that and returns `0.1.13`. |
| `raven/agent/loop/wiring.py` | imported `SkillForgeRouterConfig` from `raven.config.schema`; it lives in `raven.config.raven`. Under `TYPE_CHECKING`, so nothing raised -- the annotation simply meant nothing. |
| `scripts/skill_forge_full_e2e.py`, `skill_forge_retrieval_eval.py`, `skill_hub_e2e.py` | imported `raven.context_engine.base` and `raven.memory_engine.base`, neither of which exists. **Two of the three failed at module import** -- running them was not possible. `AssemblyContext` is in `raven.contracts.context`, `TokenBudget` in `raven.contracts.assembled`. |

The other eighteen were the environment, not the code:

- **Plugin roots (5).** The three agent products keep their plugin package beside them rather than under `plugins-dist`, so `code_flow` and `research_flow` are not on the path a plain run sees. Their roots are declared in `[tool.ty.environment]`, the way `plugins-dist` is declared in ruff's `src` for the same reason.
- **lxml (5).** lxml ships a compiled `etree` with no type information. `lxml-stubs` is a dev dependency now -- real types rather than five silenced lines. It does surface one stub imprecision (`itertext()` is typed as yielding `str | bytes`, because one element API also serves byte-mode trees), suppressed by rule name at the line that reads it.
- **UNO (3).** `uno` and `com.sun.star.*` are suppressed per line: that worker runs inside LibreOffice's own interpreter and is never imported by the host process, so they are unresolvable from raven's environment by design.

`unresolved-import` comes off the ignore list. The gate is now **109 of ty's 133 rules**; ten remain off with their backlog recorded, fourteen ship disabled upstream and are untouched.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `make lint-types` -> All checks passed, with `unresolved-import` enabled
- Gate proven live: an `from raven.nonexistent.module import thing` probe turns it red; removing the probe turns it green
- The three scripts now import cleanly (checked by loading each as a module; two raised `ModuleNotFoundError` before)
- `HarnessNode.current_core_version()` returns `0.1.13`, was `"unknown"`
- New test `tests/test_evolver_node_core_version.py` mutation-checked: restoring the old import path fails both cases, reverting passes them
- Tests: 184 passed across evolver, ppt-engine publish/provenance, agent loop, design-engine face and the two import guards

## Risk

Two behaviour changes, both turning a silent failure into the intended value: nodes are stamped with the real raven version instead of `"unknown"`, and two scripts that could not start now start. The rest is imports moved to the modules that hold the names, one dev-only stub package, and three suppressions that each say why. Rollback is a revert of the single commit.

## Related Issues

N/A


gloryfromca (commented):

Reviewed `github/chore/type_check_gate...c220a7e6a5f3`, including the late docs-only follow-up. I covered the repository rules in AGENTS.md/CLAUDE.md and CONTEXT-MAP.md, the complete diff, surrounding version/wheel and plugin-loading callers, relevant history, backward compatibility for stored evolver nodes, dependency/lockfile changes, test strength, and the Runtime architecture constraints.

Verification:
- `uv run --frozen --python 3.12 --all-extras ty check raven evolver agents plugins-dist scripts`: passed.
- Focused import/config/evolver/PPT/Skill Forge tests: 248 passed.
- Importing all three repaired scripts with `runpy`: passed.
- Full `uv run --frozen --python 3.12 --all-extras pytest -q`: 22040 passed, 55 skipped, 2 failed. The failures are in untouched QQ event-loop and Rich theme tests; the QQ test passed on focused rerun, while the unchanged Rich assertion still failed. Neither affected source/test nor its relevant dependency changed in this revision.

No inline findings.

gloryfromca (commented):

Reviewed the new `c220a7e6a5f3...59660d1b1442` delta and rechecked it in the full target diff. The added change makes the QQ factory guard run inside an event loop, matching the production call from `QQChannel.start()`, while preserving the original instance assertion and adding no skip or weaker condition.

Covered the repository rules, changed test and production caller, rewritten commit history, backward compatibility, test strength, and architecture boundaries. No new domain term or production behavior was introduced.

Verification:
- `uv run --frozen --python 3.12 --all-extras pytest -q tests/test_channels_qq.py`: 35 passed.
- Explicitly clearing ambient loop state before running `_make_bot` inside `asyncio.run`: passed and bound the client to the running loop.

No inline findings.

gloryfromca (commented):

This revision is a rebase onto the refreshed target. `git range-diff` shows both PR commits patch-identical to the previously reviewed versions; the QQ test-only commit is now in the target branch and no longer in this PR. The full target diff was rechecked for repository rules, callers/history, backward compatibility, test strength, and architecture constraints.

Verification on the new base:
- `uv run --frozen --python 3.12 --all-extras ty check raven evolver agents plugins-dist scripts`: passed.
- Focused evolver, provenance, context-factory, and skill-router suite: 76 passed.

No inline findings.

gloryfromca (commented):

Reviewed the new coverage commit and rechecked the full target diff. The production edits retain the same fetched tool and invoke the same broker setter without an intervening await or lookup; the QQ and tracing tests add coverage without removing or weakening existing assertions. `git range-diff` shows the two previously reviewed PR commits remain patch-equivalent.

Covered repository rules, the changed callers and history, backward compatibility, test strength, and architecture constraints. At this late round, no candidate issue met the required combination of being introduced here, ordinarily reachable, and having no working escape.

Verification:
- Configured production type check: passed.
- Affected QQ, tracing, deep-research, and gateway tests: 189 passed.
- GitHub unit shards 1/4 through 4/4: passed.

No inline findings.

gloryfromca:
Folded into #416, which now carries the whole stack: the gate, unresolved-import and invalid-return-type, as three commits. Closing this so there is one review, one approval and one merge.

PR #423, opened 2026-09-15
fix(*): make the declared return types true, and turn that rule on

gloryfromca:
## Summary

Stacked on #422, which is stacked on #416.

`invalid-return-type` was off with 39 diagnostics. **One stale annotation held eight of them**: `AgentLoop.run_turn` and the `_run_turn` behind it were declared `-> LoopOutcome` while all three of their `return` statements build a `TurnOutcome` -- a different dataclass with no field in common.

```
TurnOutcome(usage: Usage, explicit_reply: bool)          # raven/spine/runner.py
LoopOutcome(status: str, checkpoint_id, edited_files)    # raven/agent/loop/_shared.py
```

Nothing was broken at runtime: the spine worker reads `.usage` / `.explicit_reply` off what it actually gets, and `LoopOutcome` never leaves `_run_agent_loop`, whose own consumer reads `.status` / `.checkpoint_id`. The annotation was the only thing that was wrong, and fixing it cleared `run_turn`'s two spine callers and the runner with it.

### The other 31, by kind

| Kind | Count | What was done |
|---|---|---|
| Annotation narrower than the code | 9 | Five `return tuple(<genexpr>)` typed as fixed-arity tuples (a generator-built tuple is `tuple[T, ...]`) now unpack to the arity they declare; an inner `one()` in the deck reviewer still declared the two-tuple it returned before a third element was added; `image_block` returns the `ImagePart` TypedDict, so the two callers that wrap it say `ImagePart` too |
| Declaration refusing a value the function returns | 5 | Four tool `execute` overrides said `-> str` while returning `ToolResult` on their error paths (the base is `str \| ToolResult`); a confirm helper said `-> MenuReply` while returning `None` on two paths its only caller already types optional |
| Invariant living in a `# type: ignore` comment | 7 | "chain always non-empty", "`_do_refresh` sets it or raises", "the constructor refuses an empty key" -- now asserts, stated where both a reader and the checker meet them |
| Honest widening | 10 | A shaping dict that really carries a `None`, a delivered path that really can be absent, an api-key source that really is `str \| None`, and an onboarding prompt that really can return the untyped `_BACK` sentinel |

Two notes on the edges of that:

- **Unpacking changes a failure mode, for the better.** `tuple(<genexpr>)` of the wrong length used to build a wrong-shaped tuple silently; unpacking raises. In `_as_size` that raise lands inside the `ValueError` its caller already catches and turns into the message it was written to give.
- **`_BACK` is an untyped `object()` sentinel** returned from twelve sites; the flagged function was the only one declaring something precise enough to notice. It now says `Any`, like its four siblings. Giving the sentinel a type of its own is a separate change across all twelve.

The note in `raven/utils/images.py` saying "CI runs no type checker, so this annotation does not gate a merge" is no longer true, and no longer says it.

The gate is now **110 of ty's 133 rules**: nine off with their backlog recorded, fourteen shipped disabled upstream and untouched.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `make lint-types` -> All checks passed, with `invalid-return-type` enabled (39 -> 0)
- Gate proven live: a function returning a str under `-> int` turns it red; removing the probe turns it green
- Tests: **9035 passed, 44 skipped, 2 failed** across every suite matching the touched areas (providers, routing, config, cli, agent loop, subagent dag, media_gen, render, sentinel, spine, filesystem, ppt-engine, research-flow, trajectory)
- Both failures are **pre-existing on `c87622da4`**, checked by a real checkout of `github/main` in this same environment, where both fail identically:
  - `test_a_missing_key_is_said_once_rather_than_per_query` asserts the Serper key is absent; this machine has `SERPER_API_KEY` set, so it is environment-dependent and should pass in CI
  - `test_a_ceiling_as_large_as_the_window_still_leaves_room_for_history` fails on `reserved_tools = 6044 > 6_000`, and also fails under `--extra dev`, so it is not extras-dependent. That one looks like a real budget overrun on `main` and is not addressed here

## Risk

No intended behaviour change. The seven asserts state invariants the code already relied on; the tuple unpackings raise on a shape that previously produced a wrong-arity tuple, which no caller could have used correctly anyway. Everything else is a declaration moved to match what the function already returned. Rollback is a revert of the single commit.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the full target diff plus affected callers and relevant history. Coverage included AGENTS.md/CLAUDE.md, the canonical Turn Runner terms and layer constraints in CONTEXT-MAP.md/CONTEXT.md, backward compatibility and public exports, runtime effects of the tuple unpacking and assertions, and test integrity; this PR patch does not change or weaken tests. The return annotations match the reachable values, and the AgentLoop boundary now matches the established TurnOutcome protocol while preserving the LoopOutcome re-export.

Verification:
- The exact underlying `make lint-types` command passed on this rebased head: all 1,085 files checked successfully. The Makefile wrapper itself could not run because `make` is absent from this environment.
- `uv run pytest` on the immediately preceding patch-identical head completed with 1 failed, 22,041 passed, and 55 skipped. The lone failure was `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`; the same ANSI output reproduces with the target branch's original `_theme.py` under this dependency environment, and this patch only rewrites the unrelated OSC reply tuple construction.
- `git diff --check` passed. Both subsequent rebases retained the identical stable patch ID.

No inline findings.

gloryfromca (commented):

This is a patch-identical rebase of the previously reviewed change; the stable patch ID is unchanged. Rechecked the repository rules and architecture constraints, the target diff, affected callers and history, backward compatibility, and test integrity; the PR still changes no tests. The exact-head type gate passes, `git diff --check` passes, and all four exact-head GitHub unit-test shards pass. No new findings.

gloryfromca (commented):

Reviewed the new BenchmarkCache commit and rechecked the repository rules, full target diff, callers and history, backward compatibility, test integrity, and architecture constraints. Returning the freshly built data directly removes the swallowed-assert path while preserving the stale-cache, snapshot, in-memory, and background-refresh behavior. The added tests strengthen rather than weaken coverage.

Verification: the exact type gate passes locally; `uv run pytest tests/test_routing_benchmark_cache.py -q` passes 3 tests; `git diff --check` passes; and all four exact-head GitHub unit-test shards pass. No new findings.

gloryfromca:
Folded into #416, which now carries the whole stack: the gate, unresolved-import and invalid-return-type, as three commits. Closing this so there is one review, one approval and one merge.

PR #424, opened 2026-09-15
fix(*): the shipped config says whether the rerun runs

ZuyiZhou:
## Summary

The dead-end rerun ships on, and it ships on by a class default the delivered file does not
mention. Reading `agents/raven-research/config.json` does not tell you that a dead turn is
run again, and the day `DeadEndRetryConfig`'s default moves the shipped product changes with
not one character altered in the file. The block is written out at the value that ships
today: enabled, one retry, no reason filter.

Nothing runs differently, and that is asserted rather than claimed: the slice with the block
and the slice without it resolve to the same `DeadEndRetryConfig`. That same equality is what
reddens if the class default ever moves away from the file, which turns a silent change into
a decision about whether the product follows.

The vendored fork writes no such key, so this lands as a `TWIN_DRFLOW_PRODUCT_ONLY` row
rather than a parity break: the fork bounds a dud from outside the turn and carries no in-run
rerun, and its schema would refuse the key. The parity gate refused the key before the row
existed, which is what makes the row evidence rather than decoration, and it will redden
again if the fork ever gains one.

The flow label does not move. It names what the model reads, and neither a retry budget nor a
bound on the turn is model-visible text; `finalShape.reportDepth`, the one model-visible knob
in the neighbourhood, is already on and unchanged.

This is the last of the three unabsorbed commits from the research fork. The upstream
original also wrote out every inherited nested knob so the shipped file would be fully
explicit, and that half is deliberately not here: the product slice is pinned value for value
against a fork fixture frozen at `dr@3.5`, so each newly written knob would need its own
product-only row against a fork that is sealed and can never catch up. The table's contract is
that a row cannot outlive the divergence it records, and those rows never could. That is a
decision for the maintainer, not a change to make in passing.

Stacked on #410, which carries the rerun this configures. It cannot go to `main` first:
`DeadEndRetryConfig` does not exist there.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
mv agents/raven-research/.env /tmp/            # two launcher tests otherwise find a key
  22133 passed, 43 skipped, 109 warnings in 364.81s
uv run pytest -q tests/ -k research
  949 passed
```

```
make lint-python / lint-imports / lint-deps      pass
make check-source-language / check-large-files / check-commits   pass
```

Coverage is not evidence here, and saying so is better than quoting a number that measures
nothing. This change touches no file the gate measures, so it reports a pass by vacuum:

```
COVERAGE_BASE_REF=origin/feat/research_dead_end_verbatim make coverage-diff
  Diff coverage: 100.00% (0/0 executable changed lines)

git diff --name-only origin/feat/research_dead_end_verbatim...HEAD \
  -- ':(glob)raven/**/*.py' ':(glob)plugins-dist/**/*.py'
  (empty)
```

What stands in for it is mutation. Two mutants, each run against the new test:

```
MUTANT  drop deadEndRetry from the shipped config
  killed - test_the_shipped_rerun_is_written_out_and_not_inherited
           test_the_flow_slice_is_the_vendored_twins_drflow_verbatim
MUTANT  shipped maxRetries 1 -> 2, away from the class default
  killed - test_the_shipped_rerun_is_written_out_and_not_inherited
```

The second mutant is the reason for the second commit. It died on the first run, but of
`ModuleNotFoundError` rather than of the assertion: the test imported the plugin without
putting it on the path, so it had been passing only behind whichever test imported it first,
and under `pytest-randomly` that is a flake. With the path inserted the way the rest of the
file does it, the assertion fires and prints both configs. The equality it pins had never
actually been exercised before that.

This machine has no search-provider key, so no brief can be run here. Nothing in this change
is on a retrieval path.

No user-facing surface changed. The file states what it was already doing.

## Risk

Behaviour is unchanged today, and the test asserts that rather than assuming it. The risk the
change removes is the other direction: with the knob unwritten, a move to
`DeadEndRetryConfig`'s default silently changed what ships.

The one thing to know is what it now pins on: a dead turn is run again once, from the original
question, and the turn's wall clock is shared across both attempts, so the cost ceiling is one
budget plus at most one iteration rather than two budgets.

Rollback is the `deadEndRetry` block plus its `TWIN_DRFLOW_PRODUCT_ONLY` row, which returns
the product to inheriting the default.

## Related Issues

N/A


gloryfromca (commented):

The shipped `deadEndRetry` block makes the existing enabled/one-retry/no-filter policy auditable without changing runtime behavior. The parity exception is correctly product-only, and the regression test checks both that the key remains explicit and that its written value still matches today's inherited behavior. No findings.

I covered the repository rules and canonical terminology (`AGENTS.md`, `CONTEXT-MAP.md`, `CONTEXT.md`, and `agents/README.md`), the refreshed effective diff, the config render/load caller path through the research-flow plugin, relevant history and the merged dependency PR, backward compatibility, test-strength changes, and the plugin/loop ownership boundary.

Verification:
- `uv run pytest tests/test_agents_research_launcher.py tests/test_agents_research_flow_parity.py -q`: 51 passed.
- `uv run pytest tests/ -k research -q`: 949 selected tests passed, but the command exited 1 during collection with 83 missing optional-package imports in the fresh environment.
- `uv run --all-extras pytest tests/ -k research -q`: 949 passed.
- `uv run --all-extras pytest tests/test_agents_research_launcher.py::test_the_shipped_rerun_is_written_out_and_not_inherited -q`: 1 passed independently.
- `git diff --check github/feat/research_dead_end_verbatim...HEAD`: clean.

PR #426, opened 2026-09-15
feat(knowledge): four data sources for a knowledge base, recall test, and per-base settings

### LINKED ISSUE #397: fix: Can not create a `knowledge base`
### Summary

Image

### Steps to reproduce

```shell
$ npm run --prefix ui-web build && python3 ui-web/build.py
$ uv run raven web --stop && uv run raven web  --foreground

### Expected behavior

Successfully created one and uploading files.

### Actual behavior

Blocking knowledge base usages.

### Environment

Raven: 5b7566b498e10a6591ff0beed8fee8752123d5b5

### Logs or screenshots

```text

```

KevinHuSh:
## Summary

image

Close #397.

Turns the knowledge page into something a reader can actually run a knowledge base from: put material into it four ways, look at what is in it, test what it answers with, and configure it.

**Data sources.** The panel had one button that uploaded one file. It now offers file, folder, note and URL behind one menu, plus a drop target over the whole panel. A picked folder is walked by the browser; a dropped one is walked here, because a drop hands over a directory entry rather than files. Either way the files are filtered to what this build can index, capped, and uploaded one at a time: each carries its bytes base64 in a single websocket frame under a 25 MB ceiling, so N at once is N of those in memory. One file failing no longer takes the rest of the folder with it.

A note and a fetched page are stored as the markdown they were captured as, in the same blob store as every upload, so nothing downstream learns a second shape. Reading a page follows Cherry Studio's own capture: Jina Reader, asked for JSON so the page title comes back as a field rather than as a preamble to parse out of the text, and through `validate_url_target` first, because the URL arrives from a browser and without that check the gateway is a request forger.

**Recall Test.** Asks a base a question and shows what comes back: the chunk, where in its document it sat, its score beside its rank, and what the search cost. An explicit button rather than search-as-you-type, because every keystroke used to spend a billed embedding call. Past questions are kept in localStorage and nowhere else.

**Settings.** Each base carries its own Top K and chunking, with a sentence on every field saying what moving it costs. The embedding model is shown and not offered: the collection is sized to its vector width at creation, so changing it is a rebuild of every vector rather than a setting, which is what the stale-base check exists to detect. Letting a reader change it is a separate piece of work.

**Two defects found while measuring the above.**

SiliconFlow refuses an entire request when one input is over the model's token limit, naming neither which input nor what was wrong with it. A document of eighteen chunks where seven were long indexed none of the other eleven, and four documents in a real base were failing this way. The cause is the chunker's token estimate -- utf-8 bytes over four, an English-prose rule that reads a CJK character as three quarters of a token when it is one, and a URL as a quarter of what it costs. Binary-searching the longest accepted prefix of real documentation pages against the endpoint put the true figure at 2.1 to 2.3 characters per token. A vendor client now cuts each input to fit before sending; the four documents embed whole.

Search reported one number for its cost. It now reports two, because they answer different questions: embedding is a round trip to whatever endpoint is configured and runs to hundreds of milliseconds, while the store query is the index doing its job. A surface reporting one figure as the search time wants the second, or it is describing the provider.

**Deliberately not wired yet.** File processing lists the processors it will offer with every row disabled and saying why it is unavailable; smart chunking, separator, chunk size and overlap are stored on the base but not yet read by the chunker. Both are placeholders with their state stated rather than controls that would be silently ignored.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run on this branch rebased onto `main` at `4f5d7dac`:

- `uv run pytest -q` -- 22151 passed, 45 skipped
- `npx vitest run` in `ui-web` -- 1738 passed (103 files), of which 92 cover this page
- `npx vitest run` in `ui-tui` -- 2072 passed
- `npx tsc --noEmit` in `ui-web` and `ui-tui` -- clean
- `uv run ruff check raven/ tests/` and `ruff format --check` -- clean
- `node scripts/check-page.mjs`, `check-css.mjs`, `check-class-namespace.mjs`, `count-shared-globals.mjs` -- all OK
- `make check-large-files` -- OK
- The SiliconFlow client was measured against the live endpoint, not mocked: batch size is not the limit (64 short inputs pass), 500 CJK characters pass and 512 do not, and all four previously failing documents in a real base now embed whole.
- Every panel was checked in the browser in both light and dark themes.

## Risk

User-visible behaviour changes, all additive:

- A search that names no `top_k` now answers with the base's own setting instead of a hardcoded 5. New bases default to 6.
- Documents carry an `origin` and bases carry six settings. Every field is defaulted, so a registry written before them loads with the behaviour it already had; an older gateway answering without them is read as those same defaults. A field from a build *after* this one is ignored rather than losing the row, which is the same protection in the other direction.
- The SiliconFlow client truncates an over-long input rather than failing the request. A cut chunk's vector speaks for its beginning; the chunk is still stored and shown in full when retrieved. The proper fix is chunking to the model's real limit, which this does not do.
- Four brand-logo SVGs are added under `ui-web/src/assets/providers/` (16 KB total) and spared the dark-mode inversion that monochrome vendor marks get.
- Previewing an office file retains a suffixed copy of its bytes in the render cache, because the converter picks its input filter from the extension and a blob is stored under a bare id. Deleting the document or its base now removes that copy, and the cache sweep reaches it as a backstop for one whose owner never got the chance.

Rollback is a `git revert` of the squash commit; nothing migrates on disk.

A registry written by this branch stays readable by the previous build, and that took work: its loader hands every stored key to a dataclass constructor and drops the row when one is a keyword that constructor does not take, so a new field written beside the record would make every base and document vanish from a rolled-back binary, with the records still on disk and nothing listing them. The settings and the document origins are therefore written under two sibling keys of `bases` and `documents`, which a loader that does not know them does not read; a record that uses none of them is written exactly as before, down to its keys. What a rollback loses is the settings, which that build has no use for. What it keeps is every base and every document. A test pins the stored shape against the field set the shipped loader accepts, rather than against our own reader.

Security: the URL source takes an arbitrary address from a browser and fetches it server-side, so it goes through the same `validate_url_target` guard the web tools use, and there is a test asserting a refused target never reaches the network.

## Related Issues

N/A


raven/knowledge/_records.py
```diff
@@ -53,6 +66,27 @@ class KnowledgeBaseRecord:
...
     updated_at: str
     description: str = ""
+    #: The settings a reader can change after the base exists. Every one of
+    #: them is defaulted, so a registry written before they existed loads with
+    #: the behaviour it already had.
+    #:
+    #: At most this many chunks come back from one search of this base. A
+    #: property of the base rather than of each call: how much context this
+    #: material is worth is a fact about the material.
+    top_k: int = DEFAULT_TOP_K
```
gloryfromca, line 153: Blocking: Preserve knowledge records across the documented rollback.

`RecordStore._save()` now serializes these new defaulted fields for every base (and the new origin fields for every document). The target-main loader passes all stored keys directly to its older dataclass constructors; they reject `top_k` / `origin` with `TypeError`, which the loader handles by dropping that row. I reproduced the rollback path with a registry saved by this revision: the old base constructor raises `unexpected keyword argument 'top_k'`. Consequently, rolling the binary back makes every saved base and document disappear from listings, contrary to the PR's stated rollback behavior.

raven/rpc/knowledge_preview.py
```diff
@@ -0,0 +1,118 @@
...
+
+def _alias_for(record: "KnowledgeDocumentRecord", blob: Path) -> Path:
+    suffix = named(record).suffix
+    alias = _alias_dir() / f"{record.id}{suffix}"
+    if alias.is_file() and alias.stat().st_mtime == blob.stat().st_mtime:
+        return alias
+    alias.parent.mkdir(parents=True, exist_ok=True)
+    alias.unlink(missing_ok=True)
+    try:
+        os.link(blob, alias)
```
gloryfromca, line 108: Blocking: Remove retained office source bytes when deleting their document or base.

After an office preview, this creates a hard link (or full copy) under `pdf-preview/sources`. Document deletion unlinks only the blob, base deletion likewise visits only blobs, and the preview sweeper scans only root-level PDFs, so no path removes this alias. I reproduced the hard-link path: after deleting the blob, the alias still existed and returned the original bytes. This means deleting a document or base does not actually remove its uploaded office content and the retained files can grow indefinitely.

gloryfromca (commented):
Blocking: The documented rollback drops knowledge records, and deletion retains office source aliases.

I left two blocking findings inline.

Named nonblocking follow-up (explicitly deferred by the author in the PR description): per-base `smart_chunking`, `separator`, `chunk_size`, and `chunk_overlap` are persisted and editable but are not consumed by indexing. I reproduced a base saved as `false / | / 64 / 0` still using the manager's fixed `512 / 50` chunker (one actual chunk where the configured size produces five). I am not holding this PR on an item the author has scoped to follow-up.

Coverage: I reviewed the complete diff, surrounding callers, and branch history; checked AGENTS.md and the knowledge contracts in CONTEXT.md; traced the web-to-runtime RPC boundary and generated schemas; tested persistence, deletion, and target-main rollback compatibility; and inspected test changes for weakened assertions or new skips (none found).

Verification:
- `uv run --frozen --python 3.12 --extra dev pytest -q`: 1 failed, 21886 passed, 65 skipped. The sole failure is the untouched `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`, and it fails identically in isolation.
- Knowledge/config/RPC targeted tests: 224 passed.
- `ui-web` full suite: 1738 passed; build, type-check, RPC generation check, page/CSS/global/class checks passed.
- `ui-tui` RPC/i18n lint and type-check passed.
- Relevant Ruff lint and format checks, `git diff --check`, commit-message ASCII scan, and large-file check passed.
- The source-language gate failed on six newly added CJK fixture lines in `tests/test_knowledge_embedding.py`; those lines are outside an allowed AGENTS.md exemption. This is a mechanical gate result, not a separate inline review finding.

gloryfromca (commented):
Blocking: The two open rollback and retained-preview-source findings remain unresolved.

This revision rebases onto the current target and adds one branch-owned fixture repair. The repair replaces each CJK literal with generated ideographs at the same character count, so the token-estimation and truncation assertions retain their prior boundary conditions while satisfying the source-language rule. I found no new defect in that delta.

I rechecked both open findings against the new head and target: their implementation and callers are unchanged, and the target loader still rejects the new serialized record keys. Neither thread is settled, so both remain open.

Coverage this round: revision delta and branch history; the current `github/main...HEAD` inventory; AGENTS.md source-language requirements; the changed test's assertion strength; the prior findings' deletion callers, cache sweep, persistence format, and target-main rollback path. The broader feature architecture and CONTEXT.md contract are unchanged from the reviewed revision.

Verification:
- `uv run --frozen --python 3.12 --extra dev pytest tests/test_config_raven_loader.py tests/test_knowledge_embedding.py tests/test_knowledge_manager.py tests/test_knowledge_records.py tests/test_knowledge_sources.py tests/test_rpc_files.py tests/test_rpc_knowledge.py -q`: 224 passed.
- Source-language check, Ruff check, Ruff format check, and `git diff --check github/main...HEAD`: passed.
- Worktree remained clean.

gloryfromca (commented):
Blocking: The two existing backend findings remain unresolved.

This revision adds the multi-file selection, reindex, and deletion flow in the web UI. I traced selection lifetime, navigation, optimistic state, sequential reindexing, bulk-delete settlement, reload pruning, RPC adapters, translations, and the new tests. I found no new plain error in that delta and no weakened assertion.

The two open findings are unchanged: neither affected backend file appears in this revision's delta, and neither thread has a reply or a scope decision. They therefore remain open.

Coverage this round: the complete previous-to-current diff, current `github/main...HEAD` inventory, changed UI callers/tests, generated RPC/i18n consumers, AGENTS.md checks, backward-compatibility and deletion paths underlying the standing findings, and the unchanged architecture/CONTEXT.md constraints from the prior review.

Verification:
- `npm test -- --run` in `ui-web`: 1747 passed across 103 files.
- Web type-check, RPC generation, page, shared-global, CSS, and class-namespace checks: passed.
- TUI i18n generation and type-check: passed.
- Source-language check and `git diff --check github/main...HEAD`: passed.
- Worktree remained clean.

raven/knowledge/_records.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in this revision. The writer now keeps the legacy record fields under `bases` / `documents` and puts only new values in sibling maps the target loader ignores; the settings/origins still round-trip in the new build. The rollback-shape regression test passes, so I am closing this thread.

raven/rpc/knowledge_preview.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in this revision. Document and base deletion now remove their retained office aliases, and the cache sweep bounds aliases left by interrupted or older paths. The document, base, stale-source, and live-source regression tests pass, so I am closing this thread.

gloryfromca (commented):

Both previously blocking findings are fixed, tested, replied to, and resolved. The rollback writer now preserves the exact field set accepted by the target loader while round-tripping new settings separately. Document/base deletion removes retained office aliases, and the cache sweep bounds orphaned aliases.

Named nonblocking follow-up: `pdf_preview.forget_source` interpolates the RPC-supplied `document_id` into `Path.glob`. I reproduced `../../../config` matching and deleting `RAVEN_HOME/config.json` even though no such document exists. This revision introduces that path interaction, but ordinary UI calls use server-generated hexadecimal IDs; exploitation requires a deliberately fabricated raw RPC request. It therefore fails the required ordinary-use condition for a fourth-round blocker and does not hold this PR. A later hardening change should validate the ID or avoid treating it as a glob pattern.

Coverage this round: complete revision delta and callers, target-main rollback format, deletion and cache ownership paths, tests for weakened assertions, AGENTS.md and unchanged CONTEXT/architecture constraints, backward/forward record compatibility, and the new filesystem input boundary.

Verification:
- `uv run --frozen --python 3.12 --extra dev pytest tests/test_knowledge_records.py tests/test_rpc_files.py tests/test_rpc_knowledge.py -q`: 127 passed.
- Source-language check, relevant Ruff and format checks, and `git diff --check github/main...HEAD`: passed.
- Direct traversal reproduction: `sources_dir().glob('../../../config.*')` matched and removed the Raven-home config fixture.
- Worktree remained clean.

gloryfromca (commented):

This revision only replaces the knowledge-base Settings text button with a labelled settings glyph. The button retains keyboard/button semantics, gains an accessible name and pointer title, and the adjusted tests preserve the prior enabled-control assertion while adding glyph and empty-text coverage. I found no new defect or weakened test in the delta.

The named cache-path hardening follow-up from my previous review is untouched, so the overall stance remains nonblocking; I am not repeating that already-reported item here. Both former blocker threads remain correctly resolved.

Coverage this round: complete revision delta and history, affected component/test/CSS callers, current `github/main...HEAD` inventory, accessibility behavior, AGENTS.md checks, test-strength comparison, and the unchanged architecture, compatibility, and CONTEXT.md constraints from prior rounds.

Verification:
- `npm test -- --run src/features/knowledge/KnowledgePage.test.tsx`: 101 passed.
- Web type-check, RPC generation, page, CSS, and class-namespace checks: passed.
- Source-language check and `git diff --check github/main...HEAD`: passed.
- Worktree remained clean.

gloryfromca (commented):

This revision adds base-row rename and delete controls. I traced the menu lifecycle, row opener, rename validation and failure recovery, current-base deletion, RPC adapter, generated translations, modal state, and new tests. I found no new ordinary-use error and no weakened assertion.

The previously named cache-path hardening follow-up is untouched, so the overall stance remains nonblocking; I am not repeating that already-reported item. Both former blocker threads remain correctly resolved.

Coverage this round: complete revision delta and history, current `github/main...HEAD` inventory, component/store/live-RPC callers, accessibility and dialog behavior, test-strength comparison, AGENTS.md checks, and unchanged compatibility, architecture, and CONTEXT.md constraints from prior rounds.

Verification:
- `npm test -- --run` in `ui-web`: 1754 passed across 103 files.
- Web type-check, RPC generation, page, shared-global, CSS, and class-namespace checks: passed.
- TUI i18n generation and type-check: passed.
- Source-language check and `git diff --check github/main...HEAD`: passed.
- Worktree remained clean.

gloryfromca (commented):
Blocking: Rebase onto the current main tip and resolve the two i18n catalogue conflicts.

The new commit correctly removes a deleted document's cached PDF rendering as well as its retained source alias; I found no defect or weakened test in that delta. The previously named cache-path hardening follow-up remains unchanged and nonblocking.

The revision itself is not currently mergeable. After refreshing `github/main` to the live tip, `git merge-tree --write-tree HEAD github/main` reports content conflicts in `i18n/messages.json` and `ui-tui/src/i18n/messages.generated.ts`; GitHub independently reports `mergeable: false` and `mergeStateStatus: DIRTY`. For the late-round bar: the PR changes both conflicting files, every attempted merge reaches the conflict, and there is no working merge path until the branch is rebased and the catalogues regenerated/resolved.

Both former finding threads remain correctly resolved; this is a revision-level integration blocker, not a reopened code finding.

Coverage this round: complete new delta and callers, cache-key/deletion behavior, changed tests, current `github/main...HEAD` inventory after fetching the live target, AGENTS.md checks, test-strength comparison, compatibility, and prior architecture/CONTEXT.md constraints.

Verification:
- `uv run --frozen --python 3.12 --extra dev pytest tests/test_rpc_files.py tests/test_rpc_knowledge.py -q`: 99 passed.
- Source-language check, relevant Ruff and format checks, and `git diff --check github/main...HEAD`: passed.
- Target integration dry-run: failed with the two conflicts named above.
- Worktree remained clean.

gloryfromca (commented):

The previous target-main integration blocker is fixed. This head contains the current `github/main`; the merge-tree dry run succeeds, the resolved catalogue retains both upstream and knowledge keys, and the generated TUI catalogue is current. I reviewed the merge-resolution delta and its history, repository rules, generator/callers, backward compatibility, test changes, and the affected architecture boundaries; I found no new plain error. The previously named cache-path hardening follow-up remains unchanged and nonblocking.

Verification: `uv run --frozen --python 3.12 --extra dev pytest tests/test_rpc_files.py tests/test_rpc_knowledge.py tests/test_knowledge_records.py -q` (128 passed); TUI `npm run lint:i18n && npm run type-check`; web focused test (108 passed), type-check, and generated-client check; source-language gate and `git diff --check`. The web test emitted its existing aborted iframe-fetch diagnostics but exited successfully.

gloryfromca (commented):

The rebase preserves the previously reviewed patch series. The one dropped config commit is now supplied by target `main`, and the only new branch change correctly declares MinerU's dark-theme tone and refreshes the generated RPC method count; the generation check confirms 176 methods. I covered the repository rules, current target diff and range history, affected callers, backward compatibility, test changes (none weakened), and the relevant UI/RPC architecture boundaries. I found no new plain error. The previously named cache-path hardening follow-up remains unchanged and nonblocking.

Verification: merge-tree dry run passed; `uv run --frozen --python 3.12 --extra dev pytest tests/test_rpc_files.py tests/test_rpc_knowledge.py tests/test_knowledge_records.py -q` (128 passed); focused web tests (124 passed), web type-check and RPC generation check; TUI type-check and catalogue check; source-language gate and `git diff --check`. The focused web suite emitted its existing aborted iframe-fetch diagnostics but exited successfully.

gloryfromca (commented):

The new test-only fixture removes accidental calls to a local Ollama daemon while reproducing LiteLLM's exception fallback field-for-field. No test directly exercises that patched method, so this does not weaken intentional coverage; it only makes the existing model-picker tests deterministic. I covered the repository rules, delta and history, affected test callers, backward compatibility, test-strength implications, and architecture boundaries (production code is unchanged). I found no new plain error. The previously named cache-path hardening follow-up remains unchanged and nonblocking.

Verification: `uv run --frozen --python 3.12 --extra dev pytest tests/test_rpc_model.py tests/test_provider_rates.py -q` (271 passed); target merge-tree dry run, source-language gate, and `git diff --check` passed. The worktree remains clean.

0xKT (approved):

PR #429, opened 2026-09-15
feat(rpc): a client can have a playbook written for it

Handsome-wzw:
## Summary

The last thing the library could not do from a client. `playbooks.create` binds
the composer the conversational entry binds, and takes a description of the work
-- **never a spec**. Generation, validation and repair stay inside
`PlaybookGenerator`, which is what keeps a caller from writing an arbitrary
graph into the library through this door; neither entry offers a field for the
nodes, and that is the same rule rather than a coincidence.

**Two checks refuse before a generation is spent**: a name that is not
kebab-case, and a name the library already holds. Both are single facts, and a
collision discovered after the composer has run costs a minute of model time and
throws the answer away. The wire check below shows both refusing in 0.00s.

**A generation that fails does not refuse.** `PlaybookGenerationError` joins its
reasons into one message and keeps the list beside it, so a refusal would carry
the sentence and drop the list -- and a caller that has just spent that time is
owed the reasons one by one. `created` tells the halves apart and `errors` is
non-empty exactly when it is false. The same argument the `validate` handler
already makes on this surface: an error code carries one string where the thing
being reported is a list.

**The answer is structured rather than the prose the model reads.** This is the
one place it differs from `run`, and deliberately: `run` relays the *executor's*
plan, which the CLI and the tool both consume, so relaying it verbatim is the
faithful thing. Creation has no such shared object -- the tool assembles its own
sentence for the model, and the CLI assembles a different one from `path` and
`notes`. Both derive their output from the same structured facts, so those facts
are what this answers.

`adopted` is reported rather than assumed: written-but-unloadable is a third
state, and answering only "created" would send a caller to run a name that
cannot resolve.

The name guard splits. `_kebab_name` checks the shape; `_known_name` adds the
existence check on top. The two callers want opposite answers about existence --
every handler that acts on a playbook needs the name to resolve, creation needs
it not to -- and sharing the shape rule rather than the whole guard is what keeps
those from drifting apart.

The generation budget is now **read** from the tool's `timeout_seconds` instead
of written again here, so the two entries cannot disagree and whoever
reconsiders the number does not have to know there is a second copy. That turns
out to matter: see the note in Risk.

Stacked on the run branch, which is stacked on the library branch. Until those
merge, the diff here shows their commits too.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/test_rpc_playbooks.py                          80 pass
pytest -k "playbook or rpc or subagent_dag or agent_loop"  3526 pass
cd ui-web && npm run gen:check                              matches the contract (173 methods)
cd ui-web && npx tsc --noEmit -p . / npx vitest run         clean / 1678 pass
cd ui-tui && npm run lint:rpc                               in sync (sha256 0a39a038a17e)
cd ui-tui && npm run lint:i18n / npx tsc --noEmit -p .      pass
scripts/check_source_language.py upstream/main..HEAD        exit 0
```

Fifteen new cases, every one mutation-checked. Eight mutations, eight caught:

- flatten the composer's reasons into one joined string -> red
- move the already-exists check after the generation -> red
- drop the name shape check -> red
- report `adopted` as always true -> red
- remove the budget, so a wedged generation hangs forever -> red (the case hangs
  rather than failing, so it was run under a bounded runner)
- claim a write when the name was taken while generating -> red
- generate from an empty workflow -> red
- pin blank skill entries -> red

One of those mutations is the code I first wrote: `PlaybookGenerationError`'s
`args` is the *joined* message and the list lives on `.errors`, so reading
`args` would have flattened the list this method exists to return. The test
asserts the reasons come back one per entry.

**Driven over the real gateway's WebSocket**, on a build restarted onto this
head while the machine was idle.

Refusals, all before any generation -- the elapsed column is the point:

```
name '../victim'   0.00s  playbook names are kebab-case (^[a-z0-9][a-z0-9-]*$)
name 'a/b'         0.00s  same
name 'Upper'       0.00s  same
name '-lead'       0.00s  same
name ''            0.00s  name is required
workflow '   '     0.00s  workflow is required: the generator sees only this text
```

An unknown-shaped name answers a typed `config_validation_error` rather than
method-not-found, which is what says the method is registered.

The **timeout branch was exercised for real, not against a fake**: a genuine
generation was started over the wire, ran to the budget, and answered
`created: false` with `errors: ["generation did not finish within 180s"]`,
`path: ""`, nothing written and nothing adopted.

The host was left as found: the library is empty again and
`~/.raven/config.json` hashes to the backup taken before the check.

No UI change here -- this is the interface only.

## Risk

Security: the shape of this method is the security argument. It accepts a
description and never a spec, so a caller cannot write a graph of its own
choosing into the library -- the composer is the only thing that produces one,
and it validates and repairs before anything is saved. The name is matched
against `NAME_RE` before it reaches the store, so it cannot name a directory
outside the library. `workflow` is free text that reaches a model, which is the
same untrusted-input boundary the conversational entry already stands on; it
reaches no shell, no path and no credential.

**The success path could not be completed on the verification host, and the
reason is not in this change.** A clean generation there -- `generated in 1
round(s)`, no repairs -- takes 338s against a budget of 180s. The endpoint
answered in 2.0s and a one-message provider round trip in 2.4s, so the model
path is healthy; the budget is simply shorter than one clean generation. That
makes the conversational `create_playbook` unusable on that host too, which is
why it is filed separately rather than patched here: raising it means changing
the model-facing tool, and the number to raise it to should come from
understanding why one round costs that much, not from doubling one host's
timing. Reading the budget from the tool (rather than copying it) is this PR's
part of that: whoever fixes the number fixes both entries at once.

Backward compatibility: one new method. Nothing existing changes shape. The name
guard split is internal and the existing behaviour of every handler that used it
is unchanged, which the suite covers.

Rollback: reverts cleanly.

## Related Issues

N/A


raven/rpc/methods/playbooks.py
```diff
@@ -352,14 +385,364 @@ async def playbooks_oauth_clear(params: dict) -> dict:
...
-def register_playbooks_methods(dispatcher: Dispatcher) -> None:
+# ---------------------------------------------------------------------------
+# The library as a thing a person changes, not only reads.
+#
+# Everything below has been reachable from `raven playbook` since the library
+# shipped; what it has not been is reachable from anything else. A surface that
+# can draw a playbook as disabled and offer no way to enable it is showing a
+# state it cannot act on, which is the gap these close.
+#
+# Deliberately not here: `run` and `create`. Both take model time -- a graph to
```
gloryfromca: No blockers; suggestions only, and they are marked inline.

This phase comment says `playbooks.run` and `playbooks.create` are deliberately absent, but the final file defines and registers both below, so the module documents the opposite of its behavior. AGENTS.md also says not to leave transient task context in source; please remove or update this block.

gloryfromca (commented):

The implementation is coherent across the stacked library-management, run, and create changes. The canonical-name guard closes the store path traversal before lookup; run dispatch captures the named session's workdir, model binding, and task-local DAG origin; create preserves the generator/store/adopt path and handles collisions and generation failures without weakening the wire contract.

Named follow-up: RPC-started playbook nodes still lack the turn path's session usage binding, so their token usage can remain unattributed to the session. The author previously scoped this as accounting-only; it does not change which model/workdir executes or where the run reports.

I covered the full `github/main...HEAD` diff; AGENTS.md/CLAUDE.md and the canonical Playbook, RPC Protocol, Session, session-key, model-binding, and workdir terms; the store/config/runtime/executor/DAG/RPC transport callers; relevant stacked history; backward compatibility; generated contracts; and test integrity. Existing assertions were not weakened.

Verification: `uv run pytest tests/test_rpc_playbooks.py tests/test_playbook_executor.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py -x` passed 502 tests. Ruff check and format-check passed; source-language and large-file checks passed. Local TypeScript regeneration could not start because this worktree lacks `json-schema-to-typescript`; the current GitHub `ui rpc contract`, `TUI checks`, and all other checks pass on this head.

raven/rpc/methods/playbooks.py
Handsome-wzw: Agreed on both halves, and there was a third you did not have to name. Fixed in 6bf4bca1.

The block claimed `run` and `create` were deliberately absent while both are defined and registered below it, so the module read as documenting the opposite of itself — that is the half you caught. It is also transient task context, which is the AGENTS.md half.

The third is the one that made this worth more than a tidy-up: the *reason* the comment gave was false. It said a JSON-RPC call occupying the socket for minutes blocks every other call on it. This transport dispatches every frame as its own task (`asyncio.create_task` per frame in `ws.py`), so a slow handler blocks nothing else on that connection. I found that while designing `run` and corrected it in the PR description at the time, and then left the copy in the source — which is exactly how a false constraint outlives the belief that produced it. A stale comment misleads once; one carrying a wrong claim about the transport invites the next change to be shaped around a limit that is not there.

What replaces it is the rule the handlers actually keep, which does not expire with the next addition: each reaches the library through the door that already exists for it rather than around it — enabling writes the deny list the CLI writes, running goes through the runtime the model's own tool goes through, creation binds the composer both creation entries bind. A second path to the same library would be a second place for its rules to live.

I also swept the rest of the module for the same shape of claim and found none.

Gates on this head: `pytest -k "playbook or rpc or subagent_dag or agent_loop"` 3526 passed; ruff check and format-check both pass; `ui-web` `gen:check` matches the contract at 173 methods and `ui-tui` `lint:rpc` is in sync — a comment change should move neither, and it did not.

On the usage-attribution follow-up you named again: it is filed as its own issue with the mechanism and the three options, so it does not depend on anyone remembering this thread.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified in 6bf4bca1: the stale phase claim is gone, and the replacement accurately documents the shared CLI/RPC entry paths. I also agree with the concurrency correction; requests are scheduled independently rather than one slow handler occupying the socket. The thread is settled.

Verification: 502 focused tests passed; the touched file passes Ruff lint and format checks; all four GitHub unit shards passed.

gloryfromca (commented):

I reviewed the fix delta and rechecked it in the full pull-request context. The replacement removes the transient phase note and matches the actual dispatcher, transport concurrency, and shared CLI/RPC entry paths. This revision changes documentation only, so public behavior and backward compatibility are unchanged; tests were not weakened.

Coverage included the repository rules, the diff, affected callers and history, backward compatibility, test integrity, and the relevant RPC architecture. Verification: `uv run pytest tests/test_rpc_playbooks.py tests/test_playbook_executor.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py -x` passed all 502 tests; Ruff lint and format checks passed; all four GitHub unit shards and the other completed checks passed (the aggregate coverage job was still pending when reviewed).

gloryfromca (commented):

I reviewed the changes since 6bf4bca1 and rechecked the full `github/main...HEAD` diff. The added fixes consistently pass the cleaned name to the store, turn malformed YAML into validation findings, keep deletion on the user layer, and validate agents against the same live table a run uses. I found no new defect after checking the repository rules and domain vocabulary, the diff, affected callers and history, backward compatibility, test integrity, and the RPC/playbook architecture. Tests were added rather than weakened.

Verification: 507 focused RPC/playbook tests passed; Ruff lint and format checks passed; the source-language, large-file, and commit-message checks passed for the branch range. GitHub's four unit shards, coverage, RPC contract, and frontend checks are green. The repository-files job is red only because this branch is behind current `main` and its two-dot scan sees target-only CJK additions; `git merge-tree --write-tree HEAD github/main` is clean, so refreshing the branch base should remove that unrelated gate failure.

gloryfromca (commented):

This revision is a clean rebase onto current `main`. The range-diff shows all nine PR commits are patch-identical to the previously reviewed revision, and `github/main` is now the merge base, so the earlier base-sync gate failure is resolved. I rechecked the repository rules, full diff, affected callers and history, backward compatibility, test integrity, and the RPC/playbook architecture; the new target commits introduce no conflicting interaction.

Verification: 507 focused RPC/playbook tests passed, and Ruff lint and format checks passed. GitHub reports all four unit shards, coverage, repository-files, RPC contract, frontend, and remaining required checks green.

gloryfromca (commented):

This revision rebases the remaining six commits after the library-management stack merged into `main`. The range-diff shows those six patches are unchanged, while the three dropped commits are now supplied by the target branch. I rechecked the repository rules and domain vocabulary, full diff, affected callers and target history, backward compatibility, test integrity, and the RPC/playbook architecture. The newly merged session-isolation work is confined to the design-engine plugin and does not alter the session workdir/binding path used by `playbooks.run`.

Verification: 507 focused RPC/playbook tests passed, and Ruff lint and format checks passed. GitHub's repository, contract, frontend, lint, and three unit-shard checks are green; the fourth unit shard was still pending with no reported failure when reviewed.

Handsome-wzw:
Restacked onto current main (`73d6bb70`); force-pushed, no review had been given so nothing was dismissed.

This branch was stacked on #419, and #419 landed as a squash (`2f631a6f`). Replaying only this branch's own commits showed that main had absorbed more than the one commit I expected: `94e00b81` (a run is placed in its session) and `bdbc3dd4` (a check and a run judge a step's agent against one table) were both already in main, so both replayed empty and were dropped. That is why the commit count went from 6 to 3 while the diff did not shrink.

What is left is exactly this PR's own subject:

- `feat(rpc): a client can have a playbook written for it` - `playbooks.create`
- `refactor(rpc): the generation budget is read from the entry that declares it`
- `docs(rpc): the playbook section says what the module does`

One conflict, in `tests/test_rpc_playbooks.py`: main already carried the two `playbooks.validate` agent-table tests that `bdbc3dd4` had added, and the incoming side of the hunk was empty. I kept main's side and then verified that this branch's own three run-context tests (`_loop_with`, `test_the_run_carries_the_named_sessions_workspace_and_model`, `test_the_contexts_are_gone_once_the_call_is_over`, `test_a_session_whose_workspace_will_not_resolve_is_refused`) are all still present rather than assuming the auto-merge had placed them.

Verified after the restack: `tests/test_rpc_schema_match.py` + `tests/test_rpc_playbooks.py` 439 passed, `ui-web` gen:check (173 methods) and `ui-tui` lint:rpc both in sync, ruff check and format clean.

gloryfromca (commented):

This revision rebases the change after the prerequisite `playbooks.run` stack merged to `main`. Range-diff shows the surviving create, refactor, and documentation patches are unchanged from the reviewed revision, while the former prerequisite commits are now supplied by the target branch. I found no integration or compatibility regression.

I covered the repository rules and domain terms, the full diff, affected callers and history, backward compatibility, test integrity, and the RPC architecture. Creation still enters through the shared generator/store/adoption path, and the Python models, OpenRPC schema, and generated clients remain aligned. The test diff adds 201 lines and deletes none, so no test was weakened. The previously settled thread remains resolved.

Verification: `git diff --check github/main...HEAD` passed; 507 focused RPC/playbook tests passed; Ruff lint and format checks passed. GitHub's TUI, RPC contract, Python lint, pre-commit, repository, bridge, page, title, commit-message, wheel, and Windows checks passed; unit shards 1-3 passed and shard 4 was still running when this review was posted.

gloryfromca (commented):

This revision restacks the feature onto current `main`. Range-diff shows the implementation, refactor, and documentation patches are unchanged; the only adaptation is the generated web RPC method count moving from 173 to 177 to include the three methods added on `main` plus `playbooks.create`. The overlapping RPC model, schema, and generated-client changes compose cleanly, with no compatibility or architecture regression.

I covered the repository rules and canonical terms, the full target diff, affected callers and history, backward compatibility, test integrity, and the shared generator/store/adoption architecture. The test change still adds 201 lines and removes none. The owned thread remains resolved.

Verification: `git diff --check github/main...HEAD` passed; 515 focused RPC/playbook tests passed; Ruff lint and format checks passed. All GitHub checks, including the four unit shards, RPC contract, and coverage gates, passed.

0xKT (approved):

PR #430, opened 2026-09-15
fix(agent): a turn with a rerun coming is not salvaged first

ZuyiZhou:
## Summary

The terminal seam fires at the end of every answerless attempt, including one the turn
is about to run again. The gate behind it in the shipped research chain may spend 240
seconds writing an answer, and both ends that answer can meet are wrong.

- The rerun discards it, having charged the turn's own clock for the minutes it took.
  A 240-second purchase against a one-hour turn is survivable; against the budget left
  late in a turn it is the rerun itself, thrown away for an answer that was also thrown
  away.
- When the attempt ended answerless without erroring, the salvage is worse than wasted.
  The dead-end reading asks whether the trajectory ended on the model's own prose, and
  a salvage ends it on exactly that. The trigger is emptied, the rerun never happens,
  and the turn ships the manufactured answer: salvage winning by suppressing the thing
  measured to beat it. `dead_end.py` states that ordering as the policy already -- 23
  of 25 real duds produced content, gold hit rate 0 -- so this is the module's own rule
  reaching the code path that overrides it.

The loop now asks the caller, once per attempt, whether a rerun is coming, and skips the
seam when the answer is yes. The rerun budget belongs to the caller, so the caller is who
is asked; a loop nobody budgeted a rerun for asks no one and salvages exactly as before,
which is every agent but the one that asked for the budget.

Three decisions worth naming for review:

- **One evaluation, not two.** The retry loop acts on the answer the seam already got
  rather than working the same question out again. The two askings sit seconds apart,
  with a checkpoint commit and a persist between them, and a clock crossing its budget
  in that gap would turn a skipped salvage into a refused rerun -- a turn ending with
  neither. A spent clock is therefore a no to the rerun and a yes to the seam, in one
  place.
- **The loop keeps its own bound.** `while retries_left > 0 and pending["rerun"]` reads
  one counter for two questions: the loop asks how many reruns are left to run, the seam
  asks whether the attempt it is ending has one after it. Dropping the first half leaves
  a loop whose only exit is a value a callback writes, which no reader of the loop can
  check -- and the mutant that dropped it did not terminate.
- **The reading drops recovery scaffolding first.** It used to run after the attempt
  returned, on the list the loop had already dropped its own nudges and prefills from.
  It runs inside the attempt now, so it does that drop itself; otherwise a product that
  re-runs turns ending on the question is told they ended on a prefill the loop wrote.

Raised as a nonblocking follow-up on #410, whose branch this stacks on.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Full suite, Python 3.12, all extras:

- `uv run --frozen --python 3.12 --all-extras pytest -q`: 22,140 passed, 43 skipped.

Revert to red, production file only, tests kept:

- `git checkout <pr base> -- raven/agent/loop/turn_path.py` then the retry module:
  3 failed, 29 passed. The three are the three claims above that are behaviour changes.
  The other new tests are controls and pass on both sides by design: the last attempt
  still being salvaged, and an agent that budgets no rerun being salvaged as before.

Mutation, on the retry module (baseline 33 passed):

| mutant | result |
|---|---|
| the seam stops asking whether a rerun is left | 2 failed |
| a spent clock stops denying the rerun | 2 failed |
| the seam stops reading the answer | 3 failed |
| the caller stops being asked at all | 11 failed |
| the reasons stop dropping recovery scaffolding | 1 failed |
| CONTROL: a sibling line in the same block | 1 failed |

The scaffolding mutant survived the first round; the test that kills it was written for
that survivor and is the third commit.

Coverage, against the base branch:

- `make coverage-diff COVERAGE_BASE_REF=origin/feat/research_dead_end_verbatim`:
  100.00% (24/24 executable changed lines), threshold 90.00%.
- `make coverage-ratchet`: line 88.67% vs baseline 87.30% (+1.36pp), branch 81.59% vs
  79.43% (+2.16pp).
- `make coverage-baseline-check`: monotonic, delta +0.000000pp on both.

Lint and repo gates:

- `ruff check .` and `ruff format --check .`: clean, 2024 files.
- `lint-imports`: 10 contracts kept, 0 broken.
- `check_source_language.py`, `make check-large-files`, `git diff --check`: clean.

## Risk

Behaviour changes only for an agent that budgets a dead-end rerun, which today is the
research product alone; every other agent reaches the seam on exactly the paths it
reached before. For that one agent, an attempt with a rerun coming no longer produces a
salvaged answer -- which is the point, since that answer was either discarded or was
suppressing the rerun. The last attempt of a turn, and any attempt the clock has already
denied a rerun, still get the seam.

The loop hook surface is unchanged: `terminal_answerless` keeps its contract, its
context fields and `FACTORY_LOOP_SURFACE_VERSION`. What moved is when the loop offers it
a turn. `_run_agent_loop` gains one keyword-only-in-practice parameter defaulting to
`None`, so every existing caller is byte-identical in behaviour.

Rollback is the revert of the three commits; nothing is persisted or migrated.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the refreshed target diff, all three commits, the surrounding `_process_message` / `_run_agent_loop` control flow, `dead_reasons`, the production research hook's terminal salvage, direct callers, and the checkpoint/persistence ordering. I specifically checked retry bounds (including multiple reruns), wall-clock exhaustion, filtered reasons, transient recovery scaffolding, and the no-rerun compatibility path.

I also checked the repository rules and Runtime terminology, and verified that the tests were extended in the existing matching file without weakening or skipping prior coverage. No dependency or asset changes are in this PR.

Verification: `uv run pytest tests/test_agent_loop_dead_end_retry.py tests/test_agent_loop_iteration_hooks.py -q` passed 53 tests; pytest also reported four non-failing wall-clock-ceiling warnings for unmarked tests. `uv run pytest tests/test_agents_research_flow_finalize.py tests/test_agent_loop_dead_end.py tests/test_agent_loop_empty_recovery.py -q` passed 107 tests. `git diff --check github/feat/research_dead_end_verbatim...HEAD` passed, and every GitHub check is green.

PR #431, opened 2026-09-15
fix: isolate design artifacts and task state by session

daoxize1:
Give each design session its own working directory under designs using a slug and a hash of the full session key. Restore the same directory when the session resumes, including turns that skip the user inbound hook.

Keep review ledgers and auxiliary files in the session directory and provide workdirPerSession=false for the existing shared-directory behavior. Preserve existing shared files and task state without guessing ownership.

Verify concurrent sessions, resumed turns, directory name collisions, and completion isolation with regression tests and a live ACP rendering run.

## Summary

Design sessions using the same working directory previously shared Task State and output files. Give each session its own directory under `designs/`, named with a session slug and a hash of the full session key. Reuse that directory on resume and keep review ledgers and auxiliary files there.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```bash
uv run --no-sync pytest -q tests/test_design_engine_*.py tests/test_agents_design_launcher.py tests/test_ppt_engine_plugin.py tests/test_agent_workdir.py -x
```

Result: 265 passed. One existing PPT test produced a non-failing timing warning.

```bash
make check-large-files check-source-language COMMIT_RANGE=origin/main...HEAD
git diff --check
```

Result: passed. Pre-commit checks also passed for all seven changed files.

A live ACP scenario exercised two concurrent sessions, actual model calls, Chrome rendering, process restart, and session recovery. Both sessions generated independent HTML, PDF, PNG, review ledgers, and Task State. Revising session A left all nine session B artifact and state file hashes unchanged. PDF image pixels matched the corresponding previews.

The final verification command rechecked completed concurrent outputs and ran the two recovery turns:

```bash
DESIGN_E2E_OUTPUT=/tmp/raven-design-live-e2e-hx9Ohs/run3 DESIGN_E2E_RESUME=1 uv run --no-sync pytest -q -s /tmp/raven-design-live-e2e-hx9Ohs/test_design_session_e2e.py
```

Result: 1 passed. This was a local live test harness outside the repository. EverOS memory was disabled; Web UI interaction was not covered.

## Risk

## Related Issues

N/A


plugins-dist/design-engine/raven_design/plugin/hook.py
```diff
@@ -269,19 +271,36 @@ def name(self) -> str:
...
+        """Keep state and artifacts together across resumed and concurrent turns."""
+        if not self._cfg.workdir_per_session:
+            return bound
+        name = mint_slug(session_key.rpartition(":")[2], max_chars=48) or "session"
+        digest = hashlib.sha256(session_key.encode("utf-8")).hexdigest()[:16]
+        dirname = f"{name}-{digest}"
+        if bound.name == dirname and bound.parent.name == DESIGNS_DIRNAME:
+            return bound
+        own = bound / DESIGNS_DIRNAME / dirname
+        own.mkdir(parents=True, exist_ok=True)
```
gloryfromca: Blocking: validate the computed session directory before making it the live workdir.

`mkdir(..., exist_ok=True)` accepts a pre-existing symlink to a directory, and `workdir.repoint(own)` then promotes that symlink target to the live allowed root. I reproduced this with `restrict_to_workspace=True`: placing the predictable `designs/<session-slug>-<hash>` path as a symlink to a sibling directory caused `ReadFileTool(..., allowed_dirs=(workspace,)).execute("secret.txt")` to return the sibling's contents instead of rejecting the out-of-workspace read. Please preserve the original workspace fence when deriving this child so a pre-existing symlink cannot expand tool access.

gloryfromca (commented):
Blocking: the new session-directory handoff can escape the configured workspace through a symlink.

I reviewed the full diff and modified tests, traced the mutable workdir through AgentLoop hook ordering and the filesystem, shell, render, Task State, and checkpoint consumers, and compared the established ppt-engine mechanism and relevant history. I also checked AGENTS.md/CONTEXT-MAP.md constraints, naming, backward compatibility and migration behavior, and confirmed the tests were extended rather than weakened. The isolation, collision resistance, resume behavior, and `workdirPerSession=false` compatibility path otherwise look coherent.

Verification:
- `uv run --all-extras pytest -q tests/test_design_engine_plugin_face.py -x`: 28 passed.
- `uv run --all-extras pytest -q tests/test_design_engine_*.py tests/test_agents_design_launcher.py tests/test_ppt_engine_plugin.py tests/test_agent_workdir.py -x`: 263 passed, 2 skipped because Chromium could not load the host's missing `libatk-1.0.so.0`; one existing PPT timing warning.
- A plain `uv run pytest -q tests/test_design_engine_plugin_face.py -x` first failed at collection because the default environment does not install the local `raven_design` extra; rerunning with the repository's all-extras setup produced the passing result above.
- `git diff --check`, `check_source_language.py`, and `check_large_files.py`: passed. (`make` is unavailable in this environment, so I invoked the two gate scripts directly.)

plugins-dist/design-engine/raven_design/plugin/hook.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in fffa635744b6: the hook now resolves the caller root, rejects symlinks at both the designs directory and session leaf, restores the original root on refusal, and short-circuits before selector, model, or tools. I re-ran the former exploit path through both hook phases and the real ACP subprocess coverage; outside reads and writes remain refused.

gloryfromca (commented):

The new commits close my prior finding: unsafe session-directory redirects now fail closed before selector, model, tools, Task State, or artifact writes, while normal and symlinked caller roots retain the intended session isolation. I re-read the complete diff and changed tests, traced the hook short-circuit contracts and workdir consumers, checked the relevant history and ppt-engine precedent, and rechecked AGENTS.md/CONTEXT-MAP.md rules, backward compatibility, architecture boundaries, and that coverage was added without weakening existing assertions. No new findings.

Verification:
- `uv run --all-extras pytest -q tests/test_design_engine_plugin_face.py tests/integration/test_design_session_e2e.py -x`: 47 passed.
- `uv run --all-extras pytest -q tests/test_design_engine_*.py tests/integration/test_design_session_e2e.py tests/test_agents_design_launcher.py tests/test_ppt_engine_plugin.py tests/test_agent_workdir.py -x`: 282 passed, 2 skipped because Chromium cannot load the host's missing `libatk-1.0.so.0`; one existing PPT timing warning.
- `git diff --check`, source-language, large-file, and commit-message checks: passed.

My prior thread is resolved.

plugins-dist/design-engine/raven_design/plugin/hook.py
0xKT, line 1: The per-session directory mechanism now exists twice, and the two copies do
not agree about the hazard this PR names.

The README says this follows the deck agent's mechanism, and it does --
closely enough that both are a method called `_own_folder(self, bound,
session_key)` behind a `*_per_session` flag, both mint a directory under a
sibling-named folder, both `mkdir(parents=True, exist_ok=True)` and both
`workdir.repoint(own)`:

    plugins-dist/ppt-engine/raven_ppt/plugin/hook.py:280      (deck_per_session)
    plugins-dist/design-engine/.../plugin/hook.py:278         (workdir_per_session, new here)

Where they part is the reason this PR gives for its own checks:

    # The live workdir becomes a tool access root; child links must not
    # redirect it.

That statement is about `workdir.repoint`, not about design. The deck agent
repoints the same way and has no such check: in `_own_folder`
(ppt-engine/hook.py:280-298) the strings `resolve` and `symlink` appear zero
times, and the body goes straight from building the path to `mkdir` to
`repoint`.

Measured, one temp workdir whose `designs/` is a symlink pointing outside it:

    design-engine (this PR)  -> refused, ValueError: Symlinked design session
                                directory is not allowed
    design-engine, flag off  -> returns the caller's directory unchanged

I did not execute the deck agent's half: the acceptance venv has no
python-pptx, so `raven_ppt.plugin.hook` will not import here. Its absence of
a check is read from source, not run, and that is the one claim above I am
handing over unexecuted.

So one of two things is true, and both are worth a moment:

  - the hazard is real, in which case the deck agent has it today and this
    PR is the commit that makes that knowable; or
  - it is not, in which case the new checks and the turn-stopping refusal
    they carry are cost without a buyer.

This PR's side is the hardened one, so nothing here argues against merging
it. What it argues is that the guard belongs where both callers reach it.
`workdir.repoint` is the shared thing being protected, and neither plugin
owns it -- so a third plugin that repoints will start from whichever copy
its author reads first.

Not blocking. Also not a request to fix the deck agent in this PR: the fix
that closes it properly is one owner for "repoint the turn at a session
folder, safely", which is a different change from this one and should not be
smuggled into it.

One smaller thing from the same reading, filed here rather than on its own
thread because it is the same shape. The four mutants I ran against the new
guard:

    _own_folder returns `bound` unconditionally   -> 25 failed   killed
    drop the first symlink check                  -> 11 failed   killed
    drop the second symlink check (after mkdir)   -> 47 passed   SURVIVED
    EngineConfig.workdir_per_session default True->False -> 47 passed  SURVIVED

The fourth is not a defect: `from_slice` carries its own `True` literal
(config.py:297), mutating THAT one kills 25 tests, and a field default beside
a parser default is this file's established shape (`enabled` at :94/:106 and
every other flag). The PR is matching its neighbours.

The third I could not prove non-equivalent. The second check exists for a
symlink that appears after the first one passes, and I could not build that
window deterministically without also tripping the first check -- so I am
recording it rather than claiming it. If the window is real, one test that
swaps the directory between the two checks would pin it; if it is not, the
check can go.

gloryfromca (commented):

This revision only merges the current `github/main` into the previously reviewed feature head: the first parent is exactly `fffa635744b6`, the feature diff against the second parent is unchanged, and the merge has no combined-diff conflict-resolution edits. I rechecked the full PR diff, callers and architecture assumptions, AGENTS.md/CONTEXT-MAP.md constraints, backward compatibility, and test coverage; none changed from the prior clean revision.

The remaining nonblocking follow-up is the one @0xKT already named inline: give safe per-session workdir repointing one shared owner and migrate the pre-existing ppt-engine copy in a separate change. Their thread explicitly does not request that expansion here.

Verification on the merged tree:
- `uv run --all-extras pytest -q tests/test_design_engine_plugin_face.py tests/integration/test_design_session_e2e.py -x`: 47 passed.
- `uv run --all-extras pytest -q tests/test_design_engine_*.py tests/integration/test_design_session_e2e.py tests/test_agents_design_launcher.py tests/test_ppt_engine_plugin.py tests/test_agent_workdir.py -x`: 282 passed, 2 skipped because Chromium cannot load the host's missing `libatk-1.0.so.0`; one existing PPT timing warning.
- `git diff --check`, source-language, large-file, and commit-message checks: passed.

My own prior thread remains resolved; I did not resolve @0xKT's thread.

plugins-dist/design-engine/raven_design/plugin/hook.py
daoxize1: Could you please approve this PR？

gloryfromca: No blockers; suggestions only, and they are marked inline.

My review verdict on this exact head is nonblocking: it can merge as far as I am concerned. I cannot submit GitHub’s formal APPROVE event from this reviewer role; the review instructions reserve that action for a human. Verification on this head was 282 passed and 2 environment-related Chromium skips, with the diff, source-language, large-file, and commit-message checks passing.

daoxize1:
Could owner please approve this PR？

0xKT (approved):

PR #306, opened 2026-08-12
feat(agent): add MiniMax text-to-video backend

octo-patch:
Reason: Raven's video_generate tool lacks a MiniMax text-to-video backend with regional endpoint and async result handling.

## Summary

- Add MiniMax V2 text-to-video submission and task polling for MiniMax-H3.
- Add MiniMax V1 text-to-video submission, task polling, file retrieval, and download handling.
- Resolve global and China API endpoints from media or provider configuration.
- Keep download credentials scoped to the configured API origin.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- [x] Relevant tests pass locally
- [x] Relevant lint / type checks pass locally
- [ ] User-facing docs or screenshots are updated when needed

- `uv run pytest tests/test_media_gen.py tests/test_tool_registry_timeout.py -x` (11 passed)
- `uv run pytest tests/test_provider_auth_method.py -x` (42 passed)
- `uv run ruff check raven/agent/tools/media_gen.py raven/config/schema.py tests/test_media_gen.py` (passed)
- `uv run ruff format --check raven/agent/tools/media_gen.py raven/config/schema.py tests/test_media_gen.py` (passed)

No user-facing documentation update is required because the tool and configuration schema describe the new behavior.

## Risk

- [x] Security impact considered
- [x] Backward compatibility considered
- [x] Rollback path is clear for risky changes

Existing video behavior remains unchanged unless a MiniMax text-to-video model is selected. Reverting this commit restores the previous behavior.

## Related Issues

N/A

PR #433, opened 2026-09-15
feat(agent): add MiniMax image-to-video generation

octo-patch:
Reason: Raven's video_generate tool cannot submit a first-frame image to MiniMax video generation.

## Summary

- Add first_frame_image input with MiniMax-H3 V2 content mapping and MiniMax V1 image-to-video submission.
- Poll tasks, retrieve V1 output files, and save completed videos in the workspace.
- Use explicit MiniMax video credentials, regional endpoints, and origin-scoped download authorization.
- Document configuration, image sources, and supported models.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run --python 3.12 pytest tests/test_media_gen_minimax.py tests/test_media_gen_tool.py tests/test_tool_registry_timeout.py -x`: 117 passed.
- `uv run ruff check raven/agent/tools/media_gen.py raven/config/schema.py tests/test_media_gen_minimax.py`: passed.
- `uv run ruff format --check raven/agent/tools/media_gen.py raven/config/schema.py tests/test_media_gen_minimax.py`: passed.
- `make check-large-files check-source-language`: passed.
- `git diff --check`: passed.

HTTP requests were mocked; live paid generation was not exercised.

## Risk

MiniMax image animation requires an explicitly configured MiniMax video model and key.
Invalid image sources and incompatible model overrides return errors before submission.
Revert the commit to remove this integration.

## Related Issues

#306


raven/agent/tools/media_gen.py
```diff
@@ -1114,6 +1149,177 @@ async def execute(
         logger.info("video_generate: {} bytes via {} -> {}", len(data), model_id, path)
         return json.dumps({"success": True, "model": model_id, "path": str(path)}, ensure_ascii=False)
 
+    @staticmethod
+    def _resolve_key(config: "MediaToolConfig | None") -> str:
```
LivXue, line 1153: **Blocker - ask the concrete tool for the credential, not the base class.** This is the
first `_resolve_key` override in the module, and `raven/agent/tools/capabilities.py:283`
resolves the credential through a hardcoded base class:

    return _OpenRouterMediaTool.has_key(_resolved_media(cap, config))

`has_key` is a classmethod that calls `cls._resolve_key(config)`, so pinning `cls` to
`_OpenRouterMediaTool` means this override is never consulted. The divergence is specific
to the environment-variable route, which is one of the two `docs/minimax-video.md` offers:
a section that carries `apiKey` still resolves through the base ladder and reports
correctly, but a section that names only the model and relies on `MINIMAX_API_KEY` does
not.

Reproduced at this head in a read-only export
(`git archive ee9b6291a188d1eda8e81703a9cc1e30b962599e`), with `MINIMAX_API_KEY` set,
`OPENROUTER_API_KEY` unset, and `tools.media.video.model = "MiniMax-H3"`:

    _OpenRouterMediaTool.has_key(section) -> False     # what capabilities.py asks
    VideoGenerateTool.has_key(section)    -> True      # what this override answers
    has_credential(cap, config)           -> False
    configured_from(cap, config)          -> ''
    tool.api_key at call time             -> 'mm-secret'

So generation works and the diagnostic contradicts it. Before this PR the two answers
could not diverge, because nothing overrode the ladder; this PR is what turns that
hardcoded reader stale. Resolving through the capability's own tool class rather than the
base would close it for any later override as well.

raven/agent/tools/media_gen.py
```diff
@@ -1114,6 +1149,177 @@ async def execute(
...
+        options = params or {}
+        is_v2 = model_id == _MINIMAX_V2_MODEL
+        if is_v2:
+            if not prompt.strip() or len(prompt) > 7000:
+                return json.dumps({"error": "MiniMax-H3 requires a non-empty prompt of at most 7000 characters."})
+            duration = options.get("duration", 5)
+            if type(duration) is not int or not 4 <= duration <= 15:
+                return json.dumps({"error": "MiniMax-H3 duration must be an integer from 4 to 15."})
+            if options.get("resolution", "2K") != "2K":
+                return json.dumps({"error": "MiniMax-H3 resolution must be 2K."})
```
LivXue, line 1192: **Nonblocking - accept `768P` here, or say in the message why only `2K` is offered.** The
guard refuses every `resolution` except `2K` before any request is made, but MiniMax
documents `768P` and `2K` as the supported values for `MiniMax-H3`: the V2 create
reference lists both in the `resolution` property and repeats them in the model
description. A caller asking for the cheaper of the two documented resolutions gets a
client-side refusal rather than a video, and the `params` text at line 1056 repeats the
restriction to the model, so the model will not try `768P` either.

The neighbouring guards are right where the vendor puts them -- the 4-15 second duration
range and the 7000 character prompt cap both match, and the `type(duration) is not int`
test correctly rejects a bool -- so this is the only one stricter than the API. Checked
against https://platform.minimax.io/docs/api-reference/video-generation-v2-create ; the
code path itself is code-reading, not executed. Either widen the accepted set to the two
documented values, or keep the restriction and give the reason in the error so a deployer
is not left guessing.

raven/agent/tools/media_gen.py
```diff
@@ -1030,13 +1044,18 @@ class VideoGenerateTool(_OpenRouterMediaTool):
...
+                "type": "string",
+                "description": "MiniMax first-frame image: public HTTP(S) URL or image data URI",
             },
             "params": {
                 "type": "object",
                 "description": (
-                    "Optional extra provider params merged into the request, e.g. "
-                    '{"duration": 5, "aspect_ratio": "16:9"} (3-15s; 16:9/9:16/1:1)'
+                    "Optional generation settings. MiniMax-H3 uses duration (4-15 seconds), "
+                    "resolution (2K), ratio, and callback_url. MiniMax V1 supports duration, "
```
LivXue, line 1057: **Nonblocking - drop `ratio` from the advertised options, or say that MiniMax ignores
it.** Every path through this integration requires a first-frame image -- the guard at
line 1082 refuses the call without one -- and MiniMax documents that for image-to-video
the aspect ratio is taken from the input image and `ratio` is always `adaptive`: another
legal value is accepted and then ignored rather than rejected. So this sentence points
the model at a knob that cannot do anything on any call this tool is able to make, and
`ratio` sits in the V2 `allowed` set at line 1193, so it is forwarded rather than
refused.

Checked against https://platform.minimax.io/docs/api-reference/video-generation-v2-create
("the aspect ratio is determined by the input image and `ratio` is always `adaptive`");
code-reading on this side.

raven/agent/tools/media_gen.py
```diff
@@ -1030,13 +1044,18 @@ class VideoGenerateTool(_OpenRouterMediaTool):
...
+            "first_frame_image": {
+                "type": "string",
+                "description": "MiniMax first-frame image: public HTTP(S) URL or image data URI",
             },
             "params": {
                 "type": "object",
                 "description": (
-                    "Optional extra provider params merged into the request, e.g. "
-                    '{"duration": 5, "aspect_ratio": "16:9"} (3-15s; 16:9/9:16/1:1)'
+                    "Optional generation settings. MiniMax-H3 uses duration (4-15 seconds), "
```
LivXue, line 1056: **Nonblocking - keep the OpenRouter options in this description; the default model still
uses them.** This replaces the whole `params` text with MiniMax-only guidance, but
`default_model` at line 1035 is still `kwaivgi/kling-v3.0-std`, so every deployment that
has not configured MiniMax -- which is all of them until someone does -- now reads a
description that names no option it can use. The text this removed named `duration`,
`aspect_ratio`, the 3-15s range and the three aspect ratios.

The capability itself is intact: the OpenRouter branch still forwards `params`
unfiltered at line 1099, so `aspect_ratio` keeps working. What changed is that the model
is no longer told it exists. The tool `description` at line 1036 also drops "(async;
takes a while)", which applied to both backends and is the only place the call's latency
was stated. Code-reading only. Naming both backends in the one description would keep the
default path documented.

raven/agent/tools/media_gen.py
```diff
@@ -1114,6 +1149,177 @@ async def execute(
...
+                "content": [
+                    {"type": "text", "text": prompt},
+                    {"type": "image_url", "image_url": {"url": first_frame_image}, "role": "first_frame"},
+                ],
+                "resolution": options.get("resolution", "2K"),
+                "duration": options.get("duration", 5),
+            }
+            if httpx.URL(api_base).host != "api.minimaxi.com":
+                allowed.discard("aigc_watermark")
+            body.update({key: value for key, value in options.items() if key in allowed})
```
LivXue, line 1205: **Nonblocking - tell the caller when a `params` key was dropped.** This filter and its V1
twin at line 1210 keep only the keys in their `allowed` set and discard the rest in
silence, and the call then returns `success: true`, so a caller who passes a key this
integration does not carry gets a video that ignored it with no way to find out. This
tool's own `params` description advertised `aspect_ratio` until this PR rewrote it, so
that exact key is a realistic input rather than a hypothetical one.

Note the asymmetry with the OpenRouter branch, which forwards `params` unfiltered at line
1099 (`body.update(params)`): the same call shape is permissive on one backend and
silently lossy on the other. Code-reading only. Either return the dropped keys in the
result, or refuse an unknown key the way the resolution guard already refuses an unknown
value.

raven/agent/tools/media_gen.py
```diff
@@ -1114,6 +1149,177 @@ async def execute(
...
+                    )
+
+                video_origin = httpx.URL(video_url)
+                api_origin = httpx.URL(api_base)
+                same_origin = (video_origin.scheme, video_origin.host, video_origin.port) == (
+                    api_origin.scheme,
+                    api_origin.host,
+                    api_origin.port,
+                )
+                download_headers = headers if same_origin else None
```
LivXue, line 1277: **Nonblocking - route this through `_api_headers_for` rather than re-encoding the
rule.** Lines 1272-1276 re-implement the scheme/host/port comparison that
`_is_api_origin` already performs, and `headers if same_origin else None` on this line
re-implements `_api_headers_for`, which exists for exactly this decision: whether a URL
the provider handed back may carry the bearer token. That rule is now encoded twice in
one file, and only the older encoding is covered by the existing tests.

The copy also skips the warning the wrapper emits --
`logger.warning("media_gen: {} is not the configured API origin; fetching it without
credentials", url)` -- so a token deliberately withheld from a third-party CDN leaves no
trace in the log. Worth noting the MiniMax poll calls pass `headers` directly rather than
through the wrapper too; those URLs are built from `api_base` so they are same-origin by
construction, but the OpenRouter `_poll` routes even those through `_api_headers_for`.

The behaviour as written is correct and withholding the header is the right call. This is
about the second encoding drifting from the first. Code-reading only.

raven/agent/tools/media_gen.py
```diff
@@ -1114,6 +1149,177 @@ async def execute(
...
+        self,
+        prompt: str,
+        model_id: str,
+        params: dict[str, Any] | None,
+        first_frame_image: str,
+    ) -> str:
+        api_base = self._minimax_api_base()
+        headers = {"Authorization": f"Bearer {self.api_key}", "Content-Type": "application/json"}
+        options = params or {}
+        is_v2 = model_id == _MINIMAX_V2_MODEL
```
LivXue, line 1184: **Question - what happens when the next V2 model is added?** Routing into this
integration is a membership test against `_MINIMAX_IMAGE_VIDEO_MODELS`, but the protocol
choice on this line is equality against a single constant. MiniMax documents
`MiniMax-H3-Max` as the other model on the same V2 endpoint, with its own limits
(`480P`/`768P` and explicitly no `2K`, and 5-15 seconds rather than 4-15), so adding it
to the frozenset -- the one obvious way to support it -- would route it to
`/v1/video_generation` with a V1 body, and nothing would say so: the V1 branch has no
model check, so the request would go out and fail at the vendor rather than here.

I did not file the missing model itself as a defect, since nothing in the PR claims to
cover it; the question is only whether the two lists should be able to disagree. Deriving
`is_v2` from a set of V2 models rather than from one name would make them move together.
Code-reading only.

raven/agent/tools/media_gen.py
```diff
@@ -66,6 +66,20 @@
     from raven.config.schema import MediaToolConfig
 
 _DEFAULT_BASE = "https://openrouter.ai/api/v1"
+_MINIMAX_IMAGE_VIDEO_MODELS = frozenset(
```
LivXue, line 69: **Question - should "this deployment is MiniMax" have one definition?** It is written
twice in this PR: this frozenset of seven model names, which decides routing, and a
`("MiniMax-H", "I2V-")` prefix test at `raven/config/schema.py:2469`, which decides
whether the OpenRouter chat key is lent to the section. The two agree on today's seven
names, but they are different predicates over different inputs: a later `MiniMax-H...`
name that reaches the config rule and not this set would stop borrowing the OpenRouter
key and then fail with a missing-key error instead of routing to MiniMax.

Worth flagging that #432 introduces a third spelling of the same question in the same
function -- `is_minimax_image_base(tool.api_base)`, keyed on the base URL rather than on
the model name -- so whichever of the two lands second has to reconcile them. A single
predicate that both the tool and the config rule import would remove the class of
mismatch. Code-reading only.

LivXue:
Blocking: 1 finding needs correction before this revision can merge, and 2 questions are
open for the author; see the inline notes.

Scope: I reviewed only the `raven/agent/**` portion of this PR -- 1 of 4 changed files
(`raven/agent/tools/media_gen.py`, +211 -5). The remaining 3 files
(`tests/test_media_gen_minimax.py`, `docs/minimax-video.md`, `raven/config/schema.py`)
were read as context but are not covered by this verdict.

This is a well-fenced addition and it reads as one. The MiniMax branch sits entirely
ahead of the existing OpenRouter path without changing a line of it, and the AST symbol
delta is 8 added, 0 removed: nothing was re-homed, so there are no stranded readers to
hunt. All four client-side refusals fire before any request goes out. The endpoint paths,
the model names, the `role: first_frame` content shape, both poll routes, the status
vocabularies and the `task_id`/`file_id` handoff all match MiniMax's published schemas.
The download drops the Authorization header when the returned URL is not on the
configured API origin, which is the right call and not one every integration makes. And
the new credential is explicit rather than borrowed, with that intent carried in both the
tool and the config rule.

What I examined: the in-scope diff at `ee9b6291a188`; the credential ladder
(`_resolve_key` / `has_key` / `api_key`) and its readers in
`raven/agent/tools/capabilities.py`; the tool schema and both request bodies against the
MiniMax V2 create and query references and the V1 image-to-video reference; the two
option filters against the documented parameter sets; tool registration, roster and i18n
surfaces for the new schema field; and the merge interaction with #432.

Confirmed findings, inline:

- `:1153` The capability report resolves credentials through a hardcoded base class, so
  this PR's new `_resolve_key` override is never consulted and `raven doctor` reports a
  working MiniMax video section as having no credential. Specific to the
  `MINIMAX_API_KEY` route. Reproduced.
- `:1192` `resolution` is restricted to `2K`, but MiniMax documents `768P` as supported
  for `MiniMax-H3`.
- `:1057` The `params` description advertises `ratio`, which MiniMax ignores on every
  image-to-video call this tool can make.
- `:1056` The same description drops every OpenRouter option while `default_model`
  remains `kwaivgi/kling-v3.0-std`, so the majority path is now undocumented to the model.
- `:1205` Unknown `params` keys are discarded silently and the call still reports success.
- `:1277` The download re-encodes the origin rule that `_api_headers_for` already
  implements, and skips the warning it logs.
- `:1184` The protocol split keys on one model name while routing keys on a set of seven.
- `:69` "This deployment is MiniMax" has two definitions in this PR, and a third in #432.

Two findings have no line in this diff to anchor to:

- `raven/agent/tools/capabilities.py:130-137`: the static `video_generate` capability row
  still carries `env_var="OPENROUTER_API_KEY"` and `cost_note="Billed per call; needs
  prepaid OpenRouter credit to run at all."`. `raven doctor` prints both verbatim, so a
  deployer following this PR's own doc page is told to buy OpenRouter credit for a call
  that will never reach OpenRouter. Same root cause and same fix area as the blocker
  above. The file is untouched by this PR, so there is no hunk to anchor to.
- `raven/config/schema.py:2469` is outside the reviewed package: this PR and #432 rewrite
  the same return statement in `borrows_openrouter_key`. `git merge-tree <merge_base>
  <head432> <head433>` reports two files changed in both and exactly one conflict, here;
  `media_gen.py` is touched by both and merges cleanly. Against current `main` this branch
  alone merges cleanly. Whichever lands second has to keep both guards, and they are
  written against different inputs (model name here, base URL in #432).

Verification:

- Reproduced the credential mismatch at this head in a read-only export
  (`git archive ee9b6291a188d1eda8e81703a9cc1e30b962599e`), with `MINIMAX_API_KEY` set and
  `OPENROUTER_API_KEY` unset.
- Ran `tests/test_media_gen_minimax.py`, `tests/test_media_gen_tool.py` and
  `tests/test_tool_capabilities.py` at this head: 145 passed. The two capability findings
  are therefore untested rather than newly broken; no existing assertion states the
  base-class `has_key` semantics.
- Ran `git merge-tree` against #432's head for the conflict claim above.
- Computed the AST symbol delta between the merge base and this head for the changed file.
- Checked the resolution, duration, prompt-length, `ratio` and model-availability claims
  against MiniMax's published V2 create reference.
- Everything else in these notes is code-reading, and each note says so.

- The completed video is fetched with a bare `client.get` rather than `guarded_fetch`. The
  existing OpenRouter video path does the same at the merge base, so the twin is
  pre-existing and this is disclosed rather than filed. It is the sharpest thing I swept
  clean.
- `_minimax_error` reads only the V1 `base_resp` envelope. On V2 it is inert and
  `raise_for_status` plus `_format_http_error` carries the vendor's message instead, so
  errors still reach the caller with the vendor's text.
- `first_frame_image` is not in `required`. It is genuinely optional for the OpenRouter
  backend the same schema serves, and the guard returns a clear error, so this is not a
  defect.
- V1's own limits (prompt length, `duration` 6 or 10, per-model resolution) are not
  validated where the V2 model's are. They are model-dependent in the vendor's table and
  the API answers a readable 400 that `_format_http_error` surfaces, so this is a
  degradation rather than a silent failure.
- `_minimax_api_base` strips a configured endpoint suffix longest-first, so a base already
  carrying `/v1/video_generation` normalises correctly rather than half-way.
- `timeout_seconds = 660.0` is tight against the worst case (120 submit + 600 poll + 180
  download), but the existing OpenRouter path has the same structure, the same shared
  `_POLL_TIMEOUT_S` and the same ceiling, so it is inherited rather than introduced.
- `callback_url` is forwarded. It is a documented MiniMax parameter and MiniMax calls it,
  not this process, so the SSRF-shaped reading does not apply on this side.
- Tool registration, rosters and i18n: `first_frame_image` appears nowhere else in the
  tree, and `agents/*/config.json`, `raven/acp/tool_kinds.py` and `i18n/messages.json` all
  key on the tool name `video_generate`, which does not change. Clean. The one
  registration-shaped surface left stale is the capability table above.
- Added lines are ASCII throughout and no comment was added to `media_gen.py`, both
  matching the repo's rules for this file.
- `docs/minimax-video.md` is not listed in the `## Index` of `docs/README.md`, which
  otherwise names every top-level entry in that directory. A one-line omission.

LivXue (changes requested):
Requesting changes on the `raven/agent/**` slice of this PR: 1 blocker and 2 open questions are filed as inline notes on `raven/agent/tools/media_gen.py`. The scope statement, the verification log, and two findings that have no line in this diff to anchor to are in the PR-level summary comment.

PR #434, opened 2026-09-16
feat(plugins): hosted memory backends (Mem0, Zep, MemOS)

gloryfromca:
## Summary

Three hosted memory services -- Mem0 Platform, Zep Cloud and MemOS Cloud -- join Raven as memory backends, one distribution each (`plugins-dist/mem0-memory`, `zep-memory`, `memos-memory`), every one a manifest, a backend and an onboarding factory: the shape `docs/memory-plugin-architecture.md` always described for a second backend. They are the first plugins that are pure clients: an HTTP call per contract method over the `httpx` the host already carries, no vendor SDK, no process, no local storage, and no use of the host's embedding endpoint. Each wheel is four files and depends on `raven` and `httpx` alone. The MemoryBackend seam now holds both a local service the plugin spawns and thin cloud clients.

Key decisions:

- What the three share lives in the host next to the contract tests, for any author of a hosted backend: `raven.memory_engine.HttpMemoryBackend` owns the contract's failure semantics (a fault costs one call, never the turn), the recall budget (the client gives up at 2.5s, before the host's 5.0s), and the write semantics: `store` returns `True` when the service accepts the batch -- all three services extract asynchronously, and nothing polls a status endpoint on the turn path. `raven.memory_engine.ApiKeyOnboardStep` is the matching wizard screen (one key, one probe, one slice). A plugin's backend fills in its service's requests and how to read the answers.
- Each backend reads the config slice keyed by its own contribution name (`plugins.config.mem0` / `.zep` / `.memos`), so the three API keys never share a slot and switching `memory.backend` leaves the other slices in place. `<NAME>_API_KEY` in the environment beats the slice; a key that came from the environment is not copied into the file. Identity comes from `ServiceLocator` only.
- The owner on the wire is the owner of the call: `recall` / `recall_session` use their `user_id`, `store` uses `metadata["user_id"]` (a sub-agent's scope) and falls back to the host's `ServiceLocator.user_id`; Zep creates the user and thread for whichever owner writes. A flat service has no agent track, so `agent_id` calls are answered empty without a request.
- Zep caps a write at 30 messages per request and 4,096 characters per message; the importer batches 100. A store splits into as many requests as needed, a long message into consecutive pieces of the same role, and the fake refuses what Zep refuses.
- `feedback` is a no-op for all three (flat backends answer the agent track empty, so the host never hands them skill ids). `delete` takes the native id the backend's own recall put in `metadata`.
- The plugins reach the engine through `raven.memory_engine` only; the face guard now scans `plugins-dist/` too (and routed the design engine's two direct submodule imports it found through the face). The two new credential readers are argued into the provider-auth allowlist as a hosted service's own key.
- The one change to host behaviour: with more than one memory plugin installed, step 4 of `raven onboard` opens with a single choice (each backend by name, plus Off, cursor on the recorded backend) and runs only the chosen screen. One installed backend behaves exactly as before.

Not in this PR, deliberately: a settings-page card for these backends (the existing card is bound to EverOS by name), the memory browser (its wire is EverOS's), `install.sh` pulling the new wheel (the release builds it; install with `uv tool install --with`), and any dataset ingestion path for evaluation.

Stacked on #414: the base branch is `refactor/memory_backend_seams`, which carries the contract this plugin implements. Retarget to `main` once #414 lands.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Automated (fakes of each service over `httpx.MockTransport`: auth, per-user buckets, asynchronous ingestion, one-shot faults):

```
uv run pytest tests/test_hosted_memory_backend_contract.py tests/test_mem0_memory_backend.py tests/test_zep_memory_backend.py tests/test_memos_memory_backend.py tests/test_hosted_memory_plugins.py tests/test_hosted_memory_onboard.py tests/test_plugin_boundary.py tests/test_cli_onboard_commands.py tests/test_core_plugin_stack.py tests/test_i18n_boundary.py tests/test_rpc_settings.py tests/test_everos_backend_contract.py tests/test_memory_engine_face.py tests/test_provider_auth_method.py tests/test_design_engine_*.py -q
810 passed, 2 skipped (design engine: no system chromium) in 9.01s
```

The shared `MemoryBackendContractTests` / `LifecycleContractTests` run against all three, nothing overridden. Seventeen mutations (drop the owner XOR guard, pin recall to the host owner, ignore the metadata owner on store, send a Zep batch unsplit, drop the two auth allowlist entries, let a plugin import an engine submodule, poll the status endpoint after `store`, raise the recall timeout, swap the auth scheme, replace instead of merge the config slice, probe without a key, drop the native id, walk the thread instead of the graph, drop the preference list, classify 429 as missing, log the auth header, return True on a 4xx, remove the chooser) each turned their covering tests red; files restored byte-for-byte.

Real services, run by hand with the three keys exported (this file is outside CI's collection and fails, rather than skips, without keys):

```
uv run pytest tests/integration/test_hosted_memory_real_cloud.py -q -n 0
9 passed in 61.98s
```

The first real run caught what the fakes could not: MemOS refuses `memory_ids` and `user_id` together on `/delete/memory` (code 40071). The payload and the fake were corrected; the second run is the one above. Ingestion latency observed: Mem0 about 9s, Zep 45-60s, MemOS about 2s.

Manual acceptance in an isolated `RAVEN_HOME` (wizard and TUI driven through tmux, WebUI through playwright-cli): the chooser appears with everos / mem0 / memos / zep / Off; choosing mem0 records `memory.backend` and a `plugins.config.mem0` slice holding only `base_url`; re-running the wizard and choosing zep keeps the mem0 slice; `raven doctor` prints `mem0: ok https://api.mem0.ai` (exit 0) and, with the variable unset, `mem0: missing  no API key: set MEM0_API_KEY or run raven onboard` (exit 2); a fact told to the TUI (with the agent's request to write a local profile file denied) is recalled verbatim in a fresh TUI session, and Mem0's own search shows it stored under the configured user with `run_id` = the TUI session id; the WebUI memory page explains that long-term memory runs on mem0 and this page reads EverOS's store only; `raven import` without a key stops at the health gate with nothing read; none of the three keys appears in any log or workspace file.

Gates: `ruff check` / `ruff format --check`, `make check-source-language`, `make check-large-files`, `make lint-imports` (10 contracts kept), `make lint-deps`, `make check-commits` all clean.

## Risk

User-visible: `raven onboard` step 4 shows a backend chooser when more than one memory plugin is installed. Installs with only `everos-memory` see no change. Nothing changes for a configured EverOS setup; `everos` stays the default `memory.backend`.

Data leaves the machine only when a user selects one of these backends: every `store` sends the conversation slice to that vendor. Keys are read from the environment or the plugin's own slice, sent only as the vendor's auth header, and never logged. Each plugin ships as its own wheel and none is pulled by `install.sh`, so no existing install gains one without an explicit `uv tool install --with`.

Rollback: set `memory.backend` back to `everos` (or run `raven onboard` and choose it), or uninstall the distribution. The host change to the wizard is a single function in `onboard_commands.py` with the old single-screen path intact; the two `raven.memory_engine` helper modules are unused unless a plugin imports them.

## Related Issues

N/A (stacked on #414)


raven/memory_engine/http_backend.py
```diff
@@ -0,0 +1,336 @@
...
+    ENV_KEY: ClassVar[str]
+    SIGNUP_URL: ClassVar[str]
+    ROLES: ClassVar[frozenset[str]] = frozenset({"user", "assistant"})
+
+    def __init__(self, ctx: PluginContext, *, client: httpx.AsyncClient | None = None) -> None:
+        self._config: dict[str, Any] = dict(ctx.config or {})
+        self._logger = ctx.logger
+        # Identity comes from the host, never from the slice: a second copy of
+        # the owner id is how store and recall silently drift apart.
+        self._user_id: str = ctx.services.user_id
```
gloryfromca, line 130: Blocking: honor the owner supplied on each memory call instead of pinning every request to the host owner.

The contract gives `recall`/`recall_session` a per-call `user_id`, and gives `store` explicit `metadata["user_id"]` / `metadata["agent_id"]` overrides for sub-agent writes. This value is captured once from `ServiceLocator`, every request builder uses it, and the call arguments are used only for XOR validation. A focused Mem0 reproduction with `ServiceLocator.user_id="host-owner"` and `recall(user_id="requested-owner")` put `host-owner` on the wire and failed to return the requested tenant's fact. The ordinary sub-agent trace path (`prime_from_turn`) also supplies a different owner in store metadata; ignoring it files that agent's content under the host and makes it eligible for normal host recall. The default can still come from `ServiceLocator`, but the per-call override has to control the request when present (with flat backends continuing to return empty for the agent track).

plugins-dist/cloud-memory/raven_cloud_memory/raven-plugin.toml
```diff
@@ -0,0 +1,40 @@
+[plugin]
+id                 = "cloud-memory"
```
gloryfromca: Blocking: isolate each vendor credential under the canonical plugin configuration model.

Raven's public config contract and `CONTEXT.md` key `plugins.config` by manifest plugin id, and `_resolve_plugin_config_slice` deliberately prefers that canonical id over the contribution-name fallback. Grouping Mem0, Zep, and MemOS under `cloud-memory` therefore makes `plugins.config["cloud-memory"].api_key` feed all three factories. If an operator follows the canonical contract for Mem0 and later selects Zep, the Mem0 token is sent in a request to Zep before being rejected. `test_the_plugin_id_is_never_a_shared_slice` actually demonstrates this behavior; a manifest comment and onboarding's fallback keys cannot enforce the claimed separation. A nested per-backend canonical slice, separate plugin ids/manifests, or an explicit host seam for contribution-specific configuration would avoid the cross-vendor disclosure.

plugins-dist/zep-memory/raven_zep/backend.py
```diff
@@ -0,0 +1,142 @@
...
+        return False
+
+    def _store_call(self, session_id: str, messages: list[dict[str, Any]]) -> Call:
+        rows = []
+        for m in messages:
+            row: dict[str, Any] = {"role": m["role"], "content": m["content"]}
+            if m.get("timestamp"):
+                row["created_at"] = m["timestamp"]
+            rows.append(row)
+        return Call("POST", f"/api/v2/threads/{session_id}/messages", json={"messages": rows})
```
gloryfromca, line 85: Blocking: split or reject Zep writes before exceeding its request limits.

This forwards the whole normalized `store` batch in one `thread.add_messages` request. Raven's importer batches up to 100 messages / 30,000 characters, while Zep documents a maximum of 30 messages per call and 4,096 characters per message: https://help.getzep.com/v3/adding-messages#message-limits. A 31-message reproduction produced one 31-message request because the fake does not model the vendor limit; the real service returns 400, after which the store pipeline retries the identical payload and `raven import` leaves the source failed. The backend needs to respect both limits (or return a clear unsupported result before sending) so the advertised import path works for normal conversation histories.

gloryfromca (commented):
Blocking: per-call owner isolation, vendor credential isolation, and Zep request sizing must be corrected.

Three concrete blockers are marked inline. There is also a repository-rule blocker in the commit history: both commits use `Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>`, while AGENTS.md requires `Co-authored-by: Claude (<actual-model-id>) <noreply@anthropic.com>`. That has no diff line to anchor, but it needs correction before merge.

Verification: `uv run --extra dev pytest tests/test_cloud_memory_backend_contract.py tests/test_cloud_memory_mem0.py tests/test_cloud_memory_zep.py tests/test_cloud_memory_memos.py tests/test_cloud_memory_plugin.py tests/test_cloud_memory_onboard.py tests/test_plugin_boundary.py tests/test_cli_onboard_commands.py tests/test_core_plugin_stack.py tests/test_i18n_boundary.py -q` passed (575 tests). A first run without `--extra dev` in a fresh environment failed with 7 collection errors and 11 failures because the optional plugin distributions were not installed; that is the real setup-dependent result, not a test failure in the change. Ruff, source-language, large-file, and commit-message scripts passed. Real-cloud tests were not run because no service credentials were available.

Covered: AGENTS.md/CLAUDE.md and CONTEXT-MAP/CONTEXT rules, the complete diff, runtime/import/sub-agent/onboarding callers, target and branch history, backward compatibility, test changes (including whether existing tests were narrowed), and plugin/admission/assembly architecture.

plugins-dist/cloud-memory/raven_cloud_memory/raven-plugin.toml
gloryfromca: Blocking: the remaining owner, Zep batching, architecture-test, public-face, and attribution issues still need changes.

This credential-isolation finding is fixed. Mem0, Zep, and MemOS now have distinct manifest plugin IDs and distinct contribution-name fallback keys, so both supported config lookup paths keep each vendor token in its own slice. Closing this thread; the other findings remain independent.

raven/memory_engine/__init__.py
```diff
@@ -98,6 +100,10 @@
     "upsert_section",
     "LifecycleContractTests",
     "MemoryBackendContractTests",
+    "ApiKeyOnboardStep",
```
gloryfromca, line 110: Blocking: complete and use the public Memory Engine face for the hosted plugin API.

The architecture contract says consumers enter through `raven.memory_engine` so the underlying modules remain rearrangeable, and this revision explicitly adds these helpers to that face. Nevertheless, all three separately shipped plugins import `raven.memory_engine.http_backend` and `raven.memory_engine.api_key_onboard` directly. The existing face guard misses this because it scans only `raven/`, not `plugins-dist/`; a Raven release can therefore move an internal module while preserving the face and make every installed hosted-plugin wheel fail at import. The face is also incomplete for the current consumers: Mem0/MemOS need `clamp_score`, and Zep needs `STORE_TIMEOUT_S`, but neither is exported here. Export the supported helper surface (or define another documented public module), route the plugins through it, and extend the boundary test to scan shipped plugin sources.

raven/memory_engine/api_key_onboard.py
```diff
@@ -0,0 +1,110 @@
...
+                    choice = ui.failure_choice(options, non_interactive=non_interactive)
+                    if choice == "rekey":
+                        continue
+                    return StepOutcome.DISABLED
+                ui.console.print(ui.t("  [green]✓ {name} connected.[/green]", name=cls.NAME))
+            self._record(api_key if not from_env else None)
+            return StepOutcome.CONFIGURED
+
+    def configured(self) -> bool:
+        return bool(os.environ.get(self._cls.ENV_KEY) or self._ctx.config.get("api_key"))
```
gloryfromca, line 80: Blocking: restore the provider-auth architecture test for the new credential readers.

The full suite and current CI both fail `test_only_the_auth_module_decides_configuredness_from_a_key`: it reports this line, `_record` at line 106, and `http_backend.py:117` as unaccounted key reads. These are hosted-service/plugin credentials rather than LLM-provider configuredness, so they appear to be legitimate exceptions, but this repository requires every such reader to be named and argued in that invariant's allowlist. Add the narrow, documented exemptions (or route the reads through an already-approved owner) so the architecture gate is green; suppressing or skipping the test would not address the ledger.

gloryfromca (commented):
Blocking: honor per-call memory owners, enforce Zep write limits, restore the architecture gates, use the public engine face, and fix the required attribution trailers.

The vendor-credential issue is fixed by splitting Mem0, Zep, and MemOS into distinct plugin IDs; that thread has been closed. The prior owner-routing and Zep-limit threads remain open because those implementations are unchanged. Two additional findings are inline: the separately shipped plugins bypass the Memory Engine face introduced for them, and the new credential readers leave the provider-auth invariant red.

All four branch commits still use `Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>`. AGENTS.md requires the exact `Co-authored-by: Claude (<actual-model-id>) <noreply@anthropic.com>` trailer, so the history still needs correction with the actual contributing model ID.

Coverage this round: repository rules in AGENTS.md, CLAUDE.md, and CONTEXT.md; the complete target diff plus the `fc4fa29a5f1a..471d9941bc63` delta; callers in onboarding, plugin config resolution, store/import paths, and the MemoryBackend contract; commit history and backward compatibility; test changes; and the documented plugin/Memory Engine boundaries. The new 180-second real-cloud poll ceiling is supported by recorded Zep timings and does not suppress a correctness assertion, so it is not a finding.

Verification:
- Focused hosted-memory/onboarding/boundary suite: 545 passed (two asyncio cleanup warnings).
- `uv run --extra dev pytest tests -q --ignore=tests/integration/test_hosted_memory_real_cloud.py`: 2 failed, 22,096 passed, 65 skipped. The relevant failure is `test_only_the_auth_module_decides_configuredness_from_a_key`; CI reproduces it. The other failure, `test_bold_accent_renders_styled_not_bare`, is in untouched theme code and passed in CI, so it is not attributed to this change.
- Real-cloud file: 9 setup errors because `MEM0_API_KEY`, `ZEP_API_KEY`, and `MEMOS_API_KEY` are unavailable; nothing was skipped.
- Source-language and large-file checks passed when their scripts were invoked directly (`make` is unavailable in this environment); `git diff --check` passed.

raven/memory_engine/http_backend.py
gloryfromca: Fixed in 9376e6229. recall and recall_session now put the owner of the call on the wire; store uses metadata["user_id"] / ["userId"] (a sub-agent's scope) and falls back to ServiceLocator.user_id; the hook signatures carry the owner explicitly. Zep creates the user and thread for whichever owner is writing. Covered by test_recall_uses_the_owner_of_the_call, test_store_files_under_the_owner_named_in_metadata and test_recall_session_uses_the_owner_of_the_call in tests/_hosted_memory_cases.py (all three services), plus test_a_write_under_another_owner_creates_that_user_and_thread for Zep. Pinning recall back to the host owner turns the first one red.

plugins-dist/zep-memory/raven_zep/backend.py
gloryfromca: Fixed in 9376e6229. ZepBackend._store_calls splits a store into requests of at most 30 messages, and a message over 4,096 characters into consecutive pieces of the same role, so the importer's 100-message batches land. The fake now returns 400 for either limit, and test_the_fake_refuses_what_zep_refuses pins that so test_store_splits_at_zep_request_limits can actually fail (it does, when the split is removed).

raven/memory_engine/__init__.py
gloryfromca: Fixed in 9376e6229. The face exports the helper surface the plugins use (clamp_score, RECALL_TIMEOUT_S, STORE_TIMEOUT_S alongside Call, Reply, HttpMemoryBackend, ApiKeyOnboardStep); all three plugins import from raven.memory_engine only. test_memory_engine_face.py now scans plugins-dist/ as well as raven/ (test_the_scan_covers_the_shipped_plugins pins the coverage) -- which also turned up the design engine's two direct imports of skill_local.registry; both routed through the face in the same commit.

raven/memory_engine/api_key_onboard.py
gloryfromca: Fixed in 9376e6229. raven/memory_engine/http_backend.py and api_key_onboard.py are added to the allowlist in tests/test_provider_auth_method.py with the argument beside them: a hosted memory service's own credential, read to put it on that service's Authorization header, to refuse to send without one, and to probe/record it during onboarding -- never a verdict on whether a Raven provider is set up. The test is green again; removing the two entries turns it red.

raven/memory_engine/http_backend.py
gloryfromca: Blocking: correct the four remaining malformed attribution trailers.

Verified in `9376e6229`. The per-call user now reaches recall and session request construction, store metadata selects the write owner with the host identity as fallback, and Zep creates owner-specific users and threads. The shared owner tests pass for all three services, including the focused alternate-owner store and recall cases. This finding is settled.

plugins-dist/zep-memory/raven_zep/backend.py
gloryfromca: Blocking: correct the four remaining malformed attribution trailers.

Verified in `9376e6229`. Zep now slices content to 4,096 characters, groups at most 30 resulting messages per request, and requires every request to be accepted. The fake independently rejects either limit, and the split tests pass. This finding is settled.

raven/memory_engine/__init__.py
gloryfromca: Blocking: correct the four remaining malformed attribution trailers.

Verified in `9376e6229`. The complete helper surface used by the hosted plugins is exported through `raven.memory_engine`, all hosted-plugin imports now use that face, and the guard scans shipped plugin sources and catches the pre-existing design-engine submodule imports that this commit also reroutes. This finding is settled.

raven/memory_engine/api_key_onboard.py
gloryfromca: Blocking: correct the four remaining malformed attribution trailers.

Verified in `9376e6229`. Both hosted-service credential readers are now explicitly justified in the invariant ledger, and `test_only_the_auth_module_decides_configuredness_from_a_key` passes without suppressing or narrowing its scan. This finding is settled.

gloryfromca (commented):
Blocking: correct the four older malformed Claude attribution trailers.

The four code findings are fixed in `9376e6229`: per-call owners now control the hosted requests, Zep writes respect both vendor limits, shipped plugins use the complete Memory Engine face, and the credential readers are documented in the provider-auth invariant. Each corresponding thread has been replied to and resolved. No new code finding survived refutation at the fifth-round blocker bar.

The remaining blocker is unchanged from the earlier review. Commits `fe787cece167`, `206fb8490e6b`, `fc4fa29a5f1a`, and `471d9941bc63` still contain `Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>` instead of AGENTS.md's required `Co-authored-by: Claude (<actual-model-id>) <noreply@anthropic.com>` form. The new fix commit uses the correct form, but it does not repair those four existing messages.

Coverage this round: the full target diff and `471d9941bc63..9376e622992e` delta; AGENTS.md/CLAUDE.md and the relevant CONTEXT memory/plugin rules; MemoryBackend, importer, store-pipeline, sub-agent, onboarding, and plugin-face callers; branch history and backward compatibility; architecture boundaries; and whether tests were weakened. The new tests strengthen the owner, vendor-limit, face-scan, and credential-ledger checks; no existing correctness assertion was relaxed.

Verification:
- Focused hosted-memory, onboarding, sub-agent, provider-auth, and boundary suite: 637 passed, with one asyncio cleanup warning.
- GitHub CI: all four Python unit shards, Python lint, pre-commit diff, commit-message check, kernel wheel smoke, and the other completed gates passed; coverage was still running when checked.
- Source-language, large-file, and `git diff --check` checks passed.
- Real-cloud tests were not rerun because the three vendor credentials are unavailable.

gloryfromca (commented):
Blocking: rebase the hosted-memory commits onto current main and correct the four malformed Claude attribution trailers.

The new `b28cd00df589` code is sound: MemOS now sends the documented `memory_limit_number`, the 8s client timeout remains below the 10s host ceiling, and the revised test reads both values so it still enforces that ordering rather than merely following one changed literal. No new runtime finding survived refutation.

The retarget to `main` does introduce a merge blocker. This branch still carries the 38 pre-squash memory-seams commits even though their result is already on `main` as `e70a868e9`. GitHub reports the PR as conflicting, and `git merge-tree --write-tree HEAD github/main` produces 24 conflict paths, including config, onboarding, contracts, EverOS, RPC, and their tests. At this sixth review round this meets all three blocker conditions: the retarget without replaying only the hosted-memory commits introduced it; every ordinary merge attempt reaches it; and there is no working merge path until the branch is rebased/reconstructed on current `main` or those conflicts are resolved.

The prior history blocker also remains: `fe787cece167`, `206fb8490e6b`, `fc4fa29a5f1a`, and `471d9941bc63` still use `Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>` instead of AGENTS.md's required `Co-authored-by: Claude (<actual-model-id>) <noreply@anthropic.com>` form. Replaying the hosted commits onto `main` is also an opportunity to correct those messages.

Nonblocking follow-up while resolving the branch: `docs/memory-plugin-architecture.md` still says the hosted client gives up before the host's 5s budget; this revision changes that host budget to 10s. Keep the architecture text synchronized when the conflict in that documentation is resolved.

Coverage: AGENTS.md/CLAUDE.md and relevant CONTEXT rules; the full `github/main...HEAD` diff and 43-commit history; the focused `9376e622992e..b28cd00df589` delta; recall callers and timeout layering; backward compatibility; architecture boundaries; test changes and whether assertions were weakened; and the main/head merge graph. The older seam stack was identified as already-squashed target history rather than treated as 38 new feature commits.

Verification:
- Focused memory, onboarding, sub-agent, provider-auth, boundary, and segment suite: 681 passed.
- `git merge-tree --write-tree HEAD github/main`: failed with 24 conflicts.
- Source-language, large-file, and `git diff --check` checks passed.
- GitHub has no check rollup for this conflicting head yet.
- Real-cloud tests were not run because the vendor credentials are unavailable.

MemOS field reference: https://memos-docs.openmem.net/memos_cloud/mem_operations/search_memory/

gloryfromca (commented):
Blocking: rebase the hosted-memory commits onto current main and correct the four malformed Claude attribution trailers.

This revision does not clear the two standing blockers. After refreshing `github/main` to `de3870a5b`, GitHub still reports the PR as conflicting and `git merge-tree --write-tree HEAD github/main` still produces 24 conflict paths. The four hosted-memory commits identified in the prior review still carry the malformed attribution form. The new `c26f0bd23581` commit changes neither issue.

The new benchmark-query extraction otherwise does what its focused test claims: the advertised envelope is reduced to its problem statement, while a message without that field is recalled whole. No new blocker survived the seventh-round bar.

Nonblocking follow-up: `_problem_statement` treats any occurrence of the literal `problem_statement:` as a benchmark envelope. An ordinary coding prompt such as `Please rename problem_statement: to issue_text` therefore recalls only `to issue_text`, discarding the leading intent. Gate this on the complete harness envelope or pass the benchmark query as structured data when the branch is rebuilt. The new branch-specific source comment should also be rewritten as stable product rationale; AGENTS.md section 1.1 explicitly rejects transient branch context.

Coverage this round: AGENTS.md and the Runtime context vocabulary; the full current `github/main...HEAD` diff and history; the focused `b28cd00df589..c26f0bd23581` delta; callers that populate `AssemblyContext.current_message`; backward compatibility for ordinary messages; the memory/plugin architecture boundaries; test assertions and whether they were weakened; and the refreshed main/head merge graph. The benchmark test strengthens coverage and no existing assertion was relaxed.

Verification:
- Focused memory, onboarding, sub-agent, provider-auth, boundary, and segment suite: 683 passed.
- The first launch did not collect because an ignored stale `plugins-dist/cloud-memory` cache directory matched the workspace glob; moving that cache aside temporarily and restoring it allowed the unchanged suite to run.
- `git merge-tree --write-tree HEAD github/main`: failed with 24 conflicts.
- Source-language, large-file, and `git diff --check` checks passed.
- Real-cloud tests were not run because vendor credentials are unavailable.

gloryfromca (commented):
Blocking: preserve MemOS preference recall, rebase onto current main, and correct the malformed Claude attribution trailers.

One new blocker is marked inline. The MemOS skill request and `skill_value` parsing match the official schema, but selecting only `detail_factual` and `skill` removes the preference category that MemOS otherwise returns by default. The fake does not enforce the request's view allow-list, so the revised test passes while asserting a response the real service will not send.

This satisfies the eighth-round blocker bar: `2816140a1681` introduces the category loss; ordinary MemOS use reaches it whenever a relevant stored preference should guide a response; and the plugin exposes no configuration or alternate recall path that lets an operator restore preferences.

The two standing blockers are unchanged by this commit: after refreshing `github/main` to `de3870a5b`, the dry merge still has 24 conflict paths and GitHub reports the PR as conflicting; the four hosted-memory commits previously identified still use the malformed attribution trailers. The earlier parser follow-up also remains nonblocking and is not repeated here.

Coverage this round: AGENTS.md and Runtime context rules; the full current target diff and branch history; the focused `c26f0bd23581..2816140a1681` delta; MemoryBackend callers; MemOS's official request and response contracts; backward compatibility for facts and preferences; test changes and fake fidelity; architecture boundaries; and the refreshed main/head merge graph.

Verification:
- Focused memory, onboarding, sub-agent, provider-auth, boundary, and segment suite: 685 passed.
- `git merge-tree --write-tree HEAD github/main`: failed with 24 conflicts.
- Source-language, large-file, and `git diff --check` checks passed.
- Real-cloud tests were not run because vendor credentials are unavailable.

MemOS contract references: [search category behavior](https://memos-docs.openmem.net/memos_cloud/mem_operations/search_memory/) and [category names/defaults](https://memos-docs.openmem.net/memos_cloud/introduction/memory_types/).

plugins-dist/memos-memory/raven_memos/backend.py
```diff
@@ -0,0 +1,197 @@
...
+from typing import Any
+
+from raven.contracts.memory import Memory
+from raven.memory_engine import Call, HttpMemoryBackend, Reply, clamp_score
+from raven.plugins import PluginContext
+
+RATE_LIMITED_CODE = 40309
+#: Views a recall draws from. ``preference`` is left out: it holds standing
+#: likes and dislikes about a person, which a coding task has no use for.
+SEARCH_VIEWS = ("detail_factual", "skill")
```
gloryfromca: Blocking: preserve MemOS preference recall when opting into skills.

`include_memory_view` is an allow-list. MemOS documents that omitting it recalls facts plus preferences, while supplying it returns only the named categories. This tuple therefore makes every real request drop `preference_detail_list`, even though `_parse_hits` and the revised test still expect preferences. The fake hides the regression because it emits preferences unconditionally instead of honoring the request.

This commit introduces the loss on every MemOS recall, stored response/style preferences are ordinary backend data, and there is no operator setting or alternate path to retrieve them. Include `"preference"` alongside facts and skills and make the fake filter by the requested views (or provide an equivalently general policy rather than hard-coding the coding-benchmark subset).

Official behavior: https://memos-docs.openmem.net/memos_cloud/mem_operations/search_memory/

gloryfromca (commented):
Blocking: preserve MemOS preference recall, rebase onto current main, and correct the malformed Claude attribution trailers.

The open MemOS finding is not fixed by `4d74360bcba3`. `USER_VIEWS = ("detail_factual",)` still omits `preference`, while MemOS documents `include_memory_view` as an allow-list. The fake still emits `preference_detail_list` unconditionally, so `test_hits_keep_the_views_asked_for_and_drop_the_rest` continues to assert a response the real service will not send. That thread remains open.

The new agent-track hook is a nonblocking follow-up, not a ninth-round blocker: MemOS officially supports writing conversations with `agent_id` and searching independent Agent memory by `agent_id`, but this implementation never sends `agent_id` on store and translates Raven's configured agent id into MemOS `user_id` on recall. With distinct `memory.userId` / `memory.agentId`, the skill source therefore searches a different user bucket than normal turns write. Keeping the two IDs equal is the existing default and a working escape, so this does not meet the no-way-out condition; if the agent path remains, align its write/read identity with MemOS's actual `agent_id` contract or document and enforce the same-id view-partition model.

The standing merge/history blockers are unchanged: against refreshed `github/main` at `de3870a5b`, the dry merge still has 24 conflict paths and GitHub reports `CONFLICTING`; the four hosted-memory commits previously identified still carry the malformed attribution trailers.

Coverage this round: AGENTS.md and the Runtime/MemoryBackend contract; the full current target diff and branch history; the focused `2816140a1681..4d74360bcba3` delta; user- and agent-track callers, write/read ownership, backward compatibility, MemOS's official identity/view contracts, whether shared tests were weakened or skipped, architecture boundaries, and the refreshed merge graph.

Verification:
- Focused memory, skill-router, context-factory, onboarding, sub-agent, provider-auth, boundary, and segment suite: 732 passed, 1 skipped. The skip is the shared "agent track sends no request" case for MemOS, replaced by its service-specific agent-track cases.
- `git merge-tree --write-tree HEAD github/main`: failed with 24 conflicts.
- Source-language, large-file, and `git diff --check` checks passed.
- Real-cloud tests were not run because vendor credentials are unavailable.

MemOS references: [search/view semantics](https://memos-docs.openmem.net/memos_cloud/mem_operations/search_memory/), [add-message agent identity](https://memos-docs.openmem.net/memos_cloud/mem_operations/add_message/), and [multi-agent isolation](https://memos-docs.openmem.net/memos_cloud/introduction/isolation_filters/).

gloryfromca (commented):
Blocking: rebase the branch onto current main, preserve MemOS preference recall, and correct the malformed attribution trailers.

The `3921fba5f259` delta changes only the raven-code launcher and its tests. The new defaults do make backend agent-track hits visible through push discovery while preserving explicit discovery/local-weight choices; I found no additional functional blocker in that delta.

No prior blocker was addressed here: the open MemOS preference thread still applies, the refreshed `github/main` merge remains dirty with 25 conflict paths (merge base `09a5f7f8`; GitHub reports `CONFLICTING/DIRTY`), and `fe787cece`, `206fb8490`, `fc4fa29a5`, and `471d9941b` still carry the malformed attribution trailer form.

Nonblocking cleanup while resolving those items: the new source comment records a dated canary result and a branch-local "verdict D5"; AGENTS.md section 1.1 excludes transient task context from source comments, so retain the durable pull/push constraint and remove the transient provenance.

Verification: `uv run --frozen --python 3.12 --extra dev pytest` over the launcher plus hosted-memory, architecture, onboarding, sub-agent, segment, and context-engine suites: 805 passed, 1 skipped (the shared no-agent-track case is intentionally replaced for MemOS). `check_source_language.py`, `check_large_files.py`, and `git diff --check` all passed.

Covered this round: AGENTS.md and domain/architecture constraints, the head delta and full PR diff, relevant context-engine callers, target/history topology, backward-compatible explicit config behavior, and test changes (no weakening found).

gloryfromca (commented):
Blocking: settle the existing MemOS preference thread and rebuild the branch cleanly on main with compliant attribution trailers.

The `ea4716f768c6` delta is correctly wired: `skillForge.router.topK` controls the no-gate push injection count, the default becomes 20 only when a memory backend is configured, and an explicit value still wins. I found no new blocking defect in this delta.

Nonblocking follow-up: phase-A skill bodies have no aggregate prompt-budget clamp, so injecting 20 full backend hits can consume a large share of the model context. This does not meet the late-round blocker bar because an operator can set a smaller `skillForge.router.topK`, but the product default should eventually be bounded by tokens rather than rows.

The existing record is otherwise unchanged: the MemOS preference finding remains open, the refreshed `github/main` merge still produces 25 conflict paths and GitHub reports `CONFLICTING/DIRTY`, and the four previously named hosted-memory commits still use the malformed attribution form. This delta addresses none of those items.

Verification: the focused launcher, hosted-memory, architecture, onboarding, sub-agent, segment, and context-engine suites passed with 805 passed and 1 skipped (the intentional MemOS agent-track replacement). `check_source_language.py`, `check_large_files.py`, and `git diff --check` also passed.

Covered this round: AGENTS.md and domain/architecture constraints, the head delta and full PR state, runtime callers and prompt-budget behavior, target/history topology, explicit-config backward compatibility, and test changes (no weakening found).

gloryfromca (commented):
Blocking: settle the open MemOS preference defect and rebuild this branch cleanly on current main with compliant attribution trailers.

The two commits since the last assessed revision restore the EverOS case approach and isolate request-level HTTP failures while scaling import write timeouts. Both behaviors are correctly wired, preserve existing fallback/config behavior, and are already represented by equivalent fixes on `main`; I found no new blocking defect in their implementation or tests.

They do not settle the standing review record. The MemOS preference thread remains open, and after refreshing `github/main` the histories are now 51/50 commits apart with 24 conflict paths; GitHub reports `CONFLICTING/DIRTY`. Replaying already-upstream fixes on this branch increases that divergence rather than resolving it.

Verification: the EverOS backend/plugin/admission suites plus hosted-memory and architecture regressions passed with 508 passed and 1 skipped (the intentional MemOS agent-track replacement). `check_source_language.py`, `check_large_files.py`, and `git diff --check` passed.

Covered this round: AGENTS.md and architecture constraints, both head commits and the full PR state, affected state-machine/import callers, backward compatibility, target/history topology, and test changes (no weakening found).

raven/memory_engine/http_backend.py
```diff
@@ -0,0 +1,398 @@
...
+            # in whichever arm ran first, and nothing takes it back: there
+            # is no per-run delete, and a later run cannot tell an ingested
+            # row from one its predecessor left behind. Recall is untouched.
+            #
+            # The refusal sits here rather than on the turn enqueue so the
+            # write pipeline keeps working as it does in production (queue
+            # ceiling, retries, drain accounting, health alarm). What
+            # changes is the answer at the wire, which every caller of
+            # store() already handles as a write that did not land.
+            return False
```
gloryfromca, line 264: Blocking: make read-only stores settle as intentionally handled instead of reporting a failed delivery.

`StorePipeline._store_once` treats only an explicit `False` as failure. With this return, every completed turn in read-only mode enters the 2/5/15/30-second retry ladder; a focused pipeline reproduction observed attempts `[0, 1, 2, 3, 4]`, one failure callback, no success callback, and `dropped == 1`. After three turns the loop also emits `memory.health` false, and CLI teardown says the memory service was unavailable.

This change introduces the behavior, every completed benchmark turn takes this ordinary path, and there is no operator escape that preserves recall: disabling the backend removes the read side too. Keep the request off the wire but acknowledge intentional suppression as settled, or carry a distinct read-only outcome through the pipeline. Exercise it through `StorePipeline`; the current shared test calls `store()` directly and asserts `False`, so it pins the regression instead of exposing it. The equivalent EverOS branch at `raven_everos/backend.py:1204` needs the same semantics.

gloryfromca (commented):
Blocking: correct read-only delivery semantics, restore MemOS preference recall, and rebuild the branch cleanly on current main with compliant attribution trailers.

Reviewed the three-commit delta from `cd4a54fd005c` through the repository rules, the diff, callers and history, backward compatibility, test-strength changes, and the Memory Engine/plugin architecture. The new read-only failure is inline. The earlier MemOS preference finding remains open and its files are untouched. The standing attribution issue is unchanged. After refreshing `github/main` to `027637b0711d`, this branch is 51 commits behind and 53 ahead; `git merge-tree --write-tree HEAD github/main` still reports 24 conflicts, matching GitHub's `CONFLICTING` / `DIRTY` state.

Verification: the focused raven-code and memory suites passed (407 passed, 1 skipped). The full suite completed with 22,125 passed, 66 skipped, and one failure in untouched `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`; the same test failed again in isolation, so it is reported but is not attributed to this delta. `git diff --check`, source-language, and large-file checks passed.

plugins-dist/everos-memory/raven_everos/backend.py
```diff
@@ -232,6 +278,24 @@ def _headers(self) -> dict[str, str]:
             return {"Authorization": f"Bearer {self._api_key}"}
         return {}
 
+    def _scope(self, app_id: str | None = None, project_id: str | None = None) -> dict[str, str]:
+        """The ``app_id`` / ``project_id`` pair to send, caller's overriding ours.
+
+        Every route that touches stored rows takes the same pair. Sending it on
```
gloryfromca, line 284: Blocking: pass the configured EverOS scope to session retrieval.

The claim that every stored-row route takes this pair is not true for `get_session`: its `/api/v2/memory/get` body still contains only the owner, memory type, session filter, and page size. A focused reproduction with `app_id="raven"` and `project_id="pool"` captured neither field on that request. EverOS 1.2.3's `GetRequest` therefore defaults both to `"default"`, and `GetManager` compiles those defaults into the row predicate.

This revision introduces named-pool configuration, `collect_memories` ordinarily calls `recall_session` for every completed sub-agent record, and the caller has no scope parameter or alternate retrieval path. A named-pool operator consequently gets an empty audit record even though ordinary recall finds the pool. Apply `_scope()` to the `/get` body and include that route in the wire-shape test; the new “every route” test currently checks only search, add, and flush.

gloryfromca (commented):
Blocking: scope EverOS session reads, settle read-only writes, restore MemOS preferences, and rebuild the branch cleanly on current main with compliant attribution trailers.

Reviewed the `f6a840b9a1a7..c5afe29036d0` delta through the repository rules, diff, callers and history, backward compatibility, test-strength changes, and the Memory Engine/plugin architecture. The new namespaced-session defect is inline. The open read-only and MemOS preference findings are untouched and remain standing; the attribution issue is also unchanged.

After refreshing `github/main` to `45203a4a624e`, the branch is 53 commits behind and 54 ahead. `git merge-tree --write-tree HEAD github/main` reports 25 conflicts, matching GitHub's `CONFLICTING` / `DIRTY` state. Focused EverOS and sub-agent memory suites pass (211 passed); `git diff --check`, source-language, and large-file checks pass.

Nonblocking follow-up: declare `app_id` and `project_id` in the EverOS manifest's `config_schema`; otherwise correct named-pool configuration is admitted only with an unknown-key warning and receives no type validation.

plugins-dist/everos-memory/raven_everos/backend.py
```diff
@@ -131,15 +142,52 @@ def _jsonify(obj: Any) -> Any:
...
+# track on a hosted pool is a different operation: it fuses agent_case and
+# agent_skill through a cross-encoder, and the skill router over-fetches 2x, so
+# one turn asks for 40 rows. Measured against the hosted service on six unseen
+# queries: 10.2 / 15.0 / 19.9 / 20.8 / 26.0 / 33.9 s. Every one of those was a
+# timeout, and a timeout here is silent -- recall returns empty, the # Skills
+# segment renders nothing, and the arm scores as if it had no memory at all.
+# 45 clears the measured worst case with headroom. It is deliberately not the
+# store budget: a read that overruns costs the turn its memory, and that is the
+# cost being bought back here.
+_RECALL_TIMEOUT_S: float = 45.0
```
gloryfromca, line 155: Blocking: keep EverOS recall inside Raven's caller budgets instead of raising the shared timeout to 45 seconds.

This constant controls both tracks. User recall is wrapped by `MemorySegmentBuilder`'s 10-second `wait_for`; with 45 seconds here, the host cancels first, so `EverosBackend.recall` never reaches its timeout handler or demotes the service and later turns can pay the same 10-second delay again. Agent recall is awaited by `BackendSkillSource` and `SkillsSegmentBuilder` without an enclosing budget, so the measured 10.2-33.9-second calls now hold phase-A context assembly before the model call.

The repository already pins the required ordering, and the focused suite now fails `test_the_plugin_budget_is_strictly_inside_the_framework_one` with `45.0 < 10.0` (218 other tests pass). This commit introduces the inversion, both ordinary recall paths use it, and there is no per-track timeout setting or bounded fallback; reducing `top_k` does not cap a stalled request, while disabling the backend removes the feature. Keep the plugin timeout below the framework ceiling and address slow agent-track retrieval separately, for example with smaller/fanned-out requests or a distinct bounded agent policy.

gloryfromca (commented):
Blocking: restore the EverOS recall budget invariant, settle the three open memory defects, and rebuild the branch cleanly on current main with compliant attribution trailers.

The `c5afe29036d0..2d27c2ed8883` delta only raises EverOS's shared recall timeout from 4 to 45 seconds. The resulting budget regression is marked inline. The open MemOS preference, read-only delivery, and namespaced EverOS session findings are untouched and remain standing; the previously reported attribution issue is also unchanged.

After refreshing `github/main` to `c2c5f279e600`, the branch is 54 commits behind and 55 ahead. `git merge-tree --write-tree HEAD github/main` reports 25 conflicts, matching GitHub's `CONFLICTING` / `DIRTY` state.

Verification: `uv run --frozen --python 3.12 --extra dev pytest tests/test_segments.py tests/test_everos_backend.py -q` produced 1 failed, 218 passed; the failure is the existing plugin-inside-framework timeout invariant (`45.0 < 10.0`). `git diff --check`, source-language, and large-file checks passed.

Covered this round: AGENTS.md/CLAUDE.md and the relevant context rules, the head delta and full PR state, user- and agent-track callers, cancellation/state behavior, backward compatibility, tests and whether they were weakened, the Memory Engine/plugin architecture, commit history, and the refreshed target merge graph.

PR #436, opened 2026-09-16
fix(permissions): an approval prompt reads as an action, not as a payload

### LINKED ISSUE #435: fix(permissions): the approval dialog shows a tool's raw payload instead of the action
## Summary

The approval dialog for every tool except `exec` shows a truncated JSON dump of
the call's parameters. For `spawn` that means the whole prompt is pasted into
the prompt asking whether to allow it — cut off mid-sentence at 200 characters,
and on the page it appears twice, once in the title line and again in the body
below it.

A person looking at that cannot answer the question the dialog is asking. What
they need to decide is *which sub-agent is being started, to do what, with what
reach*. What they get is the machine payload that would have been sent.

## Where it comes from

`raven/permissions/gate.py:117` picks between two renderings:

```python
description = described.description if described else f"Approve this action: {action_line(tool_name, params)}"
```

The first branch is the human-readable one, and exactly one ruling ever produces
it: the shell-command safety classifier, which carries an
`_APPROVAL_DESCRIPTIONS` sentence per reason code
(`raven/permissions/builtin.py:216-222`). Everything else falls through to
`action_line`, whose own docstring states the intent it does not meet for those
callers (`raven/permissions/builtin.py:136-145`):

```python
def action_line(tool_name: str, params: dict[str, Any]) -> str:
    """The action as a human should read it in an approval prompt."""
    if tool_name == "exec" and isinstance(params.get("command"), str):
        ...
        return params["command"]
    compact = json.dumps(params, sort_keys=True, ensure_ascii=False, default=str)
    if len(compact) > 200:
        compact = compact[:200] + "..."
    return f"{tool_name} {compact}"
```

So `exec` got a considered rendering and the fallback was never revisited as
other tools started reaching the gate. `spawn` is the sharpest case because
`prompt_template` is one of its required fields
(`raven/agent/subagent/dag_tool.py:323`), so the prompt *is* the payload.

## Why it is worth fixing rather than truncating harder

Three separate problems, and shortening the string only touches the first:

1. **It does not support the decision.** 200 characters of a prompt, cut
   mid-word, tells a reader neither which agent will run nor what it may touch.
   A dialog that cannot be acted on gets approved out of habit, which is the
   failure mode a permission gate exists to prevent.
2. **It puts the content on screen twice**, in a dialog that is often the one
   thing visible during a demo or a screen share. The prompt may carry material
   the author would not have chosen to display; the gate did not ask.
3. **It leaks the shape of the call** into a surface meant for people. The
   parameter names, the JSON, the truncation marker — that is a log line, not a
   question.

## Suggested fix

The mechanism already exists; it just has one producer. Let a tool describe its
own action for the gate, the way the shell classifier does, and keep the JSON
dump strictly as a last resort for a tool that offers nothing.

For `spawn` a description would name the agent and the job rather than quoting
the brief — for example "Start sub-agent *X* on a background task", with the
node id, and, if the gate wants to show more, the brief behind a disclosure
rather than in the question itself.

Two smaller things worth doing whichever way the above goes:

- Do not render the same string in both the title and the body; one of the two
  is enough.
- If the payload must be shown at all, show it on the detail surface rather than
  in the sentence being answered.

## Reproducing

Configure `spawn` into the ask tier and let an agent dispatch one. The approval
prompt reads `Approve this action: spawn {"node_id": ..., "prompt_template":
"<the whole brief, cut at 200 chars>...`, and the same text appears again
underneath.

Handsome-wzw:
## Summary

The prompt asking whether to allow a call was the call's own arguments,
serialised and cut at a fixed width. For a tool whose brief is one of its
arguments -- a sub-agent dispatch is the plain case, since `prompt_template` is a
required field -- the question "may I send this?" was mostly the thing being
sent, truncated mid-sentence, with the fields that actually decide it past the
cut.

Before and after, on the same call:

```
spawn {"agent": "Raven-PPT", "node_id": "intro-deck", "prompt_template": "Make a deck introducing the assis...

spawn agent='Raven-PPT' node_id='intro-deck' prompt_template=<1560 chars>
```

Three things go wrong at once in the first, and shortening the string fixes only
the first of them:

- **It does not support the decision.** A reader cannot tell which agent will run
  or what it may touch, because those fields are past the cut. A dialog that
  cannot be acted on gets approved out of habit, which is the failure a
  permission gate exists to prevent.
- **It puts the material on screen** in the one window most likely to be visible
  while someone is presenting or sharing, and the gate never asked whether it
  should be.
- **It reads as a log line**, not a question: parameter names, JSON punctuation,
  an ellipsis.

Arguments are now summarised. A value short enough to read is shown; anything
longer is named by its size or its shape, so the reader is told a brief is there
and how big it is and the text stays out of the question. Arguments that do not
fit are counted rather than dropped, because a prompt that silently omits one is
worse than a long one.

`exec` keeps the rendering it always had. For a shell call the command *is* the
action, and there is nothing to summarise -- that special case was the only
considered rendering on this path, and the fallback beside it was never revisited
as other tools started reaching the gate.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/test_permissions_gate.py                     213 pass
pytest -k "permission or approval or gate or spawn"      1666 pass
ruff check + ruff format --check                          both pass
scripts/check_commit_messages.py upstream/main..HEAD      exit 0
scripts/check_source_language.py upstream/main..HEAD      exit 0
```

Four new cases, each mutation-checked. Four mutations, four caught:

- serialise the arguments again (the shape this replaces) -> 3 red
- inline a long string instead of naming its size -> red
- drop the arguments that do not fit instead of counting them -> red
- send `exec` through the general summary -> red

The last one is there because `exec`'s rendering is the one thing on this path
that was already right, and a rewrite of the fallback beside it is exactly the
kind of change that takes a neighbour with it.

## Risk

Security: this is the text a person reads before granting a permission, so the
risk runs the other way from usual -- the old rendering was the one that could
mislead, by pushing the deciding fields past a truncation. Nothing about what is
allowed changes; `action_digest`, which keys the remembered decisions, is
untouched, so an approval granted before still matches the same call after.

Backward compatibility: one function's output, consumed by the approval prompt.
No wire shape and no stored value moves.

Not addressed here, and worth a separate look: the sheet renders this string in
both its title and its body, so the reader sees it twice. That is a page-side
decision about which of the two should carry the detail, and it is a smaller
problem now that the string is a summary.

Rollback: one commit, reverts cleanly.

## Related Issues

Fixes #435


raven/permissions/builtin.py
```diff
@@ -133,16 +133,69 @@ def _bound_workdir() -> str:
...
+    or its shape. The distinction is what the reader is being asked about: an
+    id, a path or an agent's name decides the call, and a paragraph of prose is
+    the material the call carries rather than the call itself.
+    """
+    if isinstance(value, str):
+        if len(value) <= _INLINE_ARG_LIMIT:
+            return repr(value)
+        return f"<{len(value)} chars>"
+    if isinstance(value, (list, tuple)):
+        return f"<{len(value)} items>"
```
gloryfromca: Blocking: structured arguments hide the action targets an approver must judge

Collapsing every list or dict to only its count hides decision-critical action targets in ordinary ask-tier tools. For example, an unknown/MCP `delete_files` call with `paths=["/etc/passwd"]` now renders as `delete_files paths=<1 items>`, so the user cannot tell what they are authorizing; unknown tools default to the ask tier specifically because their effects are not known. The shipped `run_subagent_dag` has the same reachable regression: its entire node list (IDs, selected agents, and task summaries) renders only as `nodes=<N items>`. Those values describe the action rather than bulk prose, so the formatter needs to preserve enough identifying information from structured arguments for the prompt to remain judgeable.

gloryfromca (commented):
Blocking: structured arguments hide the action targets an approver must judge

I found one concrete regression in the new generic formatter. I covered the repository rules and domain map, the complete diff, both permission-prompt callers and UI transports, relevant tool schemas, file history, backward compatibility, and whether the tests were weakened.

Verification: `uv run pytest tests/test_permissions_gate.py -x` passed all 213 tests. `uv run pytest -x` could not collect because this worktree lacks the optional `raven_everos` module (`tests/test_cli_onboard_commands.py`); seven optional channel suites were also skipped for unavailable SDKs.

raven/permissions/builtin.py
Handsome-wzw: You are right, and it was a regression my change introduced. Fixed in `1ddf1ac5`.

Both of your examples reproduced exactly:

```
delete_files paths=<1 items>
run_subagent_dag nodes=<2 items> task_summary='weekly digest'
```

The first is indefensible on its own terms: it asks someone to authorise a deletion without telling them what is deleted.

What I got wrong was treating one rule as covering two different things. Prose -- a brief, a file body -- is material the call *carries*, and pasting it into the question pushes out everything the reader needs. A container is what the call will *act on*. Shortening both by the same rule meant the second inherited the treatment the first needed.

Containers are now opened rather than counted, as far as the line allows, with whatever did not fit counted beside what did:

```
delete_files paths=['/etc/passwd']
run_subagent_dag nodes=[{id='pull', subagent='data-raven', node_summary='pull the feedback', prompt_template=<800 chars>}, +1 more] task_summary='weekly digest'
```

The step ids, the chosen agents and what each step is for all survive; the 800-character prompt inside the same node is still named rather than quoted. Keeping the leaf rule is what makes opening the container safe.

Your point about the ask tier is the one I would have missed on my own: an unknown tool is there *because* its effects cannot be inferred from its name, so it is exactly the call where hiding the arguments costs most. That is now the reason written on the test rather than my own framing.

**A second defect, found by one of the new tests rather than by reading.** The budget handed to an argument did not exclude the argument's own name, so a value rendered to fit came back one key too long and the outer loop dropped it whole -- `delete_files` with 40 paths rendered as `delete_files (+1 more)`, which says nothing at all. Fixed, and the first argument is now never dropped entirely.

Four mutations, four caught: collapse containers to a count again (3 red -- your regression), inline a long string inside a container (2 red), drop an argument that does not fit instead of shortening it (red), and stop counting the members that did not fit (red).

Gates on this head: `pytest tests/test_permissions_gate.py` 216 pass; `pytest -k "permission or approval or gate or spawn or dag"` 2349 pass; `ruff check` and `ruff format --check` both pass.

gloryfromca: Blocking: long scalar action targets are still hidden by the generic length rule

The structured-container regression is fixed on this head: the deletion target and DAG node identity now survive while nested prose stays summarized. The added wide-container case also covers the drop-whole failure you found. I am closing this thread; the remaining scalar-target issue is separate and is marked on the new revision.

raven/permissions/builtin.py
```diff
@@ -133,16 +133,102 @@ def _bound_workdir() -> str:
...
+    rather than counted, as far as the budget allows, and whatever did not fit is
+    counted beside what did -- a partial list plus a remainder still says what
+    kind of thing this call touches, where a bare count says nothing at all.
+
+    That the leaves stay short is what makes opening a container safe: a node
+    carrying an 800-character prompt renders as its id, its agent and a named
+    size, not as the prompt.
+    """
+    if isinstance(value, str):
+        return repr(value) if len(value) <= _INLINE_ARG_LIMIT else f"<{len(value)} chars>"
```
gloryfromca: Blocking: long scalar action targets are still hidden by the generic length rule

Long scalar action targets are replaced by a character count, so ordinary approvals can still omit what the call acts on. For example, a valid `write_file` call with a 104-character nested workspace path renders as `write_file content='replacement' path=<104 chars>`; the approver cannot identify which file will be overwritten. The `spawn` schema explicitly allows an unbounded `node_id`, and an 84-character ID likewise renders only as `<84 chars>`. These are decision-bearing identifiers, not payload prose. The scalar summary therefore needs to retain an identifying representation for long targets (including those nested in containers) instead of classifying every long string as material to hide.

gloryfromca (commented):
Blocking: long scalar action targets are still hidden by the generic length rule

The container fix resolves my prior finding, and I closed that thread. One related plain error remains on this revision: length alone cannot distinguish prose from scalar identifiers such as paths and node IDs.

I covered the new commit and full diff, repository rules and domain terminology, both prompt callers and transports, affected built-in and unknown-tool call shapes, history, backward compatibility, and whether tests were weakened.

Verification: `uv run pytest tests/test_permissions_gate.py -x` passed all 216 tests. `uv run pytest -k "permission or approval or gate or spawn or dag" -x` could not collect because this worktree lacks the optional `raven_everos` module; seven optional channel suites were also skipped for unavailable SDKs.

raven/permissions/builtin.py
Handsome-wzw: Right again, and both of your examples reproduced. Fixed in `8f223d5a`.

Long scalars are now cut from the middle rather than replaced by a count:

```
write_file content='replacement' path='/Users/admin/wor...nested/report.md' (91 chars)
spawn agent='Raven' node_id='node-xxxxxxxxxxx...xxxxxxxxxxxxxxxx' (84 chars) prompt_template='Make a deck. Mak...k. Make a deck. ' (1560 chars)
```

Both ends are what identify a value -- the root and the filename of a path, the prefix and the suffix of an id -- and the count stays beside it, since that is the part an elision hides.

**The trade, stated rather than buried:** a few dozen characters of a brief now reach the prompt where none did a revision ago. I think that is the right side of the line: it is a fingerprint rather than the brief, it is bounded so it cannot crowd out the fields that decide the call, and it buys back every long identifier -- including paths with spaces in them, which is the case a shape test would have got wrong. If you would rather it were narrower, the window is one constant.

**Two defects your finding uncovered downstream, both found by running the new cases rather than by reading.**

Elided members are much longer than `<800 chars>` was, so a container whose members no longer fit fell straight back to `nodes=<2 items>` -- the exact shape your previous finding removed, reintroduced by the fix for this one. `_container_summary` now cuts its first member to the room available and counts the rest, so a bare count is no longer reachable at all. There is a test asserting `"items>" not in line` for precisely that.

And a non-string scalar was being routed through the string path and quoted, so `append=False` rendered as `append='False'` -- a flag dressed as a string that happens to spell "False".

Four mutations, four caught: name a long scalar by its size again (4 red -- your finding), keep only the head so the filename is lost (2 red), let a container fall back to a bare count (2 red), quote a boolean (2 red).

Gates on this head: `pytest tests/test_permissions_gate.py` 220 pass; `pytest -k "permission or approval or gate or spawn or dag"` 2353 pass; `ruff check` and `ruff format --check` both pass.

One note on my own process, since it bit me twice in this exchange: my first measurements of the elision width were read from a stale bytecode cache and showed no change across three different values. I cleared it and re-measured before settling on the current one.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified on this head: long paths and IDs retain identifying prefixes and suffixes at the top level and inside containers, long prose remains bounded, and flags retain their scalar spelling. The focused and adjacent approval suites pass. This resolves my finding, so I am closing the thread.

gloryfromca (commented):

The long-scalar fix resolves the remaining approval-summary defect, and both threads I opened are now resolved. I covered the new delta and full diff, repository rules and domain terminology, callers and transports, built-in and unknown-tool shapes, history, backward compatibility, test strength, and the established runtime architecture boundaries.

Verification: `uv run pytest tests/test_permissions_gate.py -x` passed all 220 tests. `uv run pytest tests/test_permissions_gate.py tests/test_approval_broker.py tests/test_agent_loop_approval.py tests/test_acp_permissions.py tests/test_shell_approval.py -x` passed all 532 tests.

0xKT:
Not a blocker and not a finding. The new rendering is a clear improvement --
every shape I threw at it came out shorter than the old one, several by a lot.
This is one loose end in it, worth a line because the number it misses is part
of the safety argument rather than a formatting preference.

`_ACTION_LINE_LIMIT = 160` carries its own reason:

    The whole line, after which the remaining arguments are counted rather
    than listed. A question a person cannot read to the end is one they
    answer out of habit.

Measured against that number, the line goes past it at realistic sizes:

    parameters   line length
      5            74
     10           139
     17          163     <- over
     20          166     <- over

Seventeen is not a contrived number. It is the largest `properties` count I
could find among the tool schemas in this repo (raven/context_engine/curator.py).
Two other shapes go further -- a 300-character key renders at 314, and a
40-node container at 164 -- but I could not show either of those is real, so
the 17-parameter case is the one I would act on and the others are just where
the arithmetic leads.

The cause is in the loop rather than in the constant. `room` is decremented per
argument, but the budget handed down is `max(room - len(key) - 1,
_INLINE_ARG_LIMIT)`, so once `room` is spent the floor of 60 puts it back, and
the `if shown and ...` break cannot fire on the first argument by design. Both
choices are right on their own -- a delete whose paths did not fit would say
nothing about what it deletes, which the comment says in as many words -- they
just leave the total unbounded by the constant that claims to bound it.

Scale, so this is not read as worse than it is: at 17 parameters the overshoot
is 3 characters, and the same input through the old implementation was 213.
Nothing regressed. What is missing is that the promise and the behaviour agree.

Nothing in tests/test_permissions_gate.py mentions the limit -- I grepped for
both the constant and the number -- so the 220 tests passing says nothing about
it either way. One assertion over the shapes above would pin it, whichever way
you settle it: hold the line at 160, or let the constant say what it really is
(a per-argument budget rather than a whole-line cap) so the next reader is not
measuring against a number that was never enforced.

I am filing this as a note rather than a finding because the behaviour you
shipped is better than what it replaced on every input I measured, and a
3-character overshoot on the widest real tool is not something to hold a merge
for.

0xKT (approved):

PR #438, opened 2026-09-16
feat(ui-web): a preview can be asked to run, once, for one file

Handsome-wzw:
## Summary

Stacked on the notice PR, which is what this gives an action to. Until that
merges, the diff here shows its two commits as well.

The viewer serves HTML script-free and now says so. What it could not do is let
the reader act on that: running the page meant saving it and opening it by hand
-- the same execution by a longer route, and with none of the isolation the
viewer keeps.

**Both halves have to grant it.** The frame's `sandbox` attribute and the
response's own `Content-Security-Policy` are separate gates, and the running
view needs both: the attribute is set in the page, the header comes from asking
the route for it. Neither an edited attribute nor a crafted request runs a page
by itself.

**What it grants is what a PDF already has**: `allow-scripts` without
`allow-same-origin`. The origin stays opaque, so the page still reaches no
cookie of this one and no socket of this one -- that protection was never coming
from refusing scripts, and the docstring on `sandbox_for` has said so about PDFs
all along.

**What is newly permitted is the page making requests of its own.** That is the
part worth a deliberate act rather than a default: material an agent wrote can
carry instructions it was given, and running it is how those instructions reach
the network. The reader asking is the consent; the notice above the button is
where they are told what running means.

**Nothing remembers the answer.** The grant is held as the path it was given for
rather than a flag, so opening another file revokes it with no effect to
remember to write -- the shape that cannot leak a grant onto a page nobody
looked at. The route reads it per request and forgets.

The tab route takes the same grant. A reader who asked once and then found the
tab still refusing would go and save the file, which is the worse path to the
same place.

Two kinds may be asked: a page and a drawing. Everything else served here is
read as text or as a picture, where running nothing costs nothing.

### Why this is worth having, concretely

The shape that prompted it is not the blank page. A deck laid out at a fixed
size and scaled to the viewport by script draws its first slide either way: the
content is there, the scale never runs, and what arrives is a picture cropped
off-centre under a page counter that was written into the markup and never
moves. It reads as a layout bug in the file. That is the common case for what an
agent produces when asked for something to look at.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
pytest tests/test_rpc_files.py                          31 pass
pytest -k "rpc_files or rpc_transport or console"      179 pass
ruff check + ruff format --check                        both pass
cd ui-web && npx vitest run                            1732 pass
cd ui-web && npx tsc --noEmit -p .                      clean
cd ui-web && node scripts/check-css.mjs                 OK (tokens only)
cd ui-web && python3 build.py && node scripts/check-page.mjs   OK
cd ui-tui && npm run lint:i18n                          catalogue up to date
scripts/check_commit_messages.py / check_source_language.py    exit 0
```

Four new cases across both halves, each mutation-checked. Six mutations, six
caught:

Server:
- ignore the `run` parameter on the route -> red
- let any kind be asked to run -> red
- grant `allow-same-origin` alongside the scripts -> red

Page:
- hold the grant as a flag instead of the path, so it carries to the next file -> red
- set the frame attribute without asking the route for the header -> red
- run by default instead of on the ask -> 4 red

One of those tests was wrong before it was right: switching files needs
`store.restore()` *and* `store.sync()`, and without the second the assertion was
passing against a view that had never changed file. It failed for the right
reason once corrected.

## Risk

Security: this is the change that lowers a default, so the reasoning is above
rather than here in summary. In short: the isolation that mattered -- opaque
origin, no cookie, no socket -- is unchanged and was never provided by refusing
scripts; the new exposure is outbound requests from a page the reader chose to
run, for one file, with no memory of the choice. The two-gate design means a
single edited value on either side does not reach execution.

Worth a reviewer's judgement, and the reason this is its own change rather than
riding in with the notice: whether one click is the right weight for "execute
this", or whether it should be a confirmed action. The label and the sentence
above it say what happens; I did not add a second press, because a preview that
takes two presses to look at is a preview people stop using. If you want that
heavier, it is the button and not the plumbing.

Backward compatibility: additive on both sides. A request without `run` behaves
exactly as before, which the existing sandbox tests still assert unchanged.

Rollback: one commit on top of the notice, reverts cleanly.

## Related Issues

N/A


ui-web/src/features/workspace/WorkspacePage.tsx
```diff
@@ -310,11 +310,17 @@ function BinNote({ f }: { f: WsFile }): JSX.Element {
...
    which is why there is nothing left here to navigate with. */
 export function FileView({ ws, file = ws.file }: { ws: WsShared; file?: WsFile | null }): JSX.Element {
+  /* Held here because both the bar and the body act on it, and held as the path
+     it was granted for rather than a flag: opening another file revokes it with
+     no effect to remember to write, which is the shape that cannot leak a grant
+     onto a page the reader never saw. */
+  const [ranFor, setRanFor] = useState<string | null>(null)
   if (!store.source().canBrowse) return <div className="wsnote">{t('gui.ws.file_unreadable')}</div>
   const f = file
+  const running = !!f && ranFor === f.path
```
gloryfromca: Blocking: reopening the same path must require another run request.

`ranFor` is never cleared, and `running` compares only the path. But `makeFile()` assigns a new `seq` precisely because two opens of the same path are fresh views, and the desk keeps `FileView` mounted under the same pane id. Repro: run `/repo/game.html`, open another file, then reopen `/repo/game.html` (or let it be rewritten and reopen it); `ranFor` still matches, so the new iframe immediately receives `allow-scripts` and `run=1` without another click. That violates the one-shot consent and can execute newly changed agent output. Please scope the grant to the current view identity or clear it on every view change, and cover A -> B -> A / same-path refresh.

gloryfromca (commented):
Blocking: the one-shot grant survives a fresh open of the same path.

I found one blocking lifecycle issue inline. The server-side opt-in and its default behavior are otherwise well covered.

Nonblocking follow-up: the server advertises SVG as runnable, but the product renders SVG through `` and exposes no Run control for it, so a scripted drawing cannot currently be asked to run through the UI. This is a narrow format gap, not an additional merge block.

Coverage: I read `git diff github/main...HEAD`, the workspace callers and fresh-view `seq` contract, the file-route auth and sandbox boundary, relevant history, both stacked commits, and the PR description. I checked AGENTS.md/source-language and canonical context rules, backward compatibility for requests without `run`, that tests were added rather than weakened, and the UI-to-RPC architecture boundary.

Verification: `uv run pytest tests/test_rpc_files.py -x` (31 passed); `npm test --prefix ui-web -- src/features/workspace/WorkspacePage.test.tsx` (32 passed, with non-failing happy-dom iframe connection noise); `npm run type-check --prefix ui-web` (passed); `npm run gen:check --prefix ui-web` (passed); `npm run lint:i18n --prefix ui-tui` (passed); source-language and commit-message checks (passed); `git diff --check` (passed).

ui-web/src/features/workspace/WorkspacePage.tsx
Handsome-wzw: Fixed in 3fa45443. You were right, and the tell was one line below the bug: `FileBody` is keyed by `f.seq ?? f.path` precisely because two opens of the same path are two views, and I keyed the grant by the path anyway.

The grant now holds that same identity:

```tsx
const view = f ? String(f.seq ?? f.path) : null
const running = !!view && ranFor === view
<FileBody key={view ?? undefined} f={f} running={running} onRun={() => setRanFor(view)} />
```

One expression for both, so the remount and the revoke are one event and cannot disagree about which bytes were consented to.

Two regression tests, both covering what you named:

- `asks again when the same path is opened as a new view` - A -> B -> A; the returning frame is `sandbox=""`, `src` carries no `run=1`, and the note is back.
- `asks again when the open file is rewritten in place` - A -> A with no file in between, which is the ordinary case while the agent is still writing and the one that never passes through a revoke if the grant is keyed by name.

Mutation-checked rather than just run: keying `view` back on the path fails exactly those two and nothing else; making `running` a bare flag fails those two plus the existing next-file test. Gates: ui-web 1734 passed, tsc, check:css, check:page, gen:check, ui-tui lint:i18n, ruff both halves.

On the SVG follow-up, verified in both halves and you are right about the state of it: the server lists `.svg` in `_RUNNABLE_SUFFIXES`, but `FileBody` renders `f.kind === 'svg'` through `` (`WorkspacePage.tsx:463`), and the Run control sits in the HTML branch only, so nothing in the product can ask an SVG to run. Worth adding that an ``-referenced SVG cannot run scripts whatever the CSP says, so the server rule is unreachable rather than mistaken - it is the right answer for a client that frames SVG, and no client does today. I have left it as it is rather than growing this PR into framing SVG; say the word if you would rather the server stopped advertising what the product cannot reach, and I will open that separately.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified: the grant now uses the same `seq ?? path` identity as the `FileBody` remount, so both same-path reopen and in-place rewrite return to the script-free preview. The two focused regression tests cover those paths and pass. I accept the fix and am closing this thread.

gloryfromca (commented):

The prior blocker is fixed: the grant and `FileBody` now share the same fresh-view identity, and the new A -> B -> A and in-place rewrite tests prove that replaced bytes return to the script-free state. I replied to the author and resolved my thread.

No new findings survived refutation. I reviewed the full `github/main...HEAD` diff, the fix delta, file-view and desk-pane callers/history, AGENTS.md and context terminology, backward compatibility for requests without `run`, the authentication/CSP boundary, and whether tests were weakened. The security diff scan found no reportable candidates.

Verification: `uv run pytest tests/test_rpc_files.py -x` (31 passed); `npm test --prefix ui-web -- src/features/workspace/WorkspacePage.test.tsx` (34 passed, with non-failing happy-dom iframe connection noise); UI type-check, RPC generation check, TUI i18n generation check, source-language check, commit-message check, and `git diff --check` all passed.

gloryfromca (commented):

Re-reviewed the new head after its base advanced; the effective PR range is the nine preview-related files. No new finding survived refutation, and the per-view fix still makes replacement and revocation share the same `seq ?? path` identity.

Coverage: I checked AGENTS.md/CLAUDE.md and the context terminology, the full `github/main...HEAD` diff, file-view and desk-pane callers/history, backward compatibility when `run` is absent, whether tests were weakened, and the UI-to-RPC authentication/CSP architecture boundary. The security diff scan found no reportable candidates; Daybreak access was unknown, so protected advisory results may not have been displayable.

Verification: `uv run pytest tests/test_rpc_files.py -x` (46 passed); `npm test --prefix ui-web -- --run src/features/workspace/WorkspacePage.test.tsx` (34 passed, with non-failing happy-dom iframe connection/abort noise); web type-check, RPC generation, TUI i18n generation, commit-message, source-language, and `git diff --check` checks passed. `make` is unavailable in this environment, so I invoked the source-language gate script directly through `uv run`.

gloryfromca (commented):

This revision is a clean rebase onto a `main` that now contains the earlier preview-note work. `git range-diff` shows both remaining commits are patch-equivalent to the already reviewed run-request feature and per-view lifecycle fix, and the effective nine-file diff introduces no integration regression.

Coverage: I re-read the full `github/main...HEAD` diff and revision history, checked the file-view/desk callers, AGENTS.md/CLAUDE.md and context terminology, backward compatibility without `run`, unchanged test strength, and the UI-to-RPC authentication/CSP boundary.

Verification: `uv run pytest tests/test_rpc_files.py -x` (46 passed); `npm test --prefix ui-web -- --run src/features/workspace/WorkspacePage.test.tsx` (34 passed, with non-failing happy-dom iframe connection/abort noise); web type-check, RPC generation, TUI i18n generation, source-language, commit-message, and `git diff --check` checks passed.

0xKT (approved):

0xKT:
**Not a blocker** -- a standing run grant survives the Source/Rendered toggle, so the frame can re-run bytes written after the reader consented.

The design's own promise, in this PR's comment at `WorkspacePage.tsx:313-317`:

> held as the view it was granted for rather than a flag: anything that replaces the view -- another file, or **this path opened again over rewritten bytes** -- revokes it

That second clause does not hold for one ordinary control. `f.raw = raw` is mutated in place by the Rendered/Source segmented buttons (`:390-392`) with no new `seq`, so `view = String(f.seq ?? f.path)` is unchanged and `running` stays true. Switching to Source removes the iframe from the tree; switching back creates a new one, which issues a second GET of the same `?run=1` URL. The route is stateless about `run` and sends `Cache-Control: no-store`, so that GET is answered with whatever is on disk at that moment -- still under `sandbox allow-scripts`.

Measured against the real aiohttp route on `3ecc2402`, with the control that makes it mean something:

| request | Content-Security-Policy | body |
| --- | --- | --- |
| `GET /file?path=game.html&run=1` | `sandbox allow-scripts` | `revision one` |
| file rewritten, then the **same** URL again | `sandbox allow-scripts` | the new bytes, containing `attacker.example` |
| control: same route, no `run=1` | `sandbox` | -- |

The reader is not asked again: after the toggle there is no `.vnote` and no Run button, because `running` never went false.

Why non-blocking, and I checked this specifically because it is the line between a consent gap and an isolation one: `allow-same-origin` is never granted. It appears in this file only inside comments -- every `sandbox_for` return is `"sandbox allow-scripts"` or `"sandbox"` (`files.py:248-251`) and every iframe attribute is `"allow-scripts"` or `""` (`WorkspacePage.tsx:507,513`). The origin stays opaque, so this is the page running newer bytes than the reader saw, not the page reaching anything of the host's.

New here, not pre-existing: on `4effe8ad` the signature is `def sandbox_for(path: Path) -> str` with no `run`, and `ranFor` / `runURL` do not appear at all. The toggle itself is old (`f.raw = raw` is on the base at line 381), which is why the new grant collides with it.

The same shape reaches the tab button at `:409-411`: `window.open(runURL(f.path))` bakes `run=1` into a tab URL that stays reloadable for the life of that tab, with no revoke path. Same reasoning, same severity.

Not filed and recorded so nobody re-derives them: `_RUNNABLE_SUFFIXES` includes `.svg`, but the pane renders svg through `asImage` (``, which does not execute scripts) while the Run button only exists on the `asFrame` branch, so I could not reach a running svg from the UI -- an allowlist entry I could not make live rather than a demonstrated risk. And the two-halves claim (frame attribute plus response header) I read but did not execute a browser experiment for; it is not what this finding rests on.

PR #443, opened 2026-09-16
fix(agent): a rerun keeps the turn's live state, not just its question

ZuyiZhou:
## Summary

Fixes the blocking finding on #410: the dead-end rerun discards a failed attempt's
work, and two things that are not that attempt's work were going with it. Both
reproductions were confirmed here before anything was changed.

**A correction the reader typed mid-turn was un-asked.** A message sent with
`busy=inject` is drained into the running attempt and reaches its next model call.
The rerun opened on a snapshot taken before the attempt started, so the correction
was absent from the rerun's context and from what the turn filed -- and the inject
queue does not offer a consumed message twice, so nothing else would deliver it. The
model answered the question the reader had already corrected, and the session recorded
a turn in which the correction was never sent. Drained messages now carry a mark, and
the rerun's seed is the question plus every marked message on the finished list.

Two details in that seed worth a look. It reads the whole finished list rather than
its tail, because compaction can move the tail. And it rebuilds from the untouched
pre-attempt snapshot each time rather than appending to a growing seed, so a second
rerun carries the correction exactly once.

**A cancel during the rerun filed the wrong attempt.** The rescue path was handed
`initial_messages`, which is the first attempt's list; a rerun runs on a list of its
own. So a tool the reader watched the rerun run, and its result, reached the screen
and then nothing -- the record stopped at work that had already been discarded. The
attempt now records the list it is running on and the rescue files that one, which is
the parity the first attempt always had. The generic-exception path is fixed by the
same line, though only the cancellation path was reproduced.

The mark is a message key in the loop's existing family (`_hook_injected` and its
siblings): it persists in the transcript, where it says the message arrived mid-turn,
and the provider's `_ALLOWED_MSG_KEYS` allowlist drops it on the way to the wire.

Current `main` is merged in.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Full suite on the merged head, Python 3.12, all extras:

- `uv run --frozen --python 3.12 --all-extras pytest -q`: 23,014 passed, 108 skipped.

Revert to red, each half separately, production file only:

| reverted | result |
|---|---|
| the seed stops carrying the correction | 2 failed, 34 passed |
| the rescue stops following the live attempt | 1 failed, 35 passed |

Each half reddens only its own tests, so neither test is covering the other's fix.

Mutation, on the retry module (baseline 36 passed):

| mutant | result |
|---|---|
| the seed stops carrying the correction | 2 failed |
| the drained message stops being marked | 2 failed |
| the attempt stops recording its list | 1 failed |
| CONTROL: a sibling line in the same block | 1 failed |
| CONTROL: a no-op edit in the same block | 36 passed |

Coverage, against the PR base:

- `make coverage-diff`: 95.62% (699/731 executable changed lines), threshold 90.00%.
  The 731 is the merge of `main`; all 32 uncovered lines are in `raven/rpc/*` and
  `raven/knowledge/*` and arrive with that merge. No changed line under
  `raven/agent/loop/` is uncovered.
- `make coverage-ratchet`: line 88.75% vs baseline 87.30% (+1.45pp), branch 81.71% vs
  79.43% (+2.28pp).
- `make coverage-baseline-check`: monotonic, +0.000000pp on both.

Lint and repo gates:

- `ruff check` and `ruff format --check` over `raven tests plugins-dist scripts agents`:
  clean, 1882 files.
- `lint-imports`: 10 contracts kept, 0 broken.
- `check_source_language.py`, `check_large_files.py`, `check_commit_messages.py`,
  `git diff --check`: clean.

## Risk

Both changes are reachable only through the dead-end rerun, which today only the
research product budgets; an agent that budgets no rerun runs one attempt, so the
list the rescue follows is the list it always was and no message is ever marked.

The mark adds one underscore key to a drained user message. It reaches the session
record, like `_hook_injected` does, and the provider allowlist keeps it off the wire.
A session written before this change simply has no marked messages, so replay and
history are unaffected in both directions.

Rollback is the revert of the fix commit; nothing is persisted or migrated.

## Related Issues

N/A


raven/agent/loop/turn_path.py
```diff
@@ -2582,6 +2591,19 @@ async def _attempt(seed: list[dict], attempt: int):
...
+            """The question again, plus whatever the reader said while the attempt ran.
+
+            The rerun exists to discard an attempt's research, and a correction typed
+            mid-turn is not research: the loop took it off the inject queue, which will
+            not offer it twice, so a seed rebuilt from the question alone silently
+            un-asks it. Read off the whole finished list rather than its tail, because
+            compaction may have moved the tail, and rebuilt from the untouched snapshot
+            each time, so carrying one forward twice cannot double it.
+            """
+            return [dict(m) for m in retry_seed] + [dict(m) for m in msgs if m.get(_MID_TURN_USER_KEY)]
```
gloryfromca, line 2606: No blockers; suggestions only, and they are marked inline.

Nonblocking follow-up: this can only recover marked messages that still exist in `msgs`, but `_summarize_head` delegates to `build_compacted`, which replaces the compacted head with a new unmarked summary. If a reader injects a correction early in a long attempt, that correction later falls into the summarized head, and the attempt then dead-ends, the rerun again starts without the correction. I confirmed directly that `build_compacted` drops `_mid_turn_user` in this shape. This needs opt-in head compaction plus an early correction and a later dead end, so it is too narrow to hold the present fix; keeping the consumed inputs outside the compactable transcript would close it.

raven/agent/loop/turn_path.py
```diff
@@ -2551,6 +2555,11 @@ async def _tap_episode(_index: int) -> None:
 
         async def _attempt(seed: list[dict], attempt: int):
             pending.update(rerun=False, reasons=[])
+            # The list this attempt appends to, for the rescue paths below. A rerun
+            # runs on a list of its own, and a turn cancelled during one used to be
+            # saved from the first attempt's -- so a tool the reader watched run in
+            # the rerun, and its result, were not in what the cancel filed.
+            live["messages"] = seed
```
gloryfromca, line 2562: No blockers; suggestions only, and they are marked inline.

Nonblocking follow-up: assigning `live` to the seed does not track the loop's current list after `_run_agent_loop` rebinds `messages` during proactive/reactive compaction or image demotion. Tool calls and results produced after such a rebind append to the new list; if a later model call is cancelled or raises, `_save_broken_turn` still receives the old seed and omits that visible work. This requires a rebind, subsequent tool work, and then a failure, so it is a narrow residual of the pre-existing loss rather than a reason to block this improvement.

gloryfromca (commented):

The main paths are sound: a consumed mid-turn user message is distinguished from failed-attempt work and carried into the rerun, while cancel/failure rescue now follows the rerun seed instead of the original attempt. The focused suite passes: `uv run pytest tests/test_agent_loop_dead_end_retry.py -q` -> 36 passed. Both the feature-commit diff and the full PR diff pass `git diff --check`.

I reviewed the feature diff plus the merged-main ancestry, the injection/rerun/cancel callers, provider message sanitization, session persistence/reassembly, compaction paths, and relevant history. I also checked AGENTS/CLAUDE constraints, the runtime context vocabulary, backward compatibility, and that the tests add coverage rather than weakening existing assertions. The two inline items are narrow compaction combinations and are follow-ups under the campsite rule, not reasons to hold this improvement.

Current GitHub CI has one red unit shard even though its 6,224 tests passed: the idle-time gate measured 3.45s for `test_a_vendor_with_no_catalogue_still_reaches_the_page[baichuan]`. That test is in the merged `main` material, not this agent-loop fix; the other reported checks pass.

gloryfromca (commented):

This head only merges the updated target branch; against that target, the effective PR diff remains the same three agent-loop/test files. I rechecked the target-sync history, effective diff, prior resolved threads, project rules and runtime context, relevant callers, backward compatibility, and test integrity; there is no new finding and the prior nonblocking stance still fits. `uv run pytest tests/test_agent_loop_dead_end_retry.py -q` passes: 36 passed. `git diff --check github/feat/research_dead_end_verbatim...HEAD` also passes. Current GitHub checks are passing or still running.

PR #446, opened 2026-09-16
feat(*): raven speaks Agent2Agent 1.0, inbound and outbound

LivXue:
## Summary

Raven speaks the Agent2Agent protocol, version 1.0, in both directions. The host
agent is the peer: nothing here makes a sub-agent addressable from outside, and a
raven launched as a sub-agent refuses to serve the face at all, because it is
already reached over ACP by the host that started it.

Inbound is an aiohttp face carrying the two cards and a JSON-RPC endpoint. It
mounts on the gateway when `a2a.server.enabled` is set, and `raven a2a serve` runs
it standalone for a headless deployment, where running the command is itself the
opt-in. Outbound is a single tool, `a2a_send`, that takes a card URL and a message;
it enters the tool list only when `a2a.peers` is non-empty, so a host with nowhere
to send a message does not carry the schema. The two switches are deliberately
independent: calling out and being called are separate deployments.

Onboarding mints `a2a.server.token` and stops there. The face ships closed:
finishing an unrelated install is not consent to open a second network surface, so
`raven a2a enable` is what opens it, and it is the only writer of that flag.
Enabling mints a missing token in the same command, because a face switched on with
an empty token answers every caller with a 401 and reads as a broken deployment
rather than the closed door it is. `raven a2a disable` closes it again and keeps the
token, so reopening does not re-issue a credential callers already hold. The wizard
never writes the flag at all, not even as false, so a repeat onboard leaves a face
the operator opened by hand alone. A config nobody onboarded, and any `A2aConfig()`
assembled in process, stays inert.

What that opens is narrower than a network face. The app carrying it binds loopback
at both call sites, hardcoded rather than read from config, and every request is
checked against a freshly minted 256-bit bearer token. The minting write is also
what fixes `config.json` to owner-only, from the temporary file's first byte rather
than with a chmod afterwards; a later enable or disable adds no secret material, so
it leaves whatever mode the operator chose. Nothing else narrows that file and it
already held provider API keys, so the obligation predates this work and is merely
discharged here.

There are two cards because the protocol splits one description across two
authentication states. The public card answers an unauthenticated GET, and has to:
a caller reads it to learn which scheme to authenticate with, so requiring the
credential first would be circular. It therefore names no installed agent. The
extended card is a `GetExtendedAgentCard` RPC riding the same bearer check as every
other method, and is the only one of the two that may say what this host can
actually do.

On that extended card the sub-agent roster rides `capabilities.extensions`, not
`skills`. A skill is the protocol's word for something an agent can be asked to do,
and a peer cannot ask for one named sub-agent: it can only send a message to this
host, which then decides what to dispatch. One skill per sub-agent would advertise
call targets that do not exist. What a peer can genuinely request is the
orchestration, so that is a single skill, and the roster moves to the protocol's
own extension point under a versioned URI. That is the only place a conformant card
may carry a payload the spec does not define, because `AgentCard` is a closed set
of fourteen protobuf fields: a custom top-level key is rejected by a strict parser
and silently dropped by a lenient one. The entry declares `required` false, so a
reader that has never heard of the URI ignores it and still talks to this host.

Both cards derive `supportedInterfaces[].url` from the origin the caller arrived
on. One face answers under every name that routes to it, and a URL fixed when the
face is mounted is right for at most one of them. This matters more than it looks:
a caller that finds an interface off the card's own origin must refuse it, since a
card fetched from a trusted origin would otherwise redirect the credential
elsewhere, and a relative URL reads as off-origin to that check.

## Type

- [x] Feature

## Verification

- `make test-python` at this head: 23729 passed, 112 skipped, 1 failed, in 288 s.
  The one failure is
  `tests/test_install_script.py::test_resolve_node_dir_answers_each_case_it_exists_for`,
  which fails with the same assertion on `origin/main` by itself; this branch changes
  no install or node file.
- `uv run --frozen --all-extras pytest tests/integration/test_a2a_protocol_e2e.py -q`:
  6 passed. That directory sits outside the default pytest scope, so it is run by
  hand rather than reached by the line above.
- `make lint-imports`: 10 contracts kept, 0 broken. `make lint-types`: all checks
  passed.
- `uv run ruff check raven/ tests/`: clean. `ruff format --check`: clean.
- `make check-commits` over all 68 commits, `make check-source-language`,
  `make check-large-files`: all exit 0.
- Live, over a real network boundary rather than in process: a raven in a Docker
  container and a raven on the host, each running this branch, called each other in
  both directions. Nineteen assertions per direction covering both cards, the four
  refusal paths, and a completed turn. The container publishes its face on a
  different port than it binds, so the card has to name the port the caller used and
  not the socket the server opened; both cards did, in both directions. That run
  predates the default-off change described above, which touches no file under
  `raven/a2a/` or `raven/a2a_client/`: both packages are byte-identical to the head
  it ran against.
- Reverting each fix turns its own tests red, checked one fix at a time rather than
  all together, including the interface-URL fix mutated out inside the running
  container.

## Risk

Behaviour changes for an operator who runs `raven onboard` after this lands: a
token is minted into `a2a.server.token`, and `config.json` is narrowed to
owner-only. No listening face is opened. Opening one takes `raven a2a enable` and a
gateway restart, so the second network surface only ever exists because somebody
typed for it.

Rolling an opened face back is `raven a2a disable` and a restart. The token stays in
place and is never rotated, so callers already holding it keep working when the face
is reopened, and a later onboard will not reopen it by itself. Nothing else in the
install changes.

Costs and limits, stated rather than left to be discovered:

- A gateway that serves A2A imports `a2a-sdk`, measured at roughly 570 ms and 391
  modules on top of the gateway's own stack, sqlalchemy and the OpenTelemetry api
  among them. It is paid once at mount, not per request, and an operator who
  switches the face off pays none of it. Moving it to the first request is recorded
  as deferred in the design note rather than done here.
- `a2a_send` adds 157 tokens to the tool array of every turn, measured off a
  captured provider call, and only on a host with a peer configured.
- The task store is in memory and unbounded, and inbound concurrency is not capped.
  A local face behind a bearer token is the deployment this is written for; a
  host exposed more widely should bound both first.
- Pulling in `a2a-sdk` brings the OpenTelemetry api into the gateway's import
  graph. This is a trade accepted for the SDK's protocol types and handler, not an
  adoption of OpenTelemetry by this project.

Checked during review and deliberately not fixed here: config validation accepts an
agent row with an empty or whitespace name, which would reach the extended card's
roster as an empty entry. It is not introduced by this work, the previous shape
carried the same value, and closing it belongs with the agent schema rather than
with the card.

## Related Issues

N/A


gloryfromca (commented):
Blocking: fix the SendMessage wire envelope, declare the bearer requirement, and preserve the public HTTPS origin.

I found three protocol interoperability defects, marked inline. I tried to refute each against the installed A2A 1.0 SDK and live aiohttp reproductions before keeping it.

Coverage: I reviewed the aggregate diff and branch history, the inbound and outbound callers, SDK boundary behavior, backward compatibility in config/onboarding, modifications to existing tests for weakening, and the AGENTS.md/CONTEXT.md architecture and naming constraints. The import-layer contracts remained intact.

Verification:
- `PYTHONPATH=plugins-dist/everos-memory uv run pytest <A2A and touched test set> -q`: 620 passed.
- `uv run pytest -q`: 18,862 passed, 44 failed, 129 skipped, 81 errors. The full run was not green; its output was dominated by absent optional `raven_ppt`/`raven_design` packages and plugin discovery failures.
- `uv run lint-imports`: 10 contracts kept, 0 broken.
- `uv run ruff check ...`: passed for the changed Python surfaces.
- `scripts/check_large_files.py github/main...HEAD` and `scripts/check_source_language.py github/main...HEAD`: passed when run directly because `make` is unavailable in this environment.
- Official `JsonRpcTransport.send_message()` against this server reproduced a `ParseError` on the bare response.
- A request carrying standard `Forwarded` and `X-Forwarded-Proto: https` headers reproduced an advertised `http://` interface and failed Raven's own `same_origin` check.

raven/a2a/routes_aiohttp.py
```diff
@@ -0,0 +1,205 @@
...
+
+        if method in STREAMING_METHODS:
+            return await _serve_stream(request, handler, method_name, params, request_id, context)
+
+        try:
+            result = await getattr(handler, method_name)(params, context)
+            # Serialization stays inside the guarded region, same as the streaming branch's
+            # json.dumps: a result that fails to convert or encode is a caller-facing
+            # InternalError, not an unhandled exception that falls through to a bare 500.
+            return web.json_response({"jsonrpc": "2.0", "id": request_id, "result": _to_jsonable(result)})
```
gloryfromca, line 272: Blocking: non-streaming SendMessage replies must use the protocol response envelope.

`DefaultRequestHandler.on_message_send` returns a `Task` or `Message`, but the JSON-RPC binding must serialize that inside `SendMessageResponse.task` or `.message`. This line emits the bare object. The official `JsonRpcTransport.send_message()` then parses the result as `SendMessageResponse` and fails on a successful Raven turn with `ParseError: Message type ... SendMessageResponse has no field named id`. Any conformant client that chooses non-streaming `SendMessage` therefore cannot consume Raven's reply.

raven/a2a/card.py
```diff
@@ -0,0 +1,179 @@
...
+                protocol_binding=JSONRPC_BINDING,
+                protocol_version=PROTOCOL_VERSION,
+            )
+        ],
+        capabilities=AgentCapabilities(
+            streaming=True,
+            push_notifications=False,
+            extended_agent_card=extended_available,
+        ),
+        security_schemes={
```
gloryfromca, line 98: Blocking: the card must advertise the bearer requirement it enforces.

`security_schemes` only defines what `http_auth` means; `security_requirements` selects that scheme for contacting the agent. The serialized card currently omits `securityRequirements` entirely, while every RPC call is rejected without the bearer token. A generic peer therefore discovers no required authentication, makes an unauthenticated call, and receives 401; the extended-card contract also requires the enforced scheme to be declared by both fields.

raven/a2a/routes_aiohttp.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def _interface_url(request: web.Request, config: A2aConfig) -> str:
+    """Where this agent's JSON-RPC interface is, as *this* caller reached it.
+
+    One computation for both routes. `aiohttp.web` does not re-export `yarl.URL`
+    (there is no `web.URL`); the server's path is always absolute, so plain
+    concatenation is enough.
+    """
+    return str(request.url.origin()) + config.server.path
```
gloryfromca: Blocking: cards behind a TLS-terminating reverse proxy advertise the wrong origin.

`request.url.origin()` uses aiohttp's upstream socket scheme and does not honor `Forwarded` or `X-Forwarded-Proto`. Since Raven's gateway itself listens on plain loopback HTTP, a normal HTTPS reverse proxy makes a card fetched from `https://agents.example.com/...` advertise `http://agents.example.com/a2a`. Raven's own outbound `same_origin` guard then refuses that interface. I reproduced this with both standard forwarding headers present; the card still emitted `http://` and `same_origin` returned false. The binding needs a trusted way to retain the caller-visible HTTPS origin.

raven/a2a/routes_aiohttp.py
LivXue: Confirmed and fixed in 56afad60a.

You are right that only this method is affected, and I checked the rest against the same parser rather than assume: `get_task` and `cancel_task` parse a bare `Task`, and `on_list_tasks` already returns `ListTasksResponse` itself. So the wrap is a one-entry table beside `METHODS` and `STREAMING_METHODS`, on the claim that an absent entry means the handler already returns what the client parses.

Reverting just the wrap reproduces your error verbatim:

    ParseError: Message type "lf.a2a.v1.SendMessageResponse" has no field named "id" at "SendMessageResponse".

and the official client with `ClientConfig(streaming=False)` now completes a round trip against this face where it previously could not.

This was the rule the streaming sibling already followed. `_as_stream_response` wraps for exactly this reason and says so in its docstring; I fixed one half of the dispatch and never asked the same question of the other.

One consequence worth flagging: `tests/integration/test_a2a_protocol_e2e.py` read `result["id"]`, so it had encoded the defect. Its two `SendMessage` reads move to `result["task"]["id"]` and its `GetTask` reads stay bare, which pins the asymmetry rather than hiding it.

raven/a2a/card.py
LivXue: Confirmed and fixed in d81ffb234.

`AgentCard` does carry both fields, and we set only `security_schemes`. The card now declares `security_requirements` naming `http_auth` with an empty scope list, which is the protocol saying "this scheme, no scopes" rather than no requirement at all. Both cards carry it, since the extended one rides the same bearer check.

Verified by reading the card back through the official SDK client, which now reports the requirement rather than finding none.

Worth naming why our own tests could not have caught this: raven's client resolves a peer credential from config by origin and never consults the card's requirements, so the only consumer in the test set was one that does not read the field. The same blind spot as the envelope finding, in a different field.

A follow-up in 45e1c7008, unrelated to your point but on the same file: the rebase brought in the pinned type checker, which reads `from google.protobuf.struct_pb2 import Struct` as an unresolved import (generated at import time, no stub). The type never had to be named -- `params` is declared by `AgentExtension` -- so the extension is now built in place and the serialized card is byte for byte the same.

raven/a2a/routes_aiohttp.py
LivXue: Confirmed and fixed in 887d41c6c. I reproduced it before changing anything, and got exactly what you describe:

    caller fetched the card at : https://agents.example.com/.well-known/agent-card.json
    card advertises            : http://agents.example.com/a2a
    same_origin(card, iface)   : False

The binding now reads `Forwarded` (RFC 7239) then `X-Forwarded-Proto`, taking the hop nearest the client and ignoring any value that is not http or https. With no such header the socket scheme still stands, so a plain http caller is not handed an https card it cannot use -- that case has its own test.

On "a trusted way": I went looking for the gate that phrasing implies and concluded the trust boundary is already where this puts it, which I would rather state than assume you agree with.

- `request.url.origin()` already takes the host from the caller's own `Host` header. My reproduction shows `agents.example.com` arriving that way. The scheme is the same trust level as the host, not a new one.
- The advertised URL is read by the caller's same-origin check and by nothing on this side. `supported_interfaces` has exactly two readers in the tree, the card builder and the outbound guard, which I checked rather than assumed.
- The card is built per request, so no caller can affect the card another one is given.

If you still want it behind a configured trusted-proxy flag I will add one -- that is a config-surface decision I would rather you make than I assume.

This is also the finding I am least comfortable with, since the URL it corrects is one I changed earlier on this branch. I fixed host and port and verified across a container port mapping, which is a boundary that translates the port but not the scheme, so the half I happened to test is the half that passed.

LivXue:
All three blockers are fixed and pushed, one commit each so they can be verified independently:

- 56afad60a - non-streaming `SendMessage` carries `SendMessageResponse`
- d81ffb234 - the card declares `securityRequirements`, not only `securitySchemes`
- 887d41c6c - both cards keep the scheme the caller arrived on

Each was reproduced before being changed and reverted after, alone rather than all together. Reverting the envelope reproduces your `ParseError` verbatim; reverting the scheme read returns `http://` under standard forwarding headers and fails `same_origin`; reverting the requirement takes down both new card tests. The official SDK client with `streaming=False` now completes a round trip, which it could not before.

Two of the three were defects in code I had already touched on this branch, including the interface URL, whose host and port I had fixed one commit earlier while leaving the scheme. I verified that fix across a container port mapping, which translates the port but not the scheme, so what I tested was the half that already worked.

Two notes on things you may see and wonder about:

- `45e1c7008` removes an import of `google.protobuf.struct_pb2.Struct`. The rebase brought in the pinned type checker, which reads that generated class as unresolved. `params` is declared by `AgentExtension`, so naming the type was never necessary and the serialized card is unchanged.
- `f1b2c57cf` is a one blank line formatting repair from resolving a conflict where main and this branch had each appended a function to the end of `raven/config/update.py`.

On your full-suite run: the 44 failures and 81 errors are this repo needing `--all-extras`, without which the `raven_ppt` and `raven_design` packages never load. With the flag this tree is 23240 passed, 112 skipped, and exactly 4 failed - and those same 4 fail at `origin/main` itself, which I checked by running them in a detached worktree at 5a29ea225 rather than by reading the file list. Same four ids, 4 failed / 66 passed on both sides. They are `test_config_update_tools.py` and `test_tool_capabilities.py`, around `image_search` and the web-search vendor table, and this branch touches none of those files.

I have not resolved any of the three threads. Two of them turn on protocol behaviour I would rather you re-verify than take from me, and the forwarded-scheme one has an open question in its thread about whether you want a configured trusted-proxy gate.

raven/a2a/routes_aiohttp.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at 45e1c7008eb9: the non-streaming path now wraps Task or Message in SendMessageResponse, while GetTask and CancelTask retain their bare Task contract. The real-socket interop test parses the result with the SDK response type, and the broader A2A regression run passes. This resolves my finding.

raven/a2a/card.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at 45e1c7008eb9: both public and extended cards select http_auth in security_requirements with the protocol-correct empty scope list. The in-place extension construction also preserves the serialized roster and passes the A2A type gate. This resolves my finding.

raven/a2a/routes_aiohttp.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at 45e1c7008eb9. I accept the stated trust boundary: the scheme is restricted to HTTP(S), only changes the per-request card returned to that caller, and does not drive server-side routing. The tests cover Forwarded/X-Forwarded-Proto for both card variants and preserve plain HTTP when no forwarding header exists. This resolves my finding.

gloryfromca (commented):

The new revision fixes all three findings from my previous round: non-streaming SendMessage now uses the SDK response envelope, both cards advertise the bearer requirement they enforce, and cards preserve the caller-visible HTTPS scheme behind a forwarding proxy. I also checked the final in-place protobuf extension construction introduced after the rebase. I replied to and resolved each thread I opened.

Coverage this round included the repository rules in AGENTS.md and CONTEXT.md, the complete PR diff plus the range-diff from the reviewed revision, affected callers and official SDK parser/types, commit history, backward compatibility at the protocol boundary, test changes for weakened expectations, and the import-layer architecture contracts. I found no new plain error.

Verification:
- `uv run pytest tests/test_a2a_*.py tests/integration/test_a2a_protocol_e2e.py tests/test_agent_loop_subagent_role.py tests/test_agent_loop_token_budget.py tests/test_cli_a2a_commands.py tests/test_cli_onboard_commands.py tests/test_config_update.py tests/test_l4_entrances.py tests/test_no_otel_tracing.py tests/test_spine_scheduler.py tests/test_trajectory_store.py -q`: 635 passed, 58 SDK deprecation warnings.
- Focused card/route/interop/E2E run: 51 passed, 58 SDK deprecation warnings.
- Ruff check and format check over the A2A implementation and focused tests: passed.
- `uv run --frozen --python 3.12 lint-imports`: 10 contracts kept, 0 broken.
- `uv run --frozen --python 3.12 --all-extras ty check raven/a2a`: passed.
- The repository-wide type invocation reports one error at `raven/agent/tools/web.py:516`; that file is unchanged by this PR and the error is already present on the target branch, so it is not a regression from this revision.

gloryfromca (commented):

This head merges current `main` into the already reviewed revision. The merge tree exactly matches Git's conflict-free merge result, `git show --remerge-diff` has no conflict-resolution delta, and the A2A implementation and tests are unchanged from `45e1c7008eb9`. I found no regression or new plain error.

For this revision I checked the refreshed `github/main...HEAD` diff, the first-parent delta and merge history, the affected capability/doctor callers brought in from `main`, backward compatibility of the existing A2A paths, test expectations for weakening, AGENTS.md and CONTEXT.md constraints, and the import-layer architecture contracts. The three threads I opened earlier remain resolved.

Verification:
- Expanded A2A and merged-main regression selection: 807 passed, with 58 SDK deprecation warnings.
- `uv run --frozen --python 3.12 --all-extras ty check raven evolver agents plugins-dist scripts`: passed.
- `uv run --frozen --python 3.12 lint-imports`: 10 contracts kept, 0 broken.
- Ruff check and format check over the A2A surface and merged-main affected files/tests: passed.

gloryfromca (commented):

The delta from the previously reviewed head is test-only: it expands `tests/test_a2a_runtime.py` without changing production behavior or weakening existing expectations. The new cases correctly exercise the real loop adapter contract, protobuf request conversion, parked-task resume paths, standalone site cancellation, factory readiness, and the distinct broker rules for standalone and gateway-hosted loops. I found no new plain error.

For this revision I checked the refreshed `github/main...HEAD` diff and commit history, the complete new test delta, the runtime implementation and callers those tests cover, backward compatibility, whether assertions were weakened, AGENTS.md and CONTEXT.md constraints, and the import-layer architecture contracts. The previously settled review threads remain non-blocking.

Verification:
- Focused runtime/routes/interop/E2E selection: 44 passed, with 58 SDK deprecation warnings.
- Broader A2A and affected-caller regression selection: 644 passed, with 58 SDK deprecation warnings.
- Ruff check and format check for `tests/test_a2a_runtime.py`: passed.
- `uv run --frozen --python 3.12 lint-imports`: 10 contracts kept, 0 broken.
- `uv run --frozen --python 3.12 --all-extras ty check raven/a2a`: passed.

gloryfromca (commented):

This head is a conflict-free merge of current `main` into the previously reviewed revision. Its tree exactly matches Git's automatic merge result, `git show --remerge-diff` contains no conflict-resolution delta, and no A2A implementation or A2A test changed from `911295a35ac5`. I found no regression or new plain error.

For this revision I checked the refreshed `github/main...HEAD` diff, first-parent delta and merge history, shared onboarding/config callers touched across the histories, backward compatibility of the A2A paths, test expectations for weakening, AGENTS.md and CONTEXT.md constraints, and the import-layer architecture contracts. All threads I opened remain resolved.

Verification:
- Broader A2A and affected-caller regression selection: 644 passed, with 58 SDK deprecation warnings.
- `uv run --frozen --python 3.12 --all-extras ty check raven evolver agents plugins-dist scripts`: passed.
- `uv run --frozen --python 3.12 lint-imports`: 10 contracts kept, 0 broken.
- Ruff check and format check over the A2A surface and shared onboarding/config files: passed.

0xKT:
**Not a blocker** -- four things, none of which stops this merging. The security shape is right
everywhere I looked: off by default, an empty token refuses every caller rather than admitting
one, `hmac.compare_digest` for the comparison, sub-agent processes refused, both hostings default
to loopback, and the public card carries discovery data only while the sub-agent roster rides the
extended card behind the authenticated route.

**1. A non-ASCII Authorization header answers 500, not 401** (raven/a2a/auth.py:29)

`hmac.compare_digest` raises `TypeError: comparing strings with non-ASCII characters is not
supported` when either side is a non-ASCII `str`, and `presented` comes straight off the wire.
The call at raven/a2a/routes_aiohttp.py:244 sits between the file's two try blocks -- the first
wraps `request.json()` (230-232), the second the dispatch (265-274) -- so nothing catches it and
aiohttp answers a bare 500 with a traceback, to a caller who presented no valid credential.

It fails closed: the handler is never reached, so this is robustness and log noise, not a bypass.
Worth closing anyway because the file already states the principle for the other path
(routes_aiohttp.py:268-271: "a result that fails to convert or encode is a caller-facing
InternalError, not an unhandled exception that falls through to a bare 500"), and because
tests/test_a2a_routes.py already names two members of exactly this family -- the non-dict body
and the non-string method -- while tests/test_a2a_auth.py's six parametrised headers are all
ASCII. One `presented.isascii()` guard in `is_authorized` closes it.

**2. The import-cost docstring is wrong in both directions** (raven/a2a/gate.py:71-78)

It says building the handler behind `may_mount` "keeps a2a-sdk off the module import path, so a
gateway whose operator switched the face off never pays for it at all", and that the a2a-sdk and
OpenTelemetry cost is "paid once at mount".

Neither holds, because raven/agent/loop/wiring.py:10 imports `A2aTool` at module level, and that
reaches a2a-sdk through raven/a2a_client/tool.py and client.py. Measured on this head with the
face disabled:

    import raven.config              -> no a2a, no google, no opentelemetry   (control)
    import raven.agent.loop.wiring   -> a2a, google, opentelemetry all loaded
                                        32 a2a modules, 15 opentelemetry modules
    config.a2a.server.enabled        -> False

So an operator who switched the face off pays for it anyway, and the mount is not what loads
OpenTelemetry -- the loop wiring did, earlier. The tool itself is already conditional
(wiring.py:1026 builds it only `if self.a2a_config.peers`), so moving that import next to its use
restores exactly what the docstring claims, in one line, with no behaviour change. It would also
make `a2a-sdk` eligible to be an extra rather than one of the 29 core runtime dependencies --
which matters because it brings google-auth, googleapis-common-protos, proto-plus, json-rpc,
pyasn1 and opentelemetry-api into every install, while channels, tools, sandbox and browser are
all extras in this project.

**3. The OTEL guard errors out on the reversal its own docstring prescribes**
(tests/test_no_otel_tracing.py)

The docstring says the reversal, if the trade is judged wrong, is to drop `a2a-sdk`. Do that and
`test_the_api_present_transitively_cannot_record` cannot run: its `from opentelemetry import
trace` raises, and the file that guards the decision errors instead of passing. Wrapping that
import in the same `_absent` shape the file already has -- skip when the api is gone, since its
absence is the stronger state -- keeps the reversal a one-line dependency edit.

**4. `gateway.a2a_handler` is written and never read** (raven/rpc/transports/ws.py:113 and :530)

Initialised at :113, assigned the mount's return at :530, and nothing in the tree reads it. Either
a reader is missing or the attribute is.

For the record, the parts I checked hardest and found sound: the OTEL guard's narrowing is forced
(the lock carries `opentelemetry-api` only -- no sdk, no exporter) and is proved behaviourally
rather than asserted; `atomic_update`'s new `mode` is keyword-only and defaults to `None`, which
`_replace_unlocked` already treats as "preserve the existing mode", so all ten existing callers
are byte-identical; and tests/test_l4_entrances.py reading the forbidden-surface roster out of the
contract while pinning the four names as a floor is the right shape for that guard -- the roster
copy had already drifted past `raven.a2a`, and a read alone would have let a removal narrow the
contract and the guard in one line.

Full suite on this head against the merge base: identical failure sets, 216 names each, no
difference in either direction; +137 passing, which is this PR's own tests.

gloryfromca (commented):

The new verifier note changes my classification from clean to nonblocking, not my conclusion that this revision can merge. It names four follow-ups: a non-ASCII bearer value escaping as a bare 500, eager A2A/OTEL imports despite the gate docstring, the OTEL reversal test assuming the optional import exists, and an unread `gateway.a2a_handler` attribute. The reviewer explicitly assessed all four as non-blocking; I am carrying that scope decision rather than reopening this revision.

I independently reproduced the first item: `is_authorized` raises `TypeError: comparing strings with non-ASCII characters is not supported` for `Bearer tést`. It fails closed before dispatch, so it does not undermine the security or mergeability conclusions from my review. `uv run pytest tests/test_a2a_auth.py tests/test_a2a_routes.py -q` still reports 24 passed; the missing non-ASCII case explains why the suite remains green. My three earlier threads remain settled and resolved.

gloryfromca (commented):

The rewritten history is patch-equivalent to the previously reviewed A2A series except for a resolver-generated lockfile refresh and the intentional opt-in change at this head. Onboarding now provisions the credential without opening the gateway face; `raven a2a enable` mints before enabling, `disable` preserves the credential, repeat onboarding preserves an operator's choice, and both commands clearly require a gateway/web restart. I found no new plain error, and the SDK 1.1.4 resolution remains compatible with the implementation.

The four follow-ups already named in the existing verifier note remain outstanding and explicitly non-blocking, so this revision carries a nonblocking rather than clean verdict. I am not re-reporting them as new findings. All three threads I opened remain resolved.

For this revision I covered the refreshed `github/main...HEAD` diff, the range-diff and commit history across the rebase, the changed config/CLI callers, backward compatibility for existing enabled and disabled installs, test expectations for weakening, AGENTS.md and CONTEXT.md constraints, dependency resolution, and import-layer architecture.

Verification:
- Focused A2A, interop, onboarding, CLI, and config suite: 508 passed, with 58 SDK deprecation warnings.
- `uv lock --check`: passed, 250 packages resolved.
- Full production `ty check`: passed.
- Import contracts: 10 kept, 0 broken.
- Ruff check and format check over the affected surface: passed.
- Source-language, large-file, and `git diff --check` gates: passed.

Handsome-wzw (approved):

PR #448, opened 2026-09-16
build(deps): raise the python dependencies whose advisories have a path here

0xKT:
## Summary

Thirty-three Dependabot alerts sit on `uv.lock`, the only manifest in this repo
whose packages are installed on a user's machine, and no pull request had been
raised for any of them. The python-security group PR (#227) was auto-closed by a
history rewrite on 2026-09-12 and has not returned. This closes all thirty-three
of them.

| Package | From | To | Alerts | Why |
|---|---|---|---|---|
| pillow | 12.2.0 | 12.3.0 | 13 (10 high) | `prepare_image` in `raven/agent/tools/media.py` calls `Image.open` on bytes the agent is handed, and the design engine calls `paste()` and `crop()` -- the APIs the heap out-of-bounds advisories name |
| mistune | 3.2.1 | 3.3.4 | 10 (4 high) | One call site, rendering outbound markdown for the Matrix adapter |
| aiohttp | 3.14.1 | 3.14.3 | 3 (1 high) | The C response parser, reading answers from provider endpoints a user configures |
| json-repair | 0.59.5 | 0.63.4 | 1 (1 high) | Unbounded CPU on circular schema refs |
| h2 | 4.3.0 | 4.4.1 | 1 | Ordinary minor bump |
| hpack | 4.1.0 | 4.2.0 | - | Pulled by h2 |
| pip | 26.1.2 | 26.2.1 | 1 | Ordinary minor bump |
| mcp | 1.27.0 | 1.30.0 | 3 (3 high) | Advisories are server-side only and this repo is a client; taken to leave no alert standing |
| cryptography | 48.0.1 | 50.0.1 | 1 (1 high) | Advisory is PKCS#7 `EnvelopedData`, which nothing here decrypts; taken for the same reason |

The last two were held back in the first commit on a cost argument, and the
argument was overruled: every alert closes rather than carrying a documented
exception. It is worth recording what the exception would have said, because it
is still true of the code. `mcp`'s three high advisories are all server-side --
HTTP transports serving session requests without auth, the WebSocket server
transport, experimental task handlers -- and every import in this repo is
`mcp.client.*` or `mcp.shared.*`, with no `mcp.server` import anywhere.
`cryptography`'s advisory is PKCS#7 `EnvelopedData` decryption, while the only
import here is `cryptography.hazmat.primitives.ciphers`, the AES fallback in the
weixin adapter. Neither was reachable; both are now moot.

A major version is where an entry point disappears quietly, so that was checked
before the suite ran: all sixteen symbols this repo imports from the two still
resolve on the new versions.

Four land above the version their advisory requires -- `json-repair` at 0.63.4
for a 0.60.1 fix, `mistune` at 3.3.4 for 3.3.0, `mcp` at 1.30.0 for 1.28.1,
`cryptography` at 50.0.1 for 50.0.0 -- because `uv lock --upgrade-package`
resolves to the newest version the constraints allow. Say the word and any of
them can be pinned to the minimum fix instead.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [x] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Full suite against the upgraded lock, twice: once on the first cut and again
after rebasing onto current `main`.

```
$ uv sync --locked --all-extras --dev && uv run pytest -q
1 failed, 22930 passed, 121 skipped, 110 warnings in 229.20s
```

The single failure is
`tests/test_read_file_image.py::test_image_block_is_the_only_place_the_shape_is_written`,
and it is **not** from this change. It shells out to `grep` and filters the
result on the string `raven/utils/images.py`; the `grep` on this developer's
PATH emits `raven//utils/images.py` with a doubled separator, so the filter
misses its own file and the test reports itself as an offender. Checked against
an unmodified `main` worktree, where it fails identically. CI uses GNU grep and
is green on it.

So: 22930 passed, zero regressions from the nine bumps.

## Risk

This is the batch that reaches users: `uv.lock` is what `pip` and
`uv tool install` resolve, so unlike the npm trees these packages run on the
machine of whoever installs raven.

All nine stay inside the constraints already declared in `pyproject.toml`; no
constraint was widened and `pyproject.toml` is untouched. `pillow` and `aiohttp`
are patch-level. `mistune` and `json-repair` move a minor, and both are used in
one narrow place each, which the suite covers.

`cryptography` is the one to watch: two major versions, and it is a build
dependency of several other packages in the tree as well as a direct import
here. `mcp` moves three minors, and seven files use its client.

`pillow` ships platform wheels, so the failure mode to watch is an import error
on a platform the resolver picked differently, which shows at install rather
than at build.

Rollback is to revert the commit and `uv sync`.

## Related Issues

N/A. Closes all 33 Dependabot alerts on `uv.lock`.


gloryfromca (approved):

uv.lock
```diff
@@ -698,55 +698,52 @@ wheels = [
 
 [[package]]
 name = "cryptography"
-version = "48.0.1"
+version = "50.0.1"
```
gloryfromca, line 701: Blocking: cryptography 50 drops Intel macOS support, breaking Raven's advertised macOS installer.

`install.sh` accepts `Darwin` on `x86_64`, but cryptography 49 removed x86_64 macOS support and 50.0.1 publishes only arm64 macOS wheels. I reproduced the platform break with `uv pip install --dry-run --python-platform x86_64-apple-darwin --python-version 3.12 --only-binary :all: cryptography==50.0.1`: the resolver reports no usable wheels, whereas 48.0.1 resolves. Allowing the sdist does not preserve the clean-install promise: upstream's macOS build instructions require Xcode headers, Rust, and a non-system OpenSSL, and Raven's installer provisions none of them. This makes the official one-line install fail on a platform it still accepts. Please resolve the dependency/platform compatibility before merging.

Upstream references: [49.0.0 changelog](https://cryptography.io/en/latest/changelog/#v49-0-0), [macOS source-build requirements](https://cryptography.io/en/latest/installation/#building-cryptography-on-macos).

gloryfromca (commented):
Blocking: cryptography 50 removes Intel macOS support from an installer that still advertises x86_64 macOS.

I reviewed the complete `github/main...HEAD` lockfile diff, the release/export and installer/self-upgrade consumers, dependency history, compatibility boundaries, tests, and the repository rules. No tests were changed or weakened, and no domain/architecture source changed; the release constraints do carry all nine updated versions exactly.

- `uv lock --check`, `uv sync --locked --all-extras --dev`, and `git diff --check` pass.
- The MCP, Weixin crypto, image, and sandbox unit selection passes: 485 passed.
- All four GitHub Python 3.12 shards and the remaining CI checks pass.
- The exported-lock audit no longer reports any of the nine updated packages; it still exits nonzero only for pre-existing findings in untouched `click` and `lxml-html-clean`.
- Before the final MCP/cryptography commit, my full local suite reached 22,931 passed and 120 skipped, with one reproducible Rich-theme assertion failure whose test, implementation, and locked Rich version are byte-for-byte unchanged from `main`.

The inline platform regression must be resolved before merge.

PR #449, opened 2026-09-16
ci(*): give each shard one worker per core so the idle ceiling measures waiting

gloryfromca:
## Summary

The idle ceiling added in #428 reads wall clock minus CPU, on the premise that the difference is a test waiting on something. A process queueing for a core spends it the same way.

`addopts` sets `-n 4`, and its comment says why: *"on a 2-core runner four CPU-bound workers cost only context switching"*. That was measured when the whole suite ran on one runner. #413 then split the work across four runners, so four workers on a two-core runner are oversubscription the sharding had already made unnecessary -- and the new gate charges its cost as idle.

This is why `tests/test_rpc_model.py` trips the ceiling on CI and never locally, and why it names a **different case each run**: whichever one is executing when contention peaks.

| run | charged |
|---|---|
| `main` @ `73d6bb707`, shard 4/4 | `test_options_config_models_rank_before_common_and_dedup` 4.30s, `..._page[lanyun]` 3.02s |
| #416, shard 2/4 | one case at 3.44s |
| #416, same shard re-run | `test_model_methods_registered_via_helper` 4.44s, `test_options_lists_the_codex_models...` 3.15s |

`main` fails a shard on this today, as does every branch cut after the gate landed. Re-running does not help.

### The measurement

`tests/test_rpc_model.py`, worker count against tests over a `0.05s` idle ceiling:

| workers | tests over the ceiling | worst |
|---|---|---|
| `-n 0` | none | -- |
| `-n 2` | none | -- |
| `-n 4` | 4 | 0.27s |
| `-n 16` | **77** | 1.17s |

Serially there is essentially no idle in that file, so the wait is the workers rather than the test. Nothing in it sleeps or reaches a socket.

### The change

`coverage-shard` runs one worker per core of the runner it lands on, through `COVERAGE_SHARD_WORKERS ?= 2` so a bigger runner can override it. `addopts` keeps `-n 4`, so a local run and every other caller are unchanged.

Marking the file `slow` was the other candidate and is the wrong shape: the file is not slow by nature, the case that trips rotates, and the same contention will reach any other file whose cases sit near the line.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [x] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `make coverage-shard COVERAGE_SHARD=2/4 PYTEST_ARGS=--idle-ceiling-strict` locally -> 5253 passed, 3 skipped, in 75.94s, no idle report
- The worker-count table above was produced with `pytest tests/test_rpc_model.py --idle-ceiling 0.05 -n <N>`; `-n` on the command line overrides the `addopts` value, which is what the target relies on
- Root cause attributed by checking `main` itself: run 35076470253 on `73d6bb707` fails the same gate on the same file, so this is not introduced by any branch
- CI is the real verification for the runner-side effect, since the contention cannot be reproduced on a many-core machine -- the shards in this PR are the measurement

## Risk

Shard wall clock may move. Two workers on two cores is the shape the runner can actually serve, so the expected direction is neutral to slightly faster, but the four shard jobs in this PR are the measurement. If a shard gets slower, raise `COVERAGE_SHARD_WORKERS` for that runner rather than reverting -- the ceiling reading is what the worker count now has to respect.

Nothing outside the sharded CI path changes: `addopts` still says `-n 4` for a local `pytest` and for `make coverage`.

## Related Issues

Unblocks the unit shard on #416 and on any branch cut after #428. N/A otherwise.


Makefile
```diff
@@ -95,8 +95,19 @@ test-python:
...
+# One worker per core of the runner a shard lands on. The -n 4 in addopts was
+# measured on the whole suite running on one runner; the shards then split that
+# work four ways, so four CPU-bound workers on a two-core runner are pure
+# oversubscription. That matters beyond wall clock now: the idle ceiling reads
+# wall minus CPU, and a process queueing for a core spends that difference the
+# same way a process waiting on a socket does -- which is why tests/test_rpc_model.py
+# trips the ceiling on CI and never locally, charging a different case each run.
+# Measured on tests/test_rpc_model.py: -n 0 and -n 2 report no test over 0.05s
+# idle, -n 4 reports four, -n 16 reports seventy-seven.
+COVERAGE_SHARD_WORKERS ?= 2
```
gloryfromca, line 107: Blocking: make the worker count and rationale match the actual runner capacity.

This is a public repository using standard `ubuntu-latest`; GitHub documents that runner as 4 CPUs, not 2: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories. Consequently, `2` does not implement the source and PR claim of one worker per core, and the advice to override this on a bigger runner would lead a maintainer to restore `-n 4` on the runner already in use. The live same-revision comparison also contradicts the stated neutral-to-faster expectation: shard 2 took 336.06s with this `-n 2` override versus 238.12s with `-n 4` on current main, about 41% slower. If reserving half the CPUs is the measured requirement for reliable idle accounting, keep `2` but state that actual contract in the title, source comment, and risk analysis; otherwise pass a worker count that really follows runner capacity.

gloryfromca (commented):
Blocking: align the stated worker-per-core contract with the runner this workflow actually uses.

I found one blocking issue inline. The two-worker override works mechanically, and the targeted local checks pass, but the central runner-capacity premise and the live timing evidence do not support the contract currently documented in the title, source, and PR body.

Verification:
- `uv run pytest tests/test_rpc_model.py --idle-ceiling 0.05 -n 2`: 195 passed in 30.95s, with no idle report.
- The exact `make coverage-shard ...` wrapper could not start because this container has no `make`. Running its expanded `uv run --frozen --python 3.12 ... --shard 2/4 -n 2 --idle-ceiling-strict` recipe passed: 5253 passed, 3 skipped, 30 warnings in 201.97s.
- Live PR CI: shards 2-4 passed; shard 1 failed with an existing cron test timing out and reporting 5.23s idle, so the coverage gate was skipped. That failure is on untouched code and is not a separate inline finding, but the revision's required CI is currently red.

Review coverage: I read `git diff github/main...HEAD`, the Makefile callers in `.github/workflows/ci.yml`, pytest `addopts` and idle/shard hooks, the relevant test documentation, the introducing history for sharding and the idle gate, and the current-main merge delta. I checked AGENTS.md/CLAUDE.md compliance, caller compatibility and overrides, test weakening (none), and architecture/domain-term applicability (no architecture boundary or registered domain term changes).

gloryfromca:
Evidence from since this was opened, in case it helps prioritise.

**`main` is persistently red on this.** Its last three CI runs all failed the same gate, each naming a different case in the same file:

| run | charged |
|---|---|
| 35107785288 | `test_remove_model_reflected_in_options` 4.08s of 8.02s |
| 35106418313 | `test_model_methods_registered_via_helper` 3.77s of 7.65s |
| 35104701761 | `test_no_handler_reads_the_catalogue_on_the_event_loop[options_again]` 3.92s of 7.65s |

Nothing merges while that holds, so this is not only about the PR it was cut for.

**It has spread past one file**, which is what the description predicted: #416's latest run charged `tests/test_knowledge_manager.py::test_create_upload_index_search` at 4.56s on one shard and `tests/test_rpc_model.py` on another. Marking cases would have been whack-a-mole.

**This branch's four shards are green** with one worker per core, and no `test_rpc_model.py` case is charged on any of them. One shard first failed `tests/test_cron_service_wake.py::test_mutation_during_inflight_run_fires_once_and_leaves_no_claim` with a `TimeoutError`; that case passes three times out of three locally at both `-n 2` and `-n 4`, and the shard is green on re-run, so it was a flake rather than an effect of the worker count.

0xKT:
Closing in favour of #466, which is now on `main`.

The diagnosis here was the pointer: idle grew with the worker count and not with anything the tests did, so the ceiling was charging run-queue wait as waiting. #466 changes the measurement rather than the worker count: the ceiling now subtracts the time a test's threads spent queued for a CPU, read per thread from `/proc/self/task/<tid>/schedstat`, with each thread writing its last reading on the way out because the per-test event loop takes its executor threads with it. `-n 4`, the 3 s ceiling and the four shards stay as they are, so the shard time this branch traded is not paid.

Two full CI runs on #466 were green, all four unit shards under the strict ceiling included; the same shard command pinned to two CPUs fails the session on the previous `main` and passes with the change. The contrast numbers are in that PR's description.

Thanks for the measurement table; it is what made the cause legible.

PR #450, opened 2026-09-16
fix(providers): a caller that will not fetch is not made to wait

Handsome-wzw:
## Summary

`allow_fetch=False` promises an answer from what is already on hand. The LiteLLM tier honoured half of that promise: the flag skipped LiteLLM's import, and then called `litellm.get_model_info` anyway. That call is not a second read of memory -- for an id the static table misses it asks whoever serves the model.

Measured on `main`, with LiteLLM already imported (which is the production state):

```
resolve_context_window("huggingface/deepseek-ai/DeepSeek-V4-Pro", allow_fetch=False)
  -> None   0.72s   outbound: huggingface.co:443
resolve_context_window("ollama_chat/llama3.1", allow_fetch=False)
  -> None   0.00s   outbound: localhost:11434
```

The two callers that pass the flag are the two that cannot afford to wait: `model.options`' row builder (`rpc/methods/model.py`) and `AgentLoop` construction (`providers/binding.py`). Both could block on a network round trip that the flag was there to prevent.

**How.** Skip the ask only where the ask leaves the process. Reading the installed `_get_model_info_helper`: the `huggingface` branch fetches the model's `config.json` from huggingface.co, and the `ollama`/`ollama_chat` branch asks the local server for any id `model_cost` does not hold. For every other id the call normalises the key against the same in-memory table the exact read missed -- five spellings, no I/O -- and returns, so it is a free read and a no-fetch caller keeps it. The two families are named once (`_ASK_LEAVES_MEMORY`) beside the table they complement, and the guard is `if not allow_fetch and _ask_leaves_memory(candidate)` in both tiers.

This matters because the exact-key read in `_table_entry` misses many spellings the normalised read answers: the factory default `anthropic/claude-opus-4-5` is one. A first revision of this PR skipped the ask on every table miss and handed those models the 65536 fallback -- `claude-opus-4-6` and `4-7` from 1000000 -- which the review caught; the narrowing is the correction.

Also renamed the flag from `allow_import` to `allow_fetch` so it names what it gates, and applied the same guard to `_try_litellm_max_output`, which had the identical shape. An ollama id `model_cost` knows only under a stripped spelling is skipped with its family; the caller keeps its configured default, which is the conservative side of that line and what `resolve_context_window` already documents for an unknown id.

**A measurement I could not make.** `tests/test_rpc_model.py` calls `resolve_context_window(..., allow_fetch=False)` 62,545 times, and this change takes the file from 14.9s to 11.9s locally (two runs each). That file is also where the `--idle-ceiling-strict` gate has been failing shards intermittently, with zero test failures. I cannot claim this fixes that: locally the file reports no idle at all, even at `--idle-ceiling 0.3`, so the idle is a runner phenomenon I cannot reproduce. This PR is offered for the contract and the two user-facing callers; whether the gate settles is for CI to say.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `test_allow_fetch_false_still_reads_the_normalised_table_for_free`: asserts as a precondition that the exact-key read misses `anthropic/claude-opus-4-5`, then that the no-fetch arm equals the fetching arm in both tiers with every outbound seam trapped (`socket.getaddrinfo`, `socket.socket.connect`, `httpx.get/post/request/stream`, `httpx.Client.send`).
- `test_allow_fetch_false_does_not_ask_litellm_where_the_ask_leaves_memory`: records every `get_model_info` call for unknown `huggingface/`, `ollama/` and `ollama_chat/` ids and asserts none of the asks leaves memory; the `openrouter/` alias candidate is allowed through because that ask is answered in memory. The default path must still ask, so the tier is deferred rather than dead.
- Mutation-checked both ways: the unconditional skip fails exactly the first test; a predicate that never fires fails exactly the second.
- `ModelBinding(...).context_window` with litellm imported and every seam trapped: `claude-opus-4-5` 200000, `claude-opus-4-6` 1000000, `claude-opus-4-7` 1000000; 0 outbound attempts.
- Every id shipped in `provider-models.json` (2194 de-duplicated) through both no-fetch paths with every seam trapped: 0 outbound attempts.
- `uv run --frozen pytest tests/test_provider_rates.py tests/test_agent_loop_lazy_provider.py tests/test_litellm_setup.py tests/test_provider_resolution_invariants.py -q`: 493 passed, 96 skipped.
- `ruff check` and `ruff format --check` on the changed files: clean.
- No user-facing docs change: the flag's documented contract is what this restores.

## Risk

- Security: none. The change removes outbound requests from a path that promised none; it adds no new I/O.
- Backward compatibility: a no-fetch caller asking about a `huggingface/` or `ollama*/` id absent from `model_cost` now gets the configured default instead of waiting on the network; every other id, including the factory default, resolves exactly as before (asserted by the equality test). Fetching callers are unchanged. The internal keyword `allow_import` is renamed to `allow_fetch`; both call sites in the repo are updated.
- Rollback: revert the squash commit; nothing persists.

## Related Issues

N/A


Handsome-wzw:
CI answered the question the description left open, and the answer is clearer than the local numbers suggested.

Same runner, same shard (`unit 2/4`), same 6150 tests. The "before" column is the last green run of #408 on `main` without this change; the "after" is this PR's first run:

| | before | after |
|---|---|---|
| shard wall time | 229.14s (3:49) | 135.49s (2:15) |
| `test_options_lists_lm_studio_models_from_the_local_server` | 7.43s | 1.70s |
| `test_an_added_local_runtime_asks_for_an_address_and_takes_a_key[gpustack]` | 7.43s | 1.70s |
| slowest five in `test_rpc_model.py` | 7.75 / 7.43 / 6.51 / 6.15 / 6.06 | 6.26 / 3.66 / 2.18 / 1.70 / 1.70 |

The shard is 41% faster, and the tests the idle gate has been failing shards over went from 6-8s to 1.7s. A test whose whole call takes 1.7s cannot hold 3s of idle, which is why no `idle of` line appears in this run at all.

This also explains why I could not reproduce the idle locally, as the description says. The `get_model_info` calls this change skips do reach the network on the runner; on a developer machine a refused local connection returns immediately, so the same calls cost almost nothing. Locally the file went 14.9s -> 11.9s (21%); on CI the shard went 41%.

So the last section of the description is now out of date, and I would rather correct it here than quietly edit it: this does look like the cause of the intermittent `--idle-ceiling-strict` failures, not only a contract fix. The gate was reporting a real wait.

gloryfromca (commented):

Reviewed the full `github/main...HEAD` diff and the surrounding LiteLLM table/import logic, ModelBinding and RPC callers, relevant history, the Token Rates architecture terms in `CONTEXT.md`, and the applicable `AGENTS.md` rules. I also checked backward compatibility and test integrity: the renamed keyword is private and every repository call site is updated, while the new regression test preserves the fetching path and proves only the no-fetch path skips `get_model_info`.

The implementation keeps all already-loaded LiteLLM candidate table lookups and the OpenRouter cache-only fallback, skipping only the provider-backed lookup that violates `allow_fetch=False`. I found no concrete failure to raise.

Verification:
- `uv run --frozen --python 3.12 --all-extras pytest -q tests/test_provider_rates.py tests/test_agent_loop_lazy_provider.py`: 83 passed.
- `uv run --frozen --python 3.12 --all-extras pytest -q`: 22932 passed, 120 skipped, 1 unrelated failure in the untouched `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`; the failure reproduces in isolation.
- Ruff check and format checks passed for all four changed files.
- The source-language gate and `git diff --check` passed.

raven/providers/rates.py
0xKT, line 1: **Blocking: the premise is wrong for everything except the two prefixes it names, and the skip costs 70 shipped models their real context window -- including the factory-default model -- while saving nothing.

The new guard is justified in the docstring at `raven/providers/rates.py` with:

> ``get_model_info`` is no cheaper for the ids it cannot answer from memory: it asks the provider.

That is true only for `huggingface/*` and `ollama*`. For every other id `get_model_info` normalises keys against the in-memory `litellm.model_cost` and returns without any request. I measured it with every outbound seam trapped so that a request would raise rather than succeed: `litellm.get_model_info("anthropic/claude-opus-4-5")["max_input_tokens"]` returns `200000` with **0 outbound attempts**. The ask was already free for these ids; skipping it buys nothing and loses the answer.

### What it costs

`ModelBinding.context_window` (`raven/providers/binding.py:80`) is an `allow_fetch=False` caller, so it takes the full hit. Same worktree (this PR's head merged onto `5a29ea22`), same script, only `raven/providers/rates.py` swapped:

| process with litellm already imported | base `5a29ea22` | this PR |
|---|---|---|
| `ModelBinding("anthropic/claude-opus-4-5").context_window` | **200000** | **65536** |
| outbound attempts after the import | 0 | 0 |

`anthropic/claude-opus-4-5` is `AgentDefaults.model` at `raven/config/schema.py:176`, so this is the out-of-the-box configuration, and `context_window_tokens` defaults to `None` (schema.py:194) -- nothing overrides it. A sweep over the ids shipped in `raven/providers/data/provider-models.json` (2798 after de-duplicating `curated` + `overrides`) finds **70** whose window differs between the two arms, `claude-opus-4-6`/`4-7` among them at 1000000 -> 65536, a 15x over-trim.

"With litellm already imported" is the production shape, not a contrivance -- `binding.py` says so itself, two lines above the call:

> building a provider is what imports LiteLLM, and a lazily-built one has not done it yet [...] **By the first read there is a turn in flight and the import has happened.**

In a process where litellm was never imported both arms answer 65536, which is why a cold unit test will not show this.

### The mitigation the code relies on is gone

The same docstring continues:

> A miss is therefore not cached: the next read, once a catalogue is warm, can still find the real number.

That was true while the miss meant "the catalogue is not warm yet". The new `continue` is keyed on `allow_fetch`, not on catalogue warmth, so the miss is now permanent: a warm catalogue cannot rescue it, and `lazy.py:63` records "No production setter today" for `on_built`, so nothing else corrects it later.

### The rule

`_table_entry`, unchanged by this PR (`rates.py:160-169`):

> The table is read here to skip the ask where the ask cannot be made; it does not replace it.

This change makes the table replace the ask for every `allow_fetch=False` caller. And `rates.py:986-988` records what the rung below costs:

> 2026-09-11 it gave a 1,048,576-token model one sixteenth of its window [...] until every request carried an orphan tool result

### What would fix it

The two outbound branches are distinguishable from the id with no I/O -- `huggingface/` and `ollama`/`ollama_chat/` prefixes are the only ones `_get_model_info_helper` takes to the network. Guarding on the prefix instead of on every table miss keeps the promise this PR is making (a no-fetch caller never waits) without discarding the 70 windows that were free to read.

Worth saying: the promise itself **is** kept. With `httpx.get/post/request/stream`, `httpx.Client.send`, `socket.socket.connect` and `socket.getaddrinfo` all trapped to record and raise, and litellm already imported, I drove all 2798 shipped ids through both `resolve_context_window(..., allow_fetch=False)` and `declared_max_output_tokens(..., allow_fetch=False)`: **0** outbound attempts from those paths (1 attempt total in the run, from litellm's own start-up map fetch, which is not this code). The rename is right and the direction is right; it is the unconditional skip that overshoots.

gloryfromca (commented):
Blocking: preserve LiteLLM's free in-memory normalization for no-fetch callers.

I reviewed the new merge revision against current `main`; the PR's four-file feature diff is unchanged, and the merged `main` adds no mitigating path. I independently reproduced the existing blocker on this head: after `import_litellm()`, the exact table lookup for the factory-default `anthropic/claude-opus-4-5` misses, while `litellm.get_model_info` resolves it from the already-loaded model map as 200000 tokens. This revision's `allow_fetch=False` path returns `None`, so `ModelBinding.context_window` returns the 65536 fallback instead. The installed LiteLLM implementation confirms that the general branch performs in-memory key normalization; only Hugging Face and unknown Ollama models take the provider-backed branches.

The unconditional `if not allow_fetch: continue` therefore must be narrowed so the two network-capable families stay offline without suppressing free normalized metadata for ordinary shipped models. The existing inline thread at comment 4028156823 already captures this defect, so I am not duplicating it with another inline finding.

I covered the current diff, the merge delta and relevant history, `ModelBinding` and provider callers, LiteLLM's pinned implementation, backward compatibility, test integrity, the Token Rates architecture contract in `CONTEXT.md`, and the applicable repository rules. The private keyword rename remains compatible within the repository, but the regression test only exercises an unknown Hugging Face id and does not cover a normalized table miss such as the default Anthropic model.

Verification:
- Direct reproduction on `fbd075c60892`: `resolve_context_window(..., allow_fetch=False)` returned `None`, the fetching path returned `200000`, and `ModelBinding.context_window` returned `65536`.
- `uv run --frozen --python 3.12 --all-extras pytest -q tests/test_provider_rates.py tests/test_agent_loop_lazy_provider.py tests/test_litellm_setup.py tests/test_provider_resolution_invariants.py`: 492 passed, 96 skipped; this demonstrates the regression is currently uncovered.
- `uv run --frozen --python 3.12 --all-extras pytest -q`: 23126 passed, 120 skipped, 1 unrelated failure in the untouched `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`.

raven/providers/rates.py
Handsome-wzw: Fixed in da45d8bf. You were both right, and the measurement was the convincing part: the premise I wrote -- that the ask behind the table is a request to the provider -- holds for exactly the two families you named and for nothing else. I read the installed `_get_model_info_helper` to be sure before changing anything: the `huggingface` branch fetches `config.json` from huggingface.co, the `ollama`/`ollama_chat` branch asks the local server when the id is not in `model_cost`, and every other id goes through five in-memory key normalisations against that same table and returns. The exact-key read in `_table_entry` misses many of those spellings, so the ask was what answered them, and it was free. The unconditional skip discarded that.

The guard is now `if not allow_fetch and _ask_leaves_memory(candidate)`, in both tiers, with the two families named once beside the table they complement (`_ASK_LEAVES_MEMORY = ("huggingface/", "ollama/", "ollama_chat/")`). The docstrings that carried the wrong claim are rewritten to the narrowed one, including the honest edge: an ollama id `model_cost` knows only under a stripped spelling is skipped with the family, and the caller keeps its default -- the conservative side of that line.

Your table, on this head, litellm already imported, every outbound seam trapped to raise:

| `ModelBinding(...).context_window` | before | after |
| --- | --- | --- |
| `anthropic/claude-opus-4-5` | 65536 | **200000** |
| `anthropic/claude-opus-4-6` | 65536 | **1000000** |
| `anthropic/claude-opus-4-7` | 65536 | **1000000** |
| outbound attempts | 0 | 0 |

And the promise this PR makes is still kept: every id shipped in `provider-models.json` (2194 after my de-duplication -- your 2798 counted curated and overrides separately) driven through both no-fetch paths with `socket.getaddrinfo`, `socket.socket.connect`, `httpx.get/post/request/stream` and `httpx.Client.send` trapped: **0** outbound attempts.

Two tests now pin the two edges. `test_allow_fetch_false_still_reads_the_normalised_table_for_free` asserts as a precondition that the exact-key read misses `anthropic/claude-opus-4-5`, then that the no-fetch arm equals the fetching arm for both tiers with every seam trapped -- the normalised table miss you said was uncovered. `test_allow_fetch_false_does_not_ask_litellm_where_the_ask_leaves_memory` records every `get_model_info` call for the three unknown family ids and asserts none of the asks leaves memory; it deliberately allows the `openrouter/huggingface/...` alias candidate through, because OpenRouter is neither family and that ask is answered in memory -- my first draft forbade it and was wrong.

Mutation-checked both ways: restoring the unconditional skip fails exactly the first test; making the predicate always false fails exactly the second (1 failed, 77 passed each; 78 passed unmutated). Your four suites: 493 passed, 96 skipped. Ruff clean.

gloryfromca (commented):

The revised guard fixes the context-window regression: ordinary candidates still reach LiteLLM's in-memory normalization, while `huggingface/`, `ollama/`, and `ollama_chat/` candidates are skipped on the no-fetch path before their provider-backed branches. On this head, both `resolve_context_window(..., allow_fetch=False)` and `ModelBinding.context_window` return 200000 for the factory-default `anthropic/claude-opus-4-5`.

I reviewed the delta and full PR diff, candidate remapping, the pinned LiteLLM helper and provider-info paths, callers and relevant history, backward compatibility, test integrity, the Token Rates architecture contract, and the applicable repository rules. I found no new concrete failure. The existing review thread was opened by another reviewer, so I have not resolved it.

Verification:
- `uv run --frozen --python 3.12 --all-extras pytest -q tests/test_provider_rates.py tests/test_agent_loop_lazy_provider.py tests/test_litellm_setup.py tests/test_provider_resolution_invariants.py`: 493 passed, 96 skipped.
- Ruff check and format checks passed for all changed files.
- The source-language gate and `git diff --check` passed.

PR #451, opened 2026-09-16
fix(raven-design): deck lane on the engine's modules, defaults first, one .pptx delivered

Tchen-data:
## Summary

The hand-built deck lane (Raven-Design at medium and high) and the deck engine lane (Raven-PPT at max) as one product, measured on live runs.

- image_search for the deck lanes sits behind tools.web.search.images, off by default and switched on by the design launcher; the readiness probe for the deck route reads the target lane's own picture vendor; the build-interpreter shim has a .cmd twin for native Windows.
- Deck route on Raven-Design: page furniture comes from the engine's helper modules (ppt_layout, ppt_theme, ppt_icons, ppt_shapes, ppt_charts) written beside the build script; diagrams, picture treatment and cover/section compositions are drawn by hand. Five recorded runs had each rewritten the same 48 helpers.
- Deck route keeps no anti-pattern ledger, contract or task state: nine audited runs wrote the ledger 18 times and read it back never, and every real fix came from reading a render. The editorial skill is read alongside the deck skill; every page is read back from a render at 90 dpi or more before delivery.
- Defaults before questions: language of the request, general audience, about 20 pages, light ground, memory first; ask at most once. Two of three ask_user calls in the audit timed out unanswered.
- Formulas as pictures (matplotlib mathtext, dfrac promotion, serif scale) and inline sub/superscript runs, shared by both lanes; image_search for both lanes keyed by the selected search vendor; the raven-python interpreter shim (sh wrapper, not a symlink); all three tiers at 400 iterations.
- layouts.md gains nine compositions measured off the packaged templates' reference pages; three rules ask for full components, a hierarchy among them, and the whole page used.
- The max lane delivers the .pptx alone (D64); the PDF render stays the engine's own preview.
- ppt-engine: template gates silenced on template pages, borrowed template files kept apart from borrowed page numbers, the icon pen follows the icon's side.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
uv run --frozen pytest -q -x tests/test_agents_design_launcher.py tests/test_subagent_routing_backend.py tests/test_ppt_engine_build_tool.py tests/test_ppt_engine_build_stage.py tests/test_ppt_engine_assets_formulas.py tests/test_ppt_engine_gates_registry.py
# 299 passed
```

Live runs on a test stack (medium tier, glm-5.3-flash, the eight-character request "make a deck introducing computer vision"): with the editorial skill read, 20 pages, 18 of them carrying a searched picture or a paper figure, delivered in 26 minutes with no question asked, no task-state call, no ledger, no hand-written helper, renders at 90 dpi.

## Risk

User-visible: the max lane no longer delivers a PDF beside the deck; the design lane asks fewer questions; the host loop offers image_search only where tools.web.search.images is on. Rollback: revert the commits touching agents/raven-design, raven/agent and plugins-dist/ppt-engine/raven_ppt/stages/build.py; no data migration is involved.

## Related Issues

N/A


raven/agent/loop/wiring.py
```diff
@@ -309,29 +315,30 @@ def _lane_generates_images(self, target: str) -> bool:
...
-            or os.environ.get("SERPER_API_KEY", "")
-        )
+        def key_for(vendor: str) -> str:
+            return (
+                product_secret(target, f"{vendor.upper()}_API_KEY")
+                or self._live_vendor_key(vendor)
+                or os.environ.get(SEARCH_PROVIDERS[vendor].env_var, "")
+            )
+
+        return bool(key_for(image_search_vendor(self.web_search_provider, key_for)))
```
gloryfromca: Blocking: keep routed-lane readiness aligned with the target's actual image-search backend.

This now admits the Raven-PPT max lane when the selected host provider is Tavily and only TAVILY_API_KEY is available. Raven-PPT explicitly disables the generic image_search tool and its ppt_image_search plugin still calls Serper with SERPER_API_KEY; the launcher also bridges only PPT_SERPER_API_KEY. Consequently routing succeeds but the child has no usable picture-search tool, so an ordinary Tavily-configured deck request enters a lane that cannot satisfy its declared requirement. Please base this route check on Raven-PPT's real credential/tool contract (or make that target support the selected provider).

agents/raven-design/run.py
```diff
@@ -58,6 +65,22 @@ def state_root() -> Path:
...
+    pointing at the old one would fail exactly the way `python3` does.
+    """
+    bin_dir = root / "bin"
+    bin_dir.mkdir(parents=True, exist_ok=True)
+    shim = bin_dir / INTERPRETER_SHIM
+    staged = bin_dir / f".{INTERPRETER_SHIM}.{os.getpid()}"
+    staged.write_text(f'#!/bin/sh\nexec {shlex.quote(sys.executable)} "$@"\n', encoding="utf-8")
+    staged.chmod(0o755)
+    os.replace(staged, shim)
+    return bin_dir
```
gloryfromca, line 88: Blocking: make the new build-interpreter shim executable on native Windows.

The staged raven-python file is an extensionless POSIX #!/bin/sh script. The direct executor uses the platform shell, so on the repository's supported native Windows path cmd.exe cannot resolve or execute this file via the required plain raven-python command. Medium/high hand-built deck runs therefore fail at their first interpreter invocation. Please provide a Windows-runnable shim as well as the POSIX one (and cover that path in the launcher test).

gloryfromca (commented):
Blocking: two portability and routed-readiness defects must be fixed before merge.

I reviewed the full diff and relevant surrounding code, callers, route architecture, and commit history. I also checked the repository rules and domain terminology, backward compatibility, and whether tests were weakened; I found no weakened tests or separate rule violations worth raising. The two concrete failures are marked inline.

Verification: the directly affected Python suite passed (958 passed, 1 skipped); the broader provider, capability, plugin, publish, and materials suite passed (233 passed); the workspace UI test passed (31 passed) and TypeScript checking passed. Ruff, diff whitespace, source-language, and large-file checks also passed. The skipped Python test requires LibreOffice.

gloryfromca (commented):
Blocking: the two existing review findings remain unaddressed.

I reviewed the new commit and rechecked the full diff, relevant callers, history, route architecture, repository rules and domain terms, backward compatibility, and test changes. The new tool-filtering entries and regenerated TUI catalogue are consistent and add no new finding, but neither existing blocked code path changed, so both open threads still apply to this revision. No tests were weakened.

Verification: the affected launcher, routing, and web-tool regression set passed (314 passed); TUI i18n tests passed (8 passed), the generated-catalogue check passed, and TUI TypeScript checking passed. Ruff, diff whitespace, source-language, and large-file checks also passed. An initial direct Vitest invocation collected no tests because it bypassed the package's required pretest build; rerunning through the declared npm test script built the dependency and passed.

raven/agent/loop/wiring.py
Tchen-data: Fixed in debc00c73. The probe now asks the target lane which vendor its own picture search speaks before falling back to the host's selection: a product whose config holds the host image_search out is read on its own tools.web.search.provider (Serper when it names none, which is what the shipped engine's ppt_image_search speaks), and the key is resolved in the lane's own order (PPT_SERPER_API_KEY, the host's Serper slot, SERPER_API_KEY). A Tavily host on Tavily's key alone no longer opens Raven-PPT; a Serper key does. Covered by test_a_lane_that_searches_through_its_own_plugin_is_read_on_that_plugins_vendor in tests/test_subagent_routed_lane_readiness.py.

agents/raven-design/run.py
Tchen-data: Fixed in debc00c73. write_interpreter_shim now writes raven-python.cmd beside the POSIX script (@echo off / quoted interpreter / %*), so the plain raven-python command resolves through PATHEXT under cmd.exe; both files are staged and replaced atomically on every launch. test_the_build_interpreter_is_a_shim_on_the_exec_path asserts the .cmd twin's bytes alongside the shell script.

raven/agent/loop/wiring.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed on debc00c73aa4: readiness now reads Raven-PPT's own Serper-backed picture-search contract, and the regression covers a Tavily-only host staying closed while a Serper-equipped host opens. This resolves my finding.

agents/raven-design/run.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed on debc00c73aa4: the launcher now emits raven-python.cmd with the quoted active interpreter and argument forwarding alongside the POSIX shim, and the launcher regression pins its exact bytes. This resolves my finding.

gloryfromca (commented):

Both prior findings are fixed in debc00c73aa4, I accepted the author's answers, and I resolved both threads. For this revision I reviewed the five-file fix delta and its callers, relevant history, repository rules and domain terms, backward compatibility on POSIX and Windows, route architecture, and test changes; I found no new issue and no weakened test.

Verification on the exact PR head: 184 affected launcher, routing, and web-tool tests passed. Ruff, diff whitespace, source-language, and large-file checks also passed.

0xKT (approved):

raven/config/schema.py
```diff
@@ -1018,6 +1018,11 @@ class WebSearchConfig(Base):
     """The Serper key on the pre-vendor layout. Still honoured, read after
     ``tools.web.providers.serper.apiKey``; new writes go to the vendor slot."""
     max_results: int = 5
+    images: bool = False
```
gloryfromca, line 1021: Blocking: make the new image-search opt-in part of the capability contract.

With the new default images=false and a valid Serper key, the runtime does not register image_search, but raven.agent.tools.capabilities.is_configured and is_offered still return true because they only inspect the key. As a result, raven doctor prints a green check for a tool the agent does not have, and the extension setup surface suppresses its missing-capability guidance. The existing contract tests catch the same concrete divergence: tests/test_tool_capabilities.py now has 2 failures (42 pass). Please make the capability predicates/reporting account for this opt-in and preserve the registry invariant rather than weakening those tests.

This is introduced here, occurs in an ordinary keyed default config, and leaves the operator with neither a truthful diagnostic nor a repair path from that surface.

gloryfromca (commented):
Blocking: make the new image-search opt-in agree with capability reporting and its invariant tests.

I reviewed the actual ten-file change after separating it from the branch rebase, then rechecked the full PR diff, relevant callers and history, project rules and domain terms, backward compatibility, route/tool architecture, and test changes. The prior two fixes remain valid and their threads remain resolved. I found no other issue.

This meets the late-round blocking bar: commit 6995b7843 introduces it; an ordinary default config with a search key reaches it; and an operator who sees the green capability result but lacks the tool is given no accurate state or repair path, while the repository's invariant suite has no passing path until the contract is reconciled.

Verification: the unaffected launcher, routing, web-tool, runtime, and subagent regression set passed (292 passed). The capability suite failed (42 passed, 2 failed), both on the image_search registry/reporting mismatch. Ruff, diff whitespace, source-language, and large-file checks passed.

PR #441, opened 2026-09-16
ci(*): accept the deps-dev scope and pin dependabot's commit prefix

0xKT:
## Summary

Two dependency-bump PRs (#385, #387) cannot pass commit lint, and the four
open dependabot PRs do not agree with the twelve already on main about what a
dependency bump is called. This fixes both, on the two sides that own them.

**The scope enum lost a scope it was already using.** `scope-enum` became
enforced on 2026-09-05 (8f6ef4c0) with a literal `deps` entry. But
`chore(deps-dev)` had merged to main on 2026-06-30 (b3959c9f, #24), two months
earlier. The enum swept in the sibling and missed this one, so every dependabot
dev-dependency PR fails a gate for using a scope the repo had already accepted.
`deps-dev` is restored to the enum; it is not a new scope.

**Dependabot guesses its own prefix, and it has drifted.** Nothing pinned it, so
it infers one from recent history. The 12 bumps merged so far are `chore(deps*)`;
the 4 open ones are `build(deps*)`. That split means `git log --grep` needs two
prefixes to find dependency work, and it will move again. Every ecosystem now
pins `prefix: "build"` with `include: "scope"`, which is the documented way to
get `build(deps)` / `build(deps-dev)`. `build` is what AGENTS.md section 3.1
defines for external dependencies.

**Two npm trees had no entry at all.** The repo has five npm lockfiles;
dependabot.yml configured three. `/ui-web` and `/ui-tui/packages/hermes-ink` were
missing, and both have produced real dependabot PRs (#385 is a `/ui-web` PR right
now, #57 was a hermes-ink one). Security updates reach them through the
dependency graph regardless of this file, so they were inheriting none of the
labels, grouping or commit-message settings the other trees get. Adding them is
what makes the prefix fix actually cover every tree rather than three of five.

**Neither fix had a gate behind it, so there is one now.** Both files are edited
by hand, neither imports the other, and the drift they just had stayed invisible
until a person read a CI log. `tests/test_dependabot_scope_canon.py` pins six
things: every ecosystem pins its prefix rather than letting dependabot infer one,
the pinned prefixes agree with each other, every scope the config will emit is in
the enum, scopes nobody declared are still rejected, every lockfile tracked in the
tree has an entry, and a repo that has workflows has an ecosystem for them. It reads the enum
by evaluating `commitlint.config.cjs` with node instead of matching its text,
anchored on a scope that cannot move so a broken extraction fails loudly rather
than satisfying every membership assertion.

[...]

PR #452, opened 2026-09-16
test(*): warm the provider catalogue, not only litellm

0xKT:
## Summary

The idle ceiling has failed a unit shard on nearly every pull request opened
today, `main` included, and it names a test in `tests/test_rpc_model.py` almost
every time -- but hardly ever the same one twice. #446, #448, #441 and two
pushes to `main` all died this way with every test passing.

`_warm_the_heaviest_import` in `tests/conftest.py` exists for exactly this, and
its docstring already states the mechanism: a cold read costs seconds that no
CPU accounts for, it lands on whichever test reached it first, and "a gate
cannot be held to a moving target". It warms litellm and stops there.

Measured with that hook already active, listing every file over half a megabyte
opened from inside a test in that module:

| File | Size | Before this change |
|---|---|---|
| `litellm/.../tokenizers/anthropic_tokenizer.json` | 1.7 MB | not opened in a test (the hook covers it) |
| `litellm/.../tokenizers/9b5ad71b...` | 1.6 MB | not opened in a test |
| `litellm/model_prices_and_context_window_backup.json` | 1.4 MB | not opened in a test |
| **`raven/providers/data/models.json`** | **0.6 MB** | **opened inside tests, repeatedly** |

The repetition is the registry's own design: the three files under
`raven/providers/data` sit behind `lru_cache`s that `registry_data.reset_cache()`
drops, so each is read again after every reset, and the first of those reads in
a run is the cold one. Warming them pays that read on a line that is the same
every run instead of on whichever test got there first.

After this change the same measurement finds nothing over half a megabyte opened
inside any test in that module.

**What this does not claim.** It does not claim to end the CI failures. It closes
a gap the measurement shows is real and that the hook was already written to
close; whether that gap accounts for the whole three-to-four seconds is a
question only a CI run can answer, because the page cache on a developer's
machine is warm and the effect cannot be reproduced locally at all.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [x] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

The before and after are the same measurement, an instrumented `open` recording
every file over 512 KB opened between a test's setup and teardown, run with
`--idle-ceiling-strict` so the warm hook is active:

```
before:  raven/providers/data/models.json opened inside tests, 3 times
after:   0 files over 512 KB opened inside any test in the module
```

Suites:

```
$ uv run pytest tests/test_rpc_model.py -q
195 passed in 9.77s          (same count as before the change)

$ uv run pytest tests/test_rpc_model.py tests/test_provider_catalog.py tests/test_commit_lint.py -q
287 passed, 7 skipped in 10.21s

$ uv run pytest -q --idle-ceiling-strict
1 failed, 22940 passed, 121 skipped in 223.50s
```

The single failure is
`tests/test_read_file_image.py::test_image_block_is_the_only_place_the_shape_is_written`,
which fails the same way on an unmodified `main` worktree: it shells out to
`grep` and filters on `raven/utils/images.py`, and the `grep` on this
developer's PATH emits a doubled separator, so the filter misses its own file.
CI uses GNU grep and is green on it.

## Risk

Test-harness only; nothing under `raven/` changes. The three accessors run
inside the hook's existing `try`, which already swallows exceptions on the
grounds that a tree without the provider extras is not this hook's business, so
a checkout that cannot read the catalogue still starts its suite.

The hook only runs under `--idle-ceiling-strict`, which is CI's invocation, so a
plain local `pytest` is unaffected.

Rollback is to revert the commit.

## Related Issues

N/A.


gloryfromca (commented):

I reviewed the complete diff and the surrounding provider-registry loaders, cache resets, strict idle-ceiling hook lifecycle, xdist/CI invocation, and relevant history. I also checked the repository rules and canonical provider terminology, backward compatibility and environment restoration, and that the change adds warm-up behavior without weakening tests or crossing architecture boundaries.

Verification: `uv run pytest tests/test_provider_registry_data.py tests/test_rpc_model.py --idle-ceiling-strict -q` (219 passed); `uv run pytest tests/test_provider_catalog.py tests/test_commit_lint.py --idle-ceiling-strict -q` (92 passed, 7 skipped); Ruff check and format check on `tests/conftest.py` passed; `git diff --check` passed.

arelchan (approved):

PR #457, opened 2026-09-17
fix(*): pin line endings to lf so windows checkouts pass the byte gates

xfng-sd:
## Summary

A Windows checkout of this repo with the default `core.autocrlf=true` rewrites every
text file to CRLF, and two byte-exact gates in `ui-web` then fail on a tree that is
byte-identical to the one CI passes on Linux:

- the boot-order and open-conversation node tests slice their source on `"\n}\n"`,
  which never matches a CRLF file, so the sliced function comes out empty and 9 tests
  fail with `ReferenceError`;
- `gen-rpc-client --check` compares its own LF output against the file on disk and
  reports permanent drift in `src/rpc/generated.ts`.

`.gitattributes` pins the checkout filter so the working tree matches what the
repository already stores. The repository stores LF today, so this adds no
renormalization diff: `git add --renormalize .` over the whole tree stages nothing
beyond the two files in this PR. Only the checkout filter changes.

`ui-web/build.py` is the other half of the same defect. It assembled the page under
Python's default newline translation, so `dist/index.html` came out CRLF on Windows
even from an LF source tree, and `check-page.mjs` could no longer extract the script
payloads (its `<script>\n` anchor needs the LF to be the next byte). That artifact
ships inside the wheel, so it must not depend on which platform assembled it.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run on Windows 11 with `core.autocrlf=true`, in a clean worktree cut from `main`:

- `git add --renormalize .` after the change stages nothing (confirms no
  renormalization diff);
- `file ui-web/src/rpc/generated.ts` reports CRLF on `main` and LF after a forced
  re-checkout on this branch;
- `npm run gen:check` (ui-web): `generated.ts matches the contract (178 methods)`;
- `npx vitest run` (ui-web): 108 files, 1836 tests passed;
- `npm run --prefix ui-web build && python ui-web/build.py`: `dist/index.html` is LF;
- `node ui-web/scripts/check-page.mjs`: `OK (1,482,975 bytes, 3 scripts parse)`;
- `node ui-web/scripts/count-shared-globals.mjs`, `check-css.mjs`,
  `check-class-namespace.mjs`: all OK.

## Risk

No runtime behaviour changes for users on Linux or macOS, where the checkout already
produced LF. On Windows the working tree stops being rewritten to CRLF, so a
contributor with existing local checkouts may see one large renormalization touch on
the next checkout; the committed bytes do not move. The built `dist/index.html` is now
LF on every platform, which is what the Linux-built wheel already shipped.

Rollback is deleting `.gitattributes` and dropping the `newline=""` argument in
`ui-web/build.py`.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the full `github/main...HEAD` diff at `cd5af3c6b490`. I covered the repository rules in AGENTS.md/CLAUDE.md and CONTEXT-MAP.md, the diff and relevant history, the byte-sensitive callers, tracked file types and binary handling, backward compatibility, architecture boundaries, and whether tests were weakened (they were not changed).

I reproduced a `core.autocrlf=true` checkout locally: the base produced CRLF in `generated.ts` and `install.ps1`, while this revision produced LF only. `git add --renormalize .` remained clean, and no tracked index file reports CRLF or mixed endings.

Verification passed after installing the locked UI dependencies:
- `npm run gen:check --prefix ui-web` (178 methods)
- `npm test --prefix ui-web -- --reporter=dot` (108 files, 1836 tests)
- `npm run build --prefix ui-web`
- `uv run python ui-web/build.py`
- `node ui-web/scripts/check-page.mjs`
- `node ui-web/scripts/count-shared-globals.mjs`
- `node ui-web/scripts/check-css.mjs`
- `node ui-web/scripts/check-class-namespace.mjs`
- source-language, large-file, commit-message, and diff-whitespace checks

The assembled `dist/index.html` contained 13,534 LF bytes and zero CRLF bytes. The initial generation command failed only because `ui-web/node_modules` was absent; it passed after `npm ci --prefix ui-web`.

gloryfromca (commented):

This revision is a clean rebase onto the current `main`: `git range-diff` reports the PR commit as identical to the reviewed revision, and the old and new commits have the same stable patch ID. I re-read the full `github/main...HEAD` diff and checked the repository rules, affected callers and history, backward compatibility, architecture boundaries, and test integrity; the refreshed base creates no new interaction with the unchanged two-file patch.

Fresh verification on `77cec32ec4e1` passed: RPC generation drift check, all 108 UI test files / 1,836 tests, the production UI build, all four post-build page/CSS/namespace gates, source-language check, large-file check, commit-message check, and `git diff --check`.

.gitattributes
0xKT, line 1: **Blocking: `eol=lf` turns two passing byte assertions into failing ones on Windows -- the platform this PR exists to serve. Two writers still emit CRLF, and the repo already knows the fix.

The `.gitattributes` half pins every tracked text file to LF on checkout. That is the right thing for the three consumers the comment names. But two places in the repo compare a file they **wrote** against a tracked file they **read**, byte for byte, and only the read side is being pinned:

| | writer | assertion |
|---|---|---|
| raven-code | `agents/raven-code/run.py:233` `target.write_text(wanted, encoding="utf-8")` | `tests/test_agents_code_launcher.py:160` `seeded == CARRIED_GUIDE.read_bytes()` |
| ppt-engine | `plugins-dist/ppt-engine/raven_ppt/plugin/hook.py:182` `target.write_text((prompts / name).read_text(...), encoding="utf-8")` | `tests/test_ppt_engine_plugin.py:797` `(home / seat).read_bytes() == (prompts / name).read_bytes()` |

Both sources are tracked (`agents/raven-code/plugins/code-flow/prompts/TOOLS_CODE.md`; `plugins-dist/ppt-engine/raven_ppt/prompts/{SOUL,AGENTS,TOOLS}.md`), so `eol=lf` pins them to LF. Both writers use the default `newline=None`, which translates `\n` to `os.linesep` -- CRLF on Windows. `read_text()` on the way in gives universal newlines, so the writer is the only side that changes.

### The control

Four arms in this PR's own worktree (head merged onto `3b35e7d3`), one variable at a time, run through the repo's pytest. "writer=CRLF" is `newline="\r\n"`, which is exactly what `newline=None` does on Windows:

| source on disk | writer | `test_the_seeded_guide_is_the_carried_asset` |
|---|---|---|
| LF | platform default (LF here) | **1 passed** -- today on Linux/macOS |
| CRLF | CRLF | **1 passed** -- Windows **before** this PR |
| LF | CRLF | **1 failed** -- Windows **after** this PR |
| LF | `newline=""` | **1 passed** -- with the one-keyword fix |

The second row is the one that makes this a regression rather than a pre-existing wart: with `autocrlf=true` and no `.gitattributes`, both sides were CRLF and the assertion held. Pinning only the read side is what breaks it.

(The ppt-engine pair is the same shape read off the source; I could not execute it here because the board venv has no `pptx`, so `tests/test_ppt_engine_plugin.py` skips at collection. I am reporting it as read, not as run.)

### The fix is already in this repo

`agents/raven-design/run.py:85` writes `staged.write_text(text, encoding="utf-8", newline="")` with a comment explaining the Windows reason. raven-code and the ppt-engine hook were simply missed. Adding `newline=""` at those two call sites is the whole change, and it makes this PR's promise true instead of half-true -- which is also the principle the PR states for itself in `ui-web/build.py`:

> The artifact ships in the wheel, so it must not depend on which platform assembled it either.

### Where the PR is right

- `ui-web/build.py` has exactly one writer and this PR fixed it; nothing is left half-done inside the file it touched.
- `requires-python = ">=3.12"` (`pyproject.toml:5`), so `write_text(newline=...)` is available everywhere -- no API excuse at any call site.
- The three named consumers check out: `ui-web/scripts/check-page.mjs:41` anchors on `^<script>\n`, `ui-web/src/features/composer/open-conversation.test.ts:32` does `indexOf('\n}\n', begin)`, and `ui-web/scripts/gen-rpc-client.mjs:186` compares its own `join('\n')` output against the file on disk.
- The binary list covers the tree exactly: of 3446 tracked files, 81 are binary, in seven extensions (`.jpg` 60, `.pptx` 10, `.webp` 5, `.ttf` 2, `.pdf` 2, `.png` 1, `.xlsx` 1) -- all seven are listed. The 91 tracked `.svg` files are text and being normalized is correct for them.
- There is no CRLF churn to worry about: `git ls-files --eol` reports 3337 `i/lf`, 81 `i/-text`, 27 `i/none`, zero `i/crlf` and zero `i/mixed`. No renormalization commit is needed because there is nothing to renormalize.
- Nothing in the tree requires CRLF. The only Windows script, `install.ps1`, is already stored and checked out as LF; there is no `.bat` or `.cmd` tracked, and the one `.cmd` the repo produces is generated at runtime by `agents/raven-design/run.py:82` with literal `\r\n` and `newline=""`, which `.gitattributes` cannot reach.

gloryfromca: Blocking: the two tracked-prompt writers must emit LF before this can merge.

I independently confirmed the mechanism. All four tracked prompt sources are LF under the new attributes. Writing their universal-newline text with Windows-equivalent `newline="\r\n"` makes each destination differ byte-for-byte from its source; writing with `newline=""` restores equality. The raven-code assertion passes on this Linux host; the ppt-engine test could not collect here because `pptx` is unavailable, but its source/writer/assertion path reproduces the same mismatch directly.

This evidence changes my standing stance from clean to blocker.

gloryfromca (commented):
Blocking: the two tracked-prompt writers must emit LF before this can merge.

The new review thread identifies a Windows regression that invalidates my earlier clean stance. `.gitattributes` pins the carried prompt assets to LF, but `agents/raven-code/run.py` and `plugins-dist/ppt-engine/raven_ppt/plugin/hook.py` still write their copies with platform-default newline translation. On Windows, the resulting CRLF files fail the existing byte-equality contracts against the LF sources. Direct reproduction confirmed the mismatch for the raven-code guide and all three ppt-engine identity prompts; explicit no-translation writing preserves equality.

Relevant test result: `uv run pytest tests/test_agents_code_launcher.py::test_the_seeded_guide_is_the_carried_asset -q -o addopts=""` passed on Linux. The matching ppt-engine test was skipped at module import and the node could not be collected because this environment lacks `pptx`; I verified that path by reproducing its exact read/write byte transformation instead.

The existing open thread carries the concrete locations, so I added no duplicate inline finding and left that reviewer-owned thread unresolved.

.gitattributes
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed on `e0959298a149`: both tracked-prompt writers now pass `newline=""`, so the copied bytes no longer depend on the host platform. With the declared development extra installed, the complete relevant modules pass together: 122 tests in `test_agents_code_launcher.py` and `test_ppt_engine_plugin.py`.

My blocker is closed. I am leaving the thread itself for its originating reviewer to resolve.

gloryfromca (commented):

The new revision fixes the Windows regression at both identified writers with explicit no-translation writes. I reviewed the delta and full `github/main...HEAD` diff, the repository rules, affected callers and history, equivalent text-copy paths, backward compatibility, architecture boundaries, and test integrity; no tests were weakened or removed, and I found no remaining concrete regression from the LF checkout policy.

Verification on `e0959298a149` passed:
- `uv run pytest tests/test_agents_code_launcher.py tests/test_ppt_engine_plugin.py -q -o addopts=""` (122 passed)
- Ruff check and format checks for all changed Python files
- source-language, large-file, commit-message, and diff-whitespace checks

I closed my blocking reply in the existing thread. The thread was opened by another reviewer, so I left resolution to them.

.gitattributes
xfng-sd: Fixed at both sites on `e0959298a149`: `newline=""` on the raven-code seed writer and on the ppt-engine hook, matching `agents/raven-design/run.py:85`, which already had it. Your reading of the mechanism was exactly right, and so was the call that pinning only the read side is what turns the accident into a failure.

I could run the arm you could not. This is a Windows 11 host with `core.autocrlf=true` and `pptx` present, so `tests/test_ppt_engine_plugin.py` collects here:

| revision | source on disk | `test_the_first_turn_seeds_the_deck_identity_into_the_granted_home` |
|---|---|---|
| `main` | CRLF | 4 passed |
| this branch, before the fix | LF | **1 failed** -- `At index 6 diff: b'\r' != b'\n'` |
| this branch, after the fix | LF | 4 passed |

That is your second and third rows, observed rather than modelled, and it confirms the regression framing: the assertion held on Windows before this PR only because both sides were CRLF.

One thing I did **not** verify, so it should not be read as verified. The raven-code half does not execute on this host: `test_the_seeded_guide_is_the_carried_asset` dies before reaching the assertion, in `raven/config/product_render.py:476`, on `os.fchmod`, which does not exist on Windows. I changed that writer on the strength of the identical mechanism and your Linux run, not on my own observation.

For siblings of the same shape, I diffed the failure sets of both affected files between `main` and this branch on Windows: 45 failures on each, none new and none fixed. All 45 are that same `os.fchmod` gap, which is present on `main` and untouched here.

Worth separating out: 43 of `tests/test_agents_code_launcher.py` fail on Windows on `main` for that `os.fchmod` reason. It is out of scope for this PR and wants its own change, but it does mean the raven-code byte contract has no Windows coverage today, which is the same blind spot that let this one through.

On the branch history: `df1d08e2` landed with a broken comment that split a line and left `agents/raven-code/run.py` unparseable -- a heredoc turned an escaped newline into a real one, and I had masked ruff's exit code behind a pipe. `e0959298` repairs it, verified with unmasked exit codes. The repo squash-merges, so the pair collapses into one commit.

0xKT:
**Not a blocker -- the blocking finding is fixed at both call sites and I re-verified it. One leftover is a sentence, not code.

### The fix lands

`agents/raven-code/run.py:236` and `plugins-dist/ppt-engine/raven_ppt/plugin/hook.py:185` both now pass `newline=""`, and `python -m py_compile` is clean on both (the indentation the first comment broke is repaired by `e0959298`).

Re-run on the new head merged onto `4b390228`, one mutation at a time, tree restored:

| | `test_the_seeded_guide_is_the_carried_asset` |
|---|---|
| as shipped | **1 passed** |
| withdraw `newline=""` entirely (the pre-fix state) | **1 passed** |
| pin the writer to `newline="\r\n"` (what the default does on Windows) | **1 failed** |

The middle row is the interesting one, and it is not a gap you can close: on Linux `os.linesep == "\n"`, so `newline=None` and `newline=""` write identical bytes and the mutant is *equivalent on this platform*. That is precisely why the original defect was invisible to CI, and it is a property of the runner, not of your tests. The third row shows the assertion does catch the failure mode wherever it can actually occur.

`rerun_scope.py` against my previously graded head: 2 commits, 2 files, both of them the ones I named; nothing under `raven/` changed, so the cross-cutting answers stand as they were.

### One sentence in `.gitattributes` still points at the wrong place

The comment says:

> ui-web's boot-order and open-conversation tests slice their source on `"\n}\n"`

Only half of that is true, and I checked each half:

- `ui-web/src/features/composer/open-conversation.test.ts:32` really does `source.indexOf('\n}\n', begin)`. Correct.
- `ui-web/scripts/boot-order.test.mjs` *is* LF-fragile, but at `:77`, and its anchor is `'queueMicrotask(bootPage);\n\n})();'` -- not `"\n}\n"`. `grep -F '\n}\n'` over all of `ui-web/` returns exactly one hit, and it is the open-conversation line. The file's other two LF-mentioning expressions (`:9`, `:73`) are CRLF-tolerant.
- The actual second `"\n}\n"` slicer is `tests/test_ui_language_repaint.py:53` (`end = live.index("\n}\n", start)`), which the comment does not mention.

So a reader who follows the comment into `boot-order.test.mjs` looking for `"\n}\n"` finds nothing, and the file that really depends on that anchor goes unnamed. The conclusion the comment draws is still right -- a CRLF checkout does break `boot-order.test.mjs`, so pinning `eol=lf` is justified -- it is the stated mechanism that is off. Naming the two anchors separately would fix it.

Nothing else changed since my earlier review, and everything I confirmed in the PR's favour then still holds: the binary list covers all 81 binary files across seven extensions, `git ls-files --eol` reports zero `i/crlf` and zero `i/mixed` so there is no renormalisation to do, and nothing in the tree requires CRLF.

gloryfromca (approved):

PR #464, opened 2026-09-17
fix(skill_hub): a hub skill is removable whichever installer put it there

### LINKED ISSUE #456: A market-installed skill has no uninstall control in the web UI: ext.list recognises only one of two install markers
## Summary

A skill installed through the skill market has no uninstall control in the web UI, and never had one: the page decides whether a skill can be removed by looking for a marker file that this install path does not write. Two independent installers exist and each stamps its own marker; `ext.list` recognises only one of them.

## Steps to reproduce

1. Install a skill from the skill market (the ones in this report arrived as `ppt-personal-intro`, `html2pptx` and `resume-master`).
2. Open the Skills page and switch to the installed view.
3. Look at the card, and at the drawer behind it.

## Expected behavior

A hub-installed skill offers removal, on the card since #396 and in the drawer before it. It has a bundle on disk, and `raven skill remove` deletes it, so the affordance is not in question -- only its reachability from the page.

## Actual behavior

No uninstall control appears, on the card or in the drawer. The cards are labelled `hub`, and they are the ones that cannot be removed; the only card that offers removal is labelled `workspace`.

Both controls are guarded by the same field:

- card: `{c.hub ? <ArmRemove inline fn={() => store.removeInstalled(c)} /> : null}` (`ui-web/src/features/skills/SkillsPage.tsx:299`)
- drawer: `manage(inst?.hub, ...)` (`ui-web/src/features/skills/SkillsPage.tsx:398`)

and `hub` is filled by the presence of one file:

```python
# raven/rpc/methods/console.py:112
"hub": marker is not None and marker.is_file(),
```

where `marker` is `<skill dir>/.skillhub.json` (`raven.skill_hub.hub.MARKER`).

The label under the name is a different field (`src`, from `m.source`), which is why the page can say `hub` while `c.hub` is false.

## The two installers

| | A | B |
| --- | --- | --- |
| module | `raven/skill_hub/hub.py` | `raven/agent/tools/skill_hub.py`, `raven/context_engine/segments/skills.py` |
| lands at | `<workspace>/skills/<name>/` | `<workspace>/skills/hub/<slug>@<version>/<name>/` |
| stamps | `.skillhub.json` (`MARKER`, written at `hub.py:580`) | `.install-meta.json` (`write_install_meta`, `skill_hub/audit.py:75`) |
| `ext.list` reports `hub` | true | **false** |
| removable from the page | yes | **no** |
| removable from the CLI | -- | `raven skill remove <slug>` (`cli/skill_commands.py:157`) |

Measured on one machine, four installed skills:

```
<workspace>/skills/hub/Mr-Q526_PPTMaker-skill_ppt-personal-intro@v0/ppt-personal-intro
  .install-meta.json  SKILL.md  agents  references
<workspace>/skills/hub/openclaw_skills_html2pptx@v0/html2pptx
  .install-meta.json  README.md  SKILL.md  package.json  src  tests
<workspace>/skills/hub/openclaw_skills_resume-master@v0/resume-master
  .install-meta.json  SKILL.md  agents  assets  scripts
<workspace>/skills/openclaw-secret-scanning-maintainer
  .skillhub.json  SKILL.md  scripts
```

`.skillhub.json` is absent from the three market skills at both the bundle level and the skill level. The fourth, the only one the page offers to remove, is the only one carrying it.

## Why it is worth more than it looks

The market is the ordinary way to install a skill, so this is not an edge case: the affordance is missing for most installed skills, and present for the exception. #396 moved the control from the drawer onto the card precisely because "a reader who wanted the skill gone went to the CLI"; for these skills the control does not render in either place, so that PR could not have fixed what it set out to fix for them. Two defects stacked, and the outer one hid the inner one.

Nothing here is a data-loss risk -- the skills work, and the CLI removes them -- but a page that silently offers an action for one install path and not another reads as "hub skills cannot be uninstalled", which is not true.

## A direction, for whoever owns this

Teaching `ext.list` to accept either marker is the small fix and would make the control appear. The larger question is whether two installers should keep two markers and two layouts at all, and that is not mine to answer -- `raven/skill_hub/` and `raven/agent/tools/skill_hub.py` have owners who will know whether the split is deliberate.

Worth deciding at the same time: `skillhub.remove` (what the page's button calls) removes by name through installer A's path. If the page starts offering removal for B's bundles, that RPC has to reach them too, or the button will appear and fail.

## Environment

macOS, `raven` from the repo working tree (editable install), web UI served from `ui-web/dist/index.html` built at the current `main`.

Handsome-wzw:
## What

A skill installed through the market had no way to be removed from the page, because `ext.list` decided "was this installed from the hub" by looking for one marker file, and the installer most skills arrive by writes a different one. Closes #456.

Two installers put hub skills on disk:

| | market module (`raven/skill_hub/hub.py`) | context engine + `use_skill` (`raven/agent/tools/skill_hub.py`, `raven/context_engine/segments/skills.py`) |
| --- | --- | --- |
| lands at | `<skills>/<name>/` | `<skills>/hub/<slug>@<version>/<name>/` |
| stamps | `.skillhub.json` | `.install-meta.json` |

`ext.list` tested only for the first, so the second -- the ordinary path -- reported `hub: false`. The page gates its removal control on that flag, so it appeared for the exception and never for the rule. On the machine this was found on, three of four installed skills were the second kind; the one that showed a remove button was the one that came through the market module.

## How

Backend only; nothing under `ui-web/` or `ui-tui/` changes. The page already renders the control when `hub` is true, and already skips the market-detail fetch when `hub_id` is empty, so both halves light up on their own once the flag is right.

- `raven/skill_hub/audit.py` names the stamp it writes (`INSTALL_META`) so the two readers test for the file the installer wrote rather than a spelling of their own.
- `raven/rpc/methods/console.py` reports `hub: true` when either stamp is present. The bundle stamp carries a slug, not a market id, so that skill has an empty `hub_id` -- a state the page already handles.
- `raven/skill_hub/hub.py` lets `skillhub.remove` find a bundle by the skill name inside it, resolving the wrapper folder the same way `SkillHubClient._bundle_root` did at install time, and removes the bundle whole: the bundle is the install unit, and the CLI's `skill remove` deletes the same folder. Only a stamped directory is eligible -- a folder placed under `hub/` by hand is left alone, which is the rule `MARKER` already enforced on the other layout.

The second change is not optional given the first: reporting `hub: true` without it would put a remove button on a skill the handler then refuses, which is worse than no button.

## Verification

- `test_ext_list_calls_a_skill_hub_installed_whichever_installer_stamped_it`: three skills, one per stamp and one with neither, asserting `(hub, hub_id)` for each -- `(True, "uuid-market")`, `(True, "")`, `(False, "")`. The marker names are pinned as literals in the test so a drift between writer and reader is a failure, not a pass.
- `test_remove_reaches_a_bundle_the_other_installer_cached`: a stamped bundle is removed whole by its skill name; an unstamped folder under `hub/` is refused and left in place, and so is the neighbouring bundle.
- Mutation-checked: dropping the second stamp from `ext.list` fails exactly the first test; dropping the bundle lookup from `remove` fails exactly the second. 139 passed in the two suites on the unmutated tree.
- Dry-run against the real workspace: all four installed skills now report `hub: true`; the three bundles resolve to their `<slug>@v0` folders by skill name; the market-module skill and an unknown name resolve to none.
- Ruff check and format clean.

## Not done here, on purpose

Removal does not stop the context engine re-installing the skill on its next catalogue hit; the CLI documents the same limit and points at `skill block`. Whether the page should offer that beside removal is a UI question and out of scope while `ui-web/` is frozen. Also left as they were: the two installers keeping two layouts and two stamps -- this makes the readers agree with both writers rather than deciding which writer should win.

Co-authored-by: Claude (claude-fable-5-1) <noreply@anthropic.com>


raven/skill_hub/hub.py
```diff
@@ -624,18 +625,58 @@ def _write() -> tuple[list[str], list[str], bool]:
...
+    already enforces on the other layout.
+    """
+    hub_root = root / "hub"
+    try:
+        bundles = [d for d in hub_root.iterdir() if d.is_dir() and "@" in d.name]
+    except OSError:
+        return None
+    for bundle in bundles:
+        skill_dir = SkillHubClient._bundle_root(bundle)
+        if skill_dir.name == name and (skill_dir / INSTALL_META).is_file():
```
gloryfromca: Blocking: flat bundles are advertised as removable but the removal lookup cannot find them.

`SkillHubClient._bundle_root` explicitly supports a flat ZIP by returning the bundle directory itself. In that case `skill_dir.name` is `<slug>@<version>`, while `ext.list` sends the catalog/frontmatter skill name to `skillhub.remove`, so this comparison can never match. I reproduced this with `hub/catalog-slug@v1/SKILL.md` containing `name: display-name` and a valid `.install-meta.json`: the catalog lists `display-name` as Hub-installed, `_bundle_for(..., "display-name")` returns `None`, and removal raises `SkillHubRequestError` while the bundle remains. Please identify the bundle from its stamp/path in a way that also covers the already-supported flat layout, and add the flat case to the removal test.

gloryfromca (commented):
Blocking: flat Hub bundles need removal support before this can merge.

I found one functional blocker. I reviewed the full diff, the bundle writer and both installer call sites, catalog discovery, the CLI and web removal callers, relevant history, backward compatibility, and the repository rules in AGENTS.md/CLAUDE.md plus CONTEXT-MAP.md. I also checked that the tests were additive rather than weakened.

Verification: `uv run pytest tests/test_rpc_console.py tests/test_rpc_skillhub.py tests/test_skill_hub_client.py tests/test_cli_skill_commands.py -x` passed all 168 tests; `uv run ruff check raven/rpc/methods/console.py raven/skill_hub/audit.py raven/skill_hub/hub.py tests/test_rpc_console.py tests/test_rpc_skillhub.py` and `git diff --check github/main...HEAD` passed. A focused reproduction using `LocalSkillCatalog` demonstrates the inline failure. The head was force-updated during review, but the current and prior commits have the identical tree hash, so those results exercise the current source exactly.

Handsome-wzw:
The three red checks on this PR are `main`'s, not this change's. Recording the evidence so nobody has to re-derive it:

- `python lint` fails in the **Type check** step on `raven/agent/tools/web.py:516`; `unit 2/4` fails two tests in `tests/test_config_update_tools.py` (an extra `images` key in the `web_search.set` shape); `unit 3/4` fails two tests in `tests/test_tool_capabilities.py` (`image_search` declared but not gated).
- None of those files is touched here. All three are byte-identical to `upstream/main` on this branch (`git diff --quiet upstream/main HEAD -- <file>` for each).
- `main`'s own latest CI run, on `5a29ea22`, is red on exactly the same three jobs.
- All of it arrived with #451, which added `image_search` to `web.py` and `schema.py` without updating the three gates that pin those shapes.

The fix is already up in #463 and covers all three. Once it lands I will rebase this branch and the remaining checks should clear on their own; the nine checks this PR does own are green, including `commit messages` after the scope fix (`skillhub` -> `skill_hub`, applied to both the commit and the title, since the gate lints the prospective squash header too).

raven/skill_hub/hub.py
Handsome-wzw: Fixed in 61003547. You were right, and the repro was the useful part: I had matched on the directory name, which fits the wrapped layout and can never fit the flat one, while `ext.list` reads the stamp beside `SKILL.md` and reports both -- so a flat bundle was advertised as removable and refused on the click, the exact failure the second half of this PR exists to prevent.

The lookup no longer looks at any folder name. It resolves the name through the skill table -- the running loop's, which is the one `ext.list` read when it drew the name, else a cold `LocalSkillCatalog` over the same workspace, built the way `ext.list` builds its own fallback -- takes `SkillMeta.path`, and accepts the `hub/<slug>@<version>` ancestor of that path when the stamp is beside it. Both surfaces now answer "which directory is this name" from one source.

Your repro is in the test verbatim: `hub/catalog-slug@v1/SKILL.md` with `name: display-name` and a stamp, alongside the wrapped bundle and the unstamped folder. `remove("display-name")` deletes `catalog-slug@v1`; the wrapped and hand-placed cases behave as before.

Mutation-checked: putting the directory-name comparison back fails that test and only that test (1 failed, 138 passed); the unmutated suites are 139 passed. Against the real workspace this was found on, the cold-table path resolves the three flat-or-wrapped bundles by skill name and returns None for the market-module skill and for an unknown name. Ruff clean; the wider `skill/console/hub/catalog` sweep is 1278 passed.

The three red CI jobs remain `main`'s (#451 fallout, fix in flight in #463), as noted in the PR comment.

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed at `61003547d0d4`: the lookup now derives the bundle from the catalog path, so the original flat-layout reproduction succeeds, while resolved-root containment and the adjacent install stamp still prevent local or hand-placed content from being selected. The exact-head RPC/console/client/CLI sweep passed all 168 tests, and Ruff passed. I accept the fix and am resolving this thread.

gloryfromca (commented):

The previous flat-bundle blocker is fixed: removal now resolves the displayed skill name through the catalog path and retains the Hub-root containment and install-stamp checks. I re-reviewed the full change, revision delta, writers, catalog, RPC and web callers, backward compatibility, tests, repository rules, and the Runtime/UI RPC boundary; I found no new plain error or blocker. The tests remain additive rather than weakened.

Exact-head verification in a disposable checkout: `uv run --offline pytest tests/test_rpc_console.py tests/test_rpc_skillhub.py tests/test_skill_hub_client.py tests/test_cli_skill_commands.py -x` passed 168 tests; Ruff and `git diff --check` passed. The unrelated required-check failures inherited from `main` remain an external merge condition, not a defect in this patch.

Risk assessment: merge / human review required. Impact is moderate because this path deletes persisted bundle directories; regression likelihood is low, protection is strong, recovery is managed by reinstall/revert, and confidence is high.

gloryfromca (commented):

This head only merges current `main`; the PR's own five-file diff is unchanged from the clean revision, and the merge includes the separate fix for the inherited CI failures. I rechecked the full comparison and confirmed my prior thread remains resolved.

Exact-head verification after the merge: `uv run pytest tests/test_rpc_console.py tests/test_rpc_skillhub.py tests/test_skill_hub_client.py tests/test_cli_skill_commands.py -x` passed 168 tests; Ruff and `git diff --check github/main...HEAD` passed.

Handsome-wzw:
The remaining red, `coverage gates`, was mine: the PR diff coverage step runs only once every shard has produced `coverage.json`, so it stayed skipped while `main`'s reds were inherited and appeared when they cleared. It read 72.13% (44/61) against the 90% threshold, and every uncovered line was in this PR -- the loop-factory branch of `_catalogue` (the path production takes on every removal), its fallbacks, and the skip arms of `_bundle_for`. None of my tests passed a loop factory, so the production path was never exercised. That is a real gap the gate caught, not a padding exercise.

3dc2b6ca adds seven tests, each for a contract: the running loop's table is preferred (proven by making the cold table unbuildable, so only the loop can have answered); a loop that raises or has no table falls through to the cold one; no table at all refuses rather than guessing from a folder name; an enumeration that raises reads as empty; and a row with no path, a row in the market module's own layout, and an unstamped row under `hub/` are each passed over. Plus the ImportError arm of `_install_meta_name`, on the same terms as the existing test for the first stamp.

Reproduced the gate locally with the same script and threshold (`coverage_gate.py diff --base-ref main --threshold 90`) over these two test files: 98.36% (60/61). The one line left is `write_install_meta`'s body, which these two files do not drive and CI's full run does -- it was not on CI's uncovered list -- so the CI number should read 61/61. Mutation-checked the key test: making `_catalogue` ignore the factory fails the loop-first test and the two that depend on the loop's rows (3 failed, 48 passed); 146 passed unmutated.

gloryfromca (commented):

This revision is tests-only and closes the PR-owned coverage gap without changing runtime behavior. I checked that the new cases distinguish the running-loop catalog, cold fallbacks, fail-closed paths, enumeration failure, and rejected row shapes rather than merely executing lines. The prior resolved finding remains fixed.

Exact-head verification: the RPC/console/client/CLI sweep passed 175 tests; the two changed suites passed 146 tests with coverage; `coverage_gate.py diff --base-ref github/main --threshold 90` passed at 98.36% (60/61); Ruff and `git diff --check` passed.

0xKT:
**Not a blocker -- the removal this PR set out to fix does work, and I verified it. Three things stand, none of them holds the merge button. Two are one-line shapes; the third is a word.

### 1. `remove()` sanitises the wire name with a path sanitiser, then uses it as a table key

`raven/skill_hub/hub.py:709` does `name = _safe_name(skill_name)`. That is right for the branch below it, which builds a filesystem path from the name. It is not right for the bundle branch at `:719`, which passes the same sanitised string to `_bundle_for`, where `:684` compares it against the registry's raw `meta.name`:

```python
if getattr(meta, "name", None) != name:
    continue
```

`_safe_name` (`:290-293`) keeps only `[A-Za-z0-9-_.]`, so the two sides can never match for a skill whose name carries anything else. A flat hub zip with no `name:` in its SKILL.md falls back to the directory name (`registry.py:463`), and that directory is `<slug>@<version>` (`client.py:259`) -- the `@` is stripped, so the lookup misses. A space does it too.

What makes this worth a note rather than a shrug: the same PR flips `ext.list`'s `hub` flag from "has a MARKER file" to `from_market or from_bundle` (`console.py`), and `raven/rpc/models.py:2807` documents that field as

> `Installed from the skill hub, so skillhub.remove can uninstall it.`

`ui-web/src/features/skills/SkillsPage.tsx:299` gates the Uninstall control on `c.hub` alone. So for this class of name the page now offers a button that always fails, where before it showed none.

Fix shape: pass `skill_name` raw to `_bundle_for`, keep `_safe_name` on the path branch.

I could not demonstrate that the hub actually serves such a zip -- every SKILL.md in this tree declares a name that survives `_safe_name` -- which is exactly why this is not a blocker.

### 2. The rule this PR writes down has a second decider that does not honour it

`hub.py:670-673`, new in this PR:

> Only a stamped directory counts. A folder someone placed under `hub/` by hand carries no stamp and is left alone, which is the rule `MARKER` already enforces on the other layout.

`raven/cli/skill_commands.py:171`, untouched, decides the same question differently:

```python
matches = [d for d in hub_dir.glob("*@*") if d.is_dir() and d.name.rsplit("@", 1)[0].casefold() == name.casefold()]
...
shutil.rmtree(d)
```

It keys on the directory name split at `@`, matches case-insensitively, and never looks at a stamp. So a hand-placed `hub/foo@v1`: `skillhub.remove` protects it, by design and by this PR's own words; `raven skill remove foo` deletes it.

The CLI's behaviour predates this PR, so nothing is broken that was not already. What is new is the invariant being written down, and an invariant only one of two surfaces keeps is worth knowing about before it is quoted as settled.

### 3. `market` on its own

`hub.py:656` introduces "The market reaches the workspace by two installers with two layouts." `CONTEXT.md:1035-1036` lists that exact usage under `_Avoid_`:

> `_Avoid_`: "market" on its own for either one -- both surfaces are called that in prose, and the RPC groups (`plughub.*` vs `skillhub.*`) are separate.

At base the word appeared in this file only at `:4`, inside a parenthesised `raven.market.install`, so `:656` is new prose. A nit, in the one file where the ambiguity is real.

### What held up

- **The asymmetry it set out to fix is genuinely fixed**, and the fix converges rather than forks: the bundle branch resolves through the skill table's `path`, which both layouts already populate, instead of adding a second name-matching rule.
- **No wider deletion.** Path traversal, `..`, absolute paths and symlinks were driven at it and refused; the stamp check keeps unstamped folders.
- **`audit.py` keeps up**: `INSTALL_META` is promoted to a module constant beside the writer, and `console.py` reads it from there rather than re-spelling the literal.
- **Seat and direction hold.** `raven.skill_hub` is a seated inner shelf (`pyproject.toml` source_modules, `CONTEXT.md:1409`); `console.py` importing it is surface -> inner, the sanctioned direction, behind an `ImportError` guard. `lint-imports` -> **Contracts: 10 kept, 0 broken**.

gloryfromca:

I agree with all three dispositions. The sanitized-name/table-key mismatch is a real unsupported corner but has no demonstrated Hub instance and the CLI remains a working way out; the CLI provenance mismatch predates this PR; and the bare `market` wording is the documented terminology nit described here. I am carrying those as nonblocking follow-ups, so this revision can merge as it stands.

Verification on the unchanged head: `uv run pytest tests/test_rpc_console.py tests/test_rpc_skillhub.py -x` passed all 146 tests, and GitHub now shows every required check green.

Handsome-wzw:
Thank you -- all three stand, and the first one was worth more than a note. Taken in c2d18029:

**1. Sanitised name as a table key.** You are right, and the measurement is exact: `_safe_name("acme_tool@v0")` is `"acme_toolv0"`, so the bundle branch could never match a flat bundle keyed by its directory, and a declared name with a space fails the same way -- while `ext.list` reports both removable. The bundle branch is now asked with `skill_name` as sent; `_safe_name` stays on the path branch, which is the only place a path is built from the name; and the reply echoes the name asked for. A comment beside the two spellings says why they differ. The regression test builds both cases you described -- a flat bundle with no `name:` in its SKILL.md, resolved through the real registry's directory-name fallback, and a declared `name: my skill` -- and removes each by the name the table lists. Mutation-checked: handing the bundle branch the sanitised name again fails exactly that test (1 failed, 146 passed; 147 unmutated).

**3. "market" alone.** Changed to "the skill hub"; no other bare use of the word was added by this PR.

**2. The CLI decides the same question differently.** Agreed, and agreed it predates this PR: `raven skill remove` keys on the directory name split at `@`, case-insensitively, and never reads the stamp, so a hand-placed `hub/foo@v1` is protected here and deleted there. I have left `skill_commands.py` alone in this PR -- a second surface's rule is its own change -- but I would rather it be recorded than quoted as settled: if the maintainers want, I will file it as an issue naming both deciders and the two behaviours, with the stamp check as the proposed common rule.

Diff coverage on the touched tests is 98.39% (61/62), the one remaining line being `write_install_meta`'s body, which the full suite covers.

gloryfromca (commented):

The raw-name change fixes the two advertised bundle cases (`<slug>@<version>` and a declared name containing spaces), preserves the sanitized direct-layout path, and corrects the Skill Hub terminology. The added regression test exercises both corrected names without weakening prior coverage.

One residual collision is worth carrying as a nonblocking follow-up: the sanitized direct-layout lookup still runs first. With a direct-layout skill named `myskill` and a bundle named `my skill`, removing `my skill` deletes `myskill` and leaves the bundle. I reproduced that exact result locally. It requires two colliding installs and the bundle still has the CLI removal path, so at this late review round it does not meet the blocker bar. The pre-existing CLI/stamp provenance mismatch also remains the separately scoped follow-up the author acknowledged.

I covered AGENTS.md/CLAUDE.md and the canonical CONTEXT terminology, the full diff and this head's delta, the `ext.list`/web removal callers, relevant history, both install layouts and response compatibility, test-strength changes, and the RPC-to-inner-layer architecture direction. Verification at `c2d180292110`: the RPC/console/client/policy/CLI sweep passed all 211 tests; Ruff and `git diff --check` passed.

0xKT (approved):

PR #467, opened 2026-09-17
fix(*): the deck lane draws its own pages, and a dropped stream is asked again

Tchen-data:
## Summary

Two regressions in the hand-built deck lanes (medium and high on Raven-Design), and the stream failure that made every run of them a coin toss.

**The deck lane draws its own pages again.** PR #451 pointed the hand-built lane at the engine's helper modules (`ppt_layout`, `ppt_theme`, `ppt_shapes`), which were written for the max lane's script backend. Their defaults became the deck: `heading()` put a white surface band bled across every page, `card()` became the way anything was placed, text frames were top-anchored with the lower third empty, and a build script named three colours. Measured on the same request, the accent colour appeared 8 times across 20 pages. The skill, route note and assets reference go back to the pre-#451 text: the lane draws with python-pptx and takes only the icon set, `add_formula` and `math_runs` from the engine. On the reverted lane the accent appears 54 times, the deck carries 12 real pictures and the layouts differ page to page.

**Colour is named beside the type sizes.** The skill listed sizes for every role and said nothing about colour, so a lane with the right palette still drew in one colour. One paragraph now says what colour is for; the layouts note no longer reads as one accent per page, which a lane took as a ceiling. Where `image_search` is not offered, the fallback no longer tells the lane to give the pictures up: `web_search` for the page, `web_fetch` it, and download where the fetch backend keeps image links.

**A stall in a silent think is asked again.** Across two measured runs the lane's turn failed 28 times on `Upstream idle timeout exceeded`, every time in the one round that writes the whole build script into a tool argument. `stream_llm_call` refuses to retry after output a watcher has seen, and counted reasoning deltas and half-built tool-call fragments as that output, though neither is ever shown as the reply. So the configured retry ladder was never reached. `rendered()` now asks what a watcher could have seen; every retry path resets the buffers first (a slot left standing merged with the next attempt into a call the model never made); a stall handed back as an error response carries the usage it spent. The in-process builtin backend, whose comment claimed a retry ladder and passed none, now receives the deployment's `llmErrorRetryDelays` and `llmRetryAfterOutput` through the SubagentManager. No packaged product takes that path; all five run as ACP agents on the loop's own call site.

**Two bundled templates edited in place**, by the maintainer's word. `beige_geometric_general_report` loses its page 20, a dark photograph cut by a circle under pale rounded cards; 23 pages become 22 and the two reference pages the lane borrows from it (17, 19) keep their numbers. `black_circuit_tech_launch` loses the 27 percent circuit texture on its master and the photograph on its Section Header layout, so its body and section pages sit on the master's plain `#000000`; cover and closing are unchanged. `templates.manifest.json` is repinned; `fetch_templates.py --verify` holds all ten pins. The registry archive was not re-cut.

No configuration default changes. The only behaviour every lane sees: a stream that fails before any reply text reached a watcher is retried on the configured ladder instead of failing the turn.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
uv run --frozen pytest tests/test_agent_loop_llm_error_patience.py tests/test_agent_loop_stream.py \
  tests/test_subagent_manager.py tests/test_subagent_routing_backend.py -q
265 passed in 6.85s   (after rebase onto main)

uv run --frozen pytest tests/test_litellm_provider_stream.py tests/test_litellm_provider_stream_end.py \
  tests/test_provider_stream_fallback.py tests/test_providers_stream_idle_budget.py -q
102 passed

uv run --frozen ruff check raven/agent/loop/main.py raven/agent/subagent/backends/raven_loop.py \
  raven/agent/subagent/manager.py raven/providers/streaming.py
All checks passed!

grep -nP "[^\x00-\x7F]" agents/raven-design/route-raven-ppt.md   -> no matches (the note is pinned ASCII)
```

Both new tests were checked against the unfixed code: `test_a_stall_during_a_silent_think_is_asked_again` fails with the old `emitted` test, and `test_what_the_failed_attempt_left_behind_is_dropped` fails with the buffer resets removed.

The failure itself was reproduced outside the runtime: the same 37-message history, six pictures and tool table, sent through litellm 1.101.0 to the same vendor, succeeds on a short answer every time and fails mid-stream one time in three when asked to write the full build script; healthy long streams show a maximum inter-chunk gap of 1.9s over 234s, so the cut is a dropped connection, not a slow model.

Decks built on the changed lane, same request as the #451 acceptance run (glm-5.3-flash via OpenRouter): medium 21 pages with 12 pictures and 7 text colours, 48 gate findings (4 `excessive_whitespace`, down from 18); high 13 pages with 13 pictures and 9 text colours, 37 findings. The high run's page count came from the host agent's forwarded brief, which stated 10-14 pages the user never asked for; that is a separate issue.

Template edits: `tests/test_ppt_engine_templates.py tests/test_ppt_engine_default_templates.py tests/test_ppt_engine_template_bands.py tests/test_ppt_engine_template_theme.py tests/test_ppt_engine_measure_geometry.py` -> 152 passed, 1 skipped; `make check-large-files` -> clean; both templates rendered and read page by page.

## Risk

User-visible: on every lane, a streamed call that fails during reasoning or mid tool-call, with no reply text yet shown, waits out the configured ladder (default 15/30/60s) and asks again instead of ending the turn. A call that already streamed reply text behaves as before. Reasoning tokens of the abandoned attempt are spent again and are now recorded in usage. Rollback: revert the `providers` commit alone to restore the old failure; the skill text and the builtin-backend plumbing are independent of it.

## Related Issues

N/A


agents/raven-design/route-raven-ppt.md
```diff
@@ -15,14 +15,66 @@ Two things before each generation, in this order.
...
+            shape.line.width = pen
+        for style in shape._element.findall(STYLE):
+            shape._element.remove(style)
+```
+
+A run whose whole span is thinner than the pen is one of upstream's dots and is painted rather than stroked, which is what the `dot` branch is for; the `p:style` element has to go or the theme stamps a drop shadow under every stroke. An icon is a mark and not an illustration: around 0.7in on a card, one weight and one colour across the deck, and no filled coloured disc behind each one. Both imports live on the interpreter raven itself runs on, which is not the `python3` that PATH resolves to: probe with `python3 -c "import raven_ppt, pptx"` and, where that fails, run the build script on the raven installation's own `.venv/bin/python`, or put that interpreter's `site-packages` first on `sys.path` at the top of the script. `icons` is the module `icons.py` and not a directory, so listing that path comes back empty whether or not the package is there -- the import is the only thing that answers. If that import is not available here, design the marks yourself -- do not generate them as pictures.
```
gloryfromca: Blocking: align the route note with the deck skill's build-interpreter contract.

This new fallback tells the lane to run the build with Raven's `.venv/bin/python` or inject that environment's `site-packages` when `python3` cannot import the packages. But this route requires `local/deck-to-pptx`, whose Mechanics section still says to run with `raven-python` and explicitly says not to look for a `.venv`. Before this change both sources agreed on `raven-python`; now every hand-built deck gets mutually exclusive execution instructions, and choosing the new route text can send the agent hunting for an installation path instead of using the provided runnable shim. Please make the route and skill state one executable interpreter contract.

gloryfromca (commented):
Blocking: align the route note with the deck skill's build-interpreter contract.

I found one blocking instruction conflict inline; otherwise the implementation looks mergeable.

Covered: AGENTS.md/CLAUDE.md and CONTEXT-MAP.md/CONTEXT.md; the complete `github/main...398118be4e68` diff and commit history; changed-code callers and construction sites; route-note assembly and the required skill; the subagent/loop import boundary; backward compatibility; and the changed tests for weakening (none found). I also exercised the embedded icon recipe.

Verification on the current head:

- Retry/stream/subagent suites: 265 passed.
- LiteLLM/fallback/idle-budget stream suites: 66 passed, with 26 dependency deprecation/serialization warnings.
- Capability/config/doctor suites: 159 passed after `uv sync --frozen --extra dev`. Before installing that declared extra, the base environment produced 87 doctor setup errors for missing `raven_everos`; a source-path-only retry still produced two JSON failures because plugin entry-point discovery correctly did not see an installed plugin.
- The new icon recipe executed verbatim against three documented icons, produced 11 shapes, and saved a PPTX.
- The GitHub unit 3/4 job currently reports failure, but its log shows 5,141 passed and two skipped; the job failed only the unrelated slow-test gate because `tests/test_rpc_model.py::test_options_unauthed_provider_marked` recorded 3.72s idle against a 3s threshold.

agents/raven-design/route-raven-ppt.md
Tchen-data: Fixed in c4376f7c1. The sentence was carried back with the pre-#451 icon section and predates the `raven-python` shim. The route note now states the skill's contract: both imports live on `raven-python`, probe with `raven-python -c "import raven_ppt, pptx"`, run the build on it, and do not look for a `.venv`, build one, or install python-pptx. `.venv` and `site-packages` no longer appear in the note; `tests/test_subagent_routing_backend.py` (44, including the ASCII pin) passes.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at c4376f7c1f83: the route note now probes and runs the build with `raven-python`, matching the required skill contract and explicitly rejecting `.venv` lookup, venv creation, and local `python-pptx` installation. A range comparison shows the five previously reviewed patches are unchanged after the rebase; this corrective commit is the only addition. After `uv sync --frozen --extra dev`, the focused routing and design-launcher suites pass (89 passed). The initial unsynced run had 79 passes and 10 setup failures because the development `raven_design` plugin was absent. This resolves my finding.

gloryfromca (commented):

The interpreter-contract finding is fixed: the route note and required deck skill now consistently probe and execute with `raven-python` and reject `.venv` discovery or local dependency installation. I replied on and resolved my settled thread.

Coverage: I refreshed `github/main`, re-read `git diff github/main...HEAD`, range-compared this revision with the previously reviewed patch series, and checked the repository rules, changed prose and code, relevant route/skill/launcher callers, history, backward compatibility, retry/stream architecture, and whether tests had been weakened. The five previously reviewed patches are unchanged after the rebase; this revision adds only the corrective documentation commit. No new finding survived review.

Verification: `git diff --check github/main...HEAD` passed. After `uv sync --frozen --extra dev`, `uv run pytest tests/test_subagent_routing_backend.py tests/test_agents_design_launcher.py -q` passed (89 passed). The initial unsynced run reported 79 passed and 10 failures, all caused by the absent development `raven_design` plugin; rerunning under the repository's locked dev environment passed.

plugins-dist/ppt-engine/templates.manifest.json
```diff
@@ -9,7 +9,7 @@
...
+  "source_of_truth": "The tracked payload under raven_ppt/assets/templates, whose bytes these pins sign: the eight re-cut light templates, corrected for a lone-punctuation text box and for card titles that collided when wrapped, plus two dark templates cut from user uploads (hidden vendor pages removed, reference pages from the other templates borrowed onto their own layouts, vendor marks stripped). This is the same payload the registry serves as ppt-templates 2026.09.09 (project 84258686). The earlier 2026.09.03 version carries only the eight-file payload and a gold template that no longer matches these pins; it stays downloadable for older checkouts. The retired fork kept its own twelve and was never the payload (its record: git history at the subagents/ removal commit) Two pages were then edited in place on 2026-09-17 without a new registry cut: beige_geometric_general_report lost its page 20 (a dark photograph under pale rounded cards), and black_circuit_tech_launch lost the circuit texture on its master and the photograph on its Section Header layout, so its body and section pages sit on the master's plain #000000.",
```
gloryfromca: Blocking: update the deck-template capacity calibration for the removed beige page.

Deleting page 20 changes the calibrated corpus, but `tests/test_ppt_engine_template_capacity.py` still requires 993 in-bound copy boxes and 211 pages; the new payload produces 989 and 210. The complete `tests/test_ppt_engine_*` run fails those two tests locally, and GitHub's required unit shard fails on the identical assertions. Please reconcile the calibration tests and their measured-count documentation with the payload (or retain a payload that satisfies the existing calibration).

At this fourth review round, this remains blocking because the new commit introduces it, ordinary PR CI always reaches it, and there is no merge path while the required shard is red.

gloryfromca (commented):
Blocking: reconcile the deck-template capacity calibration with the new 210-page payload.

The previously reviewed six-patch series is unchanged after the rebase; this review isolated the new template-payload commit. One revision-specific blocker is marked inline.

Coverage: I refreshed `github/main`, read `git diff github/main...HEAD`, range-compared the old and new patch series, and checked the repository rules, binary package contents, manifest pins, template callers and page references, history, backward compatibility, package relationships, architecture constraints, and whether tests were weakened. The two edited archives have no dangling internal relationships, load successfully, and match their new size/hash pins.

Verification: `git diff --check github/main...HEAD`, `uv run python scripts/check_large_files.py github/main...HEAD`, and `uv run python plugins-dist/ppt-engine/fetch_templates.py --verify` passed. `uv run pytest tests/test_ppt_engine_*.py -q` reported 3751 passed, 14 skipped, and 2 failed; the two failures reproduce alone. GitHub's required unit shard reports the same two failures (5955 passed, 103 skipped).

plugins-dist/ppt-engine/templates.manifest.json
Tchen-data: Fixed in 6d9fca30a. The calibration in `tests/test_ppt_engine_template_capacity.py` now counts the payload as it ships: 210 pages, and 1,110 sized boxes of which 989 sit inside the conservative bound and the same 121 between the bounds (the four boxes that left were on the removed beige page). The file passes locally (22 passed), and CI's unit shard 1/4, which failed on exactly these two pins, is re-running on this head.

gloryfromca: No blockers; suggestions only, and they are marked inline.

Verified on the live head `1f1dff47d654`: the capacity assertions now match the shipped payload, and `uv run pytest tests/test_ppt_engine_*.py -q` passes (3753 passed, 14 skipped). The focused capacity/skill run also passes (25 passed). This resolves my blocking finding; I am carrying the remaining arithmetic typo in the calibration prose as a separate nonblocking suggestion.

tests/test_ppt_engine_template_capacity.py
```diff
@@ -311,7 +311,7 @@ def test_the_templates_own_copy_is_never_contradicted() -> None:
     """The calibration, and the one result a wrong band cannot survive.
 
     The designer's own copy is the ground truth: over the ten bundled templates'
-    1,116 boxes with a size on a run, 993 sit inside the conservative bound and 121
+    1,110 boxes with a size on a run, 989 sit inside the conservative bound and 121
```
gloryfromca: No blockers; suggestions only, and they are marked inline.

The total here should be 1,112, not 1,110: the test measures 989 boxes inside the conservative bound, 121 between the bounds, and two more in `contradicted` past the generous bound (`989 + 121 + 2 = 1,112`). The assertions themselves are correct; only the explanatory total omits the two outliers described in the next sentence.

gloryfromca (commented):

The capacity-calibration blocker is fixed and its thread is resolved. The live branch advanced once more during review; I also reviewed the added page-ground guidance at `1f1dff47d654` and found no behavioral issue. One arithmetic correction remains inline as a nonblocking documentation suggestion.

Coverage: I refreshed `github/main`, read `git diff github/main...HEAD`, range-compared the prior patch series, and checked the repository rules, the recalibrated assertions against their measured buckets, affected template/skill callers, history, backward compatibility, architecture constraints, and whether the tests were weakened. The assertions now pin the new payload rather than relaxing the invariant.

Verification: `git diff --check github/main...HEAD` passed. `uv run pytest tests/test_ppt_engine_*.py -q` passed (3753 passed, 14 skipped), and `uv run pytest tests/test_ppt_engine_skill.py tests/test_ppt_engine_template_capacity.py -q` passed (25 passed). Required CI was still running when this review was posted.

tests/test_ppt_engine_template_capacity.py
Tchen-data: Right, the two gold_panel brackets past the generous bound were dropped from the sum. The docstring now says 1,112 (989 + 121 + 2) in 063469f71; the assertions were already 989 and 121.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at `063469f711f7`: the docstring now counts all three measured buckets (`989 + 121 + 2 = 1,112`), and `uv run pytest tests/test_ppt_engine_template_capacity.py -q` passes (22 passed). The prior patch series is unchanged; this resolves my suggestion.

gloryfromca (commented):

The calibration prose now matches all three measured buckets (`989 + 121 + 2 = 1,112`). I replied to the author, resolved the settled suggestion, and confirmed every thread I opened on this change is resolved.

Coverage: I refreshed `github/main`, read `git diff github/main...HEAD`, range-compared this revision with the previously reviewed patch series, and checked the repository rules, the one-line delta, its calibration assertions and callers, history, backward compatibility, architecture constraints, and whether tests were weakened. The prior series is unchanged and no new finding survived review.

Verification: `git diff --check github/main...HEAD` passed, and `uv run pytest tests/test_ppt_engine_template_capacity.py -q` passed (22 passed). Required CI was still running when this review was posted.

0xKT:
**Not a blocker -- the retry is properly bounded and I could not break it. Three things stand: one costs the maintainer refresh lane two templates, one leaves half the title untested, one is a dangling sentence.

### 1. `make fetch-templates` now deletes two of the ten bundled templates

The PR re-cuts two `.pptx` and rewrites their `size`/`sha256` pins, but `templates.manifest.json`'s `endpoint.url` still points at `ppt-templates/2026.09.09` -- the package version that serves the pre-edit bytes. `git diff` on that file matches the url line **zero** times.

`fetch_templates.py:134-135` deletes whatever fails its pin:

```python
if complaint := check_one(target, entry):
    target.unlink(missing_ok=True)
```

So the lane advertised at `Makefile:42` ("Re-pull the deck engine's ten bundled templates over the tracked copies") now ends with 8.

The rule is the manifest's own `endpoint.note`:

> a new cut is a new package version (upload, rewrite the pins, bump this url, commit the new bytes)

Four steps; this PR did the last two. And it says so itself, in the line it appends to `source_of_truth`: *"Two pages were then edited in place on 2026-09-17 without a new registry cut."*

Measured against a stand-in registry serving the 2026.09.09 bytes, driving the real `pull()` through the production branch:

| manifest | result |
|---|---|
| **HEAD** | `size mismatch: beige_geometric_general_report.pptx holds 748798 bytes, pinned 614701 (removed)` + the same for `black_circuit_tech_launch` -> exit 1, **8 left** |
| **base** (control) | `all 10 pins hold` -> exit 0, **10 left** |
| **HEAD + registry re-cut to the new bytes** (the refuting case) | `all 10 pins hold` -> **10 left** |

The third row is the one that matters: it isolates the deletion to the pins-vs-registry divergence, not to the script and not to the harness.

Not a blocker because the lane is maintainer-only (`grep .github/` for `fetch-templates` is empty), the failure is loud (exit 1 plus two explicit `(removed)` lines), and git still holds the bytes. Uploading the new cut and bumping the url closes it.

### 2. The subagent half of the title has no test

`raven/agent/subagent/backends/raven_loop.py:433-434` is the whole of "a dropped stream is asked again" on the subagent path:

```python
retry_delays=self.retry_delays,
retry_after_output=self.retry_after_output,
```

I deleted both lines and ran `tests/test_agent_loop_llm_error_patience.py tests/test_subagent_manager.py tests/test_agent_loop_lazy_provider.py`:

| | result |
|---|---|
| as shipped | **191 passed** |
| both lines deleted | **191 passed** |

The new test file exercises the AgentLoop/turn_path side only. So a future edit that drops these two lines un-does half the feature and every gate stays green -- which is the same shape as the unasserted stub I noted on #458. One test that spawns a subagent against a provider that fails once and asserts it retried would close it.

### 3. `deck-to-pptx/SKILL.md:145` still scopes a section by something this PR removed

> For what is drawn with python-pptx directly; the helper modules handle these themselves.

This PR deletes every other mention of the helper modules from that route -- the rows naming `ppt_layout` / `ppt_icons` / `ppt_shapes`, and the whole "Furniture from the modules" section. A prompt is an instruction an agent follows, and this one now scopes a list by a distinction the route no longer draws.

### What held up

- **The retry is bounded, and I went looking for the hole.** A provider that never succeeds, entered through every door this PR widens (a silent think, a half-built tool call, an already-rendered stream), stops at exactly **5 provider calls** = `1 + max_reconnects + len(retry_delays)` -- the bound written at `streaming.py:179` -- and every path **raises**. No path returns a silent half-result: the exhausted ones raise at `:355`/`:373`, and the two that return an error response are consumed loudly by `turn_path.py:1795-1824` and `raven_loop.py:452-476`. The `retry_after_output=False` case still stops at 1 call, so the pre-existing rule is preserved.
- **"Dropped" is not confused with a normal ending.** Three legitimate endings were built on purpose -- content+stop, a model that thinks and then says *nothing*+stop (exactly the shape the new `rendered()` makes retryable when it fails), and a complete tool call+`tool_calls` -- and all three make **1** call with the right payload. The criterion is `classification.retryable` off the live exception, not a guess about content.
- **The 4.75 MiB `.pptx` is covered by an existing per-file approval, and the gate was tested before being trusted.** `scripts/check_large_files.py c91f76bd..HEAD` exits 0, and feeding it three oversized copies reports exactly the unapproved ones while the pinned name stays absent. Worth one line for the reader, though: because the bytes half is delegated to a manifest this same PR rewrites, a re-cut of an already-approved template passes both gates by construction.

gloryfromca:

The verifier note adds three explicit nonblocking follow-ups: the stale template-registry refresh lane, missing regression coverage for the subagent retry wiring, and the dangling helper-module sentence. None changes mergeability, but together they make my prior `clean` label stale; my whole-record stance is now `nonblocking`. All required checks on `063469f711f7` are currently green.

Tchen-data:
Thanks for the three, all addressed on the branch:

1. **Registry cut.** The ten templates are uploaded to the generic registry as `ppt-templates 2026.09.17` (the 2026.09.09 package is untouched) and the manifest endpoint now names that version. Your third row reproduced: a real `fetch_templates.py` pull against the new url into a scratch directory reports `all 10 pins hold`, and `--verify` on the tracked payload holds as before.

2. **The subagent half now has a test.** Two tests in `tests/test_agent_loop_llm_error_patience.py` stand on the streamed path a spawn takes when a caller watches the reply form: a connection dropped twice before the first delta is answered on the third call only because the backend's ladder reached `stream_llm_call`, and a stream that showed a word and then dropped is asked again only when the backend was built with `retry_after_output`. I deleted the two lines at `raven_loop.py` and ran them: both fail, and both pass with the lines back.

3. **The dangling sentence** in `deck-to-pptx/SKILL.md` is gone; the pitfalls list applies to everything this route draws.

One note on the wording of the test: the fixtures raise `ConnectionError`, not `TimeoutError`, because a stall takes the helper's `except TimeoutError` arm and returns an error response without touching the ladder, so a TimeoutError would test the backend's waited-for fallback rather than the wiring you named.

gloryfromca (commented):

This revision closes the three recorded follow-ups: the manifest now names the 2026.09.17 registry cut, the subagent streamed path has regression coverage for both the deployment retry ladder and `retry_after_output`, and the obsolete helper-module qualifier is gone. The new tests exercise the production `RavenLoopBackend` path and would fail if either forwarded setting disappeared.

Coverage: I refreshed `github/main`, read `git diff github/main...HEAD`, inspected the new commits and surrounding retry/stream/template machinery, and checked the repository rules, callers, history, backward compatibility, architecture constraints, merge compatibility with current `main`, and whether tests were weakened. The merge-tree check is clean, all threads I opened remain resolved, and no new finding survived review.

Verification: `git diff --check github/main...HEAD`, `uv run python scripts/check_large_files.py github/main...HEAD`, and `uv run python plugins-dist/ppt-engine/fetch_templates.py --verify` passed. `uv run pytest tests/test_agent_loop_llm_error_patience.py tests/test_subagent_manager.py tests/test_agent_loop_lazy_provider.py tests/test_ppt_engine_templates.py tests/test_ppt_engine_skill.py -q` passed (208 passed). I could not independently repeat the private registry pull because this environment has no GitLab token; the maintainer reports a real pull of all ten pinned templates from the new endpoint. Three unit shards were still running when this review was posted; the completed checks were green.

gloryfromca (commented):

The new opener-image policy is consistent with the route and asset instructions: search for real subject photography first, generate only for imagery that does not exist (texture, light, abstraction), and fall back to typography/grid/colour when image credentials are unavailable. The previously reviewed patch series is unchanged, and no new finding survived review.

Coverage: I refreshed `github/main`, read `git diff github/main...HEAD`, range-compared this revision with the previously reviewed series, and checked the repository rules, changed skill prose, route and asset callers, history, backward compatibility, capability fallbacks, architecture constraints, merge compatibility, and whether tests were weakened. All threads I opened remain resolved.

Verification: `git diff --check github/main...HEAD`, `uv run python scripts/check_source_language.py github/main...HEAD`, and the merge-tree check against current `github/main` passed. `uv run pytest tests/test_subagent_routing_backend.py tests/test_agents_design_launcher.py -q` passed (89 passed). Four unit shards were still running when this review was posted; completed checks were green.

0xKT (approved):

PR #468, opened 2026-09-17
refactor(agent): read the charter and the window's shrinking through the roles

ypflll:
## Summary

The agent loop already names four strategy roles (Memory, Planning, Capability, Action), but two judgements a turn makes still lived in the shell around them. This moves both onto the roles, in two dependent commits, with no behaviour change intended.

**1. The charter reads through Memory and Action.** A dispatched sub-agent's playbook (`_meta["raven.playbook"]`) was read from four places inside the loop and the tool registry. `prompt` and `stopWhen` are now read by `DefaultMemory.assemble()`, which fills `task_brief` and `task_done_when` on the `TurnContext` the identity segment renders. `checks` and `code` are read by `DefaultAction.judge()`, which `ToolRegistry.execute` asks once per proposed call, the way it already asks the permission gate. `tools` stays in `withheld_names()` and the registry still decides what a refusal does, so a replaced role withholds nothing it could not already withhold.

**2. Mid-turn window shrinking seats on Memory.** The five recoveries the loop ran inline go through one method, `MemoryModule.shrink(messages, pressure, state, model)`, with `WindowPressure`, `WindowState` and `ShrinkResult` as the carriers:

- proactive compaction, once the last billed reading crosses the trigger
- the standing image window
- a provider's overflow error
- a picture refused inside a tool result
- pictures refused for size

The retry mechanism stays in the shell. The six `continue` sites and the iteration rollback did not move, so a hook sees a retried call exactly as it saw the first. `WindowState` is a per-turn object the shell owns and hands to each call, because the role outlives the turn; `image_window` has no default, since 0 is a real value meaning "no picture stays".

The pure pieces both sides need move to a new package, `raven/agent/window/`, which the import contracts treat as cargo that may not import the loop shell. `REASONING_EFFORT_LADDER` moves to `contracts/llm_provider.py` because the loop's empty-response descent and the window's head summary both read it and neither package may import the other.

Contract papers: `CONTRACTS_VERSION` 26 to 28, `CONTRACTS_LINE_CEILING` 3,040 to 3,200, each bump with its own review paragraph in `tests/test_kernel_budget.py`.

A follow-up PR carries the third step, which gives a sub-agent nine verbs in place of the six hook phases. It is kept separate because it touches every bundled plugin.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [x] Refactor
- [ ] Other

## Verification

Gates, run at each of the two commits:

```
uv run lint-imports
uv run pytest -q
```

Import contracts: 10 kept, 0 broken at both commits. The suite is green at both; the failures this machine shows are the same ones it shows on an unmodified `main` (missing browser and optional packages), and neither commit adds one. The loop's own e2e files, which the default selection leaves out, were run too (charter, playbook, truncation, direct chat, fallback chain, ACP stdio): no failure that `main` does not also show.

Behaviour preservation. Each commit was run against a worktree of its own parent under the same scripted provider, outputs normalised for nonces and temp paths, then byte-compared. Four scenarios, eight comparisons, all identical:

- a whole turn with no playbook: host turn, spawned sub-agent, tool calls and refusals
- a playbook-on worker lane in process, including two charter refusals
- six window-shrink scenarios: proactive compaction, the standing image window, overflow, a refused picture, oversized pictures, the effort ladder
- the five-plugin hook seam, every phase's decision and metadata per iteration

Mutation check on the second commit: breaking each of the five moves inside `DefaultMemory.shrink` turns exactly its own scenario red, so the comparison is not vacuous.

Real model, Sonnet 4.5 through OpenRouter, at both commits:

- playbook on, in-process lane: both workers receive the charter, write their files and finish
- playbook on, ACP fork lane: the charter prompt crosses the process boundary, and a charter check blocks a `write_file` outside the fence

## Risk

No user-visible behaviour change is intended, and the comparisons above are byte-identical on every path they cover.

For anyone replacing a harness role: `MemoryModule` gains `shrink`, so a replacement written against the old protocol needs it. `DefaultMemory.bind()` names the missing member at assembly rather than failing inside a turn.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the current head after its merge from main. I covered the repository rules and canonical runtime vocabulary, the full diff, the affected charter/window callers, both feature commits and surrounding history, compatibility of the moved internal helpers and expanded harness contracts, test changes for weakening, and the cargo-versus-loop architecture boundary. The refactor preserves the prior retry/state transitions while moving policy behind Memory, and the charter prompt/check paths remain joined end to end.

Verification: the focused changed-area run passed 247 tests; the PPT ledger passed 13 tests under the all-extras environment; `uv run lint-imports` kept all 10 contracts. The full all-extras run reported 23,127 passed and 120 skipped, with one unrelated `test_cli_theme` failure caused by this runner's `NO_COLOR=1`; that exact test passed when rerun with `NO_COLOR` unset. GitHub's four Python test shards are also green on this head.

raven/agent/tools/registry.py
```diff
@@ -383,6 +389,35 @@ def set_withheld_source(self, source: "Callable[[], frozenset[str]] | None") ->
...
+        a module reaching into the shell's state.
+
+        Caught here and nowhere inside: this is a boundary onto a *replaceable*
+        role, and the shell does not get to assume what is behind it. The
+        default implementation guards nothing of its own -- a defect in rule
+        evaluation is meant to be loud -- but a replacement that raises must
+        cost its own judgement rather than the turn, which is the same bargain
+        the plugin gates beside it are held to.
+        """
+        if self._verifier_provider is None:
```
LivXue: **BLOCKING.** This early return silently converts every registry that was not handed a `verifier_provider` from "enforces the charter" to "takes no opinion".

At the merge base, `execute` called the module-level `_charter_refusals(name, params)` unconditionally (base `registry.py:888`), so every registry enforced it. There are 10 `ToolRegistry(...)` construction sites at head; exactly two pass a provider (`loop/main.py:366` and `subagent/backends/raven_loop.py:320`).

`context_engine/segments/curator.py:357` builds a bare `ToolRegistry()`, and `curator.py:313` dispatches on it (`await registry.execute(...)`). That runs inside the `charter_scope(charter)` opened at `loop/main.py:922`, because the scope wraps `_run_turn`, which reaches context assembly at `turn_path.py:1905` -- and the curator slow path is context assembly. So a charter `check` or `code` judge that refused a curator call at base is a no-op at head.

The asymmetry it leaves is the part I would most like changed regardless of blast radius: `_charter_withheld` (via `withheld_names()` at `:434`) and `_record_call` are still module-level and unconditional on that same registry. The charter still narrows the tool set and still records the calls, but can no longer refuse them. The comment added at `:291-296` asserts the opposite -- that the curator's registry is built "never to dispatch a turn's calls" -- and `curator.py:313` is that dispatch.

Nothing pins the wiring either: `grep -rn verifier_provider tests/` returns zero hits.

Worth saying that `raven_loop.py:320` shows you were tracking exactly this question -- its comment says the charter's `checks` should "hold on this lane as they do on the forked one". This reads as an incomplete sweep of the construction sites, not a design decision.

Suggested fix, in order of preference:
1. Make the curator registry pass a provider the same way `raven_loop.py` does; or
2. If that registry genuinely must not carry a verifier, say so at the construction site and make `withheld_names()`/`_record_call` consistent with that, so the charter is wholly absent rather than half-applied; and either way
3. Consider making `verifier_provider` a required keyword, so a new construction site has to answer the question instead of inheriting the permissive default. That is what would have caught this one.

A test that constructs a bare registry under an active `charter_scope` with one `CheckRule` and asserts the refusal would keep it caught.

LivXue (commented):
## Review: verified against `f69e6cf7f` vs merge-base `3b35e7d34`

This is a high-quality refactor. `turn_path.py` loses 537 lines, the contract paper
is written with more care than most changes of this size get, and the verification
method in the description (per-commit byte-compare under a scripted provider, plus a
mutation check to prove the comparison is not vacuous) is the right one.

Two findings are, in my reading, blocking. Both are small to fix.

**B1. The charter's `checks`/`code` enforcement went from unconditional to opt-in,
and one registry that dispatches inside a live charter scope did not get the opt-in.**
This contradicts "no behaviour change intended", and it moves a security boundary in
the permissive direction. Detail inline at `registry.py:407`.

**B2. Four of the five tests this PR adds live in a directory the default run excludes,
so the guard it wrote for its own new failure mode never executes.** Detail inline at
`test_subagent_charter_e2e.py:215`.

Four more I would like resolved in the same change, each one or two lines:
`contracts/context.py:80` (positional-slot shift on a contract-tier dataclass),
`harness/__init__.py:52` (no `bind_action` beside `bind_memory`),
`contracts/harness.py:13` (the paper contradicts itself), and `CONTEXT.md:93`
(AGENTS.md section 6: three new contract-tier terms and a new ruled package are
undefined, and the charter is renamed to "playbook", which CONTEXT.md already
defines as a different feature).

Nine further comments are non-blocking.

**On the nine: five of them are pre-existing code this PR relocates verbatim**
(`images.py:77`, `images.py:90`, `memory.py:217`, `memory.py:262`, `shrink.py:237`).
I verified them byte-identical to the base after normalising the `_name` -> `name`
de-privatisation. They are not defects introduced here; they show up as `+` lines
to anyone scanning added hunks, and creating the file is the one moment the fix is
free. Fixing them is your call, not a condition of merge.

One side note outside this diff: promoting `REASONING_EFFORT_LADDER` to `contracts`
did not reach `rpc/methods/console.py:843`, which still validates
`agents.defaults.reasoningEffort` against four of the ladder's six rungs, so the
console rejects `xhigh`/`max` while config files and the Anthropic transport accept
them. Worth a separate PR.

Nothing here argues against the shape of the refactor. The policy/mechanism split is
the right seam, and the shell keeping the six `continue` sites and the iteration
rollback is what makes it safe.

tests/integration/test_subagent_charter_e2e.py
```diff
@@ -164,23 +164,116 @@ def test_a_turn_with_no_charter_staged_takes_none(workspace) -> None:
...
+    briefed = asyncio.run(segment.build(_assembly_ctx(task_brief=BRIEF, task_done_when="both tables land"))).text
 
     assert BRIEF not in plain
     assert BRIEF in briefed
     assert "both tables land" in briefed
     assert briefed.startswith(plain), "the brief is appended; the runtime facts above it stay"
 
 
+@pytest.mark.asyncio
+async def test_the_brief_only_reaches_the_identity_through_memory(workspace) -> None:
```
LivXue, line 215: **BLOCKING.** These tests never run.

`pyproject.toml:468` sets `testpaths = ["tests"]` and `:472` sets `norecursedirs = ["tests/integration", ...]`, so the default collection skips this directory. Across the workflows at this head there are two pytest invocations: `ci.yml:151` (`make coverage-shard` -> `pytest -q --shard`, default scope) and `ci.yml:260`, which names only `tests/integration/test_cli_upgrade_real_uv.py`.

So four of the five test functions this PR adds never execute: `test_the_identity_renders_the_task_it_is_handed` (:193), `test_the_brief_only_reaches_the_identity_through_memory` (:215), `test_memory_fills_the_turn_from_the_charter`, and the two registry-refusal tests further down this file. Only `test_the_window_role_answers_every_pressure_the_loop_asks_under` in `tests/test_harness_modules.py` is in the default run.

That matters most for this one specifically. Its docstring says "Memory dropping that step used to break nothing any test could see" -- which is exactly right, and is still true at head, because the test that would see it is not collected. Delete the `self._briefed(turn)` call from `DefaultMemory.assemble` and CI stays green while every dispatched sub-agent runs unbriefed. No default-run file covers it: `_briefed` appears nowhere in `tests/*.py`, and `task_brief` only in a docstring at `test_kernel_budget.py:116`.

Second, smaller point: these are misfiled under AGENTS.md section 5.2, which scopes `tests/integration/` to tests that "run against real environments". This file's `_Engine` is a local stub, `provider=lambda: None`, and :258 calls the private `DefaultMemory._briefed` directly. They are unit tests.

Both problems have the same fix: move these four into `tests/` (`test_harness_modules.py` or a sibling), where they will actually run.

raven/contracts/context.py
```diff
@@ -74,6 +74,11 @@ class AssemblyContext:
     budget: TokenBudget
     prefix: AssembledPrefix | None = None
     can_see_images: bool = True
+    # Carried through from ``TurnContext``: what this dispatch's playbook asked
+    # of the turn. Empty on every turn that carried none, which is the string
+    # every reader already appends to no effect.
+    task_brief: str = ""
```
LivXue: **Please fix before merge.** These two fields are inserted mid-dataclass here, but appended last on `TurnContext` (`:243-244`, after `describe_tool`). Same file, same pair, two different insertion strategies -- and only this one shifts positional slots.

At base, `describe_tool` sat directly after `can_see_images`. It is now two slots later. The decorator is a bare `@dataclass(frozen=True)` -- no `kw_only`, no `*` marker -- and both new fields carry `= ""` defaults, so there is no `TypeError` to catch a caller out.

`AssemblyContext` is the contract-tier object every `SegmentBuilder.build(ctx)` receives, so out-of-tree builders and fixtures construct it. A positional call like

```python
AssemblyContext('s', 'hi', None, None, None, [], budget, None, True, 'understand_media')
```

bound its tenth argument to `describe_tool` before this change and binds it to `task_brief` after: the identity renders `## This task: understand_media` and the attachment-describing instruction silently disappears.

The contract-surface digest bump (26 -> 28) does not help -- it fires identically for an append and cannot tell the two apart.

Every in-tree construction site is all-keyword, so this is latent today. Appending here, as `TurnContext` already does, costs nothing and closes it. (Adding `kw_only=True` would be the stronger fix, but that is a wider change than this PR needs to carry.)

raven/agent/harness/__init__.py
```diff
@@ -47,6 +49,8 @@ def default_harness_modules(
         context_window_tokens=context_window_tokens,
         tool_definitions=lambda: registry_provider().get_definitions(),
         system_prompt=system_prompt,
+        compaction=compaction,
```
LivXue, line 53: **Please fix before merge.** This PR makes `judge` a required `ActionModule` member, but the Action role is seated with no validation while the sibling `MemoryModule.shrink` addition did get one.

`memory=bind_memory(memory)` raises a `TypeError` naming the missing member -- and this PR correctly updated that message from "assemble and after_turn" to "assemble, shrink and after_turn". `action=DefaultAction()` gets nothing: `grep -rn bind_action` over the head tree returns nothing, and `HarnessModules(` has one construction site.

The guard is available and simply not made. `ActionModule` is `@runtime_checkable` and declares `judge`, so `isinstance(DecideOnlyAction(), ActionModule)` is already `False`.

The consequence is worse than a late failure, because `registry.py:417`'s `except Exception -> return []` converts the shape error into a permissive answer. With `Charter(checks=(CheckRule(tool='write_file', path_prefix='/srv/app', ...),))`, the default role returns `['write_file: outside the brief']` and a role missing `judge` returns `[]` plus one log line -- a `write_file` to `/etc/passwd` dispatches under a `path_prefix` charter.

That is precisely the outcome `DefaultAction.judge`'s own docstring argues against: "Catching that would turn every such defect into checks that quietly stop applying."

Latent today, since nothing in-tree supplies a non-default Action role. But it is a one-line `bind_action` mirroring `bind_memory`, and the contract you are publishing here is what invites out-of-tree roles.

raven/contracts/harness.py
```diff
@@ -7,20 +7,24 @@
 The four roles name the strategy *decisions* a turn makes, not four layers
 and not the loop itself. Memory assembles the window the model sees, Planning
 may prepare turn guidance, Capability picks the tool definitions one iteration
-exposes, and Action produces one usable model response. Everything else the
-turn does -- iteration accounting, hook phases, tool execution and approval,
+exposes, and Action produces one usable model response and judges a call
+against the dispatch's playbook. Everything else the turn does -- iteration
+accounting, hook phases, tool execution and approval,
 the three in-loop recoveries, persistence and event order -- stays with the
```
LivXue: **Please fix before merge.** The paper now contradicts itself within one file.

This paragraph still assigns "the three in-loop recoveries" to the L2 shell, "which is what makes these four replaceable at all". 130 lines below, the new `MemoryModule.shrink` (:158-165) says Memory decides "what to give up when the transcript nears the line, when a summary is worth paying for, which pictures stay".

Both cannot be true. The PR rewrapped the lines on either side of this clause (to add "and judges a call against the dispatch's playbook") and left the clause itself untouched.

The failure mode is specific: an author of a future generation reads the opening paragraph, keeps overflow / tool-image / size recovery in their own shell, and implements `shrink` as `return ShrinkResult(messages, False)`. The loop's `if shrunk.changed: continue` at `turn_path.py:938/946/958` then never fires, so every context overflow, every tool-image refusal and every oversize-picture refusal becomes a fatal turn error instead of a retry. Nothing catches it -- `bind()`'s `isinstance(memory, MemoryModule)` only checks that a `shrink` attribute exists.

The distinction your own `shrink` docstring draws is the right one and just needs to reach this paragraph: the shell keeps the retry *mechanism*, the role keeps the recovery *policy*.

CONTEXT.md
```diff
@@ -90,8 +90,9 @@ The four generation-scoped strategy roles the Agent Loop delegates to without gi
 Turn state machine: **Memory** assembles the window the model sees, **Planning** may prepare
 turn guidance, **Capability** picks the tool definitions one Iteration exposes, and **Action**
 produces one model response. The default set preserves what the loop did inline: Memory wraps
-this generation's Context Engine and adds the two turn decisions that were the shell's (which
-history slice is a candidate, and how much window the prompt may occupy), Planning passes
+this generation's Context Engine and adds the turn decisions that were the shell's (which
```
LivXue, line 93: **Please fix before merge -- AGENTS.md section 6 is a hard constraint**, and this change touches both halves of it.

**1. New domain terms are undefined.** Counted at head: `WindowState` 0 occurrences in CONTEXT.md, `ShrinkResult` 0, `agent/window` 0, `WindowPressure` 1 -- and that one is the forward reference this diff adds at :942, with no entry behind it. All three are contract-tier types on a published protocol. Section 6: "Coining a new domain term: define it in the matching CONTEXT.md in the same change, with a definition verifiable against the code".

**2. The cargo-contract member list is now stale.** `pyproject.toml:624` adds `raven.agent.window` to the "cargo does not import the loop shell" contract. The CONTEXT.md sentence that enumerates that contract's members (:1429-1430, outside this diff) still reads `tools`, `subagent`, `context`, `hook`, `personalizer`, `workdir` -- now short by two, since `harness` was already missing. That prose cites a contract it no longer matches.

**3. "Playbook" is taken.** CONTEXT.md defines **Charter** at :118 (`agent/subagent/charter.py`) and **Playbook** at :1060 (`raven/playbook/`, "a stored, reusable orchestration") as distinct terms. This PR adds about 11 new uses of "playbook" meaning the charter, across `contracts/harness.py` (1 -> 4), `contracts/context.py` (0 -> 2), `tools/registry.py` (2 -> 4), `harness/action.py` (0 -> 2), `harness/memory.py` (0 -> 1) and `segments/identity.py` (1 -> 2). `registry.py` now carries both senses: `load_playbook` at `:47`/`:773` is the other feature.

I can see where the wording comes from -- the charter does cross the wire as `_meta["raven.playbook"]`, and CONTEXT.md:122 says so. But section 6 asks for the canonical term, and a reader hitting "this dispatch's playbook" in `registry.py` has no way to tell which of the two is meant. Either use "charter" in the prose, or add an `_Avoid_` line to both CONTEXT.md entries that pins the overlap.

raven/agent/window/images.py
```diff
@@ -0,0 +1,138 @@
...
+# each reason with one of these and the filed-form parser finds the end of the facts
+# by them, so a caption may contain ". " without being cut short. A reason worded
+# past them would file the whole note into the session unnoticed, which is why the
+# composer takes its openings from here rather than spelling them out again.
+_REFUSED = "The endpoint refused"
+_ELIDED = "It was elided"
+_OUTGREW = "This turn's pictures"
+_SEEN = "You looked at it"
+_WITHDRAWAL_OPENINGS = (_REFUSED, _ELIDED, _OUTGREW, _SEEN)
+_WITHDRAWN_NOTE_FACTS = re.compile(
```
LivXue, line 90: Non-blocking, and pre-existing: byte-identical to base `_shared.py:492-550`. Flagging it because this PR creates the file whose stated job is "three readers, one grammar", which is the cheap moment to fix it.

The filed-note regex ends the facts at the first `". "` followed by a reason opening, so a caption containing that sequence is truncated mid-caption.

Reproduced against this exact regex and composer. A tool captions a picture `Slide 2. You looked at it earlier`. The live note is:

```
[image no longer in context: "Slide 2. You looked at it earlier" from render_deck at iteration 3. This turn's pictures outgrew their byte budget, ...]
```

and `filed_image_note` (called from `turn_path.py:2313` in `_save_turn`) writes this to the session JSONL:

```
[image no longer in context: "Slide 2.]
```

Unbalanced quote, no tool name, no iteration -- for all four reasons. It is written once and replayed on every resume, so the model can no longer tell which tool to ask for the picture again.

The module comment at :82-83 says the openings exist "so a caption may contain '. ' without being cut short". That holds for a bare `". "` and fails for `". "` followed by an opening word, and captions are tool-authored text up to 120 chars.

The existing test at `test_agent_loop_llm_error_patience.py:741` does not catch it: it picks a caption whose sentence break does not start an opening, and anchors its expectation past the caption -- an anchor the production regex lacks.

raven/agent/window/images.py
```diff
@@ -0,0 +1,138 @@
...
+
+    The base64 payload, not the bytes it decodes to. What the budget above it is
+    protecting is the request body, and a data URI travels encoded: counting the
+    decoded side let 11.24MB of pictures pass a 12MB budget while 16.8MB left the
+    process, which the gateway answered with an empty 200 and zero usage five times
+    over on a byte-identical payload.
+    """
+    if not is_inline_image(part):
+        return 0
+    return len(part["image_url"]["url"].partition(",")[2])
```
LivXue, line 77: Non-blocking performance note, pre-existing (byte-identical to base `_shared.py:477`).

`len(url.partition(",")[2])` materialises a full copy of the base64 payload just to take its length. Timed on a 533,358-char data URI: 19.19 us/call, against 0.13 us/call for `len(u) - u.index(",") - 1` -- 143x, byte-identical answer.

It matters because the standing pass runs every iteration on a stock install: `config/schema.py:269` defaults `image_window_budget_bytes` to 12,000,000, and `turn_path.py:525`'s `... or None` launders only a literal `0`, so `image_budget` is non-`None` and `memory.py:256`'s early return can never fire. `turn_path.py:723` sits inside the iteration loop, so `shrink.py:234` sums over every live picture every time.

Steady state under the 12 MB cap is about 22 pictures / 11.7 MB copied / 0.31 ms per iteration (roughly 18 ms over a 60-iteration deck turn), spiking to 1.79 ms on the iteration an oversized batch lands.

`shrink.py:228` also walks every message's content a first time to build `bearing`, so the parts are visited twice. Accumulating `(index, bytes)` in that one comprehension, plus the `index(",")` change, puts the whole pass under 0.05 ms.

raven/agent/harness/memory.py
```diff
@@ -109,7 +128,250 @@ async def assemble(
...
+        projected = state.last_context_used
+        if cfg.prune:
+            pruned, elided = shrink.emergency_shrink(messages)
+            if elided > 0:
+                # No server reading exists for the pruned list until the next
+                # response, so judge the summary tier by projecting the
+                # estimated savings onto the observed size (local estimates do
+                # not know the server's tokenizer; the delta is safer than the
+                # absolute).
+                saved = max(0, estimate_prompt_tokens(messages) - estimate_prompt_tokens(pruned))
```
LivXue, line 220: Non-blocking performance note. The expression is a move from base `turn_path.py:1065`, but the new `shrink.emergency_shrink` signature makes the fix local, which is why I am raising it here rather than leaving it.

`_compact_ahead` measures the pruning saving by tokenising the entire transcript twice, when `emergency_shrink` already knows which messages it replaced.

Measured with the real function under the repo venv: a 102-message tool-heavy transcript (254,489 tok) costs 65.8 ms per `estimate_prompt_tokens` call, 131.7 ms for the pair. `_compact_ahead` is `async def` (`memory.py:190`) with no `to_thread`, so the event loop is blocked for that whole window -- every other session's stream, the DAG scheduler and the web RPC stall with it.

`turn_path.py:899` re-arms `window.last_context_used` from every response, so once a near-window turn trips `should_compact` this fires roughly per iteration: about 7.9 s of stall over a 60-iteration turn, for a number derivable from the elided messages alone in about 15 ms.

Blast radius is bounded to installs that opted in (`config/schema.py:160 enabled: bool = False`). Returning the delta from `emergency_shrink`, or wrapping the two calls in `to_thread`, both close it.

raven/agent/harness/memory.py
```diff
@@ -109,7 +128,250 @@ async def assemble(
...
+
+    async def shrink(
+        self,
+        messages: list[dict[str, Any]],
+        *,
+        pressure: WindowPressure,
+        state: WindowState,
+        model: str | None,
+    ) -> ShrinkResult:
+        if pressure is WindowPressure.PROACTIVE:
```
LivXue: Non-blocking, but new API surface, so worth deciding deliberately rather than inheriting.

`WindowPressure` is declared `class WindowPressure(str, Enum)` at `contracts/harness.py:38`, so `WindowPressure.PROACTIVE == "proactive"` is `True` and `json.dumps(...)` round-trips it losslessly to a bare `"proactive"`. But all five branches here (:165/167/169/171/173) dispatch with `is`, and :175 raises `ValueError(f"unknown window pressure {pressure!r}")`.

So a value that compares equal everywhere else -- one that came back through a config value, an event payload, an RPC, or a replacement shell passing the documented string -- misses all five identity checks and hits the raise. None of the five call sites wraps it (`grep 'try:|except '` over `turn_path.py:600-1050` returns nothing), so it escapes `_run_agent_loop` and ends the turn, where the contract's whole design is that a role answers "no move left".

`tests/test_harness_modules.py:368-369` pins the raise with `pressure="sideways"`, so the raise is deliberate -- my point is only that the `str` mixin invites the case the raise then punishes.

Either drop the `str` base (nothing in the repo uses it), or coerce with `pressure = WindowPressure(pressure)` at the top of `shrink`, keeping the `ValueError` for a genuinely unknown name.

raven/agent/harness/memory.py
```diff
@@ -109,7 +128,250 @@ async def assemble(
...
+        """The standing image window, in place: pictures stay while they fit the
+        budget and collapse to the newest ``image_window`` messages when they do
+        not; a picture withdrawn once stays withdrawn (see ``shrink.window_images``)."""
+        if state.image_budget is None and state.image_window >= shrink.IMAGE_WINDOW_RECENT_MESSAGES:
+            return ShrinkResult(messages, False)
+        windowed, withdrawn = shrink.window_images(
+            messages,
+            state.image_window,
+            budget=state.image_budget,
+            reason="budget" if state.image_budget else "superseded",
```
LivXue: Non-blocking, pre-existing (expression is byte-identical to base `turn_path.py:1103`). Raising it because this PR is what publishes `WindowState` as a contract-tier type with `image_budget: int | None = None` and no validator.

The withdrawal note picks its reason by truthiness (`if state.image_budget`), while the guard six lines above tests the same `int | None` field with `is None`. A zero budget takes one branch and reports the other.

Ran it. With `WindowState(image_window=2, image_budget=0)`: `:256`'s guard admits the pass because `0 is not None`; `shrink.py:233` takes the `budget is not None` branch, where `live <= 0` is `False` for any inline picture, so every image-bearing message but the newest two is withdrawn. But `0` is falsy at `:262`, so the note is composed as `superseded`. The model reads "You looked at it when it arrived; only the 2 newest image-bearing result(s) keep their pictures" for a picture actually dropped for its size -- the wrong instruction for its next move.

Unreachable through the shipped shell, which launders it at `turn_path.py:525` (`... or None`). But the contract now invites a replacement shell to construct `WindowState` directly, and the same file documents that a `0` `image_window` is meaningful -- so `0` being meaningful for one field and sentinel-ish for the other is a trap worth closing with `is not None`.

raven/agent/window/shrink.py
```diff
@@ -0,0 +1,388 @@
...
+        for i, m in enumerate(messages)
+        if (any_role or m.get("role") == "tool" or m.get(ATTACHED_IMAGE_KEY))
+        and isinstance(m.get("content"), list)
+        and any(is_image_part(p) for p in m["content"])
+    ]
+    if budget is not None:
+        live = sum(wire_image_bytes(p) for i in bearing for p in messages[i]["content"])
+        if live <= budget:
+            return 0, 0
+    stale = bearing[:-keep] if keep else bearing
```
LivXue, line 240: Non-blocking, pre-existing (byte-identical to base `turn_path.py:522`). Same reason as the sibling comment: `WindowState.image_window` is now a required, unvalidated `int` on a published contract.

`stale = bearing[:-keep] if keep else bearing` guards `keep == 0`, but a negative `keep` silently inverts the operation.

Ran it: with `keep=-1`, `bearing[:-(-1)]` is `bearing[:1]`, so over three picture-bearing tool messages it withdrew the picture from message 0 and left messages 1 and 2 holding theirs -- the exact inverse of "keep the newest N".

The `if keep` guard exists precisely because `messages[:-0]` is `[]`, so the zero edge was already noticed; this is the neighbouring one.

Route in: `_on_images_too_large`'s `state.image_window = min(state.image_window, 1)` (`memory.py:357`) propagates a negative unchanged, since `min(-5, 1) == -5`.

Latent today -- the only production `WindowState` (`turn_path.py:523`) seeds 2 and walks 2 -> 1 -> 0, and the other two `window_images` callers pass the literal `SHRINK_KEEP_RECENT_IMAGES = 1`. A `max(0, keep)` clamp, or a `WindowState.__post_init__` validator, closes it.

raven/agent/loop/main.py
```diff
@@ -500,6 +461,11 @@ def __init__(
             model=lambda: self.model,
             context_window_tokens=lambda: self.context_window_tokens,
             system_prompt=lambda skills: self.context.build_system_prompt(skills),
+            # The window's mid-turn moves read these two live for the reason the
+            # rest are callables: a hot config apply replaces the compaction
+            # settings, and the ceiling follows the model a request goes out under.
+            compaction=lambda: self._compaction,
```
LivXue, line 469: Non-blocking, but the comment is load-bearing for a future reader and is not accurate.

It justifies `compaction=lambda: self._compaction` with a hot config apply that replaces the compaction settings. A repo-wide grep (`raven/`, `plugins-dist/`, `agents/`, `evolver/`, `scripts/`, `tests/`, `docs/`) finds exactly one assignment to the loop's `_compaction`: `main.py:277`, inside `__init__`, sourced from `engine.compaction_config` unpacked at `:206`. It is never rebound.

The two hot-apply paths that exist touch nothing else: `wiring.py:1671 apply_agents` sets `self._agent_configs` and delegates to `self.subagents`, and `apply_mcp_config` reconciles MCP servers.

The contrast with the siblings the comment groups it with is what makes this misleading: provider, model and `context_window_tokens` are properties (`wiring.py:368/379/385`), so they really do move under a live `/model` switch. An operator who edits `agents.defaults.compaction.triggerRatio` and triggers a hot apply, reading :464-466 as the promise, keeps running the startup ratio in `_compact_ahead` (`memory.py:200`) and `_on_overflow` (`memory.py:293`) until the process restarts.

The clause's second half ("the ceiling follows the model") is accurate. The callable is also fine to keep as-is -- it is only the stated reason that needs correcting, or `_compaction` needs to become a property like its siblings.

raven/context_engine/segments/curator.py
```diff
@@ -173,6 +173,12 @@ async def build(self, ctx: AssemblyContext) -> Segment | None:
             describe_tool=ctx.describe_tool,
             channel=ctx.channel,
             chat_id=ctx.chat_id,
+            # Carried through, not dropped: this turn is rebuilt from the
+            # assembly context and handed back to the engine, so a field the
+            # rebuild forgets is a field the curator's own assembly silently
+            # lacks.
+            task_brief=ctx.task_brief,
```
LivXue, line 182: Non-blocking. Two small things about the two lines added here.

**They reach no reader.** The rebuilt `TurnContext` (:169) goes into `CuratorState(session_key, ctx.session_messages, ctx.budget, turn, manifest)`, and the only reader of `CuratorState.turn` in the tree is `context_engine/curator.py:969`: `"current_user_message": _snippet(state.turn.current_message, 1200)`. So `task_brief`/`task_done_when` reach no renderer under the curator engine. The brief gets to the identity through `AssemblyContext`, which `assembler.py:141-143` already fills. Harmless, just dead.

**The comment beside them is falsified by the rebuild it annotates.** It says "a field the rebuild forgets is a field the curator's own assembly silently lacks" -- but `surface` and `selected_skills` are on both `TurnContext` and `AssemblyContext` and are still dropped here. A future reader trusts the comment, assumes the rebuild is complete, and wires a segment on `turn.surface` that silently reads `None` under the curator.

Either complete the rebuild (and the comment becomes true), or soften the comment to describe what it actually is.

tests/test_harness_modules.py
```diff
@@ -326,5 +331,58 @@ def test_a_memory_built_by_hand_is_the_same_role_the_loop_binds():
...
+        assert isinstance(result, ShrinkResult), pressure
+        assert result.messages == messages, f"{pressure} left a transcript with nothing to give up alone"
+        asked.append(pressure)
+
+    assert asked == list(WindowPressure), "every pressure the paper names is answered"
+    with pytest.raises(ValueError):
+        asyncio.run(memory.shrink([], pressure="sideways", state=WindowState(image_window=1), model=None))
+
+
+def test_the_standing_image_window_closes_a_notch_when_a_picture_is_refused():
```
LivXue: Non-blocking, but it is worth knowing that this test does not cover what its name promises, because it is the only new test in the default-run suite.

`test_the_standing_image_window_closes_a_notch_when_a_picture_is_refused` names `IMAGES_TOO_LARGE` and promises "the pressures that carry a turn's retry budget advance the state the shell holds". Its body drives `WindowPressure.STANDING` -- the one pressure that touches no counter -- and asserts only `result.changed`, `"image_url" not in str(...)` and `result.messages is shown`. The word `state` never appears in an assertion.

Concretely: delete every `state.compress_retries += 1`, `state.image_demote_retries += 1`, `state.image_strip_retries += 1`, `state.reactive_summary_tried = True` and `state.head_summary_failures += 1` from `DefaultMemory` and both new tests stay green. The per-turn bounds that stop a context overflow retrying forever have no coverage in the suite CI runs.

Compounding it: `_window_memory(*, compaction=None)` at :340 is called with no argument at both :356 and :375, so `cfg` is always `SimpleNamespace(enabled=False)` and `_compact_ahead` returns at its first guard. The prune / projection / summary logic moved out of `turn_path.py` -- the largest single block this PR relocates -- is never entered by any test in the default run.

Asserting on `state` after each pressure, and passing a real compaction config at least once, would cover both.

gloryfromca (commented):
Blocking: charter checks are bypassed by the Curator registry, and the new charter-path guards are outside default CI collection.

The new review evidence changes my prior stance. I verified that a bare `ToolRegistry()` under an active charter whose rule forbids the call still executed it and returned `RAN:blocked`; the Curator constructs that same bare registry and dispatches through it inside the turn's charter scope. I also verified that default pytest collection finds zero tests from `test_subagent_charter_e2e.py`, while the explicitly named file collects 30. `tests/test_harness_modules.py` remains green (25 passed), but does not cover the missing charter join.

The existing blocking threads already state the required fixes, so I am not duplicating inline findings or replying under another reviewer's name. My earlier clean stance is withdrawn until those open blockers are addressed.

gloryfromca (commented):
Blocking: the merge-only revision leaves the Curator charter-enforcement regression and excluded charter-path tests unchanged.

I reviewed the delta from `f69e6cf7f0f1`: this head only merges three unrelated `main` commits, so the PR's feature files and affected callers are unchanged. I also checked the full thread record; the blocking threads remain open with no replies, resolution, or maintainer deferral.

Reverified on this head: a charter-forbidden call through a bare `ToolRegistry()` still returned `RAN:blocked`; default pytest collection still found zero `test_subagent_charter_e2e` tests; and `uv run --all-extras pytest tests/test_harness_modules.py -q` passed 25 tests. I covered the revision delta and history, the relevant caller path, default test collection, the standing review record, and the repository's test/domain constraints. No new inline findings.

gloryfromca (commented):
Blocking: the unchanged charter-enforcement and default test-discovery blockers still have to be resolved.

This head only merges current `main`; compared with `bb64cb073980`, it changes the session manager and its tests but leaves the PR's own patch unchanged. I found no new defect in that merge delta, but the existing blockers still reproduce on the merged tree: a bare registry executes the charter-forbidden probe, and default collection still finds zero entries from `test_subagent_charter_e2e.py`.

Covered: the project rules (`AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md`, and `CONTEXT.md`), the full PR diff and merge delta, relevant callers and history, compatibility/API concerns, test placement and whether coverage was weakened, and the harness/role architecture. No new inline findings.

Verification: `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_harness_modules.py -q` (25 passed); explicit charter integration suite (30 passed); merged session suite (37 passed); affected loop/contracts/window unit selection (192 passed). The default-collection diagnostic returned 0 charter integration entries.

gloryfromca (commented):
Blocking: move the Memory-to-engine charter join test into default collection and finish the required Charter terminology and cargo-documentation fixes.

This revision fixes the bare-registry enforcement bypass, positional compatibility, Action-role admission, recovery ownership prose, and several non-blocking observations. I found no new defect in the fix commit.

Two established issues remain. Default collection still finds zero entries from `test_subagent_charter_e2e.py`; the new default-suite tests cover `_briefed` and identity rendering separately, but not the `DefaultMemory.assemble -> _briefed -> engine` join. Removing `self._briefed(turn)` therefore still leaves the default tests green. Also, `CONTEXT.md` now defines the window terms and adds `window` to the cargo roster, but the roster still omits `harness`, while changed runtime prose continues to use `playbook` for the separately defined Charter concept.

Covered: `AGENTS.md`/`CLAUDE.md` and the Runtime context documents, the full final-comparison diff and this revision's delta, affected callers and history, positional/API compatibility, test-strength changes, and harness/import architecture. Patch-risk result: `revise` (impact high, likelihood high, protection partial, recovery easy, confidence high).

Verification at this head: focused default-suite selection, 130 passed; explicit charter integration file, 30 passed; default collection diagnostic, 0 charter-integration entries; import-linter, 10 contracts kept; `git diff --check`, passed.

ypflll:
Thanks -- this found two real holes. Fourteen of the fifteen are addressed in ec24a4e1; the three pre-existing ones I propose to send separately, reasoning at the end.

**Both blocking items, fixed.**

The registry early return was a genuine regression, and the curator path is what makes it bite. `_verifier_refusals` no longer returns early: with no provider injected it builds the default Action role, which is what the module-level helper it replaced already was. Two regressions cover it in `tests/test_subagent_charter.py` -- a bare `ToolRegistry()` inside a charter scope refuses what the rules refuse, and an injected role is the only thing asked when one is installed.

You are right that the `tests/integration` additions never ran. The two that need no external resource moved to `tests/test_segments.py`, which is in the default selection; the registry fallback is covered where the other charter tests live.

**The four "fix before merge", fixed.**

`AssemblyContext`'s `task_brief` / `task_done_when` moved to the end of the dataclass, beside where the same pair already sits on `TurnContext`, with a comment saying why the position is load-bearing. The contract digest is re-pinned.

`DefaultAction` is admitted through a `bind()` that names a missing member, exactly as the Memory role is, and `default_harness_modules` goes through it. A test asserts a decide-only role is refused at assembly.

`contracts/harness.py` now draws the line where the code does: Memory answers what to give up, the shell owns the mechanism and the retry bounds, and a role answering `shrink` with an unchanged transcript is named as the failure that would cause.

`CONTEXT.md` defines `WindowPressure`, `WindowState` and `ShrinkResult` beside the Harness Modules entry, and the cargo-contract roster names `window`.

**Six of the nine non-blocking, fixed.**

Dispatch compares with `==` rather than `is`, with a comment naming the str-Enum reason. The withdrawal note reads `image_budget is not None`, matching the guard above it. `window_images` clamps a negative keep. The `main.py` comment no longer claims a hot apply that does not exist.

On the curator: I kept the two lines and rewrote the comment, which now says the rebuild is partial on purpose and that the brief reaches the identity through `AssemblyContext`, not through here. Dropping them would make the turn the curator hands on differ from the one the rest of the assembly saw, which is a worse thing for the next reader to discover than two fields nobody reads yet.

The misnamed test is split: one asserts what the standing pass does, and a new one drives `IMAGES_TOO_LARGE` and `TOOL_IMAGES_REFUSED` and asserts the counters and the caps -- deleting any increment now turns it red.

**Three I propose to leave.**

The filed-note regex truncation, the base64 copy in `wire_image_bytes`, and the double tokenisation in `_compact_ahead` are all byte-identical to base, as you note. This PR's claim is that the move changes nothing, and a byte-level A/B against the parent is what backs that claim; a behaviour fix inside the moved code would make "the move changed nothing" no longer checkable in one comparison. I would rather send them as their own PR against `main`, where each gets the test it deserves -- the regex one in particular is data loss in the session record and should not ride along as a footnote. Say the word if you would rather have them here.

Verification after the fixes: full suite 22898 passed, with no failure `main` does not also show; ruff, format and import contracts clean at both commits; the four scripted A/Bs against the pre-fix branch are byte-identical, so none of the above changed behaviour on a covered path.

gloryfromca:
Blocking: the default suite still needs to exercise the Memory-to-engine charter join, and the remaining Section 6 documentation mismatches need correction.

Thanks for the detailed accounting. I agree that the bare-registry bypass and the other implementation/API issues named here are fixed, and I accept the filed-note truncation, base64-copy cost, and double-tokenisation as explicitly deferred non-blocking follow-ups.

Two parts do not hold yet:

1. `test_the_brief_only_reaches_the_identity_through_memory` is still only in the excluded integration file. The new default tests exercise `DefaultMemory._briefed` directly and identity rendering separately; neither calls `DefaultMemory.assemble`. Removing `self._briefed(turn)` from `assemble` therefore still leaves the default tests green. Direct invocation of the excluded join test passes, but default collection reports `0` entries for it.

2. The documentation fix addresses the new window terms and adds `window`, but the cited cargo roster still omits `harness`, and changed prose still uses `playbook` to mean the separately defined Charter concept. Those were two explicit parts of the Section 6 finding, not the three pre-existing items deferred at the end of your response.

Verification: `uv run --all-extras pytest tests/test_segments.py tests/test_subagent_charter.py -q` passed 94 tests; the named excluded join test passed; default collection found 0 copies of it. The initial focused run without extras had 93 passes and one environment failure (`raven_everos` was not installed), then the same command with the project extras passed.

gloryfromca (commented):

This revision closes the remaining concerns from my side. The default suite now exercises the real `DefaultMemory.assemble -> _briefed -> engine` join, the cargo roster includes `harness`, and the affected prose consistently uses Charter rather than the distinct Playbook term. I found no new defect in the delta.

The three explicitly deferred, pre-existing follow-ups remain named here: filed-note caption truncation, the base64 length copy, and duplicate transcript tokenisation in `_compact_ahead`. They do not block this change. LivXue owns the existing review threads, so I have not resolved or regraded them.

Covered: repository rules and Runtime context documents, the full final-comparison diff and new delta, affected callers/history, compatibility, test strength, and harness/import architecture. Patch-risk result: `merge` / `human_review_required` (impact high, likelihood low, protection strong, recovery easy, confidence high).

Verification at this head: focused default-suite selection, 131 passed; the new join regression is collected exactly once by default; import-linter, 10 contracts kept; `git diff --check`, passed.

raven/agent/tools/registry.py
LivXue: Fixed at `1b9db7558`, and by a better route than the one I suggested.

`ec24a4e12` does not hand the curator a provider -- it makes `_verifier_refusals` fall back to `DefaultAction()` when nobody handed one in, so the default is "enforce" rather than "take no opinion". That closes the whole class rather than the one site I found, which is the right call: I had only checked the ten construction sites by hand, and a fallback does not depend on that sweep having been complete.

Verified at this head: the module docstring now states the rule ("an absent provider means 'nobody chose a role here', not 'this dispatch carries no Charter'"), and `tests/test_subagent_charter.py` pins both directions -- `test_a_registry_nobody_handed_a_role_still_refuses_what_the_charter_refuses` and `test_an_injected_role_is_asked_instead_of_the_default`. Resolving.

tests/integration/test_subagent_charter_e2e.py
LivXue: Addressed at `1b9db7558`, by a different route than moving the file, and I think the right one.

The charter join now has coverage inside the default collection: `test_the_engine_is_handed_a_turn_memory_briefed_from_the_charter` in `tests/test_harness_modules.py`, plus the new `tests/test_subagent_charter.py`. So the specific hole I raised -- delete `self._briefed(turn)` and CI stays green -- is closed where CI actually looks.

`tests/integration/test_subagent_charter_e2e.py` still holds its 30 tests and still is not collected by default. That is now a pre-existing property of the directory rather than something this PR introduced, so it is not this PR's to carry. Resolving.

raven/contracts/context.py
LivXue: Fixed at `1b9db7558`. `describe_tool` is back directly after `can_see_images` on `AssemblyContext`, and the two new fields are appended, matching what `TurnContext` already did. The positional-slot shift is gone. Resolving.

raven/agent/harness/__init__.py
LivXue: Fixed at `1b9db7558`. `raven/agent/harness/action.py` gains a `bind`, `__init__.py` imports it as `bind_action`, and the construction site is now `action=bind_action(DefaultAction())`, mirroring `bind_memory`. `test_an_action_role_missing_judge_is_refused_at_assembly` pins it. Resolving.

raven/contracts/harness.py
LivXue: Fixed at `1b9db7558`. The "three in-loop recoveries stay with the L2 shell" clause is gone, replaced by an explicit split: "The in-loop recoveries are split along that line rather than sitting on one side of it. Memory answers *what to give up* (see `shrink`); the shell owns the *mechanism*." That is the distinction the `shrink` docstring already drew, now reaching the opening paragraph. The same edit also says Charter where it used to say playbook. Resolving.

CONTEXT.md
LivXue: Fixed at `1b9db7558`, all three parts.

Counted at this head: `WindowState` 0 -> 1, `ShrinkResult` 0 -> 1, `WindowPressure` 1 -> 2, `agent/window` 0 -> 2. The cargo-contract sentence now reads `tools`, `subagent`, `context`, `hook`, `personalizer`, `workdir`, `harness`, `window` -- both missing members added, so the prose matches `pyproject.toml` again. And the playbook-for-charter uses are rolled back: `harness/action.py` 2 -> 0, `contracts/context.py` 2 -> 0, `contracts/harness.py` 4 -> 2, `tools/registry.py` 4 -> 2, with the two remaining in `registry.py` being `load_playbook`, which is the real feature. Resolving.

raven/agent/window/images.py
LivXue: Not taken, which is the option I named when I raised it -- this is pre-existing code the PR relocates, and I said fixing it was your call rather than a condition of merge. Verified unchanged at `1b9db7558`: the regex still ends the facts at the first `". "` followed by an opening.

Resolving so it does not sit on the merge gate. Worth a standalone issue if the truncated session note ever shows up in a report; nothing here needs to wait on it.

raven/agent/window/images.py
LivXue: Not taken, same standing as the sibling above -- pre-existing, and I flagged it as optional. Verified unchanged at `1b9db7558`: `len(part["image_url"]["url"].partition(",")[2])`.

The measured cost stands (19.19 us/call against 0.13 us/call on a 533,358-char data URI, about 18 ms over a 60-iteration deck turn), which is small enough that deferring it is reasonable. Resolving.

raven/agent/harness/memory.py
LivXue: Not taken. Verified unchanged at `1b9db7558`: `saved = max(0, estimate_prompt_tokens(messages) - estimate_prompt_tokens(pruned))`, still two full tokenisations and still no `to_thread`.

Keeping my own framing honest: I raised this as non-blocking because the path is factory-off (`schema.py enabled: bool = False`), and that is still true, so nothing ships with the stall. The measured 131.7 ms per fire on a 254,489-token transcript is worth revisiting before compaction is ever defaulted on. Resolving.

raven/agent/harness/memory.py
LivXue: Fixed at `1b9db7558`, by the cheaper of the two routes I offered. The `str` mixin stays, but `shrink` now dispatches with `==` instead of `is`, so `WindowPressure.PROACTIVE == "proactive"` matches and a documented string value no longer falls through to the raise. The `ValueError` still fires for a genuinely unknown name, which is the behaviour `test_harness_modules.py` pins. Resolving.

raven/agent/harness/memory.py
LivXue: Fixed at `1b9db7558`. The withdrawal note's reason is now `"budget" if state.image_budget is not None else "superseded"`, matching the `is None` test in the guard six lines above. A zero budget now reports the reason it actually acted on. Resolving.

raven/agent/window/shrink.py
LivXue: Fixed at `1b9db7558` with `keep = max(0, keep)` on the line above, which is the clamp I suggested. A negative `image_window` reaching `window_images` can no longer invert the slice. Resolving.

raven/agent/loop/main.py
LivXue: Fixed at `1b9db7558`. The comment now says what is true: "The compaction settings are fixed at construction today; the callable keeps the seam uniform and costs a lambda." That removes the hot-apply promise an operator would otherwise have read into it, and keeps the accurate half about the ceiling following the model. Resolving.

raven/context_engine/segments/curator.py
LivXue: Fixed at `1b9db7558`, on the half that mattered. The comment no longer claims the rebuild is complete; it now says the opposite explicitly -- "this rebuild is partial on purpose: the curator reads `current_message` off the turn it is handed and nothing else, so the pair reaches no renderer here" -- and gives the reason for carrying the fields anyway. A future reader can no longer be misled into wiring a segment on `turn.surface`. The two lines staying is a fine call with the comment corrected. Resolving.

tests/test_harness_modules.py
LivXue: Fixed at `1b9db7558`. The test now drives `WindowPressure.IMAGES_TOO_LARGE` and asserts the counter it was named for: `assert first.changed and state.image_strip_retries == 1, "the first refusal is paid for"`, followed by a second call with the counter forced to `MAX_IMAGE_STRIP_RETRIES`. So the per-turn retry bounds are covered in the suite CI runs, which is what the original test's name promised and its body did not do. Resolving.

gloryfromca (commented):

This head only merges the latest `main`; the feature patch is unchanged, there is no manual conflict-resolution delta, and the newly combined trajectory/replay paths remain compatible with the registry changes. All fifteen review threads are now resolved by their author. The already named deferred, pre-existing follow-ups remain outside this change and are not merge conditions.

Covered: repository rules and Runtime context documents, the final-comparison diff and merge delta, affected callers and history, compatibility, test strength, and harness/import architecture. No new findings. Patch-risk result: `merge` / `human_review_required` (impact high, likelihood low, protection strong, recovery easy, confidence high).

Verification at this head: focused harness/contract selection, 131 passed; trajectory/replay/regression selection, 150 passed; import-linter, 10 contracts kept; `git diff --check`, passed.

gloryfromca (commented):

This revision only merges the latest `main` commit. Its incoming changes are confined to the two READMEs and `tests/test_cli_onboard_commands.py`, with no overlap in the feature-modified files, so the reviewed feature patch is unchanged from `a0db8b641eb2`.

I rechecked the repository rules and terminology documentation, the final diff and its caller/history context, compatibility and test integrity, and the architecture boundaries. Verification: `git diff --check github/main...HEAD`; `uv run --all-extras pytest tests/test_cli_onboard_commands.py tests/test_harness_modules.py tests/test_subagent_charter.py tests/test_segments.py tests/test_contracts_two_tier_ledger.py -q` (463 passed); `uv run --frozen --python 3.12 lint-imports` (10 contracts kept). No new findings.

LivXue (approved):

PR #469, opened 2026-09-17
refactor(*): let a sub-agent write a conduct instead of six hook phases

ypflll:
## Summary

Follows #468 (merged as `869ad78a`), which moved the charter reads and the window's mid-turn shrinking onto the harness roles. This branch is rebased onto `main` and carries only its own three commits.

A sub-agent's product logic was an `AgentHook` subclass: six phase methods over a fourteen-field context it could write, with the turn's own state kept in the context's free-form `metadata` dict. Surveyed before this change, the five bundled plugins had 71 phase methods, 42 of them doing real work, reading fourteen context fields and using six decision fields. Every one of them fits a small set of verbs. This PR makes those verbs the contract, in three commits.

**1. A sub-agent writes a conduct instead of six hook phases.** New factory-loop-tier paper `contracts/agent_conduct.py`: `AgentConduct` with nine optional verbs (`intake`, `select_tools`, `advise`, `system_addendum`, `review`, `salvage`, `outbound`, `archive`, `observe`), answered against a read-only `StepView` of fourteen fields, with `Intake` and `Verdict` (Accept / Resample / End) as the answers and a `ConductFactory` because the host builds one conduct per turn.

It sits beside `loop_hooks` rather than replacing it. The six phases remain the loop's *timing* contract: when the loop asks, in what order, what a refusal does. This is the *judgement* contract: given one step, what does this agent say about it. `ConductHook` seats one in the other, so a plugin that still ships a plain `AgentHook` keeps working and the composite's ordering, merging and rollback are untouched.

All five bundled plugins are ported: design-engine, ppt-engine, code-flow, oncall-flow and research-flow. Research keeps its six-phase gate chain: the gates now subclass a plugin-private `Gate` and read a `GateCtx` the conduct builds from its `StepView` and its own turn-private facts, so six thousand lines of gate bodies did not move and its wrappers still see the cross-gate state they read.

The turn's conduct and the system-message addendum it splices live in the turn's own `metadata` dict, not on the hook: one hook instance serves every turn a process runs, and a system turn can overlap a user one on the same chain.

**2. The conducts seat on the roles.** `MemoryModule.intake`, `PlanningModule.advise`, `ActionModule.review` and `ActionModule.salvage` each take this turn's conducts, so a replaced role decides what a plugin's judgement does rather than the seat deciding it. The composition rules live in `raven/agent/harness/conducts.py`: the first non-Accept verdict wins, advice joins in order, the first salvage wins, an intake threads the text through each conduct. `run_turn` binds the harness for the turn the way it binds the model, and the seat falls back to the same rules when nothing is bound, which is how a plugin test drives a hook directly.

Contract papers: the conduct paper is factory-loop tier, so the contract-tier digest and `CONTRACTS_VERSION` do not move with it. `CONTRACTS_LINE_CEILING` 3,200 to 3,420, each bump with its own review paragraph in `tests/test_kernel_budget.py`.

**3. What the review found.** Eleven threads, all resolved in the third commit. The ones that changed behaviour rather than prose: `CodeFlowHook` kept its `rolls_back_iterations = False` (inheriting `True` from the new base had it hold every reply behind the draft gate instead of streaming); `StepView` now publishes deep-frozen nested values, since freezing only the outer tuple left a conduct able to write through to `ctx.messages`; a turn's conducts are collected at the role boundary, so a replaced role sees one batch of the turn's judgements rather than one call per seat; an addendum is located by its own text rather than a saved offset, which two conducts appending system text used to shift out from under each other; `Planning` and `Action` gained binders, so a replacement role missing a new member is named at assembly rather than swallowed by the composite's `except Exception`. `CONTEXT.md` carries the Agent Conduct term and the Agent Hook entry no longer describes what this PR replaces.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [x] Refactor
- [ ] Other

## Verification

Gates. CI runs these at the branch tip (17 pass, 1 skipped -- the skipped row is the disabled Claude Mention workflow). Locally, at each of the three commits:

```
uv run lint-imports
uv run pytest -q
```

Import contracts: 10 kept, 0 broken at each of the three commits. The suite is green at each, and no commit adds a failure that `main` does not already show on this machine (this checkout has the heavy extras uninstalled, so the local run carries failures `main` carries too -- CI is the authority, and it is green). The loop's own e2e files, which the default selection leaves out, were run too (charter, playbook, truncation, direct chat, fallback chain, ACP stdio): no new failure.

Behaviour preservation. The oracle for this change is the hook seam itself: all five plugins driven through every phase of a multi-iteration turn under the same scripted provider, recording each phase's decision and the observers it files. Outputs normalised for nonces and temp paths, then byte-compared.

Re-run after the review fixes at the branch tip `81e6f423` (all three commits), against `main` at `c65c2059`:

| scenario | bytes |
| --- | --- |
| five-plugin hook seam | 23,840 == 23,840 |
| a whole turn with no playbook (host turn, spawned sub-agent, tool calls and refusals) | 12,821 == 12,821 |
| six window-shrink scenarios | 133,371 == 133,371 |

These cover the tip, not each commit separately. The seam comparison is the one that matters most for the third commit: its first fix changed `rolls_back_iterations` for code-flow, which decides whether replies stream or are held, and the seam oracle records the decision at every phase, so an unintended change there shows as a diff rather than as silence.

A unit test proves the seat is a real decision point rather than a pass-through: with a lenient Action role bound, a conduct's Resample is not applied, and Planning's advice and Memory's intake are what the loop receives.

Real model, Sonnet 4.5 through OpenRouter, at the branch tip:

- playbook on, in-process lane: two workers, each receives the charter (3 checks), writes its file, finishes; 26 recorded model calls
- playbook on, ACP fork lane: the charter prompt crosses the process boundary (the fork answers ZEBRA where the task text says PONG, so the brief reached the forked model through Memory), and a charter check blocks `write_file` to `notes/hi.txt` against a `pathPrefix: out/` rule
- Raven-Research, a fork running the full gate chain: `plain-first: escalated_tool`, then `sufficiency-gate: released on the pages at iteration 4 after 2 searches / 2 pages`

## Risk

No user-visible behaviour change is intended, and the hook-seam comparison is byte-identical on every phase it covers.

For plugin authors: the six-phase `AgentHook` is unchanged and a plugin shipping one keeps working. `ConductHook` is an adapter that seats a conduct in that same chain.

For anyone replacing a harness role: the role protocols gain `intake`, `advise`, `review` and `salvage`, so a replacement written against the old protocol needs them.

## Related Issues

N/A


gloryfromca (commented):
Blocking: the conduct seam has five concrete contract and behavior regressions to resolve.

I posted five inline blockers: Raven-Code loses incremental reply streaming; `StepView` aliases mutable loop data; harness roles receive singleton conducts rather than the promised turn set; multiple system addenda accumulate; and the new canonical domain term is missing from `CONTEXT.md`.

Verification performed:
- Focused changed-area suite: 294 passed.
- Full `uv run pytest`: 22880 passed, 130 skipped, 1 failed. The sole failure was the untouched `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`; this PR changes neither that test, the theme implementation, dependency declarations, nor the lockfile, and the PR's unit CI shards are green.
- `scripts/check_source_language.py` and `git diff --check`: passed.
- Direct reproductions confirmed the mutable `StepView`, singleton role dispatch, and two-addendum accumulation.

I covered the repository rules and canonical vocabulary, the complete target diff, the two commits and pre-change implementations, loop/composite/plugin callers, harness and plugin backward-compatibility surfaces, test changes for weakening, and the hook/conduct/role architecture.

agents/raven-code/plugins/code-flow/code_flow/flow.py
```diff
@@ -108,105 +111,141 @@ def project_instructions(root: Path | None, names: list[str]) -> str:
...
-            stash = observers.setdefault(ACP_META_OBSERVER, {})
-            if isinstance(stash, dict):
-                stash[MANIFEST_META_KEY] = report
-        self._ledger.end_turn(ctx.session_key)
-        return HookDecision()
+        self._ledger.end_turn(step.session_key)
+        return {ACP_META_OBSERVER: {MANIFEST_META_KEY: report}}
+
+
+class CodeFlowHook(ConductHook):
```
gloryfromca, line 225: Blocking: CodeFlowHook now buffers every model reply despite having no rollback path.

The previous `CodeFlowHook` inherited `rolls_back_iterations = False`. Through this new base it inherits `True`, while `CodeConduct` does not override `review` and therefore can only accept. `turn_path.py` treats any true hook as `holds_drafts` and routes `on_token_delta` through `DraftGate`, so ordinary Raven-Code replies are withheld until the after-iteration hook completes instead of streaming incrementally. Please preserve the false declaration for this hook or make rollback capability explicit per conduct.

raven/agent/hook/conduct.py
```diff
@@ -0,0 +1,300 @@
...
+    @staticmethod
+    def _step(ctx: Any, *, tools_ran: bool = False) -> StepView:
+        # ``getattr`` throughout: the loop hands a full ``AgentHookContext``,
+        # a phase-level test hands the two fields it cares about.
+        meta = getattr(ctx, "metadata", None) or {}
+        return StepView(
+            session_key=getattr(ctx, "session_key", "") or "",
+            iteration=getattr(ctx, "iteration", None) or 0,
+            response=getattr(ctx, "response", None),
+            transcript=tuple(getattr(ctx, "messages", None) or ()),
```
gloryfromca: Blocking: StepView does not actually isolate conducts from the loop's mutable transcript.

`tuple(...)` freezes only the outer container; its dictionaries are the same objects held by `ctx.messages`. I reproduced a conduct assigning `step.transcript[0]["content"] = "mutated"`, after which `ctx.messages[0]["content"]` was also `"mutated"`. `history`, `tools`, and `mode_overlay` have the same shallow aliasing. A conduct can therefore bypass the return-only contract and corrupt the prompt or session data this seam says it cannot write. Publish copied/frozen nested values at this boundary.

raven/agent/hook/conduct.py
```diff
@@ -0,0 +1,300 @@
...
+        if harness is None:
+            return await compose_advice(step, [conduct])
+        return await harness.planning.advise(step, [conduct])
+
+    async def _review(self, step: StepView, conduct: AgentConduct):
+        """The verdict on this step, decided by the Action role when one is bound."""
+        harness = current_harness()
+        if harness is None:
+            return await compose_review(step, [conduct])
+        return await harness.action.review(step, [conduct])
```
gloryfromca, line 228: Blocking: Harness roles never receive the turn's conduct set promised by their new protocol.

Each `ConductHook` calls the role separately with `[conduct]`. With two conduct hooks, I measured a replacement `Action.review` receiving batch sizes `[1, 1]`, never one batch of two. Composition therefore still happens in the outer `CompositeHook`, so a replacement role cannot inspect or decide how multiple plugin verdicts compose even though the new `ActionModule` contract says it composes `Sequence[AgentConduct]`. Collect the turn's conducts at the role boundary, or otherwise make the implementation and advertised contract agree.

raven/agent/hook/conduct.py
```diff
@@ -0,0 +1,300 @@
...
+        if seat.addendum is None:
+            return
+        previous, offset = seat.addendum
+        seat.addendum = None
+        system = self._system_message(ctx)
+        if system is None:
+            return
+        content = system.get("content") or ""
+        if isinstance(content, str) and isinstance(previous, str):
+            if content[offset : offset + len(previous)] == previous:
```
gloryfromca: Blocking: Addendum removal breaks as soon as two conducts append system text.

Each seat records an absolute offset before later hooks append. On the next iteration the earlier hook removes its text first, shifting the later hook's saved offset; the later lookup then misses and leaves its old text behind. A two-hook reproduction changed `base\n\nA\n\nB` into `base\n\nB\n\nA\n\nB` on iteration two. Removal needs to remain stable under other conduct addenda, with a two-addendum regression test.

raven/contracts/agent_conduct.py
```diff
@@ -0,0 +1,197 @@
+"""AgentConduct: the narrow verbs a sub-agent implements instead of the six hook phases.
```
gloryfromca, line 1: Blocking: The new Agent Conduct domain term is absent from the canonical context map.

AGENTS.md section 6 requires a new domain term to be defined in the matching `CONTEXT.md` in the same change. The current Agent Hook entry still says an agent steers the loop with its own hook, which is precisely the architecture this PR replaces for product logic. Add the canonical Agent Conduct definition and its relationship to hooks and harness roles, and update the stale Agent Hook wording.

raven/agent/harness/__init__.py
```diff
@@ -60,7 +62,28 @@ def default_harness_modules(
     )
 
 
+_BOUND: ContextVar["HarnessModules | None"] = ContextVar("raven_harness", default=None)
+
+
+@contextmanager
+def bind_harness(modules: HarnessModules):
```
LivXue, line 71: **BLOCKING.** This PR adds three required members across two role protocols, and neither role gets a binder.

| Protocol | new required member(s) | validated at assembly? |
| --- | --- | --- |
| `MemoryModule` | `intake` | yes -- `memory.py:385 bind()`, message correctly updated to name it |
| `PlanningModule` | `advise` | **no** |
| `ActionModule` | `review`, `salvage` | **no** |

`git grep -n "def bind" raven/agent/harness/` returns exactly two hits at this head: `bind_harness` (this ContextVar, not a validator) and `bind(memory)`. `HarnessModules(` still has one construction site, immediately above, and it passes `planning=DefaultPlanning()` and `action=DefaultAction()` unchecked.

What makes this worse than the same gap in #468 is where the AttributeError lands now. `ConductHook._review` calls `harness.action.review(step, [conduct])`, and `hook/composite.py:122-129` is:

```python
try:
    decision = await method(ctx)
except Exception:
    logger.exception("hook %s.%s raised; treating as no-op and continuing", hook.name, phase)
    continue
```

So a replacement Action role written against the previous protocol does not fail loudly at assembly, and does not fail loudly at the call. Every conduct's verdict silently becomes "no opinion": ppt-engine's deck checks, design-engine's ledger check and the whole research-flow gate chain all stop firing, leaving only log lines. Same shape for `PlanningModule.advise`.

`ActionModule` and `PlanningModule` are both already `@runtime_checkable`, so the guard is a direct copy of `memory.bind`. I would add both in this PR rather than a follow-up, because between #468 and #469 this is now the second time a protocol gained a required member with no binder -- which reads as a missing habit rather than a missed line.

raven/agent/hook/conduct.py
```diff
@@ -0,0 +1,300 @@
...
+    def _system_message(ctx: Any) -> dict[str, Any] | None:
+        messages = getattr(ctx, "messages", None) or []
+        return next((m for m in messages if isinstance(m, dict) and m.get("role") == "system"), None)
+
+    def _strip_addendum(self, ctx: Any, seat: "_Seat") -> None:
+        """Take this conduct's previous addendum back out of the system message."""
+        if seat.addendum is None:
+            return
+        previous, offset = seat.addendum
+        seat.addendum = None
```
LivXue: **Please fix before merge.** `seat.addendum = None` runs *before* the check that finds the text, so a strip that cannot verify its target loses the text permanently rather than retrying.

```python
previous, offset = seat.addendum
seat.addendum = None                                    # bookkeeping cleared
system = self._system_message(ctx)
if system is None:
    return
content = system.get("content") or ""
if isinstance(content, str) and isinstance(previous, str):
    if content[offset : offset + len(previous)] == previous:   # verification
        ...
```

Every path that reaches the `if` and fails it -- a stale offset, a system message whose content changed shape between `str` and `list`, another hook having edited it, `system is None` -- leaves the previous addendum in the system message while the seat forgets it exists. `before_iteration` then calls `_splice_addendum` and adds a second copy. The next iteration repeats it. The growth is one addendum per iteration, unbounded, in the message the whole turn is prompted with.

The offset goes stale in a reachable way. `_splice_addendum` stores `(part, len(content))` -- an absolute position. Two conducts that both splice give A at `offset_A` and B at `offset_B`, with B after A. On the next `before_iteration` the composite runs A first: A strips at `offset_A` and re-appends at the end, which shifts B's part left by `len(A_part)`. B then verifies at its stored `offset_B`, misses, and splices a duplicate.

Three of the five ported plugins implement `system_addendum` -- `code-flow`, `research-flow` and `design-engine`. Whether two can share one chain is your call; `code-flow` and `research-flow` are per-agent, `design-engine` is a `plugins-dist` plugin. If they can, this is live. If they cannot, it is still worth closing, because the `content` shape check is an unconditional route to the same state.

Two independent fixes, either sufficient:
- keep `seat.addendum` set unless the removal actually happened, so a failed strip is retried rather than orphaned; or
- find the previous part by value instead of by absolute offset (`content.rfind(previous)`, or `content.index(part)` for the list shape), which removes the staleness class entirely.

Test coverage: `test_a_conducts_addendum_is_replaced_rather_than_stacked` exists, but `system_addendum` appears once in the whole new test file, so only the single-conduct path is exercised. A two-conduct variant over two iterations would pin this.

raven/agent/hook/conduct.py
```diff
@@ -0,0 +1,300 @@
...
+            return self._with_trail(
+                seat,
+                HookDecision(short_circuit_result=addendum.reply, notes=[addendum.note] if addendum.note else []),
+            )
+        if addendum is not None and addendum.text:
+            self._splice_addendum(ctx, addendum.text, seat)
+        return self._with_trail(
+            seat,
+            HookDecision(
+                modified_tools=narrowed if narrowed is not None and narrowed != offered else None,
```
LivXue, line 272: `select_tools` is documented with an invariant that nothing enforces.

`contracts/agent_conduct.py:132` states it plainly: "The tool array this iteration, narrowed. **Never wider than `offered`**." Here the answer goes straight through:

```python
narrowed = await conduct.select_tools(list(offered or []), step) if offered is not None else None
...
modified_tools=narrowed if narrowed is not None and narrowed != offered else None,
```

and downstream is `composite.py:141` (`ctx.tools = decision.modified_tools`) then `turn_path.py:773` (`tool_defs = decision.modified_tools`). No filter anywhere.

To be fair about where this sits: the hook tier never promised narrowing. `contracts/loop_hooks.py:196` says `modified_tools` "is a grant, not a halting state: in `before_iteration` it **replaces** the tool schemas the model is shown". So this is not a regression -- it is a *new, stricter* claim on a new contract, made in the one commit where honouring it costs a line.

The practical effect is bounded, and I want to be clear that I am not calling it an escalation: `ToolRegistry.execute` still consults `withheld_names()` and the charter, so a widened array shows the model a tool it cannot actually run. The cost is a confused model and a wasted iteration, not a bypass.

Still worth closing, for the reason CONTEXT.md:124 gives about the charter -- "a charter may take a tool away and **may not hand one back** -- which is what makes it safe to accept from another process". A conduct is the same kind of third-party judgement. Filtering `narrowed` to names present in `offered` is one comprehension, and a test that returns a wider array and asserts it is ignored is a few lines.

raven/agent/harness/conducts.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+async def compose_review(step: StepView, conducts: Sequence[AgentConduct]) -> Verdict:
+    accepted: list[Verdict] = []
+    for conduct in conducts:
+        verdict = await conduct.review(step)
+        if not verdict.accepted:
+            return verdict
+        accepted.append(verdict)
+    if len(accepted) == 1:
```
LivXue: An accepted verdict keeps or loses its fields depending on how many conducts are loaded.

```python
if len(accepted) == 1:
    return accepted[0]                              # every field preserved
notes = [v.note for v in accepted if v.note]
return Accept(note="\n".join(notes) or None)        # kind + note only
```

`Verdict` carries six fields (`kind`, `reason`, `inject`, `overrides`, `reply`, `note`). With one conduct the verdict is handed back as written. With two or more, `reason`, `inject`, `overrides` and `reply` are dropped. So the same conduct returning the same accept verdict behaves differently in a process with one plugin than in a process with two -- which is the kind of difference that surfaces as an unreproducible report rather than a failing test.

Harmless today, because the `Accept()` factory only takes `note` and no in-tree conduct constructs `Verdict("accept", ...)` directly. But `Verdict` is exported from the contract and is constructible, so the asymmetry is reachable from outside the tree.

Two ways out, and I do not have a strong preference: either merge notes onto `accepted[0]` with `replace(accepted[0], note=...)` so the shape is the same in both arms, or state in the docstring that only `note` survives composition on an accept and make the one-conduct arm go through the same path.

raven/agent/hook/conduct.py
```diff
@@ -0,0 +1,300 @@
...
+    def _with_trail(seat: "_Seat", decision: HookDecision) -> HookDecision:
+        """The decision with whatever this turn's conduct noted along the way."""
+        trail = seat.conduct.drain_trail()
+        if not trail:
+            return decision
+        return replace(decision, notes=[*decision.notes, *trail])
+
+    @staticmethod
+    def _decide(verdict) -> HookDecision:
+        notes = [verdict.note] if verdict.note else []
```
LivXue: `Verdict.reason` reaches no reader.

`Resample(reason, *, inject, overrides, note)` takes `reason` as a required positional argument, and all five call sites in the tree supply one (research-flow `flow.py:950`, design-engine `hook.py:394`, ppt-engine `hook.py:472/483/504`). But `_decide` renders only `note`:

```python
notes = [verdict.note] if verdict.note else []
```

and a repo-wide grep finds nothing that reads `verdict.reason` in `raven/`, `plugins-dist/` or `agents/`.

No information is actually lost -- four of the five call sites pass a `note` alongside, carrying much the same text into `HookDecision.notes`. That is also what makes this worth a line of cleanup rather than a fix: the duplication is the tell that an author reading `Resample("...")` reasonably expects the string to go somewhere, writes it, then writes it again as `note` when it does not appear.

Either fold it in (`notes = [n for n in (verdict.reason, verdict.note) if n]`, which also gives the one call site with no `note` -- research-flow's -- a diagnostic line it currently does not emit), or make `reason` keyword-optional and say in the docstring that it is for the conduct's own bookkeeping.

raven/contracts/agent_conduct.py
```diff
@@ -0,0 +1,197 @@
...
+    async def archive(self, step: StepView, reply: str | None) -> Mapping[str, Mapping[str, Any]] | None:
+        """The turn is over; file what this conduct keeps across turns, and hand
+        back what it stamps on the turn's record -- observer name to counters --
+        for the host to file with the reply."""
+        return None
+
+    async def observe(self, step: StepView) -> None:
+        """Record without deciding."""
+
+    def note(self, text: str) -> None:
```
LivXue, line 179: `note()` and `drain_trail()` write through `self.__dict__`, which quietly makes `AgentConduct` a class a conduct must inherit rather than a shape it may implement.

```python
def note(self, text: str) -> None:
    self.__dict__.setdefault("_trail", []).append(text)

def drain_trail(self) -> list[str]:
    return self.__dict__.pop("_trail", [])
```

Going through `__dict__` to spare subclasses a `super().__init__()` is a reasonable trick and I would keep it. Two consequences are worth writing down rather than leaving to be discovered:

**A conduct with `__slots__`, or a frozen dataclass conduct, raises here.** Nothing in the paper suggests a conduct may not be either -- `StepView`, `Intake` and `Verdict` on the same page are all frozen dataclasses, which makes that a natural shape for an author to reach for.

**A duck-typed conduct fails as a silent no-op.** `ConductHook._with_trail` calls `seat.conduct.drain_trail()` on every phase's return, unconditionally. An object implementing all nine verbs but not inheriting `AgentConduct` raises `AttributeError` there, and `composite.py:122-129` catches it, logs, and continues -- so the phase becomes a no-op and the plugin looks inert. Same swallowing path as B1.

The paper's opening calls this "the narrow verbs a sub-agent implements", which reads as structural typing. One sentence saying a conduct must subclass `AgentConduct` (and that the base owns `_trail`) resolves it. If it is meant to be structural instead, `drain_trail` wants a `getattr(conduct, "drain_trail", None)` guard at the call site.

LivXue (commented):
## Review: verified against `2f827373` vs its base `f69e6cf7` (the #468 branch)

**Method note, because it changes how this review should be weighed.** My review of
#468 ran a multi-angle sweep that converged 24 candidates down to 15 verified findings.
That tooling was unavailable for this PR (org spend limit), so what follows is a manual
read. It is narrower in coverage than the #468 review, not deeper. Treat an absence of
findings in an area below as "not swept", not as "clean".

Taken on its own, this is the better-built of the two. The timing/judgement split is the
right seam, the `AgentConduct` paper is precise about what a conduct may and may not
reach, all five bundled plugins are actually ported rather than shimmed, and
research-flow keeping six thousand lines of gate bodies behind a `Gate`/`GateCtx`
adapter is the correct trade. Ten new tests, and -- unlike #468 -- all ten are in
`tests/`, where the default collection actually finds them.

Three things about the stack, before the findings:

**1. This PR carries every unfixed finding from #468.** Its base is that branch, so
those two commits are in this PR's history, including the blocking one
(`registry.py:407`: the charter's `checks`/`code` enforcement became an opt-in
constructor kwarg, and `segments/curator.py:357` builds a bare `ToolRegistry()` that
does dispatch inside a live charter scope). #468 has to land first, and land fixed.

**2. `mergeStateStatus: CLEAN` here means clean against the #468 branch, not `main`.**
Worth stating because the label reads as merge-ready.

**3. The retarget will not be automatic.** The description says the base "will retarget
to `main` once it merges". Because the repo squash-merges, #468's two commits will not
be ancestors of `main`, so GitHub's retarget leaves this branch's diff re-containing all
of #468's work as if it were new. This needs
`git rebase --onto main refactor/charter_and_window_on_memory` after #468 lands, not
just a base change.

**Findings.** One I would call blocking; the rest are graded inline.

**B1. Three new required Protocol members have no assembly-time validation, and the
failure is silent and permissive.** `MemoryModule` gains `intake` and its `bind()`
message was correctly updated. `PlanningModule` gains `advise` and `ActionModule` gains
`review` and `salvage` -- and neither has a binder at all. Detail at
`harness/__init__.py:69`.

Then, in descending order: `hook/conduct.py:221` (an addendum that fails to strip is
never re-found, and the next iteration splices another copy),
`hook/conduct.py:205` (`select_tools` states an invariant nothing enforces),
`harness/conducts.py:51` (an accept verdict keeps or loses its fields depending on how
many plugins are loaded), `hook/conduct.py:130` (`Verdict.reason` is required and has no
reader), and `contracts/agent_conduct.py:173` (`note()` writes through `self.__dict__`,
and the host calls `drain_trail()` unconditionally).

None of these argues against the design. The seam is right; these are places where a
newly written contract does not yet hold itself to what it says.

gloryfromca (commented):
Blocking: the five previously reported conduct-seam defects still need fixes before merge.

At this head, all five existing blocking threads still reproduce. In particular, the new harness-role indirection continues to call each role with singleton conduct batches, so it does not settle the open role-composition finding. I found no additional defect worth raising and did not duplicate the other reviewer's open items.

Coverage: I reviewed the current diff against the updated target, the range-diff and commit history, affected callers and harness binding, backward-compatibility/admission surfaces, test changes for weakening, the stated architecture constraints, and the repository rules in AGENTS.md/CLAUDE.md plus CONTEXT-MAP.md/CONTEXT.md. The Agent Conduct context-map omission also remains.

Verification: the changed-area suite passed (328 tests). The full `uv run pytest -x` run stopped at the same untouched CLI-theme failure as the prior revision, after 11,232 passed and 13 skipped; all GitHub checks are green. `git diff --check` and the source-language check passed. Targeted reproductions reconfirmed the four behavioral findings.

agents/raven-code/plugins/code-flow/code_flow/flow.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed at this head. `CodeFlowHook` now declares `rolls_back=False` for its non-reviewing conduct, and I verified the instance reports false, so ordinary replies are no longer routed through the draft gate. Closing this thread.

raven/agent/hook/conduct.py
gloryfromca: No blockers; this can merge as far as I am concerned.

The concrete write-through is fixed: StepView rows are copied into dict-compatible read-only rows, their mutators raise, and the original assignment no longer changes the loop context. The implementation also states its shallow boundary explicitly. Closing this thread.

raven/agent/hook/conduct.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Settled by the revised contract: a role composes the conducts represented by one seat, while the hook composite retains cross-plugin composition. That accurately describes the measured `[1, 1]` calls and removes the promise this objection relied on. Closing this thread.

raven/agent/hook/conduct.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed at this head. Addenda are located by value instead of a stale absolute offset, bookkeeping is cleared only after successful removal, and the new two-seat regression stays at `base\n\nA\n\nB` across three iterations. Closing this thread.

raven/contracts/agent_conduct.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed at this head. `CONTEXT.md` now defines Agent Conduct, relates it to hooks and harness roles, and removes the stale claim that bundled agents steer the loop through their own hooks. Closing this thread.

gloryfromca (commented):

This revision settles all five findings I had open. I replied on and resolved each thread; no new candidate survived refutation. At this fourth review round, I found no defect introduced by the fix that is both reachable in ordinary use and leaves an operator without a working way out. I did not re-grade the other reviewer's threads.

Coverage: the full current diff and the new range-diff/history; AGENTS.md and its identical CLAUDE.md, CONTEXT-MAP.md and CONTEXT.md; affected callers and harness binding; backward compatibility; test changes for weakening; and the conduct/hook/harness architecture constraints.

Verification: the changed-area suite passed (333 tests). The full `uv run pytest -x` run stopped on the same untouched CLI-theme failure as the prior two revisions after 11,241 passed and 13 skipped. Ruff check and format, `git diff --check`, and the source-language gate passed. GitHub had not attached check runs to this head at review time; its `dirty` mergeability value disagreed with both the commit graph (the base is an ancestor) and a clean local `merge-tree`, so I did not treat that metadata as a code finding.

raven/agent/harness/__init__.py
LivXue: Fixed at `7a46bf6fd`. Both binders are in place and applied at the one construction site: `planning=bind_planning(DefaultPlanning())` and `action=bind_action(DefaultAction())`, alongside the existing `bind_memory`. All three roles now name a missing member at assembly instead of letting it become an AttributeError that `composite.py`'s catch converts into "no opinion". Resolving.

raven/agent/hook/conduct.py
LivXue: Fixed at `7a46bf6fd`, using both of the routes I offered rather than either alone.

The lookup is by value now (`content.rfind(previous)` for the string shape, a reverse equality scan for the list shape), so a stale absolute offset cannot exist; and the bookkeeping is cleared only after the text is actually removed, so a strip that finds nothing is retried rather than orphaned. Either change alone would have closed the growth; having both means the two-conduct ordering case and the content-shape case are covered separately.

The docstring now carries the reasoning too, including the second-seat ordering that made the offset go stale. Resolving.

raven/agent/hook/conduct.py
LivXue: Settled at `7a46bf6fd` -- and the fix went the other way from what I proposed, correctly.

The contract no longer claims narrowing. It now reads: "Usually narrower, and not required to be: a conduct may hand back an array carrying a tool its own product contributes, which is what the research flow does with its escalation tool. Handing a name back is not granting it -- `ToolRegistry.execute` still adjudicates every call, so the product's own withholding stands whatever this returns."

I had this backwards. I found that the doc and the code disagreed, which was right, and then assumed the code should be made to honour the doc. The one-comprehension filter I suggested would have broken research-flow's escalation tool. The doc was the wrong half.

The replacement is also a better safety argument than the charter analogy I reached for: the invariant that matters is adjudication at execute, not narrowing at selection, and that one is actually enforced. Nothing left open here. Resolving.

raven/agent/harness/conducts.py
LivXue: Fixed at `7a46bf6fd` with `replace(accepted[0], note="\n".join(notes) or None)`, which is the first of the two options and the one that keeps the shape identical in both arms. An accept verdict now carries whatever else it was written with whether one conduct answered or five. The comment records the reason, which is the part worth keeping. Resolving.

raven/agent/hook/conduct.py
LivXue: Fixed at `7a46bf6fd`. `_decide` now folds both fields with de-duplication: `notes = list(dict.fromkeys(n for n in (verdict.reason, verdict.note) if n))`.

That covers the case I cared about most and did not spell out well -- research-flow's `Resample` passes a reason and no note, so it previously emitted no diagnostic line at all; it now does. And the de-dup means the four call sites that pass near-identical text in both fields do not print it twice. Resolving.

raven/contracts/agent_conduct.py
LivXue: Fixed at `7a46bf6fd`, via the structural-typing route rather than by requiring inheritance.

`_with_trail` now guards the call: `drain = getattr(seat.conduct, "drain_trail", None)` / `trail = drain() if callable(drain) else []`, with a docstring that states the position directly -- "a conduct that implements the nine verbs without inheriting `AgentConduct` is a conduct, and asking it for a method it never claimed would raise into the composite's catch".

That answers the silent-no-op half, which was the one with teeth. It also settles the `__slots__` question by implication: such a conduct simply does not use the base's trail. Resolving.

gloryfromca (commented):

This is a rebase-only revision. The range-diff shows the three PR commits are patch-equivalent to the clean prior head except for adopting the target branch's canonical `Charter` terminology in harness prose. No settled finding regressed, and I found no new issue.

Coverage: the current full diff against the refreshed target, range-diff and history, AGENTS.md/CLAUDE.md and CONTEXT files, affected callers, backward compatibility, test changes for weakening, and the conduct/hook/harness architecture constraints. All five threads I opened remain resolved.

Verification: the relevant suite passed (410 tests); Ruff check and format, `git diff --check`, and the source-language gate passed. All completed GitHub checks are green; the unit shards and Windows self-upgrade were still running when reviewed.

gloryfromca (commented):

The former target has merged into `main`, and this revision retargets/merges `main` without changing the reviewed conduct implementation. The range-diff keeps the three PR commits equivalent; only the contract-budget measurements and ceiling were refreshed for the larger base, and their test passes. The merge commit's README conflict resolution does not appear in the PR diff against `main`. No settled finding regressed and I found no new issue.

Coverage: the full current diff against refreshed `github/main`, range-diff and history, merge conflict resolution, AGENTS.md/CLAUDE.md and CONTEXT files, affected callers, backward compatibility, test changes for weakening, and the conduct/hook/harness architecture constraints. All five threads I opened remain resolved.

Verification: the relevant suite passed (410 tests); Ruff check and format, `git diff --check`, and the source-language gate passed. All completed GitHub checks are green; unit shards and TUI were still running when reviewed.

gloryfromca (commented):

This revision only merges current `main`. The range-diff shows all three PR commits are exactly patch-equivalent to the prior clean head, and the merge has no manual conflict-resolution hunks. No settled finding regressed and I found no new issue.

Coverage: the full current diff against refreshed `github/main`, range-diff and history, merge result, AGENTS.md/CLAUDE.md and CONTEXT files, affected callers, backward compatibility, test changes for weakening, and the conduct/hook/harness architecture constraints. All five threads I opened remain resolved.

Verification: the relevant suite passed (410 tests); Ruff check and format, `git diff --check`, and the source-language gate passed. All completed GitHub checks are green; unit shards were still running when reviewed.

LivXue:
The description has not moved since the retarget and the rebase, and this repo squash-merges with `squash_merge_commit_message=PR_BODY` -- so whatever stands here becomes the permanent commit body on `main`. Three groups, in descending order of how much they matter.

**1. The Verification section does not cover the commit that carries the review fixes.**

The branch now has three commits of its own -- `3e426664d`, `b2c12dc51`, `62c24d387` -- but the description was written for two, and still says so in five places: "This PR makes those verbs the contract, in two commits" (line 5), "Gates, run at each of the two commits" (line 30), "10 kept, 0 broken at both commits. The suite is green at both" (line 43), "Both commits were run against a worktree of their own parent ... identical at both commits" (line 45), and "Real model, Sonnet 4.5 through OpenRouter, at both commits" (line 49).

The third commit is `fix(agent): a seat declares its own rollback, and a step is read-only`, which is where the review fixes landed. So as written, the byte-compare oracle and the real-model runs are claimed for everything except the commit a reader would most want them claimed for.

If the gates and the two behaviour-preservation runs were re-done at the third commit, this is a word change (two -> three, both -> all three). If they were not, saying so plainly is better than a number that reads as covering it. Your call which -- I cannot tell from outside.

**2. The opening paragraph is now false in all three of its clauses.**

> Stacked on #468, which moves the charter reads and the window's mid-turn shrinking onto the harness roles. Review that one first; this PR's base is its branch and will retarget to `main` once it merges.

#468 merged at 02:52:48Z as `869ad78a`; the retarget fired two seconds later; the rebase landed at 05:43Z. Verified on the current head: `869ad78a` is an ancestor and `1b9db755` is not, and the diff is back to 42 files.

Something like:

> Follows #468 (merged as `869ad78a`), which moved the charter reads and the window's mid-turn shrinking onto the harness roles. This branch is rebased onto `main` and carries only its own three commits.

**3. Four `<base>` placeholders were never substituted** (lines 37-40):

```
npx commitlint --from <base> --to HEAD --config commitlint.config.cjs
uv run python scripts/check_commit_messages.py <base>..HEAD
uv run python scripts/check_source_language.py <base>..HEAD
uv run python scripts/check_large_files.py <base>..HEAD
```

`origin/main` is the value now, and is what #468's description used.

---

Nothing here is about the code. CI is green on `main` (17 pass, 1 skipped -- the skipped row is the disabled Claude Mention workflow), the rebase is clean, all eleven review threads are resolved, and I re-verified on `fb10260c` that the six fixes from my earlier review are all still present after the two rewrites. The only merge-gate item left is an approving review.

gloryfromca (commented):

This head only merges current `main`. The range-diff shows all three PR commits are exactly patch-equivalent to the prior clean head, and the merge contains no manual conflict-resolution hunks. No settled finding regressed and I found no new issue. The existing top-level PR-description note does not change the code stance or ask this reviewer for a reply.

Coverage: the full current diff against refreshed `github/main`, range-diff and history, merge result, AGENTS.md/CLAUDE.md and CONTEXT files, affected callers, backward compatibility, test changes for weakening, and the conduct/hook/harness architecture constraints. All five threads I opened remain resolved.

Verification: the focused suite passed (333 tests); Ruff check and format, `git diff --check`, source-language, and large-file gates passed. All four GitHub unit shards and every other completed check are green; the aggregate coverage gate was still pending when reviewed.

ypflll:
All three are fair, and the answer to the open question in (1) is yes -- re-run, but at the tip rather than per-commit, which is worth stating precisely rather than as "all three".

**What was re-run, and where.** Baseline `main` at `c65c2059`, subject the branch tip `81e6f423` (so all three commits plus the three merge commits). Same scripted provider on both sides, outputs normalised for nonces and temp paths, byte-compared:

| scenario | bytes |
| --- | --- |
| five-plugin hook seam, every phase's decision and the observers it files | 23,840 == 23,840 |
| a whole turn with no playbook (host turn, spawned sub-agent, tool calls and refusals) | 12,821 == 12,821 |
| six window-shrink scenarios | 133,371 == 133,371 |

Real model, Sonnet 4.5 through OpenRouter, at the same tip:

- playbook on, in-process lane: two workers, each receives the charter (3 checks), writes its file, finishes; 26 recorded model calls
- playbook on, ACP fork lane: the charter prompt crosses the process boundary (the fork answers ZEBRA where the task text says PONG, so the brief reached the forked model through Memory), and a charter check blocks `write_file` to `notes/hi.txt` against a `pathPrefix: out/` rule
- Raven-Research: `plain-first: escalated_tool`, then `sufficiency-gate: released on the pages at iteration 4 after 2 searches / 2 pages`

The gates are covered at the tip by CI rather than by me: 17 pass, 1 skipped.

So the accurate claim is "re-run at the tip, which includes the third commit", not "at each of the three commits". The seam comparison is the one that matters most for that commit, because the first fix changed `rolls_back_iterations` for code-flow -- which decides whether replies stream or are held -- and the seam oracle records the decision at every phase, so an unintended change there would show as a diff rather than as silence.

(2) and (3) accepted as written -- the opening paragraph and the four `<base>` placeholders. Rewriting the description now with your suggested opening, `origin/main` substituted, and the Verification section saying what the table above says.

LivXue (approved):
Approving at `81e6f423`.

**What I verified myself, at this head:**

- All six findings from my review of 2026-09-17 are settled: `bind_action`/`bind_planning` are applied at the construction site; `_strip_addendum` finds its text by value and clears the bookkeeping only after the removal; `Verdict.reason` folds into the notes with de-duplication; `compose_review` keeps the first accepted verdict's fields via `replace`; `drain_trail` is asked for with `getattr`. The sixth, `select_tools`, was settled by correcting the contract rather than the code -- I had the direction wrong, and the research flow's escalation tool is why.
- Those six survive the two history rewrites: re-checked after the rebase onto `main` and again at this head.
- The rebase is clean: `869ad78a` (the #468 squash) is an ancestor, `1b9db755` is not, and the diff is back to its own 42 files.
- Eleven review threads, zero unresolved. CI at this head: 17 pass, 1 skipped -- the skipped row is the disabled Claude Mention workflow, not a check that passed.
- The description is accurate as of this head, which matters because it becomes the commit body on `main` under `squash_merge_commit_message=PR_BODY`. It now separates the gates (run per-commit locally, and at the tip by CI) from the byte-compare (tip only, stated as such), and it names the local checkout's missing extras rather than claiming a green that the machine cannot produce.

**What I am relying on others for.** My review of this PR was a manual read, narrower in coverage than the multi-angle sweep I ran on #468 -- I said so in the review and it still holds. The two findings with the most user-visible weight were not mine: the streaming regression (`ConductHook` declaring `rolls_back_iterations` on the class, so code-flow and oncall-flow inherited `True` and an ordinary Raven-Code answer stopped streaming) and the mutable `StepView` rows (the tuple froze the sequence, not the dicts inside it, so a conduct could rewrite the transcript the loop prompts with through a field the paper calls read-only). Both came from @gloryfromca's reproductions, and both are fixed in `62c24d387`.

**Not covered by this approval.** I did not independently re-run the byte-compare oracle or the real-model lanes; those numbers are the author's, stated with their baselines (`main` at `c65c2059`, subject `81e6f423`) and reproducible from the description.

One practical note: `dismiss_stale_reviews_on_push` is on for this repo, so a further `main` sync push will drop this approval. `main` has moved to `ef6afc57` since this head, and the ruleset does not require the branch to be current -- so this can merge as it stands, and re-syncing first would cost the approval.

PR #472, opened 2026-09-17
fix(context_engine): the identity line names the model the turn reaches

### LINKED ISSUE #470: The system prompt tells the agent it runs on the configured default model, so a per-session switch looks like it never happened
## Summary

After switching a session's model from the page, the agent keeps introducing itself as the previous model. The switch itself works: the session record carries the new id, the request goes to the new model, and the usage ledger bills the new model. What is wrong is the system prompt, which tells the agent it is running on `agents.defaults.model` -- the configured default -- regardless of the session's pick. The agent then repeats that line back, so from the page it looks exactly like the switch never happened.

## Steps to reproduce

1. Configure a default model (call it A) and start a conversation.
2. Switch the conversation to another model (B) from the composer's model control.
3. Ask: "what model are you running on?"

## Expected behavior

The agent names B, the model the request actually reached.

## Actual behavior

The agent names A, on that turn and on every following one. Its own reasoning trace shows why -- paraphrased from the recorded turn: *"according to the system information, I am running on model A."* It is quoting the prompt, not guessing.

Meanwhile, on the same two turns, everything else says B:

| source | model |
| --- | --- |
| `session.metadata["model"]` (written by `_remember_session_model`, `raven/rpc/methods/config.py`) | B |
| `turn_path.py:2361` -- `routed_model = session.metadata.get("model")`, the id the request is built for | B |
| usage ledger (`telemetry/usage-*.jsonl`), both calls | B |
| system prompt line the agent read | **A** |

## Where the two readings diverge

`render.identity_text` writes the line (`raven/context_engine/segments/render.py:349-351`):

```python
resolved_model = model if model is not None else _resolved_model_id()
model_line = f"\nYou are running on model: {resolved_model}." if resolved_model else ""
```

Its docstring is explicit that `model` should be "the resolved routed model id ... told to the model so it never guesses its own identity from pretraining", and that `None` "resolves it lazily from config". `_resolved_model_id()` reads `config.agents.defaults.model` (`render.py:204-`). It is session-blind by construction.

The per-turn caller never passes `model`. `IdentitySegmentBuilder.build` (`raven/context_engine/segments/identity.py:86-90`):

```python
text = render.identity_text(
    self._workspace,
    specialists=self._specialists(),
    dispatch_tools=render.live_dispatch_tools(self._get_tool_definitions),
)
```

So the lazy config fallback wins on every turn, and the switch can never reach the line.

The value is available upstream and is dropped on the way down:

- `turn_path.py` computes `routed_model` ahead of assembly precisely so "one model's verdict" does not shape "a message another model receives", and passes `model=routed_model` into `_assemble_context_messages`.
- `organ_glue._assemble_context_messages` (`raven/agent/loop/organ_glue.py:199`) uses it for one thing only, `can_see_images=self._supports_vision(model)`, and builds a `TurnContext` that has no model field.
- `TurnContext` (`raven/contracts/context.py:216`) and `AssemblyContext` (`raven/contracts/context.py:60`) both lack a model field, so no segment builder can read it.

## Why it was easy to leave this way

`IdentitySegmentBuilder.stable = True`, with the comment "SOUL.md and the identity header: the same text on every turn". The identity block is the prompt-cache stable prefix, and a model line was presumably taken as constant. A per-session pick is still constant within a session, so threading it keeps the prefix stable; a mid-session switch changes the prefix once, which is what a switch is.

## Test coverage

`tests/test_segments.py` pins the two halves of `identity_text` in isolation: an explicit `model=` wins (line 126), and the lazy fallback is used when it is `None` (130-134). Nothing asserts that the per-turn identity segment passes the session's model. The three files that mention `IdentitySegmentBuilder` (`test_segments.py`, `test_context_invariants.py`, `test_context_stable_prefix.py`) contain no model assertion, which is why this is green.

## A direction, for whoever owns the context engine

Carry the model on `TurnContext` and `AssemblyContext` (organ_glue already has it in hand), and have `IdentitySegmentBuilder.build` pass `model=ctx.model`. One regression: assemble a turn whose session carries a model different from `agents.defaults.model`, and assert the rendered identity names the session's. Whether a mid-session switch should also invalidate anything the stable prefix is cached against is the one question worth deciding alongside it.

## Environment

macOS, `raven` from the repo working tree at current `main`, web UI. Reproduced on a fresh session: two consecutive turns, both billed to B, both answering A.

Handsome-wzw:
## Summary

After a conversation switches model from the page, the agent keeps introducing itself as the model it left. The switch itself works -- the session record, the request and the usage ledger all carry the new id -- but the system prompt's `You are running on model: ...` line was read from `agents.defaults.model`, so the agent was told it still runs on the configured default and repeated that back. From the page it is indistinguishable from a switch that never happened.

The fix reads the id from the turn's active `ModelBinding` instead. The loop already opens `use_binding(binding)` around every turn (`raven/agent/loop/main.py`, around `_run_turn`), and that binding is the session's own `/model` pick when it has one (`binding_for_session` in `raven/agent/loop/wiring.py`), else the default. `render._resolved_model_id` now takes its model id from `active_binding()` when one is set and from config outside a turn, and runs either through the same storage-to-wire conversion (`providers.wire`) it always ran the default through. One source file changes; nothing is threaded through the assembly path.

This is the approach the review suggested over the earlier revision of this PR, which carried the id down as a new contract field, and it is better on every point that revision was faulted for:

- **Right spelling.** The earlier revision forwarded `session.metadata["model"]`, the storage spelling, and rendered it as-is; on Codex, Azure and every gateway install that is not the id the request carries. Here the bound id goes through `wire_model` exactly as the default does, so `openai-codex/gpt-5.1-codex` renders as `gpt-5.1-codex` and a gateway install keeps its `openrouter/` prefix, in both branches. The conversion stays on the one path that owns it, and a lazily built provider's identity `wire_model_id` is not relied on.
- **The estimation path agrees by construction.** `ContextBuilder._get_identity` (`raven/agent/context/builder.py`) sizes the system prompt for the token budget and calls the same renderer with no arguments; it is invoked from `_assemble_context_messages`, inside the binding, so it now names the same model the assembled prompt does. A field handed down the assembly path reached the segment and not this caller.
- **No contract change.** `AssemblyContext` and `TurnContext` are untouched, so there is no field position to argue about, no producer (the curator's `TurnContext` rebuild) that can forget to copy it, and `CONTRACTS_VERSION` stays where `main` has it.
- **`stable = True` stays true.** A binding is per session, constant within it; the line changes once per switch, which is what a switch is. A router's per-message pick is not reflected in the line (the router is opt-in and off by default); that is the one case a contract field would have covered, and it is also the case that made the segment's stability claim false.
- **No empty-string hole.** `ModelBinding` refuses an empty model id at construction, so the line cannot be deleted by a blank pick.

A fallback hop further down a model chain still reads a prompt naming the primary: context is assembled once, before the call. Not changed here and not a regression -- the sibling `can_see_images` field documents the same caveat.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Five new tests in `tests/test_segments.py::TestIdentityNamesTheBoundModel`:

- `test_a_switched_conversation_is_told_the_model_it_is_bound_to` -- inside `use_binding` on a gateway config, `_resolved_model_id()` names the bound model with the gateway prefix and never the configured default.
- `test_the_bound_id_goes_through_the_same_storage_to_wire_conversion` -- a codex-form id bound to the turn renders as the wire spelling the client sends, not the stored one; this is the assertion the earlier revision could not fail.
- `test_outside_a_turn_the_configured_default_still_answers` -- no binding, config default, as before.
- `test_the_segment_names_the_bound_model` -- the whole segment, through `IdentitySegmentBuilder.build`.
- `test_the_estimation_prompt_and_the_turn_prompt_agree_on_a_switched_model` -- `ContextBuilder._get_identity()` equals the segment's text inside a binding and names the switched model.

Commands and results:

- Mutation-checked: replacing the binding read with the config read fails exactly the four tests that bind a model (4 failed, 46 passed in `tests/test_segments.py`); the outside-a-turn test and the three pre-existing wire-form tests stay green.
- `.venv/bin/python -m pytest` over every test file that names `identity_text`, `_resolved_model_id`, `IdentitySegmentBuilder` or `ContextBuilder(`, plus `tests/test_contracts_two_tier_ledger.py`, `tests/test_kernel_budget.py` and `tests/test_context_stable_prefix.py` (11 files): 375 passed.
- `python scripts/coverage_gate.py diff --base-ref upstream/main --threshold 90` after those suites ran under `--cov=raven --cov-branch`: 100% (3/3 executable changed lines).
- `.venv/bin/ruff check`, `.venv/bin/ruff format --check` and `.venv/bin/ty check` on `raven/context_engine/segments/render.py`: clean.
- The contracts gates pass unchanged: this revision touches nothing under `raven/contracts/`.
- No user-facing docs change: the prompt line already documents itself as naming the model the turn runs on.

## Risk

- Security: none. The binding's model id is already in every request and in the usage ledger; nothing new is read or sent.
- Backward compatibility: outside a turn (no binding) the renderer behaves exactly as before. Inside a turn the line names the bound model instead of the configured default; for a conversation that never switched they are the same id, so the rendered prompt is byte-identical there. No wire, config or contract change.
- User-visible change: after a `/model` switch the agent names the switched model in its identity line from the next turn on. The identity prefix changes once per switch, so a prompt-cache prefix keyed on it is invalidated once per switch, not per turn.
- Rollback: revert the squash commit; one file, nothing else to reconcile.

## Related Issues

Fixes #470


raven/contracts/context.py
```diff
@@ -82,6 +82,17 @@ class AssemblyContext:
     ``tui`` channel on purpose (see ``Source.surface``), so the channel alone
     cannot say whether the reader is at a prompt or in a browser.
     """
+    model: str | None = None
```
gloryfromca: Blocking: bump the contract version and repin the surface digest for this exported field.

`AssemblyContext` is part of the frozen contract tier, so adding `model` changes its declared surface. `tests/test_contracts_two_tier_ledger.py::test_the_version_pin_matches_the_papers` now fails because `CONTRACTS_VERSION` remains `26` and `PINNED_CONTRACT_SURFACE` still carries the old digest. The contract ledger explicitly requires both to move when a field is added.

raven/contracts/context.py
```diff
@@ -82,6 +82,17 @@ class AssemblyContext:
...
+    """The id the request will actually reach, when the loop resolved one.
+
+    A session's own pick, else the router's, decided ahead of assembly (see
+    ``turn_path``). Carried so a segment that tells the model about itself
+    names this and not the configured default: the two differ for every
+    conversation that switched model, and a segment reading config would tell
+    a switched conversation it is still running on the model it left.
+    ``None`` means the loop did not resolve one, and a segment falls back to
+    what config says, as it always has.
+    """
```
gloryfromca: Blocking: restore the contracts tier to its enforced line budget.

These additions take `raven/contracts` from 3,033 lines on `github/main` to 3,049, over the 3,040-line ceiling. The full-suite run therefore fails `tests/test_kernel_budget.py::test_the_papers_stay_under_their_line_ceiling`. The new contract material needs to fit within the existing budget, or a ceiling change needs the explicit reviewed treatment described by that guard.

gloryfromca (commented):
Blocking: update the contract ledger and bring the contracts tier back under its line ceiling.

I found two branch-caused contract-gate failures, both marked inline. The model value itself follows the existing selection path correctly: I checked the full diff, the turn-selection and request-dispatch callers, renderer fallback behavior, prompt-cache implications, relevant history, backward compatibility for callers that omit the field, test changes, and the repository rules in AGENTS.md, CLAUDE.md, and CONTEXT-MAP.md/CONTEXT.md. I found no weakened tests.

Verification:
- `uv run --extra dev pytest tests/test_context_surface.py tests/test_read_file_image.py tests/test_segments.py tests/test_context_invariants.py tests/test_context_stable_prefix.py tests/test_default_context_engine.py -x`: 211 passed.
- `uv run --frozen --python 3.12 --all-extras pytest -q -x`: stopped after 6,835 passed and 3 failed. Two failures are the inline blockers. The third is the untouched `tests/test_cli_theme.py::test_bold_accent_renders_styled_not_bare`, so I did not attribute it to this change.
- Ruff check and format, `ty` on changed production files, source-language check, commit-message check, and `git diff --check`: passed.

raven/contracts/context.py
Handsome-wzw: Done in eb443604, pushed after this review was written. `CONTRACTS_VERSION` moves `26` -> `27` in `raven/contracts/__init__.py`, and `PINNED_CONTRACT_SURFACE` is repinned to `("27", "3be9d88871ea78d6511a9e77a73cd929f2c0ff97d34dcc9e0539ee8b3fcd70be")`, the digest the gate itself reported. Recomputed locally with `contract_surface_digest(raven/contracts)` on the pushed head: matches the pin; `test_the_version_pin_matches_the_papers` passes.

You are right that this is a shape change and not prose: an optional field on two contract-tier dataclasses is exactly what the ledger exists to make visible. I had read the gate's message as advice for a later step rather than a requirement of this one, and the guard is what caught that.

raven/contracts/context.py
Handsome-wzw: Done in eb443604. `raven/contracts` is back to 3,038 lines against the 3,040 ceiling (was 3,049), measured with the same formula the gate uses; `test_the_papers_stay_under_their_line_ceiling` passes. The ceiling itself is untouched -- its message says a change to it is a reviewed decision, not a bump in passing, and nothing here earns one.

What went: the two field docstrings. They carried the reasoning for the field -- why the identity line must name the turn's model and not config -- but that reasoning already lives beside the code that acts on it (`identity.py`, and the commit message), and a paper describes, it does not argue. Each field now says what it is and what `None` means, in one and two lines. The prose change does not move the surface digest, which strips docstrings, so the pin from the other thread still holds on this head.

raven/contracts/context.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed on eb443604: the contract version and pinned digest move together, and the complete contract-ledger guard passes. This resolves my finding.

raven/contracts/context.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed on eb443604: the contract tier is 3,038 lines against the 3,040 ceiling, and the complete kernel-budget guard passes. This resolves my finding.

gloryfromca (commented):

Reviewed the full revision and the delta from `37ba426f709c`. The new commit correctly bumps the contract version, repins the declared surface, and trims the contract prose without changing behavior; `raven/contracts` is now 3,038 lines against the 3,040 ceiling. Both prior blockers are fixed, I replied to the author's explanations, and I resolved both threads. No new findings.

Verification:
- Combined behavior and targeted contract slice: 213 passed.
- Complete contract-ledger and kernel-budget guard files: 14 passed.
- Ruff check and format, commit-message check, and `git diff --check`: passed.

gloryfromca (commented):

Re-reviewed the current diff and its range-diff from the previously clean revision after the rebase onto the latest `github/main`. The functional path still carries the selected model through the loop, assembler, and identity segment; the rebase also preserves the contract version/digest update and adapts cleanly to the new task-block fields. I found no blocker under the round-five bar.

Named follow-up (kept in the summary so it does not leave an open thread): append `AssemblyContext.model` after the existing optional fields instead of inserting it before `scent_text`. The adjacent contract note explicitly preserves positional compatibility for out-of-tree consumers; an old positional constructor now silently maps `scent`, `brief`, `done` to `model`, `scent_text`, `task_brief`, leaving `task_done_when` empty. This is nonblocking because all in-tree construction uses keywords and external callers have a working keyword-argument escape.

Verification:
- Relevant behavior, contract-ledger, and kernel-budget suites: 227 passed.
- Positional compatibility reproduction confirmed the shifted values described above.
- Ruff check and format, `ty`, source-language check, commit-message check, and `git diff --check`: passed.
- Both previously opened threads remain resolved.

raven/context_engine/segments/identity.py
```diff
@@ -79,9 +79,14 @@ async def build(self, ctx: AssemblyContext) -> Segment | None:
...
         # when it has to choose. A second listing here would drift against it.
+        # The model line names the id this turn's request reaches, which the
+        # loop resolved ahead of assembly. Left to the renderer's own lookup
+        # it would read config, and a conversation that switched model would
+        # be told, on every turn, that it still runs on the one it left.
         text = render.identity_text(
             self._workspace,
             specialists=self._specialists(),
             dispatch_tools=render.live_dispatch_tools(self._get_tool_definitions),
+            model=ctx.model,
```
LivXue: **Blocking: this names the stored id, not the id the request carries.**

`ctx.model` originates at `raven/agent/loop/turn_path.py:1894` as `session.metadata.get("model")`, and that value is written by `stored_model_id(...)` (`raven/rpc/methods/config.py:505`, stored at `:603`) -- the storage form. The fallback it displaces documents itself as the other form: `render._resolved_model_id()` is "The routed model id (gateway/provider prefix applied) from config. Delegates the storage-to-wire conversion to `providers.wire`". And `identity_text`'s own parameter doc promises "the resolved routed model id (full `provider/model` form)". So one rendered sentence now carries a Model Ref on this branch and a Wire Model on the fallback branch.

Measured with the repo's own `stored_model_id` / `wire_model`:

| provider | stored (rendered now) | wire (rendered before; what the request carries) | same |
| --- | --- | --- | --- |
| `openai_codex` | `openai-codex/gpt-5.1-codex` | `gpt-5.1-codex` | no |
| `azure_openai` | `azure-openai/gpt-4o` | `gpt-4o` | no |
| `anthropic` direct | `anthropic/claude-haiku-4-5` | same | yes |
| any model via a gateway | no `openrouter/` prefix | `openrouter/` prefix applied | no |

Concretely: on a Codex or Azure install a `/model` switch -- even a switch to the model the session is already on -- flips the line from `gpt-5.1-codex` to `openai-codex/gpt-5.1-codex`, a spelling the client strips before sending. On a gateway install the switch drops the `openrouter/` prefix, so the model is told it runs direct on a vendor while the request goes through OpenRouter and bills OpenRouter's account. Either way the line names an id the request does not carry, which is the property this PR sets out to establish.

Three existing tests pin the form and stay green because they call the helper rather than the line: `tests/test_segments.py:203` `test_resolved_model_id_codex_matches_wire_form`, whose docstring is "the identity line must report that wire form, not the stored one", plus the azure and gateway cases at `:211` and `:218`.

The conversion already has an owner, and the precedent is 76 lines above a line this PR edits: `raven/agent/loop/organ_glue.py:166` `_wire_output_ceiling` refuses to use the same `model` untranslated, with a docstring naming this exact difference ("`z-ai/glm-5.3-flash` against `openrouter/z-ai/glm-5.3-flash` -- and only the second reaches the OpenRouter tier"). Suggest `provider.wire_model_id(model)` before the value leaves the loop, or handing the id into `_resolved_model_id(...)` so the one conversion stays on the one path.

raven/context_engine/segments/identity.py
```diff
@@ -79,9 +79,14 @@ async def build(self, ctx: AssemblyContext) -> Segment | None:
         # nowhere else: one resident surface, where the parameter table lives,
         # where it can be narrowed per turn, and where the model is standing
         # when it has to choose. A second listing here would drift against it.
+        # The model line names the id this turn's request reaches, which the
```
LivXue: This comment asserts the property, but `stable = True` at line 40 now conflicts with the contract whenever a router is configured.

`raven/contracts/context.py:127` defines the flag as "this segment's text does not depend on what the user just said, so it reads the same on the next turn as on this one". The contract's next paragraph does license this segment interpolating "the resolved model id" -- but explicitly "per conversation, not per process". On the router path the value is per *message*: `raven/agent/loop/turn_path.py:1899` is `routed_model, fallback_models = await self.router.select_model_chain(content)`, and both routers re-rank per prompt and return `(None, [])` for some of them.

Cost when it bites: `IdentitySegmentBuilder.order == 1`, so this segment heads the stable run, and `raven/context_engine/assembler.py:215-221` places the message's only turn-surviving cache breakpoint at the end of that run. A router that changes its mind between turns rewrites the head, so that breakpoint pays a write at 1.25x the prompt rate (2.0x under `tokenWise.cacheTtl: 1h`) and never reads it back -- the waste `_stable_prefix_chars` exists to avoid. `tests/test_context_stable_prefix.py:175` pins `(IdentitySegmentBuilder, True, "SOUL.md and the identity header")` against a hardcoded expectation, so it stays green while its own justification string stops describing the segment's inputs.

Scope, fairly: `RoutingConfig.enabled` defaults to `False` (`raven/config/schema.py:699`), so this is opt-in-config only and not a default-install regression. But the honest options are to keep the value per-conversation, or to flip the flag and measure what that costs.

raven/contracts/context.py
```diff
@@ -82,6 +82,9 @@ class AssemblyContext:
     ``tui`` channel on purpose (see ``Source.surface``), so the channel alone
     cannot say whether the reader is at a prompt or in a browser.
     """
+    model: str | None = None
```
LivXue: `model` is inserted mid-list here, against the rule written four lines below it.

The comment at `:88-96`, guarding `task_brief` / `task_done_when`, gives the reason those two were appended: "Appended last, like the pair on `TurnContext`: this is the object every out-of-tree `SegmentBuilder` is handed, and the dataclass takes positional arguments, so a field inserted mid-list would rebind somebody's tenth argument rather than failing loudly." This field lands between `surface` and `scent_text`, shifting `scent_text`, `task_brief` and `task_done_when` each by one position. The `TurnContext` half of the same diff follows the rule.

Nothing in-tree breaks -- all 12 construction sites are keyword-only -- which is precisely why the rule is prose rather than a test. The population the comment names is real rather than aspirational: `raven.contracts` ships as a standalone installable artifact, and `tests/integration/test_kernel_wheel_smoke.py:42` imports it from a built wheel in a clean venv. An out-of-tree builder constructing this positionally now binds its `scent_text` argument into `model` with no `TypeError`, so a skill hint renders as the model name.

Moving the field after `task_done_when` before merge costs one digest recompute under the same version 29. Doing it after merge costs another version number.

raven/contracts/context.py
```diff
@@ -246,6 +249,8 @@ class TurnContext:
     # turn carries it at all.
     task_brief: str = ""
     task_done_when: str = ""
+    # The id this turn's request will reach, resolved by the loop ahead of assembly.
+    model: str | None = None
```
LivXue: The other production `TurnContext` producer was not given this field.

`raven/context_engine/segments/curator.py:169` rebuilds a `TurnContext` from the `AssemblyContext` and copies `current_message`, `media`, `can_see_images`, `describe_tool`, `channel`, `chat_id`, `task_brief`, `task_done_when` -- not `model`. The comment inside that call states the invariant: "Kept in step anyway, so that a later curator reading more of the turn finds the same turn the rest of the assembly saw."

Latent today: the curator reads only `current_message`, so nothing misbehaves. What it sets up is that the first curator to read `turn.model` -- the obvious use being to size against the model the conversation actually runs on, which `curator.py:129` currently answers from `resolve(None, self._fallback).model` -- gets `None`, falls back to config, and reproduces #470 one layer down. The contract digest pins the declared shape and can never see a producer omitting a field, so no gate covers this.

Calibration: `surface` and `selected_skills` are already absent there, so this joins an existing gap rather than breaking a clean record. One line, `model=ctx.model`, closes it; the alternative is to stop claiming the rebuild is kept in step.

raven/agent/loop/organ_glue.py
```diff
@@ -239,6 +239,7 @@ async def _assemble_context_messages(
                 chat_id=chat_id,
                 surface=surface,
                 selected_skills=selected_skills,
+                model=model,
```
LivXue: Two notes on the value this line forwards.

**An empty string deletes the model line entirely.** `raven/agent/loop/turn_path.py:1896` gates the session read with `if routed_model is not None:`, so `""` is treated as an explicit pick and also suppresses the router. `render.identity_text` then distinguishes `None` from `""` on one line and tests truthiness on the next (`raven/context_engine/segments/render.py:349`, `:351`), so `""` renders no `You are running on model:` line at all -- leaving the model to guess its identity from pretraining, which is what that line exists to prevent. The repo's sibling readers of the same metadata key guard for falsy rather than `None`: `raven/agent/loop/wiring.py:539-541` uses `if model:`, and `restore_session_model` returns early on `not model`. Reachability is low, since `raven/rpc/methods/config.py:478-482` rejects a falsy value, so this needs a record not written by the current RPC. `if routed_model:` there, or `model=model or None` here, makes the fallback unconditional.

**A fallback hop reads a line naming the primary.** Context is assembled once, before the call, from `routed_model` only; `raven/providers/base.py:837` then builds `model_chain = [model, *(fallback_models or [])]` and re-sends the same `messages` to the next hop with no re-assembly. So a fallback model answers a prompt asserting it is the primary. Not a regression -- before this change the line named the config default, which `select_model_chain` appends as the last-resort hop, so the old behaviour was accidentally right for that hop -- but the PR's stated property does not hold there. The sibling field this line sits beside documents exactly this caveat (`organ_glue.py:116-121`, on `can_see_images`: "A fallback further down the chain is sent the same message list"), and `model` could use the same qualifier.

tests/test_segments.py
```diff
@@ -160,6 +160,33 @@ async def test_bootstrap_renders_existing(self, tmp_path: Path) -> None:
...
+        looked like it had never happened.
+        """
+        monkeypatch.setattr(render, "_resolved_model_id", lambda: "openrouter/acme/config-default")
+
+        seg = await IdentitySegmentBuilder(tmp_path).build(_ctx(tmp_path, model="openrouter/acme/session-pick"))
+
+        assert "You are running on model: openrouter/acme/session-pick." in seg.text
+        assert "config-default" not in seg.text, "the configured default must not leak into a switched turn"
+
+    async def test_identity_reads_config_when_the_turn_resolved_no_model(self, tmp_path: Path, monkeypatch) -> None:
```
LivXue: Every new assertion on this line uses an id whose stored and wire forms are identical, so no test in this PR can detect the form change described on `identity.py:90`.

The four assertions the diff adds use `openrouter/acme/session-pick`, `openrouter/acme/config-default` and `routed/vision-model` -- all already wire-shaped, all asserting that the string passed in comes back out. A test that fed a stored id through `wire_model` and asserted the wire spelling would catch it; a round-trip assertion structurally cannot.

Separately, this particular test is subsumed: `test_identity_default_model_resolved_lazily` 13 lines below (`:194`) already pins "`model=None` resolves lazily from config" in three lines, and `tests/test_context_surface.py:131` pins the same bit a third time through the assembler. It also passes with all three production edits reverted, since it asserts the pre-existing fallback. The PR body is honest about that ("two for the ends"), but it does mean the fix's real coverage is the three link tests.

tests/test_read_file_image.py
```diff
@@ -1573,6 +1573,55 @@ async def assemble(self, session_key, session_messages, budget, *, turn):
     assert asked == ["routed/vision-model"]
 
 
+def test_the_routed_model_rides_the_turn_context_not_only_the_vision_probe(monkeypatch) -> None:
```
LivXue: This test asserts nothing about `read_file` or images, and this file's module docstring scopes it to "read_file's image branch: preprocessing, capability routing, persistence".

It is also a near-verbatim copy of `test_the_verdict_is_asked_of_the_routed_model_not_the_configured_one` 48 lines above (35 of 47 lines identical), with a two-line behavioural delta: record `turn.model` instead of the vision verdict, and `media=None`. That existing test already calls `_assemble_context_messages(..., model="routed/vision-model")` and already owns a recording `_Engine.assemble`, so `seen["model"] = turn.model` plus one assert inside it covers the same seam without a second 47-line scaffold -- and a recording engine that captures the whole `turn` makes a fourth field cost an assert rather than a fourth copy.

Four of the copied setup lines are inert once `_supports_vision` is monkeypatched: `loop._vision_ok = {}`, `loop._default_binding = ModelBinding(...)` with its import, `loop.context_engine = engine`, and `loop._last_injected_skill_ids = None`, which the method overwrites in its first statement. The `context_engine` line is the one worth removing on its own merits: it tells the next reader the loop calls `self.context_engine.assemble`, which it does not -- the method under test reads `self.harness.memory`.

`tests/test_context_surface.py` or one of the `test_agent_loop_*` files would be a better home. No AGENTS.md rule covers non-CLI test placement, so this is a placement note rather than a violation.

LivXue (commented):
## Review summary

The defect is real and the default path is fixed: with no session pick and no router, `turn.model` is `None`, the renderer keeps its config read, and nothing changes. The three links are genuinely mutation-covered -- reverting each of the three one-line edits fails exactly the expected tests -- and the contract paperwork is correct: I recomputed the surface digest independently on Python 3.12.13 and 3.13.13 and got `7bad1464...8a3eea0c`, matching the repin, with `CONTRACTS_VERSION = "29"`. Both gates pass (14 passed; `raven/contracts` measures 3,187 lines against the 3,200 ceiling).

Two findings I would gate on, both on the one prompt line this PR exists to correct.

### 1. The line names the stored id, not the id the request carries

`raven/context_engine/segments/identity.py:90`, detail inline. This is the one that matters: the fix reaches the right layer and delivers the wrong *form* of the right id, on Codex, Azure and every gateway install, and no test in the PR can detect it.

### 2. The estimation path was not threaded

`raven/agent/context/builder.py:180` is still `render.identity_text(self.workspace)` with no `model=`, while its own docstring states the invariant that breaks: "Delegates to the request path's renderer so the estimation prompt (this class) and the real per-turn prompt can never drift apart."

The path is live -- `raven/agent/loop/main.py:463` -> `raven/agent/harness/memory.py:114` `token_budget` -> `reserved_system` -> `available_history`, which decides how much history the Curator may keep. After this change the estimate names `agents.defaults.model` while the assembled prompt names the turn's pick, so the two strings differ on every switched conversation. The guard that should catch it cannot: `tests/test_segments.py:87` `test_identity_matches_legacy` builds its context with `_ctx(tmp_path)`, which sets no `model`, so it only ever exercises the branch where the two still agree. The token delta is small; the point is that a documented equality is now false and unenforced.

Three cheap items I would fold into the same round: the `AssemblyContext` field position (inline), one line in `raven/context_engine/segments/curator.py` (inline), and the sentence below.

### One line in the PR description that lands on `main`

The Verification section reads "after trimming the new docstrings under the 3,040-line ceiling". At this head `CONTRACTS_LINE_CEILING = 3_200` (`tests/test_kernel_budget.py:151`, raised 3,100 -> 3,200 on 2026-09-17 by another change), `raven/contracts` measures 3,187 lines, and the diff trims nothing: `context.py` goes 261 -> 266 with zero deletions. The "14 passed" is right; the reason given for it is not. Since this repo squashes with `squash_merge_commit_message=PR_BODY`, that sentence becomes the permanent commit body on `main`, so it is worth correcting before merge. Worth noting the same slip in commit 730dae68d's body -- "CONTRACTS_VERSION moves 26 -> 27" where the diff moves 28 -> 29 -- though squash drops individual commit bodies, so that one never lands.

### Checked and clean

Recording these so they need not be re-litigated:

- **Field insertion breaks no in-tree caller.** All 12 `AssemblyContext(` construction sites are keyword-only, including `AssemblyContext(**fields)` at `tests/integration/test_subagent_charter_e2e.py:190`. `frozen=True`, and both `dataclasses.replace` sites in the assembler preserve the new field, as does `raven/agent/harness/memory.py::_briefed`.
- **No prompt-cache corruption.** `stable_chars` is recomputed per turn from the live text and caching is content-addressed, so a switch costs one miss, not a stale offset or a poisoned key.
- **`ContextAssembler.set_provider` was not a usable alternative**, so the depth of the change is defensible: that hook fans out only from `set_default_binding` (`raven/agent/loop/wiring.py:751-753`), never from `set_session_binding`, which is the path a session-scoped `/model` actually takes.
- **Style gates clean.** `ruff check` and `ruff format --check` pass on the changed files; the longest added line is 119 against `line-length = 120`; zero non-ASCII characters in added lines.
- **No Windows exposure.** The change is string plumbing: no path handling, and `render.py`'s `if system == "Windows"` platform-policy branch is untouched. The new tests use `tmp_path` and assert no separators.

### One design question, not a blocker

`active_binding()` (`raven/providers/binding.py:90`) is open during assembly -- `raven/agent/loop/main.py:920` wraps `_run_turn` in `use_binding(binding)` -- so reading it inside `_resolved_model_id()` would be a few lines in one file, would fix finding 2 and the curator producer for free, would keep the storage-to-wire conversion on the one path that owns it, and would need no contracts version move. `raven/context_engine/segments/curator.py:96` (`context_window_tokens` -> `active_window`) is the in-tree precedent for reading a per-turn fact from a long-lived builder that way. The one case the contextvar does not cover is the router's per-turn pick, which is the real argument for a contract field -- and is also the case that makes `stable = True` false (inline). Worth a sentence in the PR either way, so the next person adding a per-turn fact knows which mechanism this subsystem wants.

Handsome-wzw:
Reworked per the design question at the end of the review: the branch now takes the `active_binding()` route instead of a contract field.

What changed on the current head:

- `render._resolved_model_id()` reads the turn's active `ModelBinding` (the session's own `/model` pick, else the default) and falls back to config outside a turn; either id goes through the existing `providers.wire` storage-to-wire conversion. One source file.
- Finding 1 (stored vs wire spelling): fixed on the one conversion path; `test_the_bound_id_goes_through_the_same_storage_to_wire_conversion` feeds a codex stored id and asserts the wire spelling.
- Finding 2 (estimation path not threaded): both renderers call the same function inside the binding, so they agree by construction; `test_the_estimation_prompt_and_the_turn_prompt_agree_on_a_switched_model` pins `ContextBuilder._get_identity()` equal to the segment on a switched model.
- The three cheap items: the `AssemblyContext` field, the curator producer and the misplaced test are all gone with the field; `raven/contracts/` is untouched and `CONTRACTS_VERSION` stays at `main`'s value.
- The stale ceiling sentence is gone with the description rewrite.

Mutation: swapping the binding read for the config read fails exactly the four tests that bind a model. Every thread has a reply and is resolved; happy to reopen any of them.

raven/context_engine/segments/identity.py
Handsome-wzw: Fixed in the rework now on the branch, and by taking your second option rather than patching the first: the contract field is gone. `render._resolved_model_id()` reads `active_binding().model` while a turn is running (the loop opens `use_binding` around every turn with the session's own pick) and `agents.defaults.model` otherwise, and feeds either through the same `providers.wire.wire_model` conversion the default always went through, so the stored spelling never reaches the line. I did not route it through `provider.wire_model_id`: `LazyProvider` answers identity until its first call, so the first turn of a session would still have rendered the stored form. `test_the_bound_id_goes_through_the_same_storage_to_wire_conversion` binds `openai-codex/gpt-5.1-codex` against a config default of `openai-codex/gpt-5.1` and asserts `gpt-5.1-codex` -- the assertion the earlier revision could not fail. It fails when the binding read is swapped for the config read.

raven/context_engine/segments/identity.py
Handsome-wzw: Holds again on the current head: the line reads the turn's binding, which is per session and constant within it, so the segment's inputs are what `stable = True` claims. The router's per-message pick is not reflected in the line (it never was); the PR description now records that trade explicitly.

raven/contracts/context.py
Handsome-wzw: No longer applies: the field is gone. `raven/contracts/` is untouched on the current head and `CONTRACTS_VERSION` stays where `main` has it.

raven/contracts/context.py
Handsome-wzw: Moot on the current head: there is no field for the curator's rebuild to miss. The value is read from the binding at render time by whichever caller renders the line.

raven/agent/loop/organ_glue.py
Handsome-wzw: The forwarding line is gone. On the two notes: `ModelBinding.__post_init__` refuses an empty model id, so the binding read cannot delete the line; the fallback-hop caveat stands, and the PR description records it beside the `can_see_images` precedent you pointed to.

tests/test_segments.py
Handsome-wzw: Those assertions are gone with the field, and so is the subsumed `test_identity_reads_config_when_the_turn_resolved_no_model`. The replacement class `TestIdentityNamesTheBoundModel` feeds a stored id through the binding and asserts the wire spelling (the codex case) and the gateway prefix, so a form change is now something the tests can see.

tests/test_read_file_image.py
Handsome-wzw: Deleted with the `organ_glue` change it pinned; nothing is added to that file on the current head.

LivXue (approved):
Approving at 08a5f7f3.

The rework takes the binding read rather than a contract field, and both findings I gated on are closed at the mechanism rather than patched around:

- **Spelling.** The bound id and the configured default now run through the same `providers.wire` conversion, so the stored form no longer reaches the line. `test_the_bound_id_goes_through_the_same_storage_to_wire_conversion` binds `openai-codex/gpt-5.1-codex` against a different configured default and asserts `gpt-5.1-codex` -- a form-divergent assertion, which is exactly what the earlier revision could not express.
- **Estimation drift.** `use_binding` wraps the whole turn tree (`raven/agent/loop/main.py`), and both renderers of this line call `_resolved_model_id()` without being told, so they agree by construction rather than by convention. `test_the_estimation_prompt_and_the_turn_prompt_agree_on_a_switched_model` pins it, which the earlier shape could not.

Your point about `wire_model_id` is right and my suggestion was wrong: `LazyProvider.wire_model_id` returns its argument unchanged while `_provider is None`, and prompt assembly runs before the first call, so every session's first turn would have rendered the stored form. Going through `providers.wire` is the correct route.

Also checked on this head, so it is on the record: the session's pick still survives a restart (`binding_for_session` falls back to `_restore_once`, which reads the session record); an unknown bound model degrades to skipping the line rather than raising, since the binding read sits inside the existing `try/except`; `raven.context_engine` and `raven.providers` are peers under the "inner layers know no surface" contract, so the new import crosses no layer; and no in-process sub-agent renders this line, so a foreign binding cannot leak into it.

The two carve-outs are recorded rather than silent: the router's per-message pick is not named (opt-in, off by default), and a fallback hop still reads a prompt naming the primary, which is pre-existing and documented beside the `can_see_images` precedent.

CI is green on this head: 14 success, 0 failures. The two cancelled runs are stale ones superseded by the current push.

PR #483, opened 2026-09-18
fix(everos-memory): give a recalled case the approach it was solved with

gloryfromca:
## Summary

An EverOS agent case carries three fields: `task_intent`, `approach` and `key_insight`. Recall built `Memory.text` from the first and the last only, so the method a case was solved with never reached the prompt. The reader saw what was attempted and what was concluded, with the steps between them missing.

The other two consumers of the same row already keep all three fields: the sub-agent memory record joins intent, approach and insight, and the memory page maps `approach` to a card's body. Only the recall path that feeds the model dropped it.

This joins the three fields in field order. Two constraints shaped how:

- Intent stays on the first line. The skill router takes this text's first line as a hit's display name and as its line in the search listing, so moving intent off line one would rename every recalled case.
- Empty fields drop out, so an EverOS that answers without an `approach` still reads as prose instead of carrying a blank paragraph.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
uv run --frozen pytest tests/test_everos_backend.py tests/test_skill_router_everos_source.py -q
182 passed in 17.69s

make lint-python lint-types
ruff check: All checks passed!
ruff format: 1960 files already formatted
ty check: All checks passed!

make check-source-language
clean, no findings
```

The case test asserts the exact joined text rather than substring membership, and it was run against the pre-change conversion first to confirm it fails there (AssertionError on the missing approach). A second test covers a row that arrives without an `approach`, which is the shape that would otherwise produce a blank paragraph.

No user-facing doc states the recall text shape. The CONTEXT.md passage that lists all three case fields describes the sub-agent memory record, which is a different code path and is unchanged.

## Risk

Each recalled case now carries one more paragraph, so the agent track's share of a prompt grows by roughly the length of one `approach` per case hit. Nothing downstream parses this text: the skill router reads the first line for a display name and passes the remainder through untouched, and that first line does not move.

Rollback is a revert of the single commit.

## Related Issues

N/A


gloryfromca (commented):

Reviewed `github/main...HEAD` and found no concrete failure. Coverage included the repository rules in `AGENTS.md` and `CONTEXT-MAP.md`, the complete two-file diff, the EverOS 1.2.3 search DTO, the backend recall path, the downstream skill-source caller, relevant history, backward compatibility with payloads that omit `approach`, and whether the tests were weakened. The exact-text assertion now proves the approach reaches the prompt in field order, while the added compatibility test proves absent fields do not create blank paragraphs. No architecture, naming, dependency, asset, or source-language constraint is violated.

Verification:
- `uv run --extra dev pytest tests/test_everos_backend.py tests/test_skill_router_everos_source.py tests/test_memory_backend_contract.py tests/test_everos_backend_contract.py -x`: 204 passed.
- `uv run --extra dev ruff check plugins-dist/everos-memory/raven_everos/backend.py tests/test_everos_backend.py`: passed.
- `uv run --extra dev ruff format --check plugins-dist/everos-memory/raven_everos/backend.py tests/test_everos_backend.py`: passed.

An initial `uv run pytest ...` attempt did not collect because a fresh default environment omits the workspace `everos-memory` dev dependency (`ModuleNotFoundError: raven_everos`); rerunning with the repository's `dev` extra produced the passing result above.

gloryfromca (commented):

Revision-specific review: this head only merges the current `main`; the two PR-authored files are unchanged from the prior revision, and the upstream commits do not intersect the EverOS recall path or its skill-source caller. Rechecking the resulting full diff against the updated base did not change the earlier conclusions on project rules, architecture, history, backward compatibility, or test strength.

`uv run --extra dev pytest tests/test_everos_backend.py tests/test_skill_router_everos_source.py tests/test_memory_backend_contract.py tests/test_everos_backend_contract.py -x` passed: 204 tests. `git diff --check github/main...HEAD` also passed.

0xKT (approved):

PR #484, opened 2026-09-18
fix(gateway): give a channel started while running its inbound dispatch

Handsome-wzw:
## What

Enabling an entrance from the page (weixin, for one) builds and starts the adapter on the spot through `ChannelManager.start_one`. The gateway command hands every channel its spine dispatch in one loop at launch, and that loop has already run by then, so a channel started this way has no `intake._submit`: it logs `no spine dispatch wired` once per message and drops every one, while the page draws it connected and the adapter's own login has succeeded.

## Where

`raven/cli/gateway_commands.py`. The outlet half of the same hazard was fixed earlier: `channels.on_started` registers an outlet so a hot-started channel can be replied to, and its comment records the replies that used to be lost. The intake half stayed a launch-time loop, and `raven/gateway/manager.py::start_one` fires `on_started` and nothing else, so nothing after launch ever calls `set_submit` for a channel born later.

## How

Compose the intake onto the hook the outlet already uses, at the point where `_inbound_dispatch` is born: `_wire_intake` sets the submit on each channel present at launch, and `channels.on_started` is wrapped so a channel started later gets the outlet it already got plus the intake. Channels present at launch are wired exactly as before; no new seam, no config, no protocol change.

## Verification

- `test_a_channel_started_while_the_gateway_runs_gets_the_inbound_dispatch`, pinned on the source the way the neighbouring mounted-relay test is, because the wiring lives inside the command with no import seam. It asserts the launch loop still wires directly and that the wrapped hook calls both `_outlet(ch)` and `_wire_intake(ch)`.
- `tests/test_cli_gateway_commands.py`: 47 passed with this change on current main.
- The drop was reproduced on a live gateway before the fix (2026-09-14): enable, disable and re-enable weixin through the page's own RPC puts the adapter on the `start_one` path, and every inbound message was logged as `no spine dispatch wired` and dropped.
- Ruff check and format clean on the two files.

Co-authored-by: Claude (claude-fable-5-1) <noreply@anthropic.com>


gloryfromca (commented):
Blocking: the source branch must use a 3-5-word snake_case description before this can merge.

`fix/channel_started_while_running_gets_its_dispatch` has seven words after the prefix, while `AGENTS.md` section 2.1 requires 3-5 and identifies its rules as hard constraints whose violations are rejected. This is branch metadata, so there is no valid diff line for an inline finding.

The code change itself looks sound. I covered the full diff; `ChannelManager.start_one`, `Intake`, `DeliveryHub`, the control-plane caller, and generation-swap wiring; relevant history; backward compatibility; test strength; and the Runtime architecture terms in `CONTEXT-MAP.md`/`CONTEXT.md`. The new callback preserves the existing outlet hook, wires both launch-time and hot-started channels, and is replaced with the current generation's hub before it is wrapped again. The source-pinned regression test follows the established pattern in this otherwise non-importable command closure and does not weaken prior coverage. I found no additional `CLAUDE.md` rules beyond `AGENTS.md`.

Verification: `uv run pytest tests/test_cli_gateway_commands.py tests/test_channels_manager.py -x` - 74 passed. `git diff --check github/main...HEAD` also passed.

PR #485, opened 2026-09-18
fix(gateway): give a channel started while running its inbound dispatch

Handsome-wzw:
## Summary

Enabling an entrance from the page (weixin, for one) builds and starts the adapter on the spot through `ChannelManager.start_one`. The gateway command hands every channel its spine dispatch in one loop at launch, and that loop has already run by then, so a channel started this way has no `intake._submit`: it logs `no spine dispatch wired` once per message and drops every one, while the page draws it connected and the adapter's own login has succeeded.

The outlet half of the same hazard was fixed earlier: `channels.on_started` registers an outlet so a hot-started channel can be replied to. The intake half stayed a launch-time loop, and `start_one` fires `on_started` and nothing else, so nothing after launch ever calls `set_submit` for a channel born later.

The fix composes the intake onto the hook the outlet already uses. A module-level `_wire_channel_intake(channels, dispatch)` sets the submit on each channel the manager holds and wraps `channels.on_started` so a channel started later gets the outlet it already got plus the intake; the command calls it at the point where `_inbound_dispatch` is born. Channels present at launch are wired exactly as before. The helper is module-level rather than a closure in the command body so that it can be executed by tests, which the command body cannot be. Nothing under `ui-web/` or `ui-tui/` changes.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `.venv/bin/python -m pytest tests/test_cli_gateway_commands.py -q` on this branch rebased onto current `main`: 50 passed. Four new cases drive `_wire_channel_intake` through fake channels and a fake manager: every channel present at launch gets the dispatch; a channel the manager starts later gets it through `on_started` while the outlet hook that was already there still runs; a manager with no outlet hook still wires the intake; and the command body is pinned by source to call the helper with `_inbound_dispatch` and to have lost the launch-only loop.
- Mutation-checked both ways: dropping `wire(ch)` from the composed hook fails exactly the two cases that start a channel late; dropping the outlet call fails exactly the outlet assertion of the hot-started case.
- `python scripts/coverage_gate.py diff --base-ref upstream/main --threshold 90` after the file ran under `--cov=raven --cov-branch`: 91.67% (11/12 executable changed lines). The one uncovered line is the call inside the `gateway` command body, which no unit test executes, the same as every other line of that body.
- `.venv/bin/ruff check` and `.venv/bin/ruff format --check` on `raven/cli/gateway_commands.py` and `tests/test_cli_gateway_commands.py`: clean.
- The drop was reproduced on a live gateway before the fix (2026-09-14): enable, disable and re-enable weixin through the page's own RPC puts the adapter on the `start_one` path, and every inbound message was logged as `no spine dispatch wired` and dropped.
- No user-facing docs change; the behaviour was always the documented one.

## Risk

- Security: none. No new input path; the dispatch a hot-started channel now receives is the same one launch-time channels already had.
- Backward compatibility: channels present at launch are wired by the same call as before; the only change is that a channel started later is wired too.
- Rollback: revert the squash commit; the hook falls back to the outlet-only lambda.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the full `github/main...HEAD` diff and the surrounding ChannelManager, Intake, DeliveryHub/outlet, gateway generation-swap, and control-plane paths. The composed hook wires both directions before `start_one` launches the adapter, preserves the existing outlet registration, and is rebuilt against the new spine on each generation bind.

Coverage included the repository rules and canonical Runtime/Generation/Channel terminology, callers and relevant history, backward compatibility for launch-time channels and the existing outlet hook, and whether the source-pinned test weakened existing coverage. I found no concrete defect or test weakening.

Verification:
- `uv run pytest tests/test_cli_gateway_commands.py tests/test_channels_manager.py -x` - 74 passed in 9.29s
- `uv run ruff check raven/cli/gateway_commands.py tests/test_cli_gateway_commands.py` - all checks passed
- `git diff --check github/main...HEAD` - clean

gloryfromca (commented):

I reviewed the revision delta from `86c81d65b4b2` and rechecked the full `github/main...HEAD` diff. Extracting `_wire_channel_intake` preserves the production callback order and generation-rebind behavior while replacing the earlier source-only coverage with direct behavioral tests for launch-time channels, hot-started channels, outlet-hook preservation, and the no-prior-hook case.

I covered the changed helper and tests, `ChannelManager.start_one` and gateway callers, relevant history, backward compatibility, test-strengthening versus weakening, and the repository's comment, test-location, Runtime terminology, and Generation architecture rules. I found no concrete defect in this revision.

Verification:
- `uv run pytest tests/test_cli_gateway_commands.py tests/test_channels_manager.py -x` - 77 passed in 6.84s
- `uv run ruff check raven/cli/gateway_commands.py tests/test_cli_gateway_commands.py` - all checks passed
- `git diff --check github/main...HEAD` - clean

0xKT (approved):

PR #488, opened 2026-09-18
feat(ui-web): bring the tasks panel and its strip onto the refactored tree

xfng-sd:
## Summary

The tasks panel, the board it draws and the strip above the composer were written against the page as it stood before the module-graph, state-layering and domain-convention steps. What they render is unchanged; what they reach for is not. This ports them onto the refactored tree so the surface arrives as a domain rather than as a page-era island.

Key decisions:

- **The seam.** There is no `tasks.*` method on the contract, so the panel had read `window.DS` directly. It reads `sources.tasks` now, and the two answers a page can get are a source module of its own: a served page gets the empty list, an offline one gets the stand-in rows, and the wiring picks between them the way it picks a transport. `liveMode()` is exported from `chooseTransport.ts` for that one decision. The rows moved out of the retired demo layer into `features/tasks/fixtures.ts`, because they answer no wire method and so have no responder to live in under `src/rpc/fixtures/`.
- **The conventions.** The store is on `makeStore`, which retires the `getState` spelling; the domain declares a manifest and registers itself in `features/manifests.ts`; and the cross-domain reaches the panel needs are pinned in the import-direction gate beside the ones the agents view it replaces had.
- **The classes.** The panel borrows the desk's list chrome and the page's state dot rather than restyling either, which the namespace gate reads as seven classes each reaching one more domain. Every one of them is page or panel vocabulary and is now named as such. The dag domain's unprefixed count rises by six and playbooks' falls by the same six: the pan-and-zoom viewport moved into `features/dag/Board.tsx` so that both surfaces could read it, and its classes moved with it.
- **Task state belongs to a pane and to a conversation.** The selection inside a task is keyed by task, which is what a pane is (the desk files them under `task:<id>`), so a pick in one pane no longer moves the others. And the panel is cleared on a session switch: `state/ws.ts` takes a reset handed in by domain rather than a fourth import into `features/`, and the wiring hands the tasks panel's in beside the three show slots it registers for the same reason.
- **The strip asks once the seam is on.** The strip renders in the first frame, before the page's wiring runs, so the ask it made for itself found no `sources.tasks`; answering that with `loaded: true` left the strip and the panel empty for the whole session. A seam that is not installed yet is unasked rather than empty, and the wiring makes the first read. The stub boot golden moves with it: the offline page now boots with the three running rows its fixtures always described.
- **The message namespaces.** The i18n gate holds one domain per namespace, so the two the port introduced are named after the domains that own them. The board's zoom labels are `gui.dag.zoom_*`, after the directory the board lives in. The desk's tasks empty state says `gui.ws.tasks_none` and `gui.ws.tasks_none_sub`, matching the two empties beside it in the same palette (`gui.ws.dlv_none`, `gui.ws.no_changes`) and the tab label above it (`gui.ws.tasks`); `gui.tasks.none` stays where the panel draws its own note.

Two things are not a straight port. `.tkempty` is named twice in the node card and no stylesheet ever defined it, so its rule lands in the domain's own `styles.css`. And the strip is a component the dock renders rather than a root mounted into markup, because `page.html` no longer carries the dock.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

CI on this branch is green: all 17 checks pass, including `page checks`, `ui rpc contract`, `TUI checks` and `commit messages`.

Locally, on Windows, against the merge base `refactor/ui_web_architecture` for comparison:

- `npm test --prefix ui-web` -- 2578 passed, 3 failed (163 files). The 3 are `KnowledgePage` failures that predate this branch in a domain it does not touch, and they are a Node-version issue: this lockfile's Vitest declares Node 20, 22 or 24+. Nothing else fails.
- `python ui-web/build.py` -- the production build and both boot snapshots pass, the stub one against the regenerated golden.
- `npm run lint --prefix ui-web` -- 0 errors, 7 warnings, identical to the base branch.
- `npm run type-check --prefix ui-web` and `npm run type-check --prefix ui-tui` -- clean.
- `npm run lint:i18n --prefix ui-tui` -- the generated catalogue is up to date after the namespace renames.
- `python scripts/check_source_language.py` and `python scripts/check_large_files.py` over the branch range -- both exit 0.

Seven of the tree-walking gates could not run on Windows at all, which is why the two i18n ownership failures were invisible locally and surfaced only on CI. They took their root from a file URL's `pathname`, so the walk landed off the repo's own tree and failed with ENOENT; and once that is fixed, `relative` hands back native separators that match none of the forward-slash lists the gates pin, so a ratchet reads as empty rather than erroring. Both answers now come from one module beside them. That fix is verified by breaking a rule on purpose: store-shape catches a reintroduced `getState` and import-direction catches an unpinned cross-domain import. On Windows the suite goes from 29 failures to 3.

## Risk

User-visible change: the desk's `agents` tab is replaced by a `tasks` tab, and a strip of running tasks appears above the composer when work is under way (it renders nothing when nothing is running). With a gateway behind the page the tasks list is empty until `tasks.list` lands on the contract; without one it shows the stand-in rows, so the surface can be opened and reviewed offline.

Rollback is the whole branch: the domain is self-contained under `features/tasks/`, and its one reach into shared wiring is the `sources.tasks` assignment in `app/install.ts`.

## Related Issues

N/A


ui-web/src/features/desk/DeskPalette.tsx
```diff
@@ -211,25 +211,23 @@ function DeliverablesNav(): JSX.Element {
...
+   The roster half of the old list is not lost with it. `AgentList` is also the
+   standalone subagents page's list, and starting an instance and opening a
+   direct chat still live there. */
+function TasksNav(): JSX.Element {
+  const state = useSyncExternalStore(tasksStore.subscribe, tasksStore.get)
+  useEffect(() => { void tasksStore.refresh() }, [])
+  /* Answered here rather than inside the view: in the desk, one nothing has to
+     look like the other two, and TasksApp's own note belongs to the panel. */
+  if (!state.rows.length) {
+    return <DeskEmpty kind="tasks" title={t('gui.tasks.none_title')} hint={t('gui.tasks.none')} />
```
gloryfromca: Blocking: make the desk/tasks namespace ownership pass the repository gate.

These new `gui.tasks.*` reads live in the desk domain while TasksPage also reads the same namespace. The full `npm test` run therefore fails `i18n-keys` with `gui.tasks is said by desk, tasks`. Split the desk-owned words or explicitly register the namespace as shared with the required reason so CI can pass.

ui-web/src/features/dag/Board.tsx
```diff
@@ -0,0 +1,259 @@
...
+        onPointerUp={onUp}
+        onPointerCancel={onUp}
+        onKeyDown={onKey}
+      >
+        <div className="gview" style={{ transform: `translate(${at.x}px, ${at.y}px) scale(${at.z})` }}>
+          {children}
+        </div>
+      </div>
+      <div className="gzoom">
+        <button className="gzb" aria-label={t('gui.graph.zoom_out')} onClick={() => zoomBy(1 / ZOOM_STEP)}>
```
gloryfromca: Blocking: register the moved graph namespace under the DAG domain.

Moving the zoom controls into `features/dag` also moved their keys from `gui.pb` to `gui.graph`, but the i18n ownership table has no `graph -> dag` alias. The full suite deterministically fails `names a namespace after the domain that owns it` with `gui.graph is features/dag/'s (ALIAS says undefined)`. Rename the namespace to the owning domain or add the required documented alias.

ui-web/src/features/tasks/TasksPage.tsx
```diff
@@ -0,0 +1,645 @@
...
+  if (task.confirm !== undefined) parts.push(t('gui.tasks.confirm', { v: String(task.confirm) }))
+  return parts.join(' · ')
+}
+
+export function TaskPane({ task, full = false }: { task: TaskRow; full?: boolean }): JSX.Element {
+  const s = useSyncExternalStore(store.subscribe, store.get)
+  /* Per pane, not per store: two open tasks are two panes, and a single
+     selected-node field would move both when the reader picked in one. Nothing
+     is selected until the reader picks, so the pane opens on the steps. */
+  const picked = s.node && task.nodes.some((n) => n.id === s.node) ? s.node : null
```
gloryfromca: Blocking: keep node selection local to each task pane.

The desk can open several task panes, but every `TaskPane` derives its selection from the singleton `store.node`. Picking a node in pane B overwrites pane A's selection; with distinct node ids A's card disappears, and with reused ids the other pane can open a card too. This directly contradicts the adjacent per-pane contract. Store the selection per task/pane (or locally) and add a two-pane test.

ui-web/src/features/tasks/TasksPage.tsx
```diff
@@ -0,0 +1,645 @@
...
+      return was
+    },
+  }
+}
+
+export function TaskRuns(): JSX.Element | null {
+  const s = useSyncExternalStore(store.subscribe, store.get)
+  /* The strip is on screen from the first paint, and is usually the first
+     thing to want the rows: the panel behind it may never be opened. */
+  useEffect(() => { if (!s.loaded) void store.refresh() }, [s.loaded])
```
gloryfromca, line 612: Blocking: reset and reload task rows when the conversation changes.

`TaskRuns` remains mounted across session switches and stops fetching once `loaded` is true. Although the new store defines `reset()` specifically for session isolation, no production caller invokes it: `state/ws.reset()` resets workspace, desk, and subagents only. After loading conversation A, switching to B or a draft therefore leaves A's task strip/list in place and does not request B. Wire tasks into the session-reset path and cover the switch.

gloryfromca (commented):
Blocking: fix the two failing i18n ownership gates and isolate task state by pane and conversation.

I covered the repository rules in AGENTS.md/CLAUDE.md, the ui-web vocabulary and domain conventions, the complete target diff, the task/desk/session callers, relevant history, backward compatibility, test changes, and the architecture gates.

Verification:
- `npm test`: 2573 passed, 5 failed. Two failures are the i18n gate blockers reported inline; three are in KnowledgePage tests under Node 23, while this lockfile's Vitest declares Node 20, 22, or 24+.
- Focused tasks/desk/DAG/playbooks run: 254 passed.
- `npm run lint`: passed with 7 warnings.
- Production build, both boot snapshots, artifact check, and generated-client check: passed.
- Source-language check against the PR target: passed.

ui-web/src/features/desk/DeskPalette.tsx
gloryfromca: Blocking: task pane selection and session isolation still have to change.

This namespace issue is fixed by moving the desk empty-state strings under `gui.ws`; the ownership gate now passes. Closing this thread while the two separate task-state blockers remain open.

ui-web/src/features/dag/Board.tsx
gloryfromca: Blocking: task pane selection and session isolation still have to change.

This namespace issue is fixed by naming the extracted board controls under `gui.dag`; the ownership gate now passes. Closing this thread while the two separate task-state blockers remain open.

gloryfromca (commented):
Blocking: keep task-pane selection local and reset task rows on session changes.

The namespace delta is correct: `gui.dag` now matches the board's owning domain, and the desk empty state now stays under `gui.ws`. I replied to and resolved those two threads. The two task-state findings from the previous round are unchanged, remain reachable through ordinary multi-pane and session-switch use, and remain open. I found no new issues in this revision.

I rechecked the current full diff, the delta and commit history, repository/domain rules, desk and session callers, backward compatibility, test changes, and architecture gates.

Verification:
- `npm test`: 2575 passed, 3 failed, all in `KnowledgePage.test.tsx` under Node 23; the two previously failing i18n gates now pass.
- Focused i18n/tasks/desk/DAG/playbooks run: 249 passed.
- `npm run lint`: passed with 7 warnings.
- Production build, both boot snapshots, artifact check, generated-client check, source-language check, and `git diff --check`: passed.

ui-web/src/features/tasks/TasksPage.tsx
xfng-sd: Fixed in 48be62f8. The selection is keyed by task now, which is what a pane is -- the desk files panes under `task:<id>`, so one entry per pane falls out of the id it already has. `pickNode` takes the task it is picking in, and the pane checks the id it finds against that task's own steps, because a refresh can land a task whose nodes no longer carry what the reader picked.

Two-pane test added, with the two tasks sharing node ids, which is the harder half of what you described: it asserts that picking in the second pane leaves the first on its board, and then that picking in the first moves neither off what it was showing. Confirmed it catches the old behaviour -- put back on one shared key it fails with `expected <div class="tkcard"> to be null`.

Also dropped the `open` verb's pre-selection of the first node. It was the one writer that set a selection nobody asked for, and it contradicted the pane it wrote for.

ui-web/src/features/tasks/TasksPage.tsx
xfng-sd: Fixed in 94178642. You were right that the strip is the part that makes this reachable: the dock is not remounted on a switch, so nothing took it down, and its effect stops asking once the rows have loaded.

On how it is wired: `state/ws.ts` already reaches into features/ three times for workspace, desk and subagents, and the import-direction gate pins each of those as debt, with its header saying a new upward edge is inverted rather than pinned. So rather than a fourth import, `state/ws.ts` now takes a reset handed in by domain and the page's wiring hands the tasks panel's in, beside the three `page.onShow` slots it registers for the same reason. Keyed by domain so a module evaluated twice leaves one reset rather than two.

Coverage is in two halves, matching how the sibling slots are covered (`state/page.test.ts` fills them "the way src/app/install.ts fills them"): `ws.test.ts` asserts the switch spends a registered reset once per domain alongside the three existing ones, and a tasks test asserts the mounted strip drops one conversation's runs and asks for the next one's. That `installActions` runs on boot is held by the boot-order gate, which pins all four lists inside `installPage`.

One thing worth flagging since it is not visible from the diff: the first two rounds of this review were the only reason those i18n gates were caught, because seven of the tree-walking gates could not run on Windows at all -- they took their root from a file URL's `pathname`, so the walk landed on `D:\D:\...` and failed with ENOENT. 94232363 fixes that separately; on Windows the suite goes from 29 failures to 3.

ui-web/src/features/tasks/TasksPage.tsx
gloryfromca: Blocking: update the boot-snapshot goldens for the new tasks strip so the required page build passes.

Verified at 48be62f8: both the board and detail pane read and write selection through the task id, and the two-pane shared-node-id regression test passes. This resolves my pane-local selection finding, so I am closing this thread. The remaining block is the separate served-page snapshot failure reported on the current revision.

ui-web/src/features/tasks/TasksPage.tsx
gloryfromca: Blocking: update the boot-snapshot goldens for the new tasks strip so the required page build passes.

Verified at 94178642: app wiring registers the tasks reset without adding a state-to-feature edge, every conversation and draft switch spends the ws reset, and the mounted strip reload regression test passes. This resolves my session-isolation finding, so I am closing this thread. The remaining block is the separate served-page snapshot failure reported on the current revision.

ui-web/src/chrome/Dock.tsx
```diff
@@ -138,6 +139,13 @@ export function Dock(): JSX.Element {
            each is its own floating glass card over the transcript, not a
            strip embedded in the box the user types in. */}
       <div className="sheets" id="sheetRack" />
+      {/* What is running while you type. Above the card and outside it: these
+          are not part of the reader's draft, they are the work already under
+          way, and a chip here is the same door the tasks tab's row is. A
+          component rather than the mount point page.html carried, because the
+          strip renders nothing at all while nothing is running -- and it says
+          in its own attributes which of its ends have more beyond them. */}
+      <TaskRuns />
```
gloryfromca, line 148: Blocking: update the boot-snapshot goldens for the new tasks strip so the required page build passes.

The dock now renders `TaskRuns` in stub mode, changing the booted DOM, but this revision leaves `scripts/__golden__/boot-stub.txt` untouched. The exact CI sequence (`npm run --prefix ui-web build` followed by `python3 ui-web/build.py`) exits here with 284 actual versus 298 golden nodes and the first difference `golden: div.dock-in`, `actual: div.runs`. That fails both required page-build jobs before their later checks run. Please regenerate and review the affected boot goldens with the strip in place.

This is introduced by this change, reached on every required CI page build, and has no working merge-path escape, meeting the fourth-round blocker bar.

gloryfromca (commented):
Blocking: update the boot-snapshot goldens for the new tasks strip so the required page build passes.

The two prior task-state blockers are fixed: selection is task/pane-keyed, and every session switch now spends the registered tasks reset without adding an inverted import. I replied on and resolved both threads.

One blocker remains inline. I reproduced it with the exact required CI page-build sequence; this revision changes the booted stub DOM but does not update the boot snapshot. This change introduces the failure, every required page build reaches it, and a failed required build has no merge-path escape, so it meets the fourth-round blocker bar.

Coverage: repository AGENTS/CLAUDE rules, ui-web CONTEXT/CONTRIBUTING architecture, the full target diff and new-head delta/history, callers and session lifecycle, backward compatibility, test changes, and the import/domain/boot constraints.

Verification:
- `npm test`: 2,580 passed, 1 failed in the untouched KnowledgePage suite under local Node 23 (`removeChild` in happy-dom/React), the same unrelated runtime-sensitive failure seen on the prior revision.
- Focused task, ws/session, and gate run: 38 files and 211 tests passed.
- `npm run lint`: passed with 0 errors and the existing 7 warnings.
- `npm run build`, `npm run gen:check`, check-page, check-css, check-class-namespace, source-language, and `git diff --check`: passed.
- `python3 ui-web/build.py`: failed deterministically because the stub boot DOM differs from its golden (284 vs 298 nodes; first difference is `div.runs` vs `div.dock-in`).

ui-web/src/chrome/Dock.tsx
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at b2c5f9af. The stub golden now records the running strip, and the same change fixes the boot-order defect exposed by that snapshot: a pre-seam read no longer marks tasks loaded-empty, while app wiring refreshes after installing the source. The exact CI page-build sequence now passes both boot snapshots and the later artifact gates. Closing this thread.

gloryfromca (commented):

The final delta fixes the boot snapshot and the underlying first-frame source race it exposed: an ask before seam installation stays unasked, app wiring refreshes after installing the task source, and the stub golden records the running strip. The exact required page-build sequence now passes. I replied to and resolved the snapshot thread; every thread I opened is resolved.

I rechecked the full target diff and final delta/history, repository and ui-web rules, callers and boot/session lifecycle, backward compatibility, test changes, and the architecture/import/domain/boot constraints. I found no new plain error.

Verification on this head:
- `npm test`: 2,576 passed, 5 failed, all in the untouched KnowledgePage suite under local unsupported Node 23; the failures varied from the prior run and remain outside this diff.
- Focused task, ws/session, and gate run: 38 files and 211 tests passed.
- `npm run lint`: passed with 0 errors and the existing 7 warnings.
- Production build, both boot snapshots (298 stub / 285 live nodes), check-page, check-css, check-class-namespace, generated-client check, source-language check, and `git diff --check`: passed.

gloryfromca (approved):

PR #489, opened 2026-09-18
feat(knowledge): read a pdf with the deepdoc models, keep tables as grids, crop each chunk

KevinHuSh:
image
image
image
image
## Summary

Everything a knowledge base could not read, and the pieces it cut them into
saying what they are.

**Reading a PDF.** A PDF records glyphs at coordinates; whether they are a
heading, a table cell or a running footer is inferred from how they look. There
are now two readers. The geometric one infers structure from type size and
position and needs nothing but the file. The other asks deepdoc's models, ported
into `raven/knowledge/parser/deepdoc`: a layout model that says what each region
is, a detector and recognizer that read a page carrying no text at all, and a
table-structure model that rebuilds a grid from a picture of one. `_pipeline`
runs the three the way RAGFlow's PDF parser runs them, with one deliberate
difference: where RAGFlow detects text boxes on every page, this uses the file's
own text layer wherever there is one and runs the recognizer only on the pages
that have nothing to read. Detecting boxes around text the file already spells
out exactly costs an inference pass per page to learn less, and recognising it
costs accuracy.

Two facts about the published weights are not in RAGFlow's code and cannot be
read out of it. The layout graph answers with a shape its own postprocessing
raises on, and its classes are DocLayout-YOLO's order rather than the list
RAGFlow carries, which disagrees on every index: read against that list every
title comes back as background and every page number as a title. Both were
established by running the model, and the label order is pinned by a test.

**Tables go in as tables.** `<table>` is the only shape that keeps what a grid
states -- which column a cell stood under, which cells are merged, which row is
the header -- and it is the shape a model reads a table in, which is what these
chunks are for. PDF, Word and PowerPoint all state their merges exactly, and
they now survive as `colspan` and `rowspan`. The flat form this replaces had to
repeat a header standing over two columns across both of them, because there was
nowhere to record a span; it had no callers left afterwards and is gone. Two
deviations from upstream, both about what the string is for: every cell's text
is escaped, where upstream interpolates it raw and a cell holding a `<` produces
markup that is no longer a table, and the span attributes are written only where
there is a span, where upstream emits `<td  >` for every ordinary cell.

**A picture of where each chunk came from.** A chunk is a few hundred words with
a page number on it, which is enough to cite and not enough to answer whether
the parser read the page correctly. So every PDF chunk that knows where it sits
carries one image: the regions it covers, cropped out of the rendered pages and
stacked. This follows `RAGFlowPdfParser.crop` in what it produces and not in how
it gets there. RAGFlow writes each region's coordinates into the chunk's own
text as marker tokens and parses them back with a regex, because its chunk is a
bare string; here they are already structured in the parser's element spans, so
the markers would only put text nobody wrote into the thing being embedded.
RAGFlow also renders every page up front and holds them, where this visits the
pages once each and releases each before the next, so the peak is one page
however long the file. The panel draws a chunk's tables as tables, parsed and
rebuilt as elements rather than injected: a person can rewrite any piece from
that same panel, so what comes back from the store is user input however it got
there.

**The word segmenter and the resources.** `raven.core.Tokenizer` is Infinity's
`rag_tokenizer`, ported: a forward and backward maximum match over a dictionary
trie, with stemming and lemmatisation on the latin path. XGBoost is not carried
over. None of the data lives in git -- the dictionary, the NLTK corpora and the
ONNX graphs run to about a hundred megabytes, so `make fetch-resources` on an
install and an optional Docker layer behind `RAVEN_RESOURCES` both point at one
script. A checkout that could not reach the network still works, and every
feature that reads this data says so when asked.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run on macOS against Python 3.12, with the deepdoc weights fetched (the model
tests skip without them):

```
uv run pytest tests/                 23697 passed, 109 skipped
npm run test --prefix ui-web          1886 passed
npm run test --prefix ui-tui          2074 passed
make lint-types                       clean (ty)
make lint-deps                        clean (deptry)
make lint-imports                     10 contracts kept
make check-source-language            clean
make check-large-files                clean
node ui-web/scripts/check-css.mjs     clean
uv run ruff check . / format --check  clean
uv run pre-commit run                 all hooks pass
```

The RPC contract was regenerated for both consumers after adding `has_crop` to
`KnowledgeChunk`: `npm run gen --prefix ui-web` and `npm run gen:rpc --prefix
ui-tui`, and `make lint-ui` verifies the generated client still matches.

## Risk

A PDF's tables, and a Word or PowerPoint file's, are now indexed as HTML rather
than as header-prefixed lines. Chunk text changes shape for those documents, so
a base indexed before this and searched after it holds both forms until its
documents are reindexed; retrieval keeps working either way, because both forms
are text. Rolling back is reverting the branch and reindexing.

The deepdoc models are optional. An install that never ran `make
fetch-resources` reads PDFs geometrically, exactly as before, and says so when
the models are asked for by name. Nothing here reaches the network at request
time.

Chunk crops are written for every PDF at index time and cost roughly 48 KB a
page on a text-heavy document, under the state directory rather than in the
repository. They are deleted with their document, with their base, and replaced
on reindex. The route serving them takes two ids and no path, validates the
chunk id before it becomes a path segment, and answers under the strictest
sandbox in the tree.

## Related Issues

N/A


raven/knowledge/_naive_chunker.py
```diff
@@ -0,0 +1,426 @@
...
+        self.overlap_size = overlap_size
+        self.table_context_size = table_context_size
+        self.image_context_size = image_context_size
+
+    async def chunk(self, sections: list[Section]) -> list[Chunk]:
+        units = self._units(sections)
+        if not units:
+            return []
+        self._lift_context(units)
+        merged = self._merge(units)
```
gloryfromca, line 310: Blocking: the naive chunker must preserve parser section boundaries.

The naive chunker merges short units across section boundaries, so the second section's heading and reading-order metadata are lost. `ChunkerBase`, `parser/__init__.py`, and `CONTEXT.md` all require one Chunk to derive from exactly one Section, but this call merges the flattened unit list globally and copies metadata only from the first unit. Two short heading sections therefore become one chunk attributed solely to the first heading. The added `test_chunks_may_span_sections` currently codifies the opposite of the repository contract; please keep packing within each Section and update that test.

raven/knowledge/_manager.py
```diff
@@ -136,6 +235,133 @@ def _client(self) -> EmbeddingClient:
...
+          one before this existed): today's client, unchanged;
+        - the base recorded its provider: that provider's endpoint, asked for
+          the base's model, because a model id does not name a credential;
+        - the base recorded a model but no provider (every base written before
+          the field existed): today's endpoint, asked for the base's model,
+          which is the last thing left to try. Whether it still serves that
+          model is not knowable from here -- if it does not, the embed fails
+          and the base answers by keyword instead.
+        """
+        configured = self._client()
```
gloryfromca: Blocking: recorded base providers must work without a global embedding pin.

A base with a recorded provider cannot use that provider after the global embedding pin is cleared. `_client_for` calls `_client()` before trying `embedding_config_for(base.embedding_provider, ...)`, so the call raises `no embedding endpoint is configured` even when that recorded provider still has valid credentials. In production this makes new documents fail indexing and silently downgrades searches to keywords, while `embedding_reach()` reports the same base as reachable. Resolve the optional configured client first, then use the base-specific provider independently.

raven/knowledge/_manager.py
```diff
@@ -429,37 +730,321 @@ async def index_document(self, document_id: str) -> KnowledgeDocumentRecord | No
...
+        """Turn pieces of one document on or off. Returns how many changed.
+
+        Off means out of retrieval entirely -- not a ranking penalty -- so the
+        filter lives in the store's own search rather than in any one caller.
+        """
+        found = self._indexed_document(document_id)
+        if found is None or not chunk_ids:
+            return 0
+        _, base = found
+        return await self._store.set_chunks_enabled(base.id, chunk_ids, enabled)
```
gloryfromca: Blocking: chunk mutations must be scoped to the requested document.

Chunk switch and delete requests can mutate chunks belonging to another document in the same base. The manager uses `document_id` only to find the collection; the store update/delete predicates contain only `chunk_id`. A caller that supplies document A with an ID previously returned for document B changes or deletes B's row, and the delete path then refreshes A's count instead of B's. Validate the IDs against the document or include `document_id` in both store predicates.

raven/knowledge/_manager.py
```diff
@@ -429,37 +730,321 @@ async def index_document(self, document_id: str) -> KnowledgeDocumentRecord | No
...
+
+        The document's recorded chunk count is rewritten from what remains, so
+        the row on the page keeps agreeing with the index -- it is the number
+        a reader uses to tell an indexed document from an empty one.
+        """
+        found = self._indexed_document(document_id)
+        if found is None or not chunk_ids:
+            return 0
+        _, base = found
+        await self._store.delete_chunks(base.id, chunk_ids)
```
gloryfromca: Blocking: manual chunk mutations must preserve positional invariants.

Manual chunk deletion and creation leave chunk indexes and totals inconsistent with the stored document. Deleting the middle of `[0, 1, 2]` leaves `[0, 2]`, and every survivor retains the old `total_chunks`; `add_chunk` similarly writes `N + 1` only on the new row while the existing N rows still say N. This violates `ChunkerBase`'s continuous-index and consistent-total contract and makes recall results report wrong positions. Renumber/update the stored chunks after these mutations.

raven/knowledge/_manager.py
```diff
@@ -72,16 +83,98 @@ class SearchOutcome:
     hits: list[VectorSearchResult] = field(default_factory=list)
     embed_ms: float = 0.0
     search_ms: float = 0.0
+    #: Bases that answered by keyword rather than by meaning, base id to the
+    #: reason their vectors could not be reached. Not an error and not a
+    #: skip -- those bases are in the results -- but a reader comparing two
+    #: sets of hits deserves to know that some of them came from words. The
+    #: reason is the same sentence the base would have failed with, so it
+    #: still says what to fix.
+    by_keyword: dict[str, str] = field(default_factory=dict)
```
gloryfromca, line 94: Blocking: keyword fallback must remain distinguishable on the RPC wire.

Keyword fallback is returned as if its BM25 score were a cosine similarity, with the fallback provenance discarded. `SearchOutcome` records `by_keyword`, but `knowledge_search` omits it and forwards each hit's raw `_score` under a schema documented as similarity; the recall UI then presents that number beside vector similarities. An unreachable embedding endpoint therefore produces apparently normal but differently-scaled recall results with no indication that retrieval changed modes. Carry the provenance through the RPC/UI and avoid labeling BM25 values as similarity.

raven/knowledge/parser/excel_parser.py
```diff
@@ -0,0 +1,399 @@
...
+    The dialect is sniffed rather than assumed: a ``.csv`` exported by a
+    spreadsheet in a locale that writes decimals with a comma is separated by
+    semicolons, and reading it as commas gives one column of joined text.
+    """
+    try:
+        text = raw.decode(encoding)
+    except UnicodeDecodeError:
+        # A spreadsheet writing "csv" on Windows still writes cp1252 as often
+        # as not, and a file that cannot be decoded is a file with no rows.
+        text = raw.decode("utf-8", errors="replace")
```
gloryfromca: Blocking: the Windows CSV fallback must decode the encoding it claims to support.

The Windows CSV fallback decodes cp1252 bytes as invalid UTF-8 and silently replaces their characters. For example, cp1252 `Caf\xe9` becomes `Caf�`, so names and values are permanently corrupted before indexing. The preceding comment identifies cp1252 as the expected fallback, but `utf-8` with replacement does not decode it; use the intended fallback encoding or reject undecodable input instead of silently changing it.

gloryfromca (commented):
Blocking: preserve section and chunk invariants and correct the embedding, fallback-search, CSV, and settings paths before merging.

I reviewed the full `github/main...HEAD` diff and relevant surrounding code, callers, branch history, and compatibility paths. I also checked AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and the Knowledge domain contract. The eight concrete failures are marked inline. I found no evidence that tests were removed or weakened merely to get green; however, the new naive-chunker test explicitly asserts behavior that contradicts the existing Chunk contract, as noted inline.

Verification:
- `uv run pytest` on all touched knowledge/parser/RPC tests: 495 passed, 2 skipped because LibreOffice is unavailable.
- `npm test --prefix ui-web`: 108 files and 1,872 tests passed.
- `npm run --prefix ui-web type-check`: passed.
- `npm run --prefix ui-web gen:check`: passed; generated RPC client matches 183 methods.
- `uv run python scripts/check_source_language.py github/main...HEAD`: passed. (`make` itself is unavailable in this environment.)
- Full `uv run pytest -x`: stopped during collection because the environment lacks the optional `websockets` package; it also reported the pre-existing optional `dingtalk_stream` skip.

ui-web/src/features/knowledge/KnowledgePage.tsx
```diff
@@ -741,19 +848,48 @@ function SettingsDialog({ base, busy }: { base: KbBase; busy: boolean }): JSX.El
     setSep(store.DEFAULTS.separator)
     setSize(String(store.DEFAULTS.chunk_size))
     setLap(String(store.DEFAULTS.chunk_overlap))
+    setTableCtx('0')
```
gloryfromca: Blocking: Restore Defaults must restore the declared context defaults.

Restore Defaults sets table and image context to zero instead of the declared defaults of 64. `store.DEFAULTS`, `KnowledgeBaseRecord`, and the new behavior all default these fields to 64, so pressing this button silently disables the context feature rather than restoring it. Set both controls from `store.DEFAULTS`, and extend the existing restore test to cover them.

raven/knowledge/_structure.py
```diff
@@ -358,28 +358,90 @@ async def chunk(self, sections: list[Section]) -> list[Chunk]:
...
+        the cut falls between them.
+
+        A single element larger than the budget still has to be split -- one
+        paragraph can be longer than any window -- and that is the one case
+        that falls through to cutting by length.
+        """
+        text = section.content.text if isinstance(section.content, TextBlock) else ""
+        spans = section.metadata.get(ELEMENTS)
+        if not isinstance(spans, list) or not spans:
+            return self._split_text(text)
```
gloryfromca: Blocking: heading prefixes must still be paid out of the chunk budget.

The heading-aware splitter ignores its reduced prefix budget for sections without element spans, producing oversized chunks. `_cut` receives `budget = chunk_size - prefix`, but this branch calls `self._split_text`, which cuts at the original `self.chunk_size`; the prefix is then prepended to every later piece. A 64-token chunker with a heading and an unbroken body reproducibly emits later chunks around 70 tokens, which can exceed the embedding input limit this budget protects. Split using the supplied budget here and in the oversized-element branch.

raven/knowledge/_naive_chunker.py
gloryfromca: Blocking: preserve section provenance, renumber migrated chunks, and honor a base's recorded provider when model names collide.

This is not settled. The patch changes the `ChunkerBase` class docstring, but the module docstring, `parser/__init__.py`, and the canonical `CONTEXT.md` Chunk definition still say chunks never span Sections. The runtime behavior still exposes the original failure too: for sections at page 2 / heading `First` and page 7 / heading `Second`, the stored chunk has internal `elements`, but `_chunk_row()` returns only page 2 and `First` for the combined `First body\nSecond body`; the RPC model does not carry `elements` or `page_end`. The document UI therefore still attributes the entire merged chunk solely to the first section. Either preserve the boundary or revise the full contract and carry the per-part provenance through the consumer that displays it.

raven/knowledge/_manager.py
gloryfromca: Blocking: preserve section provenance, renumber migrated chunks, and honor a base's recorded provider when model names collide.

This specific no-global-pin failure is fixed: `_configured_client()` now makes that lookup optional, and the recorded provider path remains usable when the pin is absent. The focused manager tests pass. I found a separate same-model/different-provider case and am reporting it independently.

raven/knowledge/_manager.py
gloryfromca: Blocking: preserve section provenance, renumber migrated chunks, and honor a base's recorded provider when model names collide.

Fixed. Both switch and delete now include `document_id` in the store predicate, and the new two-document regressions exercise the borrowed-ID case.

raven/knowledge/_manager.py
gloryfromca: Blocking: preserve section provenance, renumber migrated chunks, and honor a base's recorded provider when model names collide.

This is fixed for newly indexed rows, but not for existing collections, so the thread remains open. `_writable()` migrates old rows with `chunk_id = ''`, while `renumber()` addresses every update through `_id_predicate`, which deliberately returns `false` for an empty ID. The UI still allows appending to such a document. Against a real pre-upgrade LanceDB table containing two old chunks, appending a third and calling `renumber()` returned 3 but read back totals `[2, 2, 3]`. Please give legacy rows an update predicate (for example, document plus their old index) or prevent the mutation until reindexing.

raven/knowledge/_manager.py
gloryfromca: Blocking: preserve section provenance, renumber migrated chunks, and honor a base's recorded provider when model names collide.

Fixed. Retrieval mode now travels on every hit, fallback bases and reasons travel on the result, the schema and both generated clients are synchronized, and the recall UI labels BM25 separately. The RPC and UI regressions pass.

raven/knowledge/parser/excel_parser.py
gloryfromca: Blocking: preserve section provenance, renumber migrated chunks, and honor a base's recorded provider when model names collide.

Fixed. The fallback now decodes cp1252 rather than lossy UTF-8; I also exercised `Caf\xe9` directly and it indexed as `Café`.

ui-web/src/features/knowledge/KnowledgePage.tsx
gloryfromca: Blocking: preserve section provenance, renumber migrated chunks, and honor a base's recorded provider when model names collide.

Fixed. Restore Defaults now reads both values from `store.DEFAULTS`, and the extended UI regression verifies the nonzero defaults without saving them prematurely.

raven/knowledge/_structure.py
gloryfromca: Blocking: preserve section provenance, renumber migrated chunks, and honor a base's recorded provider when model names collide.

Fixed. Both no-span text and oversized elements now pass the reduced budget into `_split_text`; the added regressions cover both branches and pass.

raven/knowledge/_manager.py
```diff
@@ -137,6 +236,153 @@ def _client(self) -> EmbeddingClient:
...
+
+        The configured endpoint is resolved *optionally*, and that is the whole
+        point of the order: asking for it first and failing when there is none
+        made a base whose own provider still has credentials unsearchable the
+        moment the global pin was cleared, while ``embedding_reach`` went on
+        reporting that same base as reachable. Only the third case above needs
+        a configured endpoint, and only it raises when there is none.
+        """
+        configured = self._configured_client()
+        if configured is not None and base.embedding_model == configured.model:
```
gloryfromca: Blocking: a base's recorded provider must win when model names collide.

`_client_for` treats a model-name match as sufficient to reuse the configured client, without comparing `configured.provider` to `base.embedding_provider`. With a global pin of provider A / `shared-model` and a base recorded as provider B / `shared-model`, this branch returns A and never attempts B; I reproduced that directly. That can make the base fall back to keywords when A is unavailable despite valid B credentials, or query vectors through an endpoint other than the one the base records. The surrounding contract explicitly says a model id does not name a credential, and `_client_named` already uses the required provider-aware predicate. Please apply the same provider match here whenever the base has one.

gloryfromca (commented):
Blocking: preserve section provenance, renumber migrated chunks, and honor a base's recorded provider when model names collide.

Six earlier findings are fixed and their threads are resolved. Two existing blockers remain: cross-section chunks are still flattened to the first section by the RPC/UI while the canonical and parser contracts still forbid the merge, and renumbering cannot update migrated rows whose `chunk_id` is empty. One new provider-selection blocker is marked inline.

Coverage: I checked the repository rules and canonical knowledge terms; the full PR diff plus the changes since `d11723b02f5a`; affected callers across manager, LanceDB, RPC/schema/generated clients, and web UI; relevant history; backward compatibility with pre-ID collections; test changes for weakening; and the chunk/provenance architecture. I found no weakened tests.

Verification on this head: 447 focused Python tests passed and 2 LibreOffice-dependent tests skipped; 144 knowledge-page tests passed (with happy-dom's existing aborted/refused iframe-fetch diagnostics), web type-check and generated-client check passed; after installing the declared TUI dependencies, its RPC drift check and type-check passed; the source-language gate passed; and a direct cp1252 CSV check preserved `Café`.

raven/knowledge/_manager.py
gloryfromca: Blocking: the naive chunker must preserve section provenance before merge.

On the fourth-round convergence check, I am withdrawing this as a blocker. The defect remains for migrated empty-ID rows, but an operator has a working route: the document's exposed Reindex action rewrites those rows with IDs, after which append/delete renumbering works. Please retain this as a backward-compatibility follow-up; it no longer holds the revision. I am resolving this thread accordingly.

raven/knowledge/_manager.py
gloryfromca: Blocking: the naive chunker must preserve section provenance before merge.

On the fourth-round convergence check, I am also withdrawing this as a blocker. The provider collision is real, but it requires the base and global pin to name the same model through different providers, and an operator can clear the global pin or align it with the base; either route reaches the recorded provider. Please retain provider-aware matching as a follow-up. I am resolving this thread accordingly.

gloryfromca (commented):
Blocking: the naive chunker must preserve section provenance before merge.

This head only merges `main`'s bilingual README update; the affected knowledge, storage, RPC, and UI paths are byte-for-byte unchanged from the reviewed revision. One blocker remains. It meets all three fourth-round criteria: this PR introduces cross-Section merging, it is reached in ordinary indexing because `_chunker_for()` currently returns `NaiveChunker` for every base regardless of `smart_chunking`, and an operator has no exposed setting or reindex path that preserves the later Section's displayed page and heading.

Two nonblocking follow-ups remain named here: migrated empty-ID rows need reindexing before manual mutations can renumber them consistently, and same-named models on different providers require clearing or aligning the global pin so the base's provider wins. I withdrew those blockers and resolved their threads because each has a working operator escape.

Coverage this round: the `main` merge delta, the full PR diff's standing findings, the affected callers and history, backward compatibility, test changes, repository rules/canonical terms, and the chunking architecture. No tests were rerun because this revision changes only target-branch README content and none of the reviewed code, tests, schemas, generated clients, or dependencies.

gloryfromca (commented):
Blocking: the naive chunker must preserve section provenance before merge.

This revision correctly fixes the provider-aware matching follow-up when two providers expose the same model id, including the legacy empty-provider path. It does not touch the sole open blocker: the naive chunker, parser/CONTEXT contract, and RPC projection are unchanged. I reproduced the ordinary two-section path again; it still emits one chunk spanning pages 2 and 7, while `knowledge.documents.chunks` exposes only page 2 and the first heading. The existing thread therefore still stands.

At this sixth review round, I am explicitly claiming all three blocker conditions: this PR introduces the cross-section merge, every normally configured base currently uses `NaiveChunker`, and the document UI/RPC has no way to recover or display the second section's provenance.

Verification: `uv run pytest tests/test_knowledge_manager.py tests/test_knowledge_naive_chunker.py tests/test_rpc_knowledge.py -x` (205 passed); `uv run python scripts/check_source_language.py github/main...HEAD` (passed); focused runtime reproduction (failure remains). I rechecked the repository rules and canonical `CONTEXT.md`, the current delta in the full PR diff, callers and RPC/UI consumers, commit history, backward compatibility, test changes/no weakening, and the Section-to-Chunk architecture contract.

gloryfromca (commented):
Blocking: the naive chunker must preserve section provenance before merge.

This revision only removes trailing whitespace from the web type declaration. None of the chunker, canonical contract, RPC projection, or document UI paths behind the sole open blocker changed, so the existing thread still stands and there are no new findings. The previously established late-round conditions remain unchanged: the PR introduces the behavior, every normally configured base takes this path, and the operator has no way to recover the hidden section provenance.

Verification: `uv run pytest tests/test_knowledge_manager.py tests/test_knowledge_naive_chunker.py tests/test_rpc_knowledge.py -x` (205 passed); `uv run python scripts/check_source_language.py github/main...HEAD` (passed). I checked the repository rules and canonical `CONTEXT.md`, the one-line revision delta in the full PR diff, callers and RPC/UI consumers, history, backward compatibility, test changes/no weakening, and the Section-to-Chunk architecture contract.

gloryfromca (commented):
Blocking: the naive chunker must preserve section provenance before merge.

This head rewrites the trailing-space cleanup commit, but its tree is byte-for-byte identical to the last reviewed revision (`git diff --quiet 404d11c0dc71 HEAD`). The sole open provenance blocker is therefore unchanged; there are no new findings and no settled thread to resolve.

Verification: `git diff --check github/main...HEAD` (passed); `uv run python scripts/check_source_language.py github/main...HEAD` (passed); `uv run pytest tests/test_knowledge_manager.py tests/test_knowledge_naive_chunker.py tests/test_rpc_knowledge.py -x` (205 passed). I checked the repository rules and canonical `CONTEXT.md`, the rewritten commit and full PR diff, callers and RPC/UI consumers, history, backward compatibility, test changes/no weakening, and the Section-to-Chunk architecture contract.

raven/knowledge/_naive_chunker.py
gloryfromca: Blocking: per-part section provenance must reach the RPC/UI consumer.

The new `section_ordinal` values are correctly retained inside `metadata.elements`, but no production reader consumes or exposes them. The original two-section reproduction still yields one RPC row for `First body\nSecond body` with only page 2 and heading `First`; `_chunk_row`, `KnowledgeChunk`, and the web `ChunkRow` all discard `elements`. The earlier thread explicitly allowed cross-section merging if the full contract and displaying consumer carried each parts provenance, but this patch stops at stored metadata, so the concrete misattribution remains and the thread stays open.

gloryfromca (commented):
Blocking: per-part section provenance must reach the RPC/UI consumer.

The new chunker metadata and its regressions are internally consistent, but they do not close the sole open blocker: `section_ordinal` remains nested in `metadata.elements`, while the RPC projection/schema and web row still discard those elements and label the merged chunk only with its first page and heading. The original reproduction still fails at that consumer boundary, so the thread remains open. There are no additional findings.

This continues to meet the late-round blocker bar: this PR introduces the cross-section merge, every normally configured base uses it, and the operator-facing consumer has no route to the hidden provenance.

Verification: `git diff --check github/main...HEAD` (passed); `uv run python scripts/check_source_language.py github/main...HEAD` (passed); `uv run pytest tests/test_knowledge_manager.py tests/test_knowledge_naive_chunker.py tests/test_rpc_knowledge.py -x` (211 passed); focused RPC projection reproduction (failure remains). I checked the repository rules and canonical `CONTEXT.md`, the current delta in the full PR diff, chunking callers and RPC/UI consumers, history, backward compatibility, test changes/no weakening, and the Section-to-Chunk architecture contract.

raven/knowledge/_naive_chunker.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

Fixed. The original two-section reproduction now returns page 2 through 7 plus two RPC parts carrying section ordinals 0 and 1, their own pages, headings, and character ranges; the web row exposes the range and both origins. The focused Python and web regressions pass, so I am resolving this thread.

gloryfromca (commented):

I reviewed the branch delta from `f2dd796e4209` through this head, including the new slide parser/atomic-section behavior, the provenance RPC and UI path, preview navigation, their callers and generated contracts, and the merge from `main`. The remaining blocker is fixed: the original two-section reproduction now reaches the RPC and web UI with both sections own ordinals, pages, headings, and character ranges, and the open thread has been resolved. I also checked the repository rules and canonical terminology, backward compatibility, history, architecture constraints, and that the new tests exercise rather than weaken the affected behavior.

Verification: `uv run pytest tests/test_knowledge_*.py tests/test_rpc_knowledge.py -x` (487 passed, 2 skipped because LibreOffice is unavailable); `npm test --prefix ui-web -- src/features/knowledge/KnowledgePage.test.tsx` (149 passed); RPC generation drift check, TypeScript type-check, source-language check, and `git diff --check github/main...HEAD` all passed.

Nonblocking follow-ups retained: pre-upgrade rows with empty chunk IDs still require the exposed Reindex action before manual add/delete can renumber them consistently; and the canonical `CONTEXT.md` Chunk definition plus the parser package module doc still describe a never-cross-Sections contract even though `NaiveChunker` now deliberately crosses it with per-part provenance.

raven/knowledge/_lancedb.py
0xKT, line 1: **Blocking:** a knowledge base indexed before this PR answers every vector search with an error
after the upgrade, until something writes to it.

`LanceDBVectorStore.search` opens the table with `_table()` (raven/knowledge/_lancedb.py:214) and
then filters on a column that only `_writable()` adds:

    :214  table = await self._table(collection)
    :220  clauses = ["(enabled IS NULL OR enabled = true)"]

`_writable` is the only widener in the tree -- `grep -rn add_columns raven/` returns exactly one
hit, _lancedb.py:177, inside it -- and nothing calls it at boot. Every other method that touches
`enabled` goes through it: insert :183, keyword_search :277, list_chunks :317, renumber :367,
set_chunks_enabled :398, delete_chunks :405. `search` is the one that does not, and it is the
primary retrieval path.

A collection written by the previous build has no `enabled` column at all. Its schema is exactly
`vector, document_id, source, chunk_index, chunk_json, metadata_kv` (base `create_collection`,
c2c5f279:raven/knowledge/_lancedb.py:101-111), so `enabled IS NULL` cannot cover those rows -- the
predicate names a field the table does not have, and lance refuses the query rather than returning
nothing.

Reproduced here, with both controls:

    control  collection written by this head            -> OK, 2 hits
    subject  collection with the previous build's columns
             ['chunk_index','chunk_json','document_id','metadata_kv','source','vector']
                                                        -> RuntimeError: lance error: Invalid user
                                                           input: Schema error: No field named
                                                           enabled
    control  the same table after one insert (which widens through _writable)
                                                        -> OK, 3 hits

The only difference between the failing and the passing case is whether a write path widened the
table first, which is what makes this the code and not the setup.

It is a regression, not a hole in a new feature: the previous build's `search` added a `where`
only when `metadata_filter` was set, so the same on-disk collection searched fine before.

The path is the ordinary one and it fails rather than degrades:
raven/rpc/methods/knowledge.py:801 -> `KnowledgeManager.search` (_manager.py:1126) ->
`self._store.search(...)` (_manager.py:1185, no try around it) -> _lancedb.py:214, re-raised at
knowledge.py:804 as `InternalError`.

Two things in this PR already state the intent that the code misses. `_writable`'s docstring
(_lancedb.py:157-162): "the widening happens on first use rather than in a migration pass" --
search is a first use that does not widen. And the comment above the filter itself: "`enabled IS
NULL` covers the rows a build before this column wrote" -- it covers those rows only once the
column exists.

`_table` -> `_writable` at :214 closes it. Hoisting the clause and the widening into one helper
the retrieval sites share would also keep the next reader from having to notice.

0xKT:
**Not a blocker** -- three more, separate from the review thread I opened on
`raven/knowledge/_lancedb.py`. None of these holds the merge on its own.

**1. NOTICES.md promises a licence text that is not there** (NOTICES.md:43)

The file's own header says "Their copyright notices and license texts are retained in
`LICENSES/`", and every entry whose code this repo carries points at its file:

    :10  License: MIT - see `LICENSES/MIT-nanobot.txt`
    :18  License: MIT - see `LICENSES/MIT-hermes-agent.txt`
    :30  License: MIT - see `LICENSES/MIT-ink.txt`
    :82  License: Apache-2.0 - see `LICENSES/Apache-2.0-agentscope.txt`
    :43  License: Apache-2.0                       <- the new RAGFlow entry

The two entries without a pointer are the ones that say the code is not carried -- tui-use states
"Raven does NOT vendor, redistribute, or modify" it, under a heading that calls those entries
supply-chain hygiene rather than a licence requirement. RAGFlow is the opposite case: the 47 new
lines exist precisely to say which code is carried. `git diff --name-only` shows `LICENSES/` is
untouched by this PR, and `ls LICENSES/` has no ragflow file. `LICENSES/Apache-2.0-agentscope.txt`
is the nearest sibling and the template for the name.

The other two claims in that entry do hold, checked file by file: all five named modules carry the
RAGFlow origin line in their own docstring, and `grep -rn "^from ragflow|^import ragflow|from rag\.|import deepdoc" raven/`
is empty, so "rewritten rather than vendored" is accurate.

**2. The parser contract says every section carries READING_ORDER; the Markdown and HTML parser
stamps nothing** (raven/knowledge/parser/__init__.py:211)

`ParserBase`'s docstring: subclasses "must stamp every section they return with at least
`READING_ORDER`". `StructuredTextParser` (raven/knowledge/_structure.py:262) is a `ParserBase`, and
`_manager.py` registers it ahead of `TextParser` for Markdown and HTML -- text_parser.py:13 says so
in as many words. It neither imports `section_metadata` nor names `READING_ORDER`, and the sections
it returns carry `metadata={}` (_structure.py:305). `TextParser` by contrast stamps it at
text_parser.py:130.

So the contract this PR introduces has one registered claimant that does not meet it, and it is the
one serving two of the media types. Either the parser stamps it or the docstring says "parsers in
this package".

**3. Nothing pins what switching a base's embedding model does to the old vectors**
(raven/knowledge/_manager.py:612)

The behaviour is right: the manager drops the collection and recreates it at the new width
(`await self._store.delete_collection(base_id)` then `create_collection(base_id, width)`), so an
index never mixes two models' vectors. No test covers it -- the only `delete_collection` references
under tests/ are the fake store's method and two cases in test_knowledge_lancedb.py about a missing
collection. Given how quietly a mixed index would fail -- worse results, no error -- the case is
worth a test while the code is fresh.

For the record, two things I checked and found deliberate rather than defective: every one of the
five modules NOTICES names does carry its origin line in its own docstring, and an image ingested
with no vision model configured raises rather than indexing as empty content
(raven/knowledge/parser/image_parser.py:94, and the docstring at :12 says why) -- indexing a
picture as nothing would be the worse of the two.

raven/knowledge/_lancedb.py
gloryfromca: Blocking: existing pre-upgrade LanceDB collections must be widened before vector search.

Confirmed independently against the exact legacy six-column schema. Calling `LanceDBVectorStore.search()` raises `RuntimeError: No field named enabled`, and the schema remains unchanged after the failed read, so an ordinary post-upgrade search has no recovery path unless some separate write happens first. The focused LanceDB, manager, and RPC suite still passes (196 tests), which confirms the current tests do not exercise this migration path. This new evidence changes my stance on the unchanged revision from nonblocking to blocker.

raven/agent/tools/media.py
```diff
@@ -19,112 +19,48 @@
...
-            size = (max(1, int(size[0] * 0.8)), max(1, int(size[1] * 0.8)))
-
-    raise ImageTooLargeError(
-        f"image cannot be reduced below {MAX_BASE64_BYTES} bytes without dropping under {MIN_DIMENSION_PX}px"
-    )
-
-
-def to_data_uri(payload: bytes, mime: str) -> str:
-    return f"data:{mime};base64,{base64.b64encode(payload).decode('ascii')}"
+# The preparation half moved to `raven.utils.images`, beside the content-part
```
LivXue, line 27: **Nonblocking - move the preparation contract's rationale to the module that now holds it.**

This file's module docstring (lines 1-18, above the hunk) still documents the three ceilings
`prepare_image` enforces and why each has the value it does: 2000px per side, 4.5MB of base64
under the per-image cap Bedrock and Vertex enforce, and 1568 patch tokens with the note that
Anthropic applies both limits, so a pixel-only check lets an image through that still gets
resized server-side. Those constants and `prepare_image` itself now live in
`raven/utils/images.py`, whose own module docstring is unchanged by this PR and still reads
"Image content parts: sniffing, building, recognising and pricing them". So
`MAX_BASE64_BYTES = int(4.5 * 1024 * 1024)` sits at images.py:189 with nothing beside it
naming the provider cap that set the number, while the argument for it stays in a module
that no longer implements it. The next person tempted to raise that constant reads the file
that does not explain it.

Separately, the new comment at lines 27-31 gives the reason for the move as "reaching into
`raven.agent` for it would have put the two packages in an import cycle". There is no such
cycle at this head: grepping `raven/agent/` for `raven.knowledge` returns nothing, so no edge
runs back the other way, and no import-linter contract in `pyproject.toml` forbids
`raven.knowledge` from importing `raven.agent`. The relocation is still right on layering
grounds - a shared image helper belongs in `raven.utils` rather than behind an agent tool
module - but the stated constraint is not the one that applies, and a comment naming a hidden
constraint is exactly what a later reader will rely on.

Code-reading only: I read both modules at this head, grepped the repo at this head for the
reverse edge, and read the contract block in `pyproject.toml`. I ran nothing for this
finding. It does not affect the behaviour of the move, which I confirmed separately is
preserved.

Either direction closes it: carry the three-ceiling rationale across to
`raven/utils/images.py` and leave this file's docstring describing what it still does (MCP
content conversion and the model-facing image line), or keep the text here and say in it that
the constants it explains are re-exported from `raven.utils.images`. The cycle clause is
worth rewording to the layering reason either way.

LivXue:

Scope: I reviewed only the `raven/agent/**` portion of this PR - 1 of 61 changed files. The
remaining 60 files were read as context but are not covered by this verdict.

The agent slice is one file, `raven/agent/tools/media.py` (+37/-101), and it is a relocation:
`prepare_image`, `_target_size`, `to_data_uri`, `ImageTooLargeError` and the six ceiling
constants move to `raven/utils/images.py`, with a re-export left behind so existing callers
keep addressing them here. I checked that the move changes no behaviour and leaves no reader
behind.

Confirmed findings:

- Nonblocking, `raven/agent/tools/media.py`: the rationale for the three ceilings stays in
  this module's docstring while the constants it explains move to `raven/utils/images.py`,
  whose docstring is unchanged; and the new comment gives an import cycle as the reason for
  the move, which does not exist at this head. Marked inline.

Verification:

- AST identity of the ten moved symbols, base vs head: 10/10 identical. `to_data_uri` differs
  only by an added docstring line.
- Re-export completeness: every name in the new `__all__` resolves on
  `raven.agent.tools.media` at this head, and `media.prepare_image is
  images.prepare_image` is True. Every reader of the module at this head
  (`raven/agent/tools/filesystem.py`, `raven/context_engine/segments/render.py`,
  `raven/mcp/client.py`, `tests/test_read_file_image.py`) uses only names the shim carries.
- `estimate_image_tokens` is no longer reachable through `raven.agent.tools.media`. Grepped
  the repo at this head: nothing read it by that path, so the narrowing is clean.
- `uv run lint-imports`: 10 contracts kept, 0 broken, including "cargo does not import the
  loop shell". `raven/utils/images.py` imports only `raven.contracts.tool`, so the
  destination adds no cycle of its own.
- `pytest tests/test_read_file_image.py tests/test_token_estimation.py
  tests/test_knowledge_vision.py tests/test_knowledge_parser_image.py`: 147 passed, 1 skipped
  (a chmod case that does not apply as root). `pytest tests/test_mcp_client.py
  tests/test_default_context_engine.py`: 26 passed. Both run against a clean export of this
  head.

- The open blocker on `raven/knowledge/_lancedb.py` is outside the reviewed scope and carries
  no note from me. I did not re-derive it and I am not independent corroboration for it: 0xKT
  found and reproduced it, gloryfromca confirmed it separately, and it is still unresolved at
  this head.
- The nine earlier findings from gloryfromca are all resolved; I re-filed none of them.
- Whether the re-export shim should be dropped in favour of updating the callers' imports is
  a design call the comment already answers, not a defect.
- The new `__all__` changes what `from ... import *` would bring in; nothing in the repo
  imports this module that way.

gloryfromca (commented):
Blocking: existing pre-upgrade LanceDB collections must be widened before vector search.

I reviewed the three-commit delta from `6872ae295eb9` through this head and rechecked it against the full PR diff, the repository rules and domain contract, the affected callers, history, backward compatibility, architecture boundaries, and the changed tests. This head does not touch `_lancedb.py` or add a legacy-schema regression. The exact six-column pre-upgrade reproduction still raises `No field named enabled` on its first vector search and leaves the schema unwidened, so the existing blocker remains.

I found no additional blocker in the tokenizer, PDF/deepdoc, table-grid, or crop/UI delta. One nonblocking hardening follow-up is worth retaining: `crop_path()` validates `chunk_id` but uses the request's `document_id` as a directory without validating or resolving it beneath the crop root, so a crafted authenticated `/knowledge/crop` request can escape that root and serve a known `.webp` path. Validate both path segments or enforce the resolved parent.

Verification: the focused LanceDB/manager/RPC suite passed (197 tests); the KnowledgePage suite passed (156 tests, with its existing aborted preview-fetch noise); web and TUI generated-client checks, i18n generation, and type checks passed; source-language, large-file, and diff checks passed. The new PDF/crop Python suite could not be run in this host because dependency sync must compile `datrie` and no C compiler is installed; a no-sync attempt confirmed the pre-existing environment lacks `pymupdf`, so I did not count those failures as code results.

gloryfromca (commented):
Blocking: existing pre-upgrade LanceDB collections must be widened before vector search.

This head only merges the current target branch: range-diff confirms the PR commits are unchanged, and the merge does not touch the LanceDB search path. Re-running the legacy six-column reproduction at this head still fails on the missing `enabled` field, while the focused LanceDB/manager/RPC suite passes (197 tests) without covering that upgrade path. The existing open blocker therefore still holds; I found no additional blockers in the merge delta.

Review scope: target-branch diff and merge interaction, relevant callers and history, backward compatibility, test changes, repository rules/contracts, and the previously reviewed architecture constraints. `git diff --check`, source-language, and large-file checks also pass.

gloryfromca (commented):
Blocking: existing pre-upgrade LanceDB collections must be widened before vector search.

I reviewed the new PDF-region commit and its target-branch merge across the RPC/schema path, web state and rendering callers, production asset packaging, history, backward compatibility, test changes, repository rules/contracts, and the previously reviewed architecture constraints. I found no additional blocker in that delta. The exact legacy six-column reproduction still fails at this head on the missing `enabled` field, so the existing unresolved thread continues to hold the revision.

Verification: 198 focused LanceDB/manager/RPC tests and 160 focused web tests passed; web type-check, production build with both pdf.js assets, generated-client check, `git diff --check`, source-language, and large-file gates passed.

raven/knowledge/_lancedb.py
0xKT: Re-checked at 883e6c35, since the branch moved a long way from the head this finding was written
against. It still stands, untouched: `raven/knowledge/_lancedb.py` is byte-identical between
6872ae29 and 883e6c35 (sha256 2a01d993... on both), and nothing else on the search path changed --
the only file the delta touches under a name that could matter is ui-web/src/features/knowledge/
store.ts, which is the browser side.

Flagging it only because the push between those two heads is 4183 insertions across the knowledge
module, and it would be easy to read that as having swept this up. Nothing here is a complaint about
the rest of the work.

The finding above is unchanged and so is what would close it.

gloryfromca (commented):
Blocking: existing pre-upgrade LanceDB collections must be widened before vector search.

This merge-only revision accepts upstream's retirement of the Knowledge web page while preserving the backend/RPC feature. I reviewed the merge conflict resolutions, the diff and relevant callers/history, backward compatibility, test changes, repository rules and Knowledge contracts, and the affected architecture. The current target merges cleanly, generated web RPC bindings match the contract, and I found no new blocker in the merge resolution.

The exact legacy six-column LanceDB reproduction still fails on this head with No field named enabled, and the existing review thread remains unresolved without a maintainer deferral, so that standing blocker still prevents merge.

Verification: 300 focused knowledge store/manager/RPC/parser tests passed and 2 LibreOffice-dependent tests skipped; source-language, large-file, diff-check, and generated-client checks passed. The locked environment could not be rebuilt because this runner lacks gcc for datrie, so the focused suite used temporary uv overlays for the stale environment's missing pure-Python dependencies.

PR #490, opened 2026-09-18
fix(*): retire every generation's skill watcher before the process exits

LivXue:
## Summary

`raven gateway` exited with SIGSEGV (-11, or 139 in a shell) on every SIGTERM,
so a systemd or docker stop recorded a crash instead of the clean stop it was.
`raven tui` and `raven agent` had the same defect.

Root cause: `LocalSkillCatalog` auto-starts `SkillFileWatcher`, a daemon thread
that parks inside the Rust `watch()` of watchfiles. Daemon status does not make
process exit safe while the thread sits in native code -- CPython runs
`Py_FinalizeEx` under that call. Every Python shutdown step succeeds first, so
the crash lands after the graceful chain is complete and the real exit code is
masked. With faulthandler armed all three faults report `<no Python frame>`,
i.e. no Python code is on the stack.

Every owner now retires it:

- the gateway's shutdown chain, for the generation still bound at exit;
- the TUI's teardown;
- the one-shot agent's teardown;
- `RavenRuntime.dispose`, for a generation retired at a swap. `build_runtime`
  mints a fresh `AgentLoop` per generation, so each swap starts a new watcher
  and abandons the previous one; without this a reloaded gateway still
  segfaults on stop. The generation-organ roster grows the same entry, so the
  two halves of that contract stay in step;
- the gateway's shutdown again, for a candidate staged by a reload that the
  serving loop never consumed. That generation is not the bound `agent`, so
  nothing else reaches it;
- the gateway's shutdown once more, for a candidate `take()` handed out whose
  binding a cancelled unbind never finished. `_serve_generations` holds that
  candidate in a local across `await _unbind_generation()`, so a shutdown
  cancelling the await unwinds the coroutine and drops the local, leaving the
  generation staged nowhere and bound nowhere. `SwapCoordinator` now keeps it:
  `take()` records the candidate it hands out and `release()` clears it, which
  spans exactly handover to the moment the caller's own binding points at the
  new loop. The serving loop needed no new line, because it already calls both
  ends of that span.

The replay runner already stopped the same watcher for the same reason; these
were the other owners. The stop sits in the generation contract and in each
command's teardown rather than in `AgentLoop.stop`, whose other caller is the
generation swap -- that path keeps the process running and must keep skill
auto-refresh with it.

The gateway's sweep now lives in a module-level `_retire_generation_watchers`
rather than inline in the serve closure. `run()` is a 550-line closure with no
import seam, so the inline form could only be pinned by reading its source text
-- which never executes it, and so earned no coverage on the lines that matter.
The helper drains all three seats and keeps going when one stop raises, since a
sweep that stops early still leaves a watcher parked in native code.

Two comments describing a removed guard go with it: one in `agent_commands`
pointing at an exit chokepoint in `raven.cli.commands.run`, and the test-side
scaffolding that patched `os._exit` for an exit the command no longer makes.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Process-level, faulthandler armed, throwaway home and a dummy key nothing calls:

```
                                      before   after
raven gateway   SIGTERM                 -11      0     (-11 on 5 of 5 runs)
raven gateway   SIGHUP then SIGTERM     -11      0     (2 reloads after)
raven tui       PTY, then signal        -11      0
raven agent     ordinary one-shot exit  139      1     (1 is the turn's own
                                                        auth failure, which
                                                        139 had been masking)
```

Controlled pair isolating the cause, on the loop factory the TUI uses -- build
the loop, let the interpreter finalize:

```
watcher left running   exit 139
stop_file_watcher()    exit 0
```

Suites, from the branch tree:

```
uv run pytest tests/integration/test_gateway_sigterm_smoke.py -q -n 0 -m integration
  1 passed          (this is the guard test that was failing)

pytest tests/test_cli_agent_loop_parity.py tests/test_cli_gateway_commands.py \
  tests/test_cli_gateway_health.py tests/test_cli_tui_commands.py \
  tests/test_cli_agent_commands.py tests/test_gateway_spine.py \
  tests/test_generation_swap.py tests/test_plugin_services_lifecycle.py \
  tests/test_rpc_control.py tests/test_rpc_system.py \
  tests/test_skill_watcher_cleanup.py tests/test_trajectory_replay.py \
  tests/test_utils_asyncio_runner.py -q
  307 passed, exit 0

ruff check / ruff format --check over the changed files   clean
ty check over the changed files                           no diagnostics on the changed lines
scripts/check_large_files.py origin/main..HEAD            exit 0
scripts/check_source_language.py origin/main..HEAD        exit 0
commitlint --from origin/main                             exit 0
```

Second round, after review found the in-transition window; rebased onto
`origin/main` at 6d8c363:

```
pytest tests/test_cli_gateway_commands.py tests/test_core_runtime_swap.py \
  tests/test_generation_swap.py tests/test_cli_tui_commands.py \
  tests/test_cli_agent_commands.py -q
  171 passed, exit 0

every test file importing raven.core.runtime          131 passed, exit 0
real gateway: boot, SIGTERM, exit code                0 on 3 of 3, no fault text
ruff check / ruff format --check                      clean
make check-source-language / make check-large-files   exit 0
coverage_gate.py diff --threshold 90                  95.83% (23/24), passed
```

Revert check on that round: removing the `_in_transition` record, the
in-transition seat in the sweep, or the sweep's `try/except` each fails its own
test and nothing else. The cancellation window itself is reproduced against real
`SkillFileWatcher` threads, counted by thread name, and is red without the fix.

Diff coverage on changed lines was 50% (4 of 8) before that round, because the
gateway-side tests read the chain's source instead of running it. The one line
still uncovered is the call site inside `run()`, which cannot execute without
booting a real gateway; the wiring to the seam is pinned by source inspection.

Each of the four regression tests was confirmed red before its fix. The TUI and
agent stops were additionally re-confirmed afterwards by removing both again and
watching those two tests go red, then restoring them.

## Risk

User-visible change: these three commands now exit with their real status
instead of a segfault. Anything reading the exit code sees a different value --
0 where it used to see 139, and for a failed one-shot the turn's own non-zero
code rather than 139.

Rollback: revert the commits. The watcher is a cache-invalidation helper, so
removing the stops restores the previous behaviour exactly, crash included.

- The stops are not wrapped in try/except. `ContextBuilder.__init__` assigns
  `skills` unconditionally, `stop_file_watcher` is documented safe when no
  watcher was started, and the watcher's own contract is that neither `start`
  nor `stop` raises. This matches the existing production caller in the replay
  runner, which is also unwrapped.
- `raven acp` reaches finalization the same way but calls `os._exit(0)` after a
  real session, so the fault is masked there rather than fixed. Left as is;
  making it consistent is a separate change.
- No central gate was reinstated. A future command that builds a loop and exits
  normally would reintroduce this, and the source-inspection tests here only
  guard the retirement seats that exist today.
- The watcher starts in `ContextBuilder.__init__`, which is construction-time
  work, and `RavenRuntime.discard` is contract-bound to stay call-free because
  a call there would mean an organ was started before FREEZE. So a staged or
  in-transition candidate is retired by the gateway's sweep rather than by
  `discard`, and the docstring records why. Moving the watcher out of
  construction would let `discard` own it; that is a separate change.
- The shutdown-before-consumption window was not reproduced by timing: three
  runs sending SIGTERM 0.2s, 1.0s and 2.0s after a reload all completed the
  swap first and exited 0. The staged-candidate retirement is driven by
  forcing the state in a test rather than by racing for it.

Known gap in the evidence: the two `raven tui` runs were ended with SIGTERM
after Ctrl+C went unanswered, because the TUI never finished loading against a
throwaway home. The ordinary-exit case is evidenced by the agent one-shot,
which exits by itself with no signal involved.

## Related Issues

N/A


raven/cli/gateway_commands.py
```diff
@@ -783,6 +783,14 @@ async def _unbind_generation():
...
                 # dispose() cancels again, which is an idempotent no-op.
                 await agent.subagents.cancel_all()
+                # ``agent`` is still the outgoing loop here; the caller rebinds
+                # it after this returns. Its skill watcher has to go with the
+                # generation: the next one mints its own, and a watcher left
+                # behind is a daemon thread parked in native code that
+                # Py_FinalizeEx will still tear the interpreter down under --
+                # so a reloaded gateway segfaults on stop even though the
+                # teardown chain stops the live generation's.
+                agent.context.skills.stop_file_watcher()
```
gloryfromca: Blocking: ensure every built generation's skill watcher is retired, including a staged candidate cancelled before it is served.

`_request_swap()` starts the candidate's watcher inside `build_runtime()`, stages that runtime, calls `agent.stop()`, and returns while the old loop can still be in its one-second sleep. An immediate `gateway.shutdown` (or SIGTERM) can therefore cancel `_serve_generations()` before `swaps.take()` and `_unbind_generation()` run. The finalizer shown later stops only the still-bound old `agent`; `swaps._candidate` keeps the new catalog and its native watcher alive, so this reachable reload-then-shutdown sequence can still end in the SIGSEGV this change is fixing. The Generation contract puts retirement in `RavenRuntime.dispose()`/`discard()`, but neither currently stops the watcher. Please cover staged/discarded candidates as well and add a regression for shutdown before candidate consumption.

gloryfromca (commented):
Blocking: retire the watcher owned by a staged reload candidate when shutdown wins the swap race.

I found one concrete blocker, marked inline. The direct agent, TUI, gateway-shutdown, and normal generation-unbind stops are otherwise consistent with the watcher's idempotent contract.

I covered the repository rules in AGENTS.md/CLAUDE.md and the Runtime Generation/Assembly Root terms routed by CONTEXT-MAP.md; the complete diff; the LocalSkillCatalog, AgentLoop, RavenRuntime, SwapCoordinator, reload, and shutdown callers; relevant watcher/finalization history; backward compatibility; and whether the tests were weakened. The source-inspection tests follow an existing pattern, but they do not exercise the staged-candidate shutdown window.

Verification:
- `uv run pytest tests/test_cli_agent_commands.py tests/test_cli_gateway_commands.py tests/test_cli_tui_commands.py tests/test_generation_swap.py tests/test_core_runtime_swap.py tests/test_skill_watcher_cleanup.py -q`: 165 passed in 14.74s.
- `uv run pytest tests/integration/test_gateway_sigterm_smoke.py -q -n 0 -m integration`: 1 passed in 14.44s, with the suite's non-strict idle-ceiling warning.
- `git diff --check github/main...HEAD`: clean.

raven/cli/gateway_commands.py
LivXue: Confirmed and fixed in b97ed77df.

Verified before changing anything. `build_runtime` starts the watcher via
`ContextBuilder.__init__`, and neither retirement path stopped it: `dispose()`
retires plugin services, sub-agents, MCP, the loop and the backend, and
`discard()` was empty. So a candidate the serving loop never consumed kept a
daemon thread parked in `watchfiles`' Rust `watch()`, and it is not the bound
`agent`, so the teardown chain's stop did not reach it.

What changed:

- `RavenRuntime.dispose()` stops the watcher, and the generation-organ roster
  in `test_generation_swap.py` grows the same entry so both halves of that
  contract stay in step.
- The gateway's `_unbind_generation` no longer stops it itself; `dispose()`
  runs a few lines later and owns it now.
- The gateway's shutdown retires a staged-but-unconsumed candidate.

One deviation from the shape you suggested, with its reason. I did put the
stop in `discard()` first, and `test_discard_stays_a_no_op` failed: that method
is contract-bound to stay call-free, because a call there would mean an organ
was started before FREEZE. The watcher is exactly such an organ, so the honest
reading is that construction-time start is the defect and `discard()`'s
emptiness is correct about the ordering it guards. Rewriting that contract
inside a bugfix seemed wrong, so the gateway retires the staged candidate
directly and `discard()`'s docstring now records the one resource that does
exist from construction, instead of claiming there is nothing to stop. Taking
the watcher out of construction would let `discard()` own this; that is a
separate change.

Two things I could not confirm, stated rather than papered over:

- I did not reproduce the shutdown-before-consumption window by timing. Three
  runs sending SIGTERM 0.2s, 1.0s and 2.0s after a reload all completed the
  swap first and exited 0. The regression therefore forces the state instead of
  racing for it, which is also what you asked for.
- The other half of `discard()`'s population looks unreachable here: `begin()`
  holds the swap slot from claim until `release()`, and `take()` clears the
  candidate first, so a second request is refused with `swap_in_flight` and
  `stage()` never finds an existing candidate to supersede. The shutdown path
  is the one that reaches it.

Verification at this head:

- `pytest` over the CLI, gateway, generation-swap, core-runtime-swap,
  skill-watcher and trajectory files: 307 passed, exit 0.
- Boot, two SIGHUPs, then SIGTERM: exit 0. Before the generation half of this
  fix the same sequence with one reload exited -11.
- `tests/integration/test_gateway_sigterm_smoke.py -m integration`: 1 passed.
- ruff check / format over the changed files: clean.

gloryfromca: Blocking: retain and retire the candidate after `take()` if shutdown cancels generation unbinding.

The `dispose()` change fixes served generations, and the finalizer now fixes a candidate that is still stored in the coordinator. A second ownership window remains immediately after `_serve_generations()` executes `swap = swaps.take()`: `take()` clears `_candidate`, then the coroutine awaits `_unbind_generation()` before rebinding `runtime` and `agent`. If shutdown cancels any of those awaits, the local `swap` is released; the finalizer gets `None` from its own `swaps.take()` and stops only the old bound agent, so the new candidate watcher still reaches `Py_FinalizeEx` alive.

I confirmed the state directly: stage a runtime with a real `LocalSkillCatalog`, call `take()`, drop the returned local as cancellation does, then stop the current catalog; one `SkillFileWatcher` remains alive. The new shutdown test source-inspects only the pre-`take()` case and does not exercise this transition. The in-transition candidate therefore still needs ownership that survives cancellation and retires its watcher.

gloryfromca (commented):
Blocking: keep the taken reload candidate reachable through cancellable unbinding so shutdown can retire its watcher.

The revision correctly moves normal served-generation watcher retirement into `RavenRuntime.dispose()` and handles candidates that remain staged in `SwapCoordinator`. The existing thread remains open because cancellation after `take()` but before rebinding removes the candidate from both cleanup paths; I replied there with the concrete state reproduction. I found no other blockers.

I covered AGENTS.md/CLAUDE.md, the Runtime Generation and Assembly Root contracts routed by CONTEXT-MAP.md, the full `github/main...HEAD` diff and revision delta, watcher/runtime/coordinator/gateway callers and history, backward compatibility, and whether tests were weakened. The new dispose contract test strengthens the normal path, but the staged-candidate source test does not cover the post-`take()` transition.

Verification:
- Focused state reproduction: after stage -> take -> drop the in-transition candidate -> stop the current catalog, 1 `SkillFileWatcher` remained alive.
- `uv run pytest tests/test_cli_agent_commands.py tests/test_cli_gateway_commands.py tests/test_cli_tui_commands.py tests/test_generation_swap.py tests/test_core_runtime_swap.py tests/test_skill_watcher_cleanup.py tests/test_trajectory_replay.py -q`: 202 passed in 10.02s.
- `uv run pytest tests/integration/test_gateway_sigterm_smoke.py -q -n 0 -m integration`: 1 passed in 13.90s, with the existing non-strict idle-ceiling warning.
- `git diff --check github/main...HEAD`: clean.

raven/cli/gateway_commands.py
LivXue: Fixed in f7eeb67d2 (rebased onto origin/main). You were right on both counts: the window, and the reason the previous test could not see it.

**Ownership now sits in `SwapCoordinator`, not in the serve closure.** `take()` records the candidate it hands out in `_in_transition`, and `release()` clears it. That spans exactly handover to the moment the caller's own binding points at the new loop, so a cancelled `_unbind_generation()` no longer drops the only reference. `_serve_generations` needed no new line: it already calls `release()` and `take()` at the two ends of that span, so the coordinator only had to remember what it had given away.

**On the test not exercising the transition** -- agreed, and it could not: it reads source text and never runs it, which is also why diff coverage on the changed lines was 50%. The sweep is now a module-level `_retire_generation_watchers(swaps, agent)`, so tests execute it. It drains three seats (bound, staged, in transition) and keeps going if one stop raises, because a sweep that stops early still leaves a watcher parked in native code.

**Your repro is now a test.** `test_a_swap_cancelled_mid_unbind_leaves_no_live_watcher_thread` builds two real `LocalSkillCatalog` instances, stages one, calls `take()`, never stores the result (what cancellation does), runs the sweep, then counts live threads named `SkillFileWatcher`. It fails without the `_in_transition` record. This works because the autouse `_no_leaked_skill_watchers` fixture tracks watchers rather than suppressing them.

Verification:

- 171 passed across the five affected test files, post-rebase
- revert check: dropping the `_in_transition` record, the in-transition seat, or the `try/except` each fails its own test and nothing else
- real gateway: boot, SIGTERM, exit 0, 3/3, no fault text in the log
- `ruff check` and `ruff format --check` clean; `make check-source-language` and `make check-large-files` both exit 0
- diff coverage 50% to 95.83% (23/24)

The one uncovered changed line is the call site inside `run()`, which cannot execute without booting a real gateway; the wiring to the seam is pinned by a source-inspection test instead.

Leaving the thread open for you to verify and resolve.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified. `SwapCoordinator.take()` now retains the candidate in `in_transition`, `release()` clears that ownership only after the gateway has rebound and wired the new generation, and the shutdown sweep reaches bound, staged, and in-transition loops. The executable real-watcher regression covers the cancellation state that the earlier source-inspection test missed. My blocker is resolved.

gloryfromca (commented):

The cancellation-window fix holds: coordinator ownership spans `take()` through the next generation's `release()`, and shutdown sweeps the bound, staged, and in-transition generations without abandoning later seats if one stop fails. The real-thread regression reproduces the previously missed ownership state. I replied to and resolved the thread I opened.

I covered the complete `github/main...HEAD` diff and the revision delta; AGENTS.md/CLAUDE.md; the Runtime Generation and Assembly Root contracts routed by CONTEXT-MAP.md; the gateway, coordinator, runtime, catalog, and watcher callers; relevant history; backward compatibility; and whether tests were weakened. I found no remaining plain error or blocker.

Verification:
- `uv run pytest tests/test_cli_agent_commands.py tests/test_cli_gateway_commands.py tests/test_cli_tui_commands.py tests/test_generation_swap.py tests/test_core_runtime_swap.py tests/test_skill_watcher_cleanup.py tests/test_trajectory_replay.py -q`: 210 passed in 9.89s.
- `uv run pytest tests/integration/test_gateway_sigterm_smoke.py -q -n 0 -m integration`: 1 passed in 15.64s, with the existing non-strict idle-ceiling warning.
- `git diff --check github/main...HEAD`: clean.

0xKT:
**Not a blocker** -- nothing this PR fixes regresses, and the gap below predates it. But the title
says "every generation", and after this merges one entrance is still uncovered.

`RpcStack.teardown` (raven/rpc/bootstrap.py:386-458) is the fifth teardown of this shape. It owns
the loop it built (`owns_loop = agent_loop is None` at bootstrap.py:224, then
`agent_loop = build_agent_loop(channel=channel)` at :235) and retires the same organs in the same
order -- `stop_plugin_services`, `subagents.cancel_all`, the backend drain and stop, `close_mcp` --
and never calls `stop_file_watcher`.

That teardown is what `raven serve` runs: raven/cli/serve_commands.py:326 builds the stack with no
`agent_loop=`, and :378 awaits `stack.teardown()`. There is no hard exit on that path -- the only
`os._exit` in the tree is raven/cli/acp_commands.py:147, on the ACP served branch -- so `raven serve`
returns into ordinary interpreter finalization with the thread still parked in watchfiles' Rust
`watch()`, which is precisely the state this PR's own docstring (raven/cli/gateway_commands.py:213-226)
describes as the SIGSEGV.

Measured here, calling only the two expressions `raven serve` calls:

    before build_rpc_stack      : []
    after  build_rpc_stack      : ['SkillFileWatcher']
    after  stack.teardown()     : ['SkillFileWatcher']      <- still parked
    after  one stop_file_watcher(): []                       <- control

The control is the point: the property is measurable and this PR's own one-line shape closes it, so
the first three lines are the code's behaviour and not my probe's.

After this PR the rule is decided in five places -- raven/core/runtime.py:100,
raven/cli/gateway_commands.py:247, raven/cli/tui_commands.py:606, raven/cli/agent_commands.py:365,
raven/trajectory/replay.py:1118 -- and raven/rpc/bootstrap.py is the one dissenter. Either the same
line in `RpcStack.teardown`, or a single owner that every teardown already goes through, ends that.

Worth knowing while deciding: `raven web` supervises `raven serve` (raven/cli/serve_commands.py:694),
and its supervisor treats a zero exit as "the gateway exited cleanly; not restarting it"
(serve_commands.py:805-806). A crash at finalization is not zero, so the supervisor restarts an
engine it was told to leave down and spends its crash-loop budget. That behaviour exists today; this
PR is the natural place to notice it.

Things I checked and found sound, for the record:

  - The three-seat claim holds for the gateway: `SwapCoordinator` has no fourth holder of a
    candidate at shutdown, `in_transition` is set in `take()` and cleared in `release()`, and both
    halves are pinned (tests/test_core_runtime_swap.py:568 and :571).
  - The sweep's own ordering claim is tested: with the bound seat raising,
    tests/test_cli_gateway_commands.py:1101-1106 asserts the staged and in-transition seats are
    still stopped, so a raising seat does not strand the ones behind it.
  - `stop_file_watcher` is safe on a seat whose watcher is already gone -- the catalogue documents
    the start side as idempotent, and I called the stop twice against a real catalogue without an
    exception.

0xKT (approved):

PR #491, opened 2026-09-18
test(ui-web): gate the seam, the first-frame literals and the page frame classes

0xKT:
## Summary

Nine commits of gates and documentation on the long-lived `ui-web` branch: the two new gates and the two wider ones, a re-pin after the tasks panel landed under the PR, the two fixes the review asked for, and four commits closing what a red-team pass over the four gates then found. Nothing the page renders or does changes: both boot goldens, every region golden and every island snapshot are byte-identical, and the diff is six files under `ui-web/` -- four gate files, `CONTRIBUTING.md` and `README.md`.

- **Two rules that were written and not held now have gates.** `seam-assignment`: only `app/install.ts` assigns `sources.*` (22 sites), with the first frame's one assignment in `app/boot.ts` pinned to exactly one write, and `setSources` held to test files -- the regression this guards against has happened once already. The gate follows the seam under any local name a file binds it to (`import { sources as seam }`, `const s = sources`), and through `Object.defineProperty` and destructuring targets as well as plain assignment. `first-frame-literals`: the words the page shows without a `t(key)` behind them (the served `page.html` shell and 13 modules, 38 CJK runs in all) are counted per run, per file, and may only shrink; a new CJK literal anywhere in `src/` -- spelled out or as a `\uXXXX` escape -- fails with the instruction to put the words in the catalogue. The count uses the TypeScript parser so comments never count, and the same CJK ranges as the repo's `scripts/check_source_language.py`.
- **Two gates read the page frame and the shared components, which they never did.** `check-class-namespace` gains the namespaces `chrome` (`chrome/`, `App.tsx`, `main.tsx`; new classes carry `chrome-`) and `components` (new classes carry their component's file name, held to the longest matching one). Of the 176 class names there today, 10 are shared vocabulary, 2 already carry their component's name, 136 are pinned as-is in `LEGACY_CHROME` (plus 12 expression literals) and the 34 names borrowed from the domains' shared list are pinned by name in `LEGACY_BORROWED`, so a new borrow fails; `.newrun` and `.tipdn`, the two classes the frame writes that `page.css` never defines, are pinned in `UNSTYLED` rather than deleted here. The tool now reads every way a class is written -- the attribute in either quote style, string chunks inside a `className={...}` expression including its `${...}` holes, `class=` inside an HTML string, and `classList.add/remove/toggle/replace` arguments -- and asserts every non-test `.tsx` under `src/` belongs to one namespace. `state-dom-touch` counts reaches per match rather than per line, includes the page's own `$` wrapper, `getElementsBy*`, `document.activeElement` and `document.forms`, gains a `FRAME` table for the page frame's 26 imperative reaches over ten files (mostly `chrome/behaviour/*`, which exist to do that), pins the text, class and `innerHTML` writes the rule forbids, and says `lower it` when a file falls below its pin. The `state/` pin went from 58 reaches over 30 files to 79 over 34: four modules that reached only through `$` were never counted before.
- **`CONTRIBUTING.md` says what is true.** Section 4.3 no longer claims words come from `t(key)` and nothing else; it names the first-frame exception and the gate that pins it. Section 7 states the three facts the first CSS rule-mover needs: a rule moved into a domain's sheet lands after the whole of `page.css` (so same-specificity order can flip), nothing checks pixels today, and the exemplar `features/extAgents/styles.css` is empty on purpose. Section 10's step 4 adds the `ui-tui` catalogue regeneration a new namespace requires (`npm run --prefix ui-tui gen:i18n`), which the TUI CI job otherwise fails on. Sections 8 and 10 say that no gate requires a new module to carry tests. Section 11 carries the new pinned rows with their counts (and the shared and unprefixed totals corrected to what the tool prints, 84 and 357, which the tasks panel had moved), and `README.md`'s gate table has the two new gates and the two wider ones.

Every new or extended gate was mutation-checked, and then red-teamed: 48 evasions were tried against the four gates, 19 were caught and 24 slipped (6 of them plausible everyday edits: reaches or literals added on an already-counted line, template-literal holes, `classList`, HTML strings, the `$` wrapper). The last four commits close all 24 -- each mutation reproduced green before and red after -- except the ones that need a type checker or constant folding (`String.fromCharCode`, an aliased `document`, a class behind an identifier), which the gate headers now state as limits. One earlier check caught a defect in the first version of `first-frame-literals` (a scanner that stopped early on `.tsx` and counted a comment); the shipped version parses.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other

## Verification

From the repo root, on the branch head:

- `cd ui-web && npm run type-check && npm run lint` -> clean, 7 pre-existing warnings, 0 errors
- `cd ui-web && HOME=$(mktemp -d) npx vitest run` -> 165 files, 2596 tests passed, 0 skipped (the base plus two new gate files and the gates' new stale and write checks)
- `make build-ui` -> `boot-snapshot: OK (298 nodes match golden)` and `boot-snapshot: OK (285 nodes match golden)`
- `cd ui-web && node scripts/check-page.mjs && node scripts/check-css.mjs && node scripts/check-class-namespace.mjs && npm run gen:check` -> all OK; the class tool prints `919 classes and 159 expression literals across 21 domains; 84 shared, 357 unprefixed and 91 unprefixed in expressions pinned. 180 classes and 21 expression literals across the 2 namespaces beyond features/ ...; 136 unprefixed, 12 unprefixed in expressions, 34 borrowed from LEGACY_SHARED and 2 unstyled pinned`
- `uv run --frozen pytest tests/test_living_docs.py -q` -> 2 passed
- `make check-source-language` over `origin/main..HEAD` -> exit 0
- `uv run --frozen pre-commit run --from-ref refactor/ui_web_architecture --to-ref HEAD` -> all passed
- `git diff --stat refactor/ui_web_architecture..HEAD -- . ':!ui-web'` -> empty
- `git status` after the full run -> clean: no golden, snapshot or lockfile moved

## Risk

- Security: none.
- Backward compatibility: no runtime change. In-flight branches can go red on the new gates only by doing what the spec already forbids: assigning `sources.*` outside `app/install.ts`, adding a literal string the catalogue should carry, or adding an unprefixed class, a borrowed shared class, a DOM reach or a text write in the page frame. Each failure message says what to do, and every pinned count says `lower it` when the tree falls below it.
- Rollback: revert the squash merge.

## Related Issues

N/A


gloryfromca (dismissed):

gloryfromca (commented):
Blocking: the new class-namespace and seam-assignment gates both admit regressions they claim to prevent.

I found two false negatives and marked both inline. I covered the repository rules in AGENTS.md, the Web UI terms in CONTEXT-MAP.md and ui-web/CONTEXT.md, the complete target-to-head diff, surrounding source and callers, branch history, backward compatibility, whether tests were weakened, and the documented UI architecture constraints. The runtime is unchanged, and I found no separate compatibility issue.

Verification: `npm run type-check` passed; `npm run lint` passed with the seven documented warnings; the four changed gate files passed 16/16; and `node scripts/check-class-namespace.mjs` passed. Two full `npm test` attempts were not green: one ended at 2540/2544 and the other at 2541/2544, both solely in untouched `KnowledgePage.test.tsx`; that file passed 111/111 when rerun alone. The two inline false negatives were reproduced independently by mutation while the relevant gates still passed.

ui-web/scripts/check-class-namespace.mjs
```diff
@@ -402,20 +464,144 @@ for (const domain of domains) {
...
+      for (const c of tokensIn(expr)) add(exprNames, c, prefix)
+    }
+  }
+  if (!attrNames.size) {
+    fail(`${ns.reach} name no class, which means this check is reading the wrong files`)
+  }
+  nsClasses += attrNames.size
+  nsLiterals += exprNames.size
+
+  const prefixed = (c, where) => [...where].some((prefix) => carries(c, prefix))
```
gloryfromca: Blocking: the component namespace check must validate every component that names a class.

`attrNames` and `exprNames` merge a class into a set of component prefixes, but `prefixed()` returns true when only one of those prefixes matches. A wrong use in a second component is therefore hidden by the legitimate owner. I reproduced this by adding `model-tags` to `SetupSheet.tsx`: because `ModelTags.tsx` already names that class legitimately, the CLI still exited 0 with identical counts. That permits exactly the cross-component borrowing the new own-file namespace is meant to reject. Please validate each `(class, prefix)` pair (or otherwise require every recorded prefix to match).

ui-web/scripts/gates/seam-assignment.test.mjs
```diff
@@ -0,0 +1,166 @@
...
+
+  it('finds the installer it is written about', () => {
+    expect(production.filter((a) => a.rel === INSTALLER).length, `${INSTALLER} no longer installs the sources`)
+      .toBeGreaterThan(FLOOR)
+  })
+
+  it('assigns nowhere else but the installer and the pinned exception', () => {
+    const stray = production
+      .filter((a) => a.rel !== INSTALLER)
+      .filter((a) => !(EXCEPTIONS[a.rel] ?? []).includes(a.key))
```
gloryfromca: Blocking: the seam exception must be limited to the one assignment it documents.

This filter exempts every assignment with the same file and key, while the stale check only asks whether at least one exists. I duplicated `sources.rail = sessionsSource` in `app/boot.ts` in a temporary copy, and all four changed gate files still passed 16/16. Thus the claimed single first-frame exception can grow without detection. Pin the expected count or the exact assignment site so a second write fails.

gloryfromca (commented):
Blocking: the two existing gate false negatives remain unchanged after the rebase.

I refreshed `github/refactor/ui_web_architecture` to `5706bb21b1c0`, reviewed the full six-file PR diff and the revision delta, and found no new issue. The new commit correctly lowers the components expression pin from 9 to 8 after the updated base made `ok` shared. However, the logic cited in both open threads is unchanged, so those threads remain blocking and open.

Coverage included AGENTS.md, CONTEXT-MAP.md and ui-web/CONTEXT.md, the diff and surrounding callers, history, backward compatibility, test-strengthening/weakening, and the documented Web UI architecture constraints.

Verification: `npm run type-check` passed; `npm run lint` passed with 7 warnings and 0 errors; the four changed gate files passed 16/16; and `git diff --check` passed. Full `npm test` reached 165 files and ended with 3 failures in untouched `KnowledgePage.test.tsx` (2588/2591 passed); that file passed 111/111 when run alone.

ui-web/scripts/gates/seam-assignment.test.mjs
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in e3c96b417c7e. Repeating the duplicate sources.rail mutation now fails the new exact-count test and names both assignment lines, while the unmodified seam gate passes all 6 tests. The exception can no longer grow under the existing file/key pair.

ui-web/scripts/check-class-namespace.mjs
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in d68f01f121c2. Repeating the original mutation now exits 1 and reports that SetupSheet.tsx borrows .model-tags from ModelTags.tsx; the unmodified class check remains green. The per-class-and-prefix accounting closes this finding.

gloryfromca (commented):

Both prior findings are fixed and their threads are resolved. The class gate now measures each class/prefix pair independently, and the seam gate pins the boot exception to exactly one assignment. I repeated both original mutations: the cross-component class borrow now fails with the borrower and owner named, and the duplicate seam write now fails with both assignment lines named. I found no new issue in the revision delta or full PR diff.

Coverage included AGENTS.md, CONTEXT-MAP.md and ui-web/CONTEXT.md, the diff and surrounding callers, history, backward compatibility, test-strengthening/weakening, and the documented Web UI architecture constraints.

Verification: the four changed gate files passed 17/17; npm run type-check passed; npm run lint passed with 7 warnings and 0 errors; and git diff --check passed. Full npm test reached 165 files and ended with 2 failures in untouched KnowledgePage.test.tsx (2590/2592 passed); that file passed 111/111 alone, consistent with the pre-existing full-suite-only flake observed on earlier revisions.

gloryfromca (approved):

PR #492, opened 2026-09-18
feat(ui-web): take the sidebar and the conversation to the new design

xfng-sd:
## Summary

The rail, the transcript and the composer are drawn the way `ui-web/docs/design.html` draws them, and a step's sub-agent can be talked to from the step's own panel.

**The tokens.** Six of the names this needed are ones the old ramp could not say: a rail and a page were both "surface" to it, and so were a hovered row and a selected one, which is why the two drifted apart every time either was touched. `--page`, `--side-bg`, `--paper`, `--hover`, `--selected` and `--text-3` say which surface is meant, with `--bubble-me`, `--black`, `--ph`, `--ring`, `--sh-soft` and `--seg-on` for the pieces that invert.

One deliberate departure from the mock: light takes the design's own values, which this tree's light ramp already matched almost exactly, but **dark does not**. The design's dark is a neutral grey and every other surface here -- workspace, playbooks, desk -- is warm, so the dark values are derived from this tree's own ramp instead. A neutral rail against a warm page is the one result neither design wanted. Both themes are in the screenshots below.

**The rail.** White ground, a hairline seam, and one flat step of colour for hover and for selected. The lift the selected row used to carry -- a ring, a shadow, an amber glyph -- was a third treatment on a surface that now has two, and on a white rail the ring had nothing left to separate the row from. Group headings drop the mono uppercase eyebrow: a column of short names does not need a tracked-out capitalised row filing it.

**The conversation.** One turn, two shapes. What the reader wrote is a bubble that ends where the sentence ends, pulled to their side of the column; what the agent answered is a card of constant width, because a column of cards whose widths follow their word counts reads as a ragged edge. The card holds the whole turn -- the fold, the steps it holds and the answer -- so the process is part of the answer rather than loose text above it, and the footer sits under the card so what a turn delivered is above its own copy button. The column carries one gap instead of a margin per kind, which is what made a turn ending in a fold sit closer than one ending in prose. The composer is a solid pane: the blur it used to carry was reading the transcript through it, and a card that shows what is behind it competes with the one thing on the page the reader is making.

**The step's own field.** A step's executor is a stateful instance that outlives the step, so the flow tab ends in a one-line field addressed to it by name. Static, as asked: no `tasks.*` method answers it yet, so the field takes what is typed and the arrow lights when there is something to send. A step nobody started and one that was skipped never got an instance, and neither draws a field. The arrow is the shared `SendGlyph`, so this button and both real composers keep drawing one glyph.

Two pinned counts move with this. The class-namespace row rises by two names and the expression row falls by one -- the comment on that row says which names and why. And the sub-agent composer's gate now reads the card's material rather than the glass it used to be made of; what that case pins, that one rule owns the material and the pane names none of it, is unchanged.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `npm test --prefix ui-web` -- 2579 passed, 3 failed (163 files). The 3 are `KnowledgePage` failures that predate this branch in a domain it does not touch, and they are a Node-version issue: this lockfile's Vitest declares Node 20, 22 or 24+. They vary run to run within that one file and fail the same way on the base branch.
- `npm run lint --prefix ui-web` -- 0 errors, 7 warnings, identical to the base branch.
- `npm run type-check --prefix ui-web` and `npm run type-check --prefix ui-tui` -- clean.
- `npm run lint:i18n --prefix ui-tui` -- the generated catalogue is up to date after the new key.
- `python ui-web/build.py` -- production build and both boot snapshots pass, unchanged: the transcript is empty at boot, so the turn shapes do not move the golden.
- `python scripts/check_source_language.py` and `python scripts/check_large_files.py` over the branch range -- both exit 0.
- Driven in a browser against the offline fixtures, light and dark: the rail, a conversation, and a task pane's node detail. Checked that the field enables on input, that it is absent on the work-order tab, and that it is absent on a step that never ran.

## Risk

User-visible: this changes how the rail and the whole conversation look. No behaviour changes with it -- the transcript renders the same segments in the same order, and the turn wrapper is a new element around them rather than a change to what they are. The field on the flow tab sends nothing; it is markup waiting for a wire method.

The one non-cosmetic change is structural: an AI turn is now one card holding the fold, its steps and the answer, where those were three siblings. `stageRows` is what groups them, and the four kinds it cards are named there.

Rollback is the whole branch; nothing outside `ui-web/` and the message catalogue is touched.

## Related Issues

N/A


ui-web/src/features/tasks/TasksPage.tsx
```diff
@@ -274,6 +275,34 @@ function FlowTab({ node }: { node: TaskNode }): JSX.Element {
...
+  return (
+    <div className="tkchat">
+      <div className="tkchatbox">
+        <input
+          type="text"
+          value={draft}
+          onChange={(e) => setDraft(e.target.value)}
+          placeholder={t('gui.tasks.chat_on', { agent: node.agent })}
+        />
+        <button className="tkgo" disabled={!draft.trim()} aria-label={t('gui.send')}>
```
gloryfromca, line 298: Blocking: the task-step chat must not expose an enabled send action that silently does nothing.

After any non-whitespace input this button enables and is announced as `gui.send`, but it has no `onClick`, form submit, or keyboard submit path. `TasksSource` only exposes `list`, and `TaskNode` carries no stateful instance handle, so the existing `turn.send` instance path cannot be reached from here. In ordinary use, typing and clicking the new control therefore has no observable effect. Please either connect the affordance to a valid target or keep it non-actionable until that target exists.

ui-web/src/features/transcript/TranscriptPage.tsx
```diff
@@ -1029,7 +1029,7 @@ const AskView = memo(function AskView({ lane, seg }: { lane: Lane; seg: AskData
   }
   const showClip = seg.clipped && !seg.clipOpen
   return (
-    <div className="ask in">
+    <div className="turn me in">
```
gloryfromca, line 1032: Blocking: the renamed user-turn selectors must be updated so the transcript tests still exercise the rendered node.

This rename leaves `.ask` selectors at `TranscriptPage.test.tsx:1537`, `:2088`, and `:2108`. The streaming test now stores `null` and later proves only that `null === null`, so it no longer detects the remount it claims to guard; the live/replay shape helper likewise omits the user turn from both shapes. Retarget these assertions to `.turn.me` (and assert the captured element is non-null) so the changed suite cannot stay green while that behavior regresses.

gloryfromca (commented):
Blocking: the new task-step chat needs a real or non-actionable state, and the stale transcript selectors need to exercise the renamed user turn.

I reviewed the full target diff plus the surrounding task/transcript sources, the existing instance-chat caller, branch history and PR scope, Web UI architecture guidance, repository rules, compatibility, and the changed tests/gates. The structural transcript/card and styling changes otherwise look internally consistent.

Verification: the task and transcript suites pass (200/200); type-check and generated-contract checks pass; lint has 0 errors and 7 pre-existing warnings; source-language and large-file checks pass. The full web suite reports 2579 passing and 3 failing tests, all in the untouched KnowledgePage suite and matching the PR's documented baseline/unsupported-Node behavior.

PR #493, opened 2026-09-18
fix(raven-code): serve the skill route its menu advertises, and remember like the fleet

TongLi31:
## What

Two defects in the shipped `agents/raven-code` profile, both inherited from the retired vendored twin's comparison hygiene and both silent at runtime.

**The skill route was advertised but dead.** Trunk's skills are pull-discovery: a name+description menu rides the user envelope, and its footer tells the model, verbatim, *"Read one with read_skill(id); search differently with find_skill"* (`raven/context_engine/scent.py`). raven-code kept `read_skill`, `use_skill` and `find_skill` in its disable rows, so the menu taught a route that answered "unknown tool" -- while the router still ran, retrieval still happened, and every body it fetched was dropped unread, with nothing logged to say so. All three tools register off the skill registry alone (`raven/agent/loop/wiring.py`); a Hub endpoint only adds their `hub/` branch, and `hub` itself stays disabled.

**Memory was double-off.** design, oncall, ppt and research all ship `memory.backend: "everos"`; raven-code alone shipped backend `null` **plus** the `everos-memory` plugin opt-out -- the one shipped agent that could not remember. The identity was already `raven-code` in all three places (`memory.userId/agentId`, the plugin slice, `subagent.json`); only the two off-switches go.

The redundant `skillForge` block goes too: its two knobs (`rewriteEnabled`/`llmGateEnabled`) are only built under push discovery (`factory.py`), which no shipped agent runs, and the other four agents carry no such block.

**Review follow-up (second commit):** enabling everos-memory also offers the plugin's `understand_media` tool through the entry-point lane, which the hermetic face fixture cannot see -- an unledgered 14th tool in production. Same remedy as the playbook rows that board past the fixture the same way: the disable row is the pin, and `PLUGIN_LANE_WITHHELD` puts the lane on the ledger. raven-research holds the same line; design, oncall and ppt serve the tool deliberately.

## Fleet posture after this change

| | everos | read/use/find_skill | skillForge block |
|---|---|---|---|
| design / oncall / ppt / research | on | design serves all three | none |
| **raven-code (before)** | **double-off** | **all three disabled** | explicit false/false |
| **raven-code (after)** | on | all three served | none |

## Tests

The tool face moves on its ledger, not by luck (`tests/test_agents_code_launcher.py`): `SKILL_LANE_TOOLS` joins the face arithmetic, `find_skill` leaves `TRUNK_NEW_WITHHELD` (five stay withheld), `PLUGIN_LANE_WITHHELD` pins the plugin lane, and the everos posture test is rewritten from *stays-factory-off* to *ships-on*. The face itself is hermetically rebuilt from the render: 13 tools = the previous 10 + the three skill tools, verified by name and schema.

- `tests/test_agents_code_launcher.py`: 70 passed
- reviewer's set (`test_agents_code_launcher` + `test_agent_loop_skill_tools` + `test_context_scent` + `test_everos_plugin_discovery`): 110 passed
- ruff check + format: clean


agents/raven-code/config.json
```diff
@@ -49,15 +47,14 @@
...
-      "find_skill",
       "load_playbook",
       "plugin",
       "run_subagent_dag",
       "cron"
     ]
   },
   "memory": {
-    "backend": null,
+    "backend": "everos",
```
gloryfromca, line 58: Blocking: account for EverOS's contributed understand_media tool in the shipped raven-code face.

Activating `everos-memory` also admits every tool in that plugin's manifest, including `understand_media`; `build_runtime` then registers it. I reproduced the rendered raven-code runtime with the installed entry-point plugins and got the expected 13 ledgered tools plus `understand_media`. The new `_hermetic_build` assertion cannot catch this because it explicitly sets `entry_points_group` to `None`. That leaves the documented exact tool-face pin false and exposes a capability the new ledger says is absent. Please either explicitly withhold this tool or deliberately add it to the face and keep the EverOS entry-point lane live in the face test.

gloryfromca (commented):
Blocking: the EverOS activation exposes an undeclared tool outside the tested raven-code face.

I found one blocker inline. I reviewed the repository rules (`AGENTS.md` / `CLAUDE.md`), the Context Map and Runtime skill/memory terms, the full diff, config rendering and runtime/plugin call paths, fleet configurations and relevant history, backward compatibility and tool exposure, whether tests were weakened, and the documented exact tool-face architecture constraint.

I tried to refute the finding by assembling the rendered config through the production plugin-discovery lane. With the distributed plugins installed, the backend is `EverosBackend` and the visible face contains 14 tools, including `understand_media`; the authored hermetic face test passes because it removes the entry-point lane.

Verification:
- `uv run --all-extras pytest tests/test_agents_code_launcher.py tests/test_agent_loop_skill_tools.py tests/test_context_scent.py tests/test_everos_plugin_discovery.py -q`: 110 passed
- `uv run ruff check agents/raven-code/run.py tests/test_agents_code_launcher.py`: passed
- `uv run ruff format --check agents/raven-code/run.py tests/test_agents_code_launcher.py`: passed
- `PYTHONPATH=. uv run --frozen --extra dev python scripts/check_source_language.py github/main...HEAD`: passed
- `make check-source-language` could not run because `make` is unavailable; the canonical script behind that target passed directly.

TongLi31:
Blocker addressed in 73963d29.

The finding was right, and the hermetic fixture could not have caught it: it builds without the entry-point lane, so a plugin-contributed tool boards the production face invisibly. `understand_media` is now a disable row (beside the other media rows), which is the same pin this file already uses for the two playbook tools that board past the fixture the same way — and `PLUGIN_LANE_WITHHELD` puts the lane on the ledger so the face test asserts the pin explicitly.

Mechanism note: the disabled list is a registry-level dynamic filter (`registry.py`), and plugin tools register into the same registry last (`wiring.py`), so the row holds in the production lane.

Verification (same set as the review): `uv run --all-extras pytest tests/test_agents_code_launcher.py tests/test_agent_loop_skill_tools.py tests/test_context_scent.py tests/test_everos_plugin_discovery.py -q` → 110 passed; ruff check + format clean.

Fleet note: raven-research holds the same line on `understand_media`; design/oncall/ppt serve it deliberately — a per-face choice, not a fleet rule.

agents/raven-code/config.json
TongLi31: Fixed in 455f41b5 (was cd65ba2c before the rebase onto 210e2971).

You are right on both counts, including that the hermetic assertion structurally cannot catch it: `_hermetic_build` sets `entry_points_group` to `None`, so a plugin-contributed tool boards the production face invisibly to it. Chose to withhold rather than adopt -- memory boarding a media tool onto the coding face as a side effect is not a capability this product asked for, and raven-research holds the same line (design/oncall/ppt serve it deliberately).

Same remedy as the two playbook rows, which board past the fixture the same way: the disable row is the pin. `understand_media` now sits with the other media rows in `disabledTools`, and `PLUGIN_LANE_WITHHELD` puts that lane on the ledger so the face test asserts the pin explicitly instead of relying on it.

Mechanism checked rather than assumed: the disabled list is a registry-level dynamic filter (`raven/agent/tools/registry.py`), and plugin tools register into that same registry (`wiring.py`, "plugin tools register last"), so the row holds in the entry-point lane too.

Verification on your exact set: `uv run --all-extras pytest tests/test_agents_code_launcher.py tests/test_agent_loop_skill_tools.py tests/test_context_scent.py tests/test_everos_plugin_discovery.py -q` -> 110 passed; ruff check + format clean. Re-run after the rebase onto current main.

gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed. I rebuilt the rendered config through the installed entry-point plugin lane at this head: it selects `EverosBackend`, exposes exactly the 13 ledgered tools, and `understand_media` is absent. The focused 110-test set also passes. The explicit disable row and ledger assertion settle this finding.

gloryfromca (commented):

Reviewed the refreshed `github/main...HEAD` diff, the repository rules and canonical runtime terms, affected callers and plugin/runtime assembly, commit history and rebase equivalence, backward-compatible config behavior, exact tool-face architecture, and whether the tests were weakened. The prior plugin-lane blocker is fixed: the installed-entry-point build selects `EverosBackend`, exposes the intended 13-tool face, and withholds `understand_media`.

Verification: `uv run --all-extras pytest tests/test_agents_code_launcher.py tests/test_agent_loop_skill_tools.py tests/test_context_scent.py tests/test_everos_plugin_discovery.py -q` (110 passed); Ruff check and format checks passed; source-language and diff checks passed; all current PR checks pass. No new findings.

gloryfromca (approved):

PR #494, opened 2026-09-18
feat(ui-web): rebuild the settings dialog on the v0.2 prototype, with its backend

gloryfromca:
## Summary

Rebuilds the web settings dialog on the v0.2 prototype -- eight sections
(general, usage, models, skills, tools, plugins, archive, about) in place of
the thirteen tabs -- as a `features/settings/` domain in the shape this branch
gates, and adds the backend the pages need so every control reads a real
value and writes to the real place.

Backend (one commit, one test per method, each checked by a mutation run):

- `settings.usage` takes `from`/`to` and answers a per-day series;
- `session.list {archived}` and a lazy auto-archive pass inside the list
  (`sessions.autoArchiveAfterDays`: one appended record per stale unpinned
  session, none on the next list);
- `skills.manage inspect` carries the file list, `always` and the install
  stamp; `skills.manage open` opens a skill file on the gateway host;
  RPC-driven installs are stamped `trigger: "rpc"`;
- `model.add_models` (one write for a batch), `model.set_fields` (address,
  Azure deployment and API version, header patches), `model.oauth_login`
  (device flow for the page, URL + code); `model.options` gains `key_url` and
  the redacted `extra_headers`;
- `plug.retry`, `plug.revoke`, `plug.configure`;
- `skillForge.blocklist` is read live, so a switch takes effect on the next
  turn; reload-only keys return a `warning` the toast repeats.

Frontend (one commit): the domain files, the prototype's stylesheet under the
`settings-` prefix in `styles.css`, a shared props-only `components/ModelPicker`,
`gui.settings.*` in both locales, the old page and its 295 orphan keys
deleted, offline fixtures for the new methods, and gate pins moved with the
tree (`check-class-namespace` counts, two import pins, the unsent-field list).

Then a `fix(ui-web)` commit with nine faults only the real host showed (each
with a test), a `test(*)` commit pinning `toolGroups.json` to the tools the
loop registers, and a docs commit correcting two on-disk section names in
the acceptance cases.

Design record, acceptance cases and plan:
`docs/specs/2026-09-17-web-settings-page-design.md`,
`docs/specs/2026-09-17-web-settings-page-acceptance.md`,
`docs/plans/2026-09-17-web-settings-page.md`.

Reviewer notes:

- The base is this refactor branch; the six `fix(...)` commits at the bottom
  are PR #453 (still open against main) and will fall out once it merges.
- A trial merge of this branch with `main` shows three conflicts, none from
  this branch's files except one line: main's #464 rewrote
  `raven/skill_hub/hub.py::install` and no longer calls `write_install_meta`
  there, so the `trigger="rpc"` stamp this branch adds collides; take main's
  version and re-add the stamp only if a `write_install_meta` call remains.
  `ui-web/build.py` and `ui-web/src/live/120-settings.js` are this branch's
  base against main, not this PR.
- `tools.media.<kind>` takes the whole `{model, quality}` object and no null,
  so a media role is cleared by writing an empty model; the design said null.
- The dialog opens on the section the shared slot names (`usage`), as before;
  the prototype opens on general. Left as is: `state/settings.ts` is shell.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Commands run on the rebased branch (base: `refactor/ui_web_architecture`):

- `cd ui-web && npm run type-check` -- clean.
- `cd ui-web && npx vitest run` -- 175 files, 2536 tests passed, 0 failed,
  0 skipped (includes the seven convention gates).
- `cd ui-web && npm run lint` -- 0 errors (7 pre-existing warnings in other
  domains).
- `cd ui-web && npm run build && python3 build.py` -- both boot goldens
  unchanged (298 / 285 nodes).
- `uv run pytest tests/test_rpc_model.py tests/test_rpc_schema_match.py
  tests/test_settings_tool_groups.py tests/test_rpc_skills.py
  tests/test_rpc_settings.py tests/test_rpc_console.py tests/test_rpc_session.py
  tests/test_rpc_plughub.py tests/test_providers_oauth_login.py
  tests/test_skill_forge_blocklist_live.py -q` -- 935 passed.
- `scripts/check_source_language.py` and `scripts/check_large_files.py` over
  the branch -- exit 0.
- Real-host acceptance: a temp `RAVEN_HOME`, `raven serve --port 18899`, the
  built page driven by playwright-cli; 46 of the acceptance cases passed with
  every write read back from `config.json` or the session files, 1 partial
  (the offline stub page: boots with no console error, the dialog was not
  driven), 9 not run (real OAuth accounts, EverOS plugin install, catalog
  credential rewrites, three probes needing extra LLM turns). The record with
  commands, jq expectations and snapshots is kept outside the repo.

## Risk

- New RPCs only add methods and optional fields; existing methods keep their
  shapes (`model.options` gains two optional fields). The settings dialog is
  replaced whole; the previous one had no consumers outside the dialog.
- `skillForge.blocklist` is now read per turn instead of at start; a switch
  in the dialog changes the next turn's skills without a reload.
- Rollback: revert the frontend commit to restore the previous dialog; the
  backend commit can stay, nothing else calls the new methods.

- [x] Security impact considered (the settings whitelist still excludes the
  containment controls; header values and keys are redacted on every read)

## Related Issues

N/A


gloryfromca (commented):
Blocking: fix the live skill toggle, historical tool-range accounting, and the target-branch conflict before merge.

Two introduced runtime defects are marked inline. Both meet the late-round blocker bar: each is introduced by this change, reached through an ordinary Settings action, and leaves the operator without a working in-product escape. The branch also no longer merges cleanly into the current `refactor/ui_web_architecture`: `git merge-tree --write-tree HEAD github/refactor/ui_web_architecture` reports a content conflict in `ui-web/scripts/check-class-namespace.mjs`.

Coverage included the repository rules (`AGENTS.md`, `CLAUDE.md`, `CONTEXT-MAP.md`, the applicable `CONTEXT.md` files, and `ui-web/CONTRIBUTING.md`), the full target diff and current revision delta, relevant callers and history, backward compatibility paths, whether replacement tests weakened coverage, and the web source/RPC and runtime architecture constraints. The blocklist test is specifically weaker than its stated contract because it checks the registry proxy rather than the retrieval path.

Verification:
- `uv run pytest` over the 12 affected backend suites: 764 passed, 17 warnings.
- Focused web settings/model-picker suite: 81 passed.
- `npm run type-check`, `npm run lint`, and `npm run gen:check`: passed; lint reported 7 existing warnings and no errors.
- Source-language and large-file checks: passed.
- Full `npm test`: first run 2,533 passed / 4 failed; uncontended rerun 2,534 passed / 3 failed. All failures were in unchanged `KnowledgePage.test.tsx`; that file passed 111/111 in isolation, so those timing-sensitive failures are reported but are not attributed to this diff.

raven/memory_engine/skill_forge/catalog.py
```diff
@@ -70,23 +71,18 @@ def __init__(
...
-        if self._blocklist:
-            skipped = sorted({m.name for m in self._registry.list_all() if self._is_blocked(m.name)})
-            if skipped:
-                log.info(
-                    "skill blocklist: hiding blocked skill(s) from the local pool: %s",
-                    ", ".join(skipped),
-                )
-            pool_registry = _BlocklistRegistryView(self._registry, self._blocklist)
-        self._local_pool = LocalPool(pool_registry)
+        self._local_pool = LocalPool(_BlocklistRegistryView(self._registry, self._blocklist_reader))
```
gloryfromca, line 85: Blocking: make the live skill blocklist affect retrieval on the next turn.

`LocalPool` eagerly calls this proxy once and stores the resulting `_metas`/BM25 snapshot; `search()` never consults the proxy again. Reproduction on this head: search returned `codeword`, writing `skillForge.blocklist = ["codeword"]` made `catalog._is_blocked("codeword")` true, but the next `catalog.pool.search("codeword")` still returned `codeword`. The new test calls `catalog.pool._registry.list_all()` directly, so it bypasses the stale snapshot and does not exercise the runtime path. Since the context-engine blocklist is also captured at loop construction, the disabled skill can still be injected until restart. This is introduced here, disabling a skill is ordinary use, and there is no working escape besides restarting; rebuild or live-filter the retrieval path and test `pool.search()` (or the router) across both switch directions.

raven/rpc/methods/console.py
```diff
@@ -966,7 +1101,7 @@ def empty_totals() -> dict[str, Any]:
             except Exception:
                 continue
         sess_root = Path(load_config().workspace_path) / "sessions"
-        cutoff = (datetime.now() - timedelta(days=days)).timestamp()
+        cutoff = datetime.combine(frm, datetime.min.time()).timestamp()
         for p in sess_root.glob("*/*.jsonl"):
             try:
                 if p.stat().st_mtime < cutoff:
```
gloryfromca: Blocking: bound transcript fallback scans by the selected end date.

This only rejects files older than `from`. The new Usage page exposes historical custom ranges, so a transcript modified after the selected `to` date passes this check and all of its fallback tool calls are counted, while the LLM and telemetry-tool sides correctly read only the requested daily files. A normal historical range therefore reports internally inconsistent tool totals, with no UI workaround for obtaining the requested interval. This is introduced and ordinarily reachable here; add an exclusive upper bound at the start of the day after `to` (and a regression case with a later transcript mtime).

gloryfromca (commented):
Blocking: the two existing runtime findings still need fixes.

The rebase resolves the former target-branch conflict, and the CSS/model-picker/test-harness conflict resolution adds no new blocker. Both existing inline findings remain open and reproduce unchanged on this tree, so they continue to hold the stance; their details are not repeated here.

The live PR head changed during review from `96e112eeac33` to `7e9d06ec3d89`; the two commits have the identical tree (`f79c89e24706`), and this review is attached to the live head.

Coverage included the repository rules, the full target diff plus the rebase/conflict-resolution delta, relevant callers and commit history, backward compatibility, test-strength changes, and the web source/RPC and runtime architecture constraints.

Verification:
- `git merge-tree --write-tree HEAD github/refactor/ui_web_architecture`: clean; GitHub reports `mergeable: true`.
- Affected backend suites: 764 passed, 17 warnings.
- Focused settings, model-picker, CSS-balance, CSS-owner, and class-namespace suites: 91 passed.
- Type-check, lint, and generated-contract checks passed; lint reported 7 existing warnings and no errors.
- Source-language and large-file checks passed.
- Full web suite: 2,557 passed / 4 failed, all in unchanged `KnowledgePage.test.tsx`; that file passed 111/111 in isolation.

gloryfromca (commented):
Blocking: the two existing runtime findings still need fixes.

This revision only changes the Knowledge-page test wait and does not touch either open production finding: live skill blocklist changes still do not reach the cached retrieval snapshot, and historical usage fallback still lacks the selected range's upper bound. Both threads remain open.

The new wait also does not make the test deterministic in the normal full suite: `npm test` failed 3 of 2,561 tests, including the changed rename-refusal assertion timing out; the same file passed 111/111 in isolation. Backend focused tests passed 80/80, type-check passed, lint passed with 7 existing warnings, `git diff --check` passed, and the target merge-tree is clean.

Coverage: compared the one-commit delta with the previously reviewed full target diff; rechecked the applicable repository rules, affected test behavior, prior callers/history and compatibility conclusions, test integrity, and the existing architecture/runtime findings. No new inline finding is added because this revision neither introduces nor fixes the two ordinary-use, no-workaround blockers already recorded.

raven/rpc/methods/console.py
gloryfromca: Blocking: the live skill unblock path still uses the construction-time policy snapshot.

This usage-range finding is fixed. The fallback now excludes transcript mtimes at or after midnight following `to`, while retaining the inclusive selected end date, and the regression covers transcripts before, inside, and after the range. The focused backend run passed all 173 tests.

raven/memory_engine/skill_forge/catalog.py
gloryfromca: Blocking: make the Hub detail gate consult the live blocklist too.

The local BM25 snapshot and the next-turn candidate prefilter are fixed, but the re-enable direction still fails for Hub skills. A policy constructed with `blocklist=["codeword"]` and a live reader returning an empty set reports `blocked_now() == frozenset()`, yet `refusal_for_detail({"name": "codeword", ...})` still refuses it because `_refusal_for_identity()` reads `self.blocklist`. That is the method `_hydrate_bodies()` calls after the new live prefilter. Consequently, starting with a Hub skill disabled and switching it back on in ordinary settings use still cannot inject it on the next turn without restarting. This PR introduces that live switch behavior, the normal enable action reaches it, and there is no in-app recovery; use `blocked_now()` in the identity gate and cover both directions through Hub hydration.

gloryfromca (commented):
Blocking: make the Hub detail gate consult the live blocklist too.

The usage-range blocker is fixed and its thread is resolved. The local BM25 cache and live candidate prefilter are also fixed, but the same skill thread remains open because Hub hydration still calls `refusal_for_detail()`, whose identity check reads the construction-time `self.blocklist`. A Hub skill disabled at startup therefore remains refused after the operator enables it, despite `blocked_now()` seeing the empty live list. The thread reply includes the reproduction and the late-round blocker criteria: this live switch is introduced here, enabling a skill is ordinary use, and there is no in-app way out.

Verification: `uv run pytest tests/test_skill_forge_blocklist_live.py tests/test_context_scent.py tests/test_skill_segment_builder.py tests/test_skill_hub_policy.py tests/test_rpc_settings.py` passed 173/173; `uv run python scripts/check_source_language.py github/refactor/ui_web_architecture...HEAD` passed; `git diff --check` passed; and `git merge-tree --write-tree HEAD github/refactor/ui_web_architecture` completed cleanly. The direct policy reproduction returned an empty `blocked_now()` but a blocklist refusal from `refusal_for_detail()`.

Coverage: reviewed the two-commit delta against the previously reviewed full target diff; rechecked AGENTS/CLAUDE/context constraints, affected callers and policy flow, history and backward compatibility, test integrity, and the architecture boundaries. No duplicate inline finding was opened; the remaining defect stays on its original thread.

raven/memory_engine/skill_forge/catalog.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

Fixed on the live head. `_refusal_for_identity()` now consults `blocked_now()`, and the loop supplies that same live reader to read/find/use skill tools. In a detached checkout of `ef6803129`, the focused policy/tool/router suite passed 129/129 and the direct reproduction allowed, refused, then allowed the same skill as the live set changed. This settles both directions of the original retrieval finding.

gloryfromca (commented):

The two runtime findings are fixed, and both threads are resolved. The live blocklist now reaches the local index, context candidate/detail gates, and read/find/use tools in both switch directions; the transcript fallback is bounded at both ends. The UI delta opens the dialog before its first snapshot and coalesces the duplicate initial load without exposing an empty settings page.

Named nonblocking follow-up: the full web run is still flaky. `npm test` failed 2 of 2,562 tests in `KnowledgePage.test.tsx`; the edited rename test hit Vitest's enclosing 5-second test timeout, so its new 10-second `waitFor` ceiling cannot take effect under the loaded full suite. The two edited files passed 116/116 in isolation. This failure predates the latest backend fix and does not meet the late-round blocker bar, but the timeout relationship should be corrected rather than relying on isolation.

Verification on the live head: the detached `ef6803129` backend policy/tool/router run passed 129/129 and the direct live-list reproduction passed both directions. Type-check passed; lint passed with 7 existing warnings; source-language and `git diff --check` passed; the target merge-tree is clean.

Coverage: reviewed the UI delta and the subsequently pushed skill-policy delta against the previously reviewed full target diff; rechecked repository rules, affected callers and history, compatibility, test integrity, and architecture boundaries. No settled thread remains open.

gloryfromca (commented):

This revision resolves the remaining nonblocking test follow-up without weakening coverage: it creates the toast host before the action that raises the notice, restores the normal one-second `waitFor`, and retains the exact toast and open-dialog assertions. The failure mechanism is consistent with `show()` dropping notices when no `#toasts` host exists.

Verification: the full web suite passed 179/179 files and 2,562/2,562 tests; type-check passed; lint passed with 7 existing warnings; source-language and `git diff --check` passed; and the target merge-tree is clean. Both previously opened runtime threads remain resolved.

Coverage: reviewed the one-commit delta against the previously reviewed full target diff; rechecked repository rules, the toast caller/host lifecycle and history, backward compatibility, test integrity, and architecture boundaries. No new finding remains.

0xKT (approved):

PR #495, opened 2026-09-18
refactor(*): close out the participant contract the five plugins now write

ypflll:
## Summary

#469 gave a bundled sub-agent nine verbs in place of six hook phases. Reviewing it turned up a
list of things the contract said but did not yet do, and this PR closes that list. No behaviour
change is intended: the point is that the same seam becomes something a role can replace and
something a generated judgement can enter.

**1. The roles are named apart from the verbs, and `observe` is gone.** `ActionModule.review`
and the participant's `review` were two different things a line apart. The four delegating role
methods now say what they do -- they ask -- and the verbs keep their own names, so `ask_review`
asks `review`. `observe` is removed: a verb with no return value decides nothing and can only
keep state a participant is forbidden to keep. `outbound` says on its own docstring that it is
the one verb with no role seat, and `StepView` gains `phase`, so a verb asked at two moments
reads which one it is rather than inferring it from whichever other field happens to be set
(`tools_ran` becomes a property derived from it, so the two cannot disagree).

**2. The three verbs the seat was applying itself get a seat.** `system_addendum`, `archive` and
`select_tools` were rendered straight onto the hook decision, so each was the one judgement in
the set with nowhere to compose two participants, nowhere to vet what came back, and nothing a
replacement role could decide. They join the other four:
`MemoryModule.ask_system_addendum` / `ask_archive` and `CapabilityModule.ask_select_tools`. The
seat is not the point where narrowing takes effect -- `select` still is, and the registry still
adjudicates every call -- so a participant narrows after the product has spoken and never instead
of it.

**3. A participant answers with plain data.** `review`, `intake` and `system_addendum` returned
host classes (`Verdict`, `Intake`) and `salvage` returned `Any`. `Accept`, `Resample`, `End` and
`Intake` are now builders returning plain mappings, the host reads one back with `read_verdict` /
`read_intake`, and an answer it cannot read is silence rather than a refusal. That is what makes
the third invariant real and what lets a judgement compiled from source answer the same verbs a
plugin does -- it can build a `dict`; it cannot build a host class.

**4. A dispatch's judgements are a participant.** `judge` was the one verb a role answered by
reading the dispatch's own `checks` and `code` for itself. A Charter now enters as a participant
like any other, `ActionModule.ask_judge` takes the list, and the merge is a veto so the first
refusal decides. The participant list becomes the only route an external judgement takes, which
is the thing step four needs: a judgement generated for one dispatch joins the same list without
a second way in.

**5. The rename.** `AgentConduct` is `AgentParticipant`, `contracts/agent_conduct.py` is
`contracts/participant.py`, `agent/hook/conduct.py` is `agent/hook/participant.py`, and
`agent/harness/conducts.py` is `agent/harness/participants.py`. "Conduct" was already taken in
this tree -- raven-code's Coding Conduct prompts are a different thing -- and a reader hitting both
had no way to tell them apart. The rename was done symbol by symbol, not by search and replace,
for exactly that reason.

**7. Review (LivXue, at 3a1bf07c).** The `judge` disclosure had reached two of its four homes;
the other two are `ActionModule.ask_judge` on the paper and `DefaultAction.ask_judge`, and they
survived because they phrase the claim without naming any symbol the earlier fix touched. The test
meant to pin that claim could not fail for the reason it was written -- it asserted the refusal
rather than the argument, and built a `ParticipantHook` it dropped -- and now asserts what the
registry hands the role, mutation-checked by wiring a participant through. `compose_judge`
exploded a one-sentence `str` refusal into one refusal per character, because `str` satisfies
`Sequence[str]` structurally; `charter_code.run_judge` coerced it where this composition used to
live and the handling did not move with it. Two records corrected: the type-only
`raven.contracts` -> `raven.agent` edge, the first in the tree and the price of moving `Verdict`
and `Intake` off the paper, is now written beside the import contract it bends rather than left to
be discovered; and the ledger entry for the 3,200 -> 3,410 bump is restored to the names and count
it was written with. Not taken: reading `step.phase` at `research_flow/flow.py:1000` -- the gate
chain's phase vocabulary and `StepView.phase` coincide on `after_iteration` only, and that call
site wants `before_execute_tools`, which is the key `_phase` dispatches on.

**6. What the paper was claiming and not doing.** Read back over the seam afterwards rather than
over the diff. ``judge`` read as though a plugin's override were asked ahead of a dispatch's own
rules; it is not, and cannot be as written, because the party that asks the verb is
``ToolRegistry.execute``, which is not the hook chain and holds no handle on this turn's
participants. The merge order is real; nothing puts a plugin in the list. Wiring one through is a
turn-scoped register this change does not add, and it would start refusing calls that run today,
so it is named on the paper and pinned by a test instead. ``max_iterations`` had lost its
docstring to the property inserted between the field and the string documenting it. No executable
line moves in this commit: the module's AST, docstrings stripped, is identical to its parent's.

CONTEXT.md's entry is rewritten to the end state, including the two kinds of role method, the two
verbs that are exceptions, and the note that the Coding Conduct prompts are unrelated.

One item from the same list is deliberately not here: a `planning.*` block on the dispatch
parameters. It is not a contract closeout but a new model-facing field, and it needs the
generator's schema and prompt to move with it.

Contract papers: `participant.py` and `harness.py` are both factory-loop tier, so the
contract-tier digest and `CONTRACTS_VERSION` do not move. `CONTRACTS_LINE_CEILING` 3,440 to
3,470 to 3,500 to 3,540 to 3,550 to 3,570, each bump with its own review paragraph in
`tests/test_kernel_budget.py`. The chain was recomputed after rebasing onto #410, which added
seven lines to `loop_hooks.py` inside the old headroom and left one commit here with one line of
it; every "Measured at" figure in that file is the count at that commit.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [x] Refactor
- [ ] Other

## Verification

Gates, run at each of the seven commits:

```
uv run lint-imports
uv run pytest -q
```

Import contracts: 10 kept, 0 broken. The suite is green: the failures this machine shows are the
same ones it shows on an unmodified `main` in the same worktree layout (missing browser and
optional packages), and this branch adds none.

Behaviour preservation. The oracle is the hook seam itself. `main` and this branch were run under
the same scripted provider, outputs normalised for nonces and temp paths, then byte-compared.
Four scenarios, all identical:

- the five-plugin hook seam, every phase's decision and metadata per iteration (23,840 bytes)
- a whole turn with no playbook: host turn, spawned sub-agent, tool calls and refusals
- a playbook-on worker lane in process, including the charter refusals
- six window-shrink scenarios

This is the seam the change is about, and identical output across it is what "no behaviour
change" means here -- it is not a claim about paths the four scenarios do not reach.

Two claims had no assertion anywhere and now do. ``phase``, and the ``tools_ran`` derived from
it, are what three of the five bundled plugins branch on -- ppt on its close, oncall on its
counter, research on which gate runs -- and a mistyped phase string at one of the six call sites
would have changed what they do with a green suite. The read-only test now also covers
``mode_overlay``, the one frozen field it skipped. All three assertions were mutation-checked:
mislabelling a phase, deriving ``tools_ran`` from the wrong one, and dropping the freeze each
turn exactly their own test red.

New tests, each written against a claim the contract now makes rather than against the code:
a participant cannot write through the `StepView` it is shown; an answer the host cannot read is
silence, not a refusal; a participant that answers with a mapping it built itself is understood;
every verb with a seat goes through its role rather than around it, asserted as a set so a new
verb that bypasses its role fails; two seats in one chain each keep one addendum; two turns in
flight keep their own participants; a bound role decides what a participant's verdict does; a
plugin's own judgement speaks before the dispatch's, and a silent participant lets the next one
speak.

Real model, Sonnet 4.5 through OpenRouter, on this branch: a query generates a playbook, two
sub-agents are instantiated and run, each writes its `summary.md` under its own fence, and both
results are announced back to the originating session.

## Risk

For plugin authors: the six-phase `AgentHook` is untouched and a plugin shipping one keeps
working. A plugin written against #469's `AgentConduct` needs the new names and, if it answered
with `Verdict` or `Intake`, the builders instead -- the five bundled plugins are ported here and
are the worked examples.

One verb narrowed, and silently for a plugin that used the width. `salvage` was annotated
`Any | None` and `compose_salvage` returned whatever was not None; it is now `str | None` and only
a non-empty string is taken. A third-party plugin that salvages with a reply object rather than
text stops being heard rather than raising -- which is the third invariant working as written, and
worth knowing before upgrading. The narrowing is deliberate: `salvage` answers the text a turn
sends when it produced none, and `Any` was not a contract.

For anyone replacing a harness role: the four delegating methods are renamed and four more are
added, so a replacement written against #469 needs them. `bind()` names the missing member at
assembly rather than failing inside a turn.

## Related Issues

Follows #468 and #469.

Generated with Claude Code (https://claude.com/claude-code)


raven/contracts/harness.py
```diff
@@ -297,13 +335,20 @@ class ActionModule(Protocol):
...
+        participants: "Sequence[AgentParticipant]" = (),
     ) -> list[str]:
-        """Why this call must not run, or an empty list.
+        """Why this call must not run, or an empty list, composed from each
+        participant's ``judge``: the first that refuses decides, since a veto
+        needs one voice and two reasons for one refusal read as confusion.
+
+        The dispatch's own ``checks`` and ``code`` are a participant here rather
+        than something this role reads for itself, which is what lets a judgement
+        generated for one dispatch join the same list.
```
LivXue, line 357: **Please fix before merge.** The `judge` disclosure landed in two of its four homes.

3a1bf07c added the honest account to `AgentParticipant.judge` on the paper and to the CONTEXT.md
entry. This is a third surface and it still carries the wide claim: "composed from each
participant's `judge`: the first that refuses decides", and below it, that a dispatch's own rules
are "a participant here rather than something this role reads for itself, which is what lets a
judgement generated for one dispatch join the same list". A reader of the role protocol, who has
no reason to open the participant paper, is told a plugin's `judge` is composed.

Mechanism, same as the one now written on the paper: the only production call site is
`raven/agent/tools/registry.py:423`, which passes three arguments, so `participants` takes its
`()` default. The turn's participants live only in the hook context's
`metadata["raven.participant.<name>"]`, and `ToolRegistry` holds no handle on any of it.

This is the shape where correcting one home leaves the others standing, because the survivors
phrase the claim without naming anything the fix touched -- grepping the claim rather than the
symbol is what finds them. This one, and `DefaultAction.ask_judge`, noted separately.

LivXue (commented):
Reviewed at 3a1bf07c. Read the whole diff rather than a slice.

I started at ec9cccc7 and re-scoped when 3a1bf07c landed mid-review. That commit closed three of
what I had: the orphaned `max_iterations` docstring, the absence of any assertion on `phase` /
`tools_ran`, and the `judge` claim -- the last by disclosure rather than by wiring, which I think
is the right call and is argued well. `test_each_phase_names_itself_on_the_step` is a good shape:
the whole ordered list of six phases with their derived booleans, so a mistyped literal at any of
the six call sites turns exactly one test red. Everything below is against the current head only.

**Method.** "Reproduced" means I ran the branch's own code and pasted what came back; everything
else is code reading and says so. I measured the `raven/contracts` line count at every commit
rather than reading the figures. Lenses covered: rename completeness across the whole tracked
tree; the seat's phase-by-phase equivalence with the base; the merge rules base to head; the
import contracts and what the config excludes from them; every surface that states what `judge`
composes; test non-vacuity; and Windows.

**Two please-fix-before-merge**, both because the `judge` disclosure has not reached everywhere it
needs to:

- `contracts/harness.py:351` and `agent/harness/action.py:46` still carry the wide claim. Two of
  the four homes were corrected in 3a1bf07c; these are the other two, and they are the ones a
  reader of the role protocol meets first.
- `tests/test_participant_hook.py:340` cannot fail for the reason it was written. It collects
  `asked` and asserts only `refusals`, so the wiring it exists to catch would leave it green.

**Four non-blocking:** the new `raven.contracts` -> `raven.agent` edge that
`exclude_type_checking_imports` hides from lint-imports; `compose_judge` exploding a `str` refusal
into characters (reproduced); `CONTEXT.md:174` saying "a list of names" where the paper argues
definitions; and the half-updated ledger paragraph.

**One with no line to hang it on.** The Risk section does not mention that `salvage` narrowed.
`compose_salvage` now keeps only a non-empty `str`; the base returned any non-None object and
`AgentConduct.salvage` was annotated `Any | None`. A third-party plugin that salvages with a reply
object goes silent rather than erroring. The Summary mentions the change in passing, but Risk is
where a plugin author looks, and squash makes this description the commit body on main -- so the
commit body saying it is not enough.

**Merge state, as of this comment.** The branch is `dirty`. #410 landed on main after this branch
was cut, and both touch `agents/raven-research/plugins/research-flow/research_flow/flow.py` -- the
only overlapping file. The green CI at this head is therefore not evidence about the merged tree.
While that file is open for the rebase: line 983 reconstructs `"after_iteration"` from
`step.tools_ran`, which is now itself derived from `step.phase`. The string it wants is in hand.

**Checked and deliberately not filed:**

- Rename completeness. Zero hits for `AgentConduct`, `ConductHook`, `ConductFactory`,
  `agent_conduct`, `harness.conducts` anywhere in the tracked tree, markdown and prompts included.
  raven-code's own conduct vocabulary is correctly untouched and CONTEXT.md carries the
  disambiguation. Doing this symbol by symbol was the right call and it shows.
- `observe` removal. No implementation anywhere; the remaining `observe(` hits are an unrelated
  saturation counter and a browser MutationObserver.
- `tools_ran` semantics. Checked phase by phase against the base: only `after_iteration` is True,
  as before.
- `compose_review`. Equivalent to the base's `replace(accepted[0], note=...)`, including the
  case where the first participant said nothing.
- The empty-tool-array case. `compose_tools` returning `[]` survives the seat's
  `narrowed is not None` check, so narrowing to nothing is not swallowed by a truthiness test.
  Easy to get wrong; not gotten wrong here.
- Fault tolerance is genuinely better, and I do not think the description claims enough credit for
  it. The base's `review` returned `Verdict` and `compose_review` read `.accepted` off it, so a
  conduct answering None raised into the composite's catch and lost the phase. `read_verdict`
  makes that an accept.
- Windows. The diff introduces no filesystem, path, newline, encoding or subprocess surface;
  `Windows self-upgrade` is green at this head.
- CI. Green at this head. The one failing `pull request title` row is a superseded run from
  07:09:30, passing at 07:11:29.

Filed as a comment rather than changes-requested: nothing here is wrong at runtime today, and the
two please-fix items are two sentences and one assertion.

raven/agent/harness/action.py
```diff
@@ -34,13 +34,18 @@
...
         self,
         name: str,
         params: Mapping[str, Any],
         prior: Sequence[tuple[str, Mapping[str, Any]]],
+        participants: Sequence[AgentParticipant] = (),
     ) -> list[str]:
-        """Why this dispatch's Charter refuses this call, or an empty list.
+        """Why any participant refuses this call, or an empty list.
+
+        The dispatch's own judgements are one participant among them, asked
```
LivXue, line 46: **Please fix before merge** -- the fourth home of the same claim.

"The dispatch's own judgements are one participant among them, asked after whatever the caller
passes" is a sentence about callers. There is one caller, `registry.py:423`, and it passes
nothing: `participants` is `()` on every production call.

The merge order this describes is real and `test_a_plugins_own_rules_speak_before_the_dispatchs`
tests it properly. What the sentence adds, and should not, is that something arrives in that
parameter today. Same remedy direction as the `ActionModule.ask_judge` note: say what is wired,
the way the paper and CONTEXT.md now do.

tests/test_participant_hook.py
```diff
@@ -0,0 +1,558 @@
...
+        def judge(self, name, params, prior):
+            asked.append(name)
+            return ["the product refuses this one"]
+
+    ParticipantHook("product", ProductRules)  # seated is not enough; nothing carries it across
+    registry = ToolRegistry(verifier_provider=lambda: DefaultAction())
+    with charter_scope(Charter(checks=(CheckRule(tool="write_file", path_prefix="out/"),))):
+        refusals = registry._verifier_refusals("write_file", {"path": "/etc/passwd"})
+
+    assert refusals, "the dispatch's own rules must still be asked"
```
LivXue, line 345: **Please fix before merge: this test cannot fail for the reason it exists.**

The commit message says "A test pins the current behaviour, so whoever wires it has to come back
and change the prose." The only assertion here is `assert refusals`, which says the Charter still
refuses -- true before this PR, and true after any wiring anyone might add.

`asked` is declared at :328 and appended to by `ProductRules.judge`, and is never asserted on. And
`ParticipantHook("product", ProductRules)` is constructed and dropped, so the factory never runs
and `asked` could not have been non-empty whatever the production code did.

Failure scenario: someone plumbs the seat's participants through to the registry. `asked` becomes
`["write_file"]`, `refusals` becomes `["the product refuses this one"]` -- still truthy -- the test
stays green, and the prose this test was written to protect is silently false again.

Remedy direction: assert the negative the test's name promises (`asked == []`) beside the positive
one, and drive the participant through whatever path would carry it, so the assertion has
something to observe. As it stands both halves of the claim are collected and one is discarded.

raven/contracts/harness.py
```diff
@@ -46,10 +46,11 @@
 from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
 
 if TYPE_CHECKING:
-    from raven.contracts.agent_conduct import AgentConduct, Intake, StepView, Verdict
+    from raven.agent.harness.participants import Intake, Verdict
```
LivXue, line 55: Non-blocking, but I could not find anything in the repo that guards this.

This is the first `raven.contracts` -> `raven.agent` edge in the tree. At the merge base,
`git grep 'raven\.agent' -- 'raven/contracts/*'` returns one docstring cross-reference and no
import.

`pyproject.toml`'s "the kernel stands alone" contract lists `raven.agent` among the forbidden
modules for `raven.contracts`, and its comment is emphatic: "The set carries no exceptions ... an
ignore_imports entry here would be new debt, not a carry-over."

Why the pipeline does not see it: `[tool.importlinter]` sets `exclude_type_checking_imports =
true`, so `lint-imports` reports 10 kept / 0 broken with this edge present. The nearest test,
`test_the_kernel_imports_no_other_raven_package` in `tests/test_kernel_budget.py`, is scoped to
`raven/spine` and never reads `raven/contracts`.

Runtime is unaffected -- the import is type-only and `kernel wheel smoke` is green. What is lost is
the papers' type-level self-containedness, which is the stated purpose of that contract ("what
makes a raven-core wheel a build artifact rather than a refactor"): the papers now name a symbol
a papers-only wheel would not contain.

Not filed higher because the design reason for moving `Verdict` and `Intake` off the paper is
argued and I think it is right. The question is whether this edge is the intended price and
belongs written down next to that contract, or whether the shapes the shell reads belong somewhere
both sides can name.

raven/agent/harness/participants.py
```diff
@@ -0,0 +1,250 @@
...
+    """Why any participant refuses this call, the first that does deciding it.
+
+    A veto, so one refusal is enough and the rest are not asked: the sentences
+    reach the model as the call's result, and two participants' reasons for the
+    same refusal would read as one confused reason. Synchronous, because the
+    verb is.
+    """
+    for participant in participants:
+        refusals = participant.judge(name, params, prior)
+        said = [line for line in refusals if isinstance(line, str) and line] if refusals else []
```
LivXue, line 183: Reproduced. Non-blocking and latent -- no current caller reaches it -- but it sits exactly on the
seam this PR opens.

A participant whose `judge` returns one sentence as a `str` has it exploded into characters: `str`
satisfies `Sequence[str]` structurally, so neither the annotation nor a type checker objects, and
this comprehension iterates it. Run against the branch's own module:

```
class Sentence:
    def judge(self, name, params, prior):
        return "write under ./out/"

compose_judge("write_file", {}, [], [Sentence()])
-> ['w', 'r', 'i', 't', 'e', ' ', ...]      18 entries
```

At the model-sees level the refusal arrives as `"w r i t e   u n d e r   . / o u t /"`.

Two siblings already handle this shape and this one handles it neither way. `compose_tools`,
further down the same file, guards `isinstance(answer, str)` explicitly -- I checked, it returns
None for the same input. And `charter_code.run_judge:192`, the function this composition was
lifted out of, coerces: `if isinstance(out, str): return [out] if out.strip() else []`. Moving the
composition point out one layer did not bring that handling with it.

It also contradicts the third invariant stated at the top of this module -- "A malformed answer is
silence, never a refusal". Here a malformed answer is eighteen refusals.

Unreachable today: the Charter path is normalised by `run_judge` before it arrives, and no bundled
plugin implements `judge` (zero `def judge(` under `plugins-dist/` and `agents/`). Filing it
because this is the seam meant to carry judgements that are generated rather than written, and a
bare string is a plausible thing for one of those to answer.

CONTEXT.md
```diff
@@ -158,22 +164,35 @@ the last message (`append_note`), and `before_user_inbound` may rewrite the inbo
...
-a text, a narrower tool array, a note, a `Verdict`, a reply -- never a write into the loop's own
-state. The host builds one per Turn from a `ConductFactory`, so what a conduct has already done
-this turn is an attribute that dies with the turn. `ConductHook` seats one in the Agent Hook
-chain and renders its answers as the `HookDecision` the composite already merges, which is why
-the phases, their order and the rollback mechanism are unchanged by it. The four Harness Modules
-compose what the conducts of a seat say: Memory their `intake`, Planning their `advise`, Action
-their `review` and `salvage` -- so replacing a role replaces what a conduct's judgement does.
+`advise`, `system_addendum`, `review`, `salvage`, `judge`, `outbound`, `archive`. Each is asked
+about one step of a Turn, answered against a read-only `StepView` whose `phase` says which of
+the six moments is asking, and answered in plain data -- a string, a list of names, a mapping
```
LivXue: Non-blocking, but this is the canonical registry, so it is the copy that gets read as
authoritative.

"answered in plain data -- a string, a list of names, a mapping built by `Intake`, `Accept`,
`Resample` or `End`". No verb answers with names. `select_tools` answers with definitions, and
`AgentParticipant.select_tools` spends a paragraph on exactly this distinction: "Definitions
rather than names, and so deliberately out of step with the Charter's `tools` field, which is a
name tuple ... which is why the mismatch is written down here rather than papered over". `judge`
answers with sentences.

So the one place a reader checks the vocabulary contradicts the one place this PR argues that
vocabulary hardest.

tests/test_kernel_budget.py
```diff
@@ -140,10 +140,10 @@
 first bump gave.
 
 3,200 -> 3,410 (2026-09-18), and this is that review. The change is a whole
-paper: contracts/agent_conduct.py, 191 lines, the nine verbs a sub-agent
-implements instead of the six hook phases (``AgentConduct``, its ``StepView``
+paper: contracts/participant.py, 191 lines, the nine verbs a sub-agent
```
LivXue: Non-blocking: this ledger entry was updated for the rename and not for the content change.

The paragraph is the record of the 3,200 -> 3,410 bump. Renaming the symbols inside it makes it
read as a present-tense description of `contracts/participant.py`, and two of its facts are now
false of that file: "191 lines" (259 at this head), and at :145 "the ``Intake`` and ``Verdict`` it
answers with" -- `Verdict` moved off the paper to the harness in this very PR.

Either refresh it or leave the historical entry alone. Half of each is what leaves a reader unable
to tell which parts are dated measurements and which are live claims.

Separately, needing no change here. The ceiling arithmetic in three commit bodies disagrees with
the code and with these paragraphs: c1 says "3,420 -> 3,460" (3,420 was the measured count, the
ceiling was 3,440), c2 says "-> 3,500" and sets 3,490, c4 says "3,500 ->" from 3,490. Every
"Measured at" figure in the file is correct, including 3,521 at this head -- I recomputed all of
them -- and squash drops commit bodies, so nothing wrong reaches main. Noted only so the record is
straight.

ypflll:
Thanks -- this was a good read, and the two it caught that I had just written myself are the ones I
most needed caught. Pushed as `c3524fce`, on top of a rebase onto #410. Point by point.

**Taken, all three please-fix.**

`ActionModule.ask_judge` and `DefaultAction.ask_judge` now say what is wired. Your note on *why*
they survived is the part I want to keep: both phrase the claim without naming any symbol the
earlier fix touched, so grepping `judge` found the fix sites and not these. I had grepped the
symbol. Grepping the claim is the habit.

`test_a_seated_participant_is_not_asked_to_judge_a_tool_call` is gone and replaced by
`test_the_registry_asks_judge_with_no_participants_at_all`, which spies on the role and asserts
`handed == [()]`. You are right about both halves -- `asked` was collected and discarded, and the
`ParticipantHook` was constructed and dropped so the factory never ran and it could not have been
non-empty regardless. I mutation-checked the original but only against the phase literals, never
against the wiring, which is exactly the axis the test existed for. The replacement is checked the
right way: wiring a participant into `_verifier_refusals` leaves the old assertion green and turns
the new one red.

`compose_judge` coerces a bare `str` now, the way `run_judge` does, with a test and a mutation
check. Reproduced your eighteen before fixing. Your framing is the right one -- the composition
moved out a layer and the handling did not move with it.

**Taken, the four non-blocking.**

The `raven.contracts` -> `raven.agent` type-only edge is written down in two places: beside the
"kernel stands alone" contract in `pyproject.toml`, and at the import itself. You asked whether it
is the intended price or whether the shapes belong somewhere both sides can name. It is intended
-- a host class on the paper is the thing a generated judgement cannot build, which is the whole
of commit 3 -- but "intended and undocumented" was not a defensible state for the first edge in
the tree, particularly with that comment asserting no exceptions. If the papers ever ship as their
own wheel, that comment now names the one edge to resolve.

`CONTEXT.md` said "a list of names"; it says definitions now. You are right that it is the worst
possible place for that error, given the paper argues the distinction hardest.

The 3,200 -> 3,410 ledger entry is restored to the names and the count it was written with, plus a
line saying an entry is a dated record of one bump rather than a description of today's file. Your
diagnosis of "half of each" is what made the right fix obvious.

Your reading of the three commit bodies' ceiling arithmetic was correct, and the rebase onto #410
made it moot in an instructive way: #410 added seven lines to `loop_hooks.py` inside the old
headroom, so every measurement in this branch moved and one commit was left with a single line of
room. The whole chain is recomputed -- 3,470 / 3,500 / 3,540 / 3,550 / 3,570 -- and the prose
figures with it.

**Not taken, one.**

`research_flow/flow.py:1000`. The string is not in hand. That call site wants
`"before_execute_tools"`, which is the gate chain's own phase vocabulary and the key `_phase`
dispatches on; `StepView.phase` says `"execute_tools"`. The two vocabularies coincide on
`"after_iteration"` and differ everywhere else -- the plugin kept the six hook-phase names
deliberately so six thousand lines of gate bodies did not have to move. Substituting would route
that call to no gate. I have left it and said so in the commit body.

**The one with no line to hang it on: taken, and you are right about where it belongs.** `salvage`
narrowed from `Any | None` to `str | None`, and `compose_salvage` from "any non-None" to "a
non-empty `str`", so a third-party plugin salvaging with a reply object goes quiet rather than
erroring. That is now a paragraph in the Risk section, which is where you said a plugin author
looks, and it survives squash.

**On the merge state.** Rebased onto `210e2971`. The only overlapping file was the one you named,
and both conflicts in it were the same import block -- #410's `TURN_ASK_KIND_KEY` import against
my rename -- resolved by keeping both rather than either. Re-verified on the merged tree: the
five-plugin seam A/B against the new main is byte-identical (23,952 both sides), the full suite's
failure set is a strict subset of new main's (127 vs 127 failed, 88 vs 88 errors, zero branch-only
failures), and the contract gates pass at each of the seven commits individually.

raven/contracts/harness.py
LivXue: Verified at `6d82c455`. The new paragraph says it plainly -- "``participants`` arrives empty on every production call and the composition above describes an order nothing yet fills" -- with the reason a reader of this protocol needs and the alternative named (`review`). The closing sentence, that it is stated here because a reader of this protocol has no reason to open the participant paper, is what makes this durable rather than a patch on one sentence.

Resolving.

raven/agent/harness/action.py
LivXue: Verified at `6d82c455`. "There is one caller, ``ToolRegistry._verifier_refusals``, and it passes nothing: the order is real and tested, but nothing arrives in ``participants`` today."

Keeping the sentence about callers and making it true is better than deleting it -- the merge order stays documented for whoever wires the seat through.

Resolving.

tests/test_participant_hook.py
LivXue: Verified at `6d82c455`. The old test is gone (zero hits for its name) and `test_the_registry_asks_judge_with_no_participants_at_all` asserts `handed == [()]` off a `Spy` subclass of `DefaultAction`, so the assertion reads the argument the registry hands over rather than the refusal that comes back. That is the axis the claim lives on: plumbing participants into `_verifier_refusals` makes `handed` non-empty and turns this red, where `assert refusals` would have stayed green. It also keeps the positive assertion, so it cannot pass vacuously on a registry that stopped refusing altogether.

One note for anyone reading this thread later. GitHub has re-anchored my comment above onto line 345 of the current head, which is the `assert refusals` line *inside the replacement test*. "The only assertion here is `assert refusals`" was true of the deleted test and is not true of the code it now points at -- `assert handed == [()]` sits directly above it.

Resolving.

raven/contracts/harness.py
LivXue: Verified at `6d82c455`, both places. The note at the import states the edge, why the shapes sit off the paper, and that import-time closure is unaffected.

The paragraph in `pyproject.toml` is the one I would have asked for had I thought to: it sits directly under the "carries no exceptions" sentence it qualifies, names the file and the symbols, says lint-imports does not see it because `exclude_type_checking_imports` is on below, and ends "If the papers are ever built as their own wheel with type stubs, this is the one edge to resolve."

That last line is what turns an undocumented edge into a tracked one, and it answers what I asked: intended, and now findable from the contract rather than only from the import.

Resolving.

raven/agent/harness/participants.py
LivXue: Verified by re-running the same probe against `6d82c455`, in-process off the branch's own module:

```
bare str        "write under ./out/"   ->  ['write under ./out/']
whitespace str  "   "                  ->  []
list with empty ["a", "", "b"]         ->  ['a', 'b']
```

All three match `charter_code.run_judge`, the function this composition was lifted out of, so the two ends of the seam now agree instead of one coercing and the other exploding. The eighteen entries are gone.

Resolving.

CONTEXT.md
LivXue: Verified at `6d82c455`: "a string, a narrower list of tool definitions, a mapping built by `Intake`, `Accept`, `Resample` or `End`". The registry and the paper now say the same thing about the one distinction the paper argues hardest.

Resolving. (Shows as outdated because the line moved.)

tests/test_kernel_budget.py
LivXue: Verified at `6d82c455`, and the fix is better than the one I suggested. Rather than refreshing the figures, the entry is restored to the names and the count it was written with, plus a sentence saying what an entry in this ledger *is*: "an entry here is a dated record of one bump, not a description of the file today", with pointers to the later entries carrying the rename and the growth. Refreshing it would have destroyed the archive's readability as an archive; this keeps it.

I recomputed the whole chain after the rebase. Measured per commit: 3,470 / 3,500 / 3,500 / 3,540 / 3,540 / 3,550 / 3,570, which de-duplicates to exactly the five values you reported, and every commit sits 11 to 28 lines under its own ceiling. Your "one commit was left with a single line of room" checks out too: `fc5d0928` measures 3,489 against the 3,490 it carried before the rebase.

Resolving.

LivXue:
Re-reviewed at `6d82c455`. Every fix verified, with the method in each of the seven threads, all now replied to and resolved. Two things belong here rather than on a line.

**You were right and I was wrong about `research_flow/flow.py`.** I checked the reason rather than taking it, and it holds. `_phase` dispatches on the gate chain's own keys, and the two vocabularies line up like this:

```
gate chain (_phase keys)   StepView.phase
before_user_inbound        user_inbound
before_iteration           iteration
before_execute_tools       execute_tools
after_iteration            after_iteration    <- the only member they share
terminal_answerless        answerless
after_send                 sent
```

`step.phase` is `"execute_tools"` at the moment that line needs `"before_execute_tools"`, so the substitution would dispatch to a key with no gate behind it.

Naming what misled me, since that is the reusable part: I generalised from the one member where the two vocabularies coincide. That line reads `"after_iteration"` on both sides of its conditional, which made the string look already in hand, and I never checked the other five keys against the dispatch table. Two phase vocabularies overlapping on exactly one member is a nasty shape -- the shared member is the one a reader samples, and it makes the two look like one vocabulary. Your commit body is the right home for that.

**One bookkeeping note.** Your reply cites `c3524fce`. That sha is not on the branch: it is the pre-rebase commit with the same subject, and `git merge-base --is-ancestor c3524fce HEAD` says no. The commit carrying this work is `6d82c455`. Worth mapping, since a rebase stales every sha in a review trail and this one still resolves to an orphan object for now.

**State as I read it.** Rebase clean -- `210e2971` is an ancestor, `0/7` left-right, and the only overlapping file is the one you named, with both imports kept rather than either. 14 checks green at this head, two unit shards still running. `mergeable_state` is `blocked` rather than `dirty`, so the conflict is gone and what remains is the approval gate.

No further findings. I have not independently verified the two claims that need your harness -- the seam A/B at 23,952 bytes on the merged tree, and the failure-set subset at 127/127 and 88/88 -- and I am taking those as reported.

ypflll:
Thanks for the second pass -- all seven threads verified and resolved on your side, nothing new
filed, and the correction on your own comment's re-anchoring was a courtesy to whoever reads the
thread later.

State as it stands: seven commits, rebased onto `027637b0` (main moved twice more; neither commit
touched a file this branch touches), 17/17 checks green at `6d82c455`, zero unresolved threads.
The suite's failure set on this machine is identical to main's, and the five-plugin seam A/B is
byte-identical against the current base.

`reviewDecision` is still `REVIEW_REQUIRED` -- both your rounds were filed as COMMENTED, which
matches what you said the first time about nothing being wrong at runtime. If you are satisfied
after the second pass, an approval is the one thing left to unblock it. If you would rather take
another look first, or you want the roster change discussed below landed before this merges, say
so and I will hold.

For context on what comes next, since it touches this seam: the roles are still asked once per
seat with a roster of one, so the merge rules in `harness/participants.py` are idle for every verb
but `judge`. That is a behaviour change for a chain with two or more seats, so it is its own PR
rather than more commits here.

ypflll:
Correction, and thank you for catching it. `c3524fce` is not on the branch -- `git merge-base
--is-ancestor c3524fce HEAD` says no here too. It is the pre-rebase commit with that subject, and
citing it in a review trail that outlived two rebases was sloppy of me.

The trail spans three bases now, so here is the whole map rather than the one sha. Subjects are
identical down each row; only the base moved.

```
first review    #410 rebase   027637b0 rebase   subject
ec9cccc7    ->  b3f30d62  ->  f2ca709f          ask_* on the asking half
3a1bf07c    ->  0ec2c9da  ->  0f0da098          say on the paper the two things it was not doing
c3524fce    ->  --------  ->  9120476f          finish the judge disclosure
```

`git diff c3524fce 9120476f` is exactly main's two intervening commits
(`context_engine/segments/render.py` and its tests, plus the benchmark chart) and nothing of mine,
so every verification either of us did at `c3524fce` still stands at `9120476f` -- but the object
is only reachable while a reflog holds it, which is the part that matters for anyone reading this
later. From here I will cite the head sha and say what it was rebased from.

On `research_flow`: the table you built is the clearer artefact, and "two phase vocabularies
overlapping on exactly one member" is the sentence worth keeping -- the shared member is the one a
reader samples. I have added your table to the commit body so it travels with the code rather than
living only in this thread. That amend is message-only -- `git diff 6d82c455 9120476f` is empty --
and it is why the head is `9120476f` rather than the `6d82c455` you verified; the tree you read is
the tree that is there.

Noted that the seam A/B and the failure-set subset are taken as reported. Both are scripted and
the scripts are reproducible off the branch; say the word if you want them attached or re-run
against anything in particular.

gloryfromca:

The correction is accurate. `9120476f` has the same tree as the reviewed `6d82c455`; the only delta is the final commit message, which now preserves the phase-vocabulary explanation and table. The seven review threads remain resolved.

Verified on the current head:

- `git diff --exit-code 6d82c4558bca36126e7deed393e032142531d3f7 9120476f8f2d850a1a38067d9b3071dfdbf1e70b -- .` passed with no tree diff.
- `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x`: 25 passed.
- `gh pr checks 495 --repo EverMind-AI/Raven`: all 17 checks passed.

The multi-seat roster behavior described for the follow-up is a separate behavior change and does not block this closeout.

gloryfromca (commented):

Reviewed the new `54679add` revision after refreshing `github/main` to `f633127e`. This is a rebase-only revision: `git range-diff` marks all seven PR commits patch-identical to the previously reviewed series, and the three files changed by #493 no longer appear in the PR diff against the refreshed target. No review finding survives refutation because there is no new PR code delta.

Verification on this head:

- `uv run pytest tests/test_participant_hook.py tests/test_agents_code_launcher.py tests/test_kernel_budget.py -x`: 95 passed.
- All 17 GitHub checks passed.
- All seven existing review threads remain resolved; zero unresolved threads.

Coverage for this revision: repository rules and canonical context, the complete target diff via exact patch-series comparison, callers and compatibility already covered by the unchanged reviewed patches, test-strength preservation, architecture boundaries, and rebase history/integration with #493.

gloryfromca (commented):

Reviewed the revision delta. `7d6d6fb6` is a rebase onto `45203a4a`, whose only change since the prior base is `.github/CODEOWNERS`; it has no path overlap with this PR. `git range-diff` marks all seven PR commits patch-identical, so the previously reviewed code, callers, compatibility behavior, tests, and architecture boundaries are unchanged. No new finding has a concrete failure to report.

Current-head verification: `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed 25/25; all 17 GitHub checks passed; zero review threads are unresolved. Repository rules, canonical context, target diff, history, backward compatibility, test strength, and architecture were covered through the unchanged patch series plus the no-overlap base delta.

gloryfromca (commented):

Reviewed the revision delta only. `9a2a1df9` rebases the same seven patch-identical commits onto `c2c5f279`; the new base changes only `README.md` and `README.zh-CN.md`, with no overlap with this PR. There is therefore no new code, caller, compatibility, test-strength, or architecture change to raise.

Current-head checks: `git diff --check github/main...HEAD` passed; the focused participant/kernel tests passed 25/25; zero review threads are unresolved. Twelve GitHub checks are green and the four unit shards are still running at review time.

Covered for this revision: repository rules and canonical context, the target diff and history through exact `range-diff`, base/PR overlap, callers and backward compatibility inherited unchanged, test preservation, and architecture constraints.

raven/contracts/participant.py
0xKT, line 1: **Blocking:** `AgentParticipant.judge` is paper no host asks, and about a third of what the
contracts ceiling was raised for is spent on it.

Said first, because it matters: this is disclosed, not concealed. The verb's own docstring
(raven/contracts/participant.py:200-209) says "the seat does not ask this verb ... a plugin that
overrides `judge` is not yet reached", CONTEXT.md and the budget entries say it again, and
tests/test_participant_hook.py pins the unwired state. What follows is about the contract tier's
bar, which is necessity rather than honesty.

It has no production caller. The only route to a refusal is:

    raven/agent/tools/registry.py:423   role.ask_judge(name, params, prior_calls())   <- no participants
    raven/agent/harness/action.py:66    compose_judge(..., [*participants, *charter_participants()])
    raven/agent/subagent/charter.py:265 charter_participants() -> (CharterParticipant(),)
    raven/agent/subagent/charter.py:242 class CharterParticipant:                      <- inherits nothing

`compose_judge` has exactly one caller (action.py:66) and `ask_judge` exactly one
(registry.py:423), across raven/, plugins-dist/ and agents/. Nothing hands that path a
participant, and the one participant it builds does not inherit the paper's verb, so
participant.py:183 is never dispatched to.

Measured here, with the control that makes it mean something:

    baseline, five suites                        -> 115 passed
    delete the whole judge block from the paper  -> 115 passed     <- nothing notices
    delete `advise` instead (control)            -> 4 failed       <- the suite does bite a
                                                                     deleted base default

So the survival is the verb's, not the mutant class's.

The cost is on the surface this tier meters. `judge` is 29 lines on the paper, and
`ActionModule.ask_judge` grew from 15 lines to 30 on contracts/harness.py -- about 44 of the 115
lines `raven/contracts` gained in this PR. Two of the five ceiling entries (3,500 -> 3,540 and
3,550 -> 3,570, i.e. 50 of the 130) are spent on this verb and on prose explaining that it is not
wired.

The rule is the budget law's own (tests/test_kernel_budget.py:17-25): "tight means adding anything
to the kernel passes one explicit review. The booklet also names the signal that a threshold was
set wrong: loose, and the assertion never once fails; tight, and someone starts bumping the number
in passing -- the moment that happens the guard has lost its authority." And the precedent this
file sets for what a bump must carry, at :55-58: "no amount of trimming fits it under the old
number."

The alternative I could not refute: keep `DefaultAction.ask_judge` calling the charter's judge
directly, as the base did, and add the verb, `CharterParticipant`, `charter_participants()` and
`compose_judge` when the registry can actually be handed a turn's participants. It changes no
behaviour -- the deletion run above is the proof, since the charter route works with the paper's
verb absent -- costs about 44 paper lines less, and removes the disclosure burden that two of the
five ceiling entries exist to pay.

What would close this, either way:

  - wire it, so the paper describes something a host asks; or
  - defer it as above and re-measure the ceiling without it.

If publishing the verb ahead of its wiring is deliberate -- so plugin authors can write against a
stable paper -- that is a fair answer and I will take it; the tier asks that the necessity be
stated rather than assumed, and the measurement above is what it has to be stated against.

0xKT:
**Not a blocker** -- separate from the review thread on `raven/contracts/participant.py`.

**The kernel's first outward edge is forbidden by a contract the guard cannot see**
(raven/contracts/harness.py:55)

Under `if TYPE_CHECKING:` the paper now reads
`from raven.agent.harness.participants import Intake, Verdict`, used in three return annotations
(`ask_intake` at :199 and :222, `ask_review` at :375).

pyproject's import contracts include one named `the kernel stands alone`, whose `source_modules`
list contains `raven.contracts` and whose `forbidden_modules` list contains `raven.agent`. So the
repo's own machine-readable law forbids exactly this edge. Measured:

    as written, inside TYPE_CHECKING      -> lint-imports: Contracts: 10 kept, 0 broken
    the same import moved out of the block -> lint-imports: the kernel stands alone BROKEN
                                              Contracts: 9 kept, 1 broken

So the law does catch this edge; the `TYPE_CHECKING` block is what it cannot see. At the merge base
`raven/contracts` has zero imports from `raven.agent`; at this head there is exactly one.

The budget docstring discloses it, in the same breath as the reason: "the first the kernel has,
taken knowingly when `Verdict` and `Intake` moved off the paper", and "lint-imports does not see
it, so it is written down here". That disclosure is the right instinct. What is missing is that the
next such edge will also pass, silently, because nothing changed in the guard.

Two ways to close it that I could not refute, both cheaper than they look:

  - annotate what actually crosses the boundary. The two-tier ledger's own comment says the answers
    are "plain mappings" and that "the host's own read of a verdict lives in the harness, not on the
    paper" -- if the paper annotates the mapping rather than the host's read model, the import is
    unnecessary and the edge disappears;
  - or keep the two shapes on the paper. They are plain dataclasses -- `Verdict` is six fields and
    one property, `Intake` three fields and no methods -- so this costs a few dozen lines on a
    surface whose ceiling this PR was already raising.

If neither is wanted, the third option is to make the guard see what was decided here: a check that
reads TYPE_CHECKING imports too, so the second edge is a decision rather than a precedent.

For the record, the parts of this change I checked hardest and found sound:

  - The export delta is net zero: `raven/contracts` goes 98 public names to 98, three out
    (`AgentConduct`, `ConductFactory`, `Verdict`) and three in (`AgentParticipant`, `Answer`,
    `ParticipantFactory`).
  - The rename is complete. Every surviving `conduct` in the tree is a deliberate one: the community
    `CODE_OF_CONDUCT.md`, raven-code's "Coding Conduct" prompts (which CONTEXT.md:195 explicitly
    calls "a different thing with a similar name"), `pristine_conducts()` which belongs to those
    prompts, and the dated historical entries in the budget docstring -- which this PR annotates as
    "a dated record of one bump, not a description of the file today".
  - The five ceiling entries each name what they bought and each record a measurement
    (3,444 / 3,515 / 3,528 / 3,542). I measured the surface independently: 3,427 lines at the merge
    base against a 3,440 ceiling, 3,542 at this head against 3,570.
  - Of the thirteen test files this PR touches, the suite is 363 passed with four failures, and all
    four are in the pre-existing failure set this checkout carries at the merge base.

gloryfromca (commented):

Reviewed the revision delta. `15a39d63` rebases the same seven patch-identical commits onto the docs-site merge at `3dd06f6e`. The only base/PR path overlap is `pyproject.toml`: the base adds docs dependencies and configuration, while this PR retains its previously reviewed import-contract disclosure. The combined configuration is intact; I found no new code, caller, compatibility, test-strength, or architecture failure introduced by the rebase.

Current-head verification: `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed 25/25; `uv run lint-imports` kept all 10 contracts with 0 broken. Four unit shards remain pending in GitHub CI; the other reported checks pass.

The one unresolved thread is @0xKT’s blocker about publishing `judge` before production wiring. It is not my thread or judgement to re-grade, and I have left it open. Covered here: the replacement AGENTS.md rules and canonical context, target diff and history via exact `range-diff`, the `pyproject.toml` overlap, callers, backward compatibility, test preservation, and architecture contracts.

gloryfromca (commented):

Reviewed the revision delta and the prospective merge with current `main`. The seven PR commits remain patch-identical. This head is based on `1d80cd4f`, four commits behind current `main`; `git merge-tree --write-tree HEAD github/main` completes cleanly. The two overlapping paths are semantically disjoint: main’s Conversation Record entry and retry-limit wiring survive alongside this PR’s participant vocabulary changes.

Current-head verification: `git diff --check github/main...HEAD` passed; `git diff --check github/main <prospective-merge-tree>` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed 25/25. All completed GitHub checks pass, with four unit shards pending at review time.

The sole unresolved thread remains @0xKT’s blocker about publishing `judge` before production wiring. It is not my thread to re-grade or resolve. Covered here: the replacement AGENTS.md rules and canonical context, target diff and history via exact `range-diff`, current-main overlaps and prospective merge, callers, backward compatibility, test preservation, and architecture constraints.

gloryfromca (commented):

Reviewed the new revision. `d6063461` is the completed rebase onto current `main` (`32e612d5`): all seven PR commits are patch-identical, and its tree is exactly `159daa023cd5`, the prospective merge tree inspected in the prior round. There is no new integration delta to raise.

Current-head verification: `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed 25/25. All completed GitHub checks pass, with four unit shards pending at review time.

The sole unresolved thread remains @0xKT’s blocker about publishing `judge` before production wiring; it is not my thread to re-grade or resolve. Covered here: the replacement AGENTS.md rules and canonical context, exact patch history, callers, backward compatibility, test preservation, architecture constraints, and equality with the previously reviewed merge tree.

gloryfromca (commented):

Reviewed the revision delta. `b30a3d43` is a rebase onto `f10ac9f1`; all seven change commits are patch-identical, and the sole new base commit changes only documentation-site search and shell files with no path overlap. There is no new code, caller, compatibility, test-strength, or architecture delta to raise.

Verification: `git range-diff` reports seven `=` pairs; `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed (25 tests). All completed GitHub checks pass; four unit shards remain pending. I also checked the repository rules and context surfaces, commit history, callers, backward compatibility, test assertions, and the kernel boundary.

The sole unresolved thread is the existing blocker opened by @0xKT; it is not my thread to re-grade or resolve. This clean verdict is from my side.

raven/contracts/participant.py
0xKT: **Re-verified fixed** on b30a3d43. The ground I blocked on is gone: the verb is asked in
production now, and something real answers it.

What I blocked on was that `judge` was paper nobody asked -- `registry.py:423` passed no
participants, `CharterParticipant` inherited nothing, and the seat never reached the verb. The
chain on this head, every hop read at b30a3d43:

    raven/agent/tools/registry.py:423   role.ask_judge(name, params, prior_calls())
    raven/agent/harness/action.py:37-66 ask_judge -> compose_judge(..., [*participants,
                                                                   *charter_participants()])
    raven/agent/harness/participants.py:174   participant.judge(name, params, prior)
    raven/agent/subagent/charter.py:256-262   CharterParticipant.judge -> charter.judge

and it is live rather than merely wired: replacing the registry's call with `return []` turns three
tests red, among them
`test_a_registry_nobody_handed_a_role_still_refuses_what_the_charter_refuses` -- a refusal reaching
a tool call through a registry nobody handed a role to. 106 passed across test_harness_modules,
test_participant_hook, test_kernel_budget and test_subagent_charter before the mutation.

**What has not changed, and I am not re-filing it.** A seated plugin's `judge` is still not
reached: `ask_judge` takes `participants: Sequence[AgentParticipant] = ()` (action.py:41) and the
registry passes none, so only `charter_participants()` answer. That is the limitation the verb's
own docstring states (participant.py:203-209), with the reason -- a turn-scoped register that would
start refusing calls which run today. A scoped gap disclosed where the next reader meets it is a
different thing from paper nobody asks, and only the second was my blocking ground.

To be clear about what this receipt is and is not: it closes the finding, not the review. This PR
is the contracts tier and 38 files, its head has moved several times since I read it, and it has not had a
full pass at b30a3d43. Closing this unblocks it into the queue, where it gets one -- including the
budget question my finding's second half raised, which is a different argument now that the verb
carries traffic.

gloryfromca (commented):

Reviewed the new revision and the whole record. `8e46ac842` is a no-overlap rebase onto `0eb8b41be`: all seven change commits are patch-identical, while the new base commit is the unrelated configuration fix from #512. @0xKT has independently re-verified the production `judge` chain, withdrawn the ground of the prior blocker, and resolved the thread; GitHub now reports zero unresolved review threads.

Verification: `git range-diff` reports seven `=` pairs; `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed (25 tests). Every completed GitHub check passes, with four unit shards still pending. I covered the repository rules and context, diff and history, callers, backward compatibility, test strength, and architecture boundaries; this revision adds no new concern.

gloryfromca (commented):

Reviewed the revision delta. `1e8fae4b4` rebases the same seven patch-identical commits onto `b4b0604f2`; the sole new base commit is the unrelated #517 documentation showcase, and it shares no changed paths with this PR. There is no new code, caller, compatibility, test-strength, or architecture concern.

Verification: `git range-diff` reports seven `=` pairs; `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed (25 tests). All completed GitHub checks pass, with TUI, UI RPC, and four unit jobs pending. GitHub reports zero unresolved review threads. I covered the repository rules and context, diff and history, callers, backward compatibility, test assertions, and architecture boundaries.

gloryfromca (commented):

Reviewed the revision delta. `9bfbbadbd` rebases the same seven patch-identical commits onto `7ab685f60`. The new base has one shared path, `pyproject.toml`, but the changes are independent: `main` adds `jieba` to the docs dependency group, while this PR adds the import-boundary rationale later in the file. Both survive intact and there is no semantic interaction.

Verification: `git range-diff` reports seven `=` pairs; `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed (25 tests). All completed GitHub checks pass, with four unit jobs pending, and GitHub reports zero unresolved review threads. I covered the updated repository rules and context, diff and history, the overlapping file, callers, backward compatibility, test assertions, and architecture boundaries.

gloryfromca (commented):

Reviewed the revision delta. `85b2aea2f` rebases the same seven patch-identical commits onto `900fa38a3`; the sole new base commit updates an unrelated documentation showcase chart and shares no changed paths with this PR. No code, caller, compatibility, test-strength, or architecture concern is introduced.

Verification: `git range-diff` reports seven `=` pairs; `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed (25 tests). Every completed GitHub check passes, with four unit jobs pending, and GitHub reports zero unresolved review threads. I covered the repository rules and context, diff and history, callers, backward compatibility, test assertions, and architecture boundaries.

gloryfromca (commented):

Reviewed the revision delta. `58a6bc0fc` rebases the same seven patch-identical commits onto `04eb8a31d`; the two new base commits are unrelated documentation presentation changes and share no changed paths with this PR. No code, caller, compatibility, test-strength, or architecture concern is introduced.

Verification: `git range-diff` reports seven `=` pairs; `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_kernel_budget.py -x` passed (25 tests). Every completed GitHub check passes, with four unit jobs pending, and GitHub reports zero unresolved review threads. I covered the repository rules and context, diff and history, callers, backward compatibility, test assertions, and architecture boundaries.

raven/agent/harness/participants.py
0xKT, line 1: **Blocking:** the new archive composition keeps only Mapping observer values, so a bundled,
default-on flow's rendered trail stops reaching the turn's message stamp. The session record still
carries it, which is the equality this repository wrote a test to hold.

**The drop.** `compose_record` (raven/agent/harness/participants.py:220-222):

    for name, counters in stamped.items():
        if isinstance(counters, Mapping):
            filed.setdefault(str(name), {}).update(dict(counters))

Anything that is not a Mapping goes nowhere -- no log, no note. The seat one layer up still carries
the old tolerance (raven/agent/hook/participant.py:426, `observers[name] = dict(counters) if
isinstance(counters, Mapping) else counters`), but it is handed the already-filtered dict, so that
`else` is now unreachable on this path. One rule, two sites, pointing opposite ways.

**Measured, both ends, same probe and same runner.** A seated participant whose `archive` returns
one dict-valued observer and one string-valued one -- the shape the research flow produces -- then
read back from `ctx.metadata["observers"]`:

    BASE 900fa38a   {'process_appendix': {...}, 'research_trail': '## Process\n- searched x...'}
    HEAD 58a6bc0f   {'process_appendix': {...}}

The dict-valued observer arrives in both runs. That is the control: the seam ran, and the only
thing that changed is the non-Mapping value. On the base the string reaches the stamp because
`ConductHook.after_send` took `conduct.archive`'s answer unfiltered and the same `else counters`
branch assigned it; `compose_record` is new in this PR (commit "seat the three verbs the adapter
was applying itself"), and it is what stands between them now.

**It is reachable by default, and the producer is untouched by this PR.** The bundled research flow
writes exactly that shape:

    agents/raven-research/plugins/research-flow/research_flow/flow.py:428
        observers["research_trail"] = appendix
    ...where appendix is build_appendix(...)[0], i.e. trail.render(), a str
        (research_flow/support/process_appendix.py:924)
    research_flow/config.py:246   process_appendix: bool = True

`ResearchFlowParticipant.archive` (flow.py:1011-1014) hands that whole dict to the host, and
`SessionRecord.observers` is `dict[str, Any]` (state.py:63), so the record keeps the string while
the stamp now loses it. `git diff` over flow.py in this PR shows only Conduct->Participant renames;
line 428 is not touched.

**The rule is the repo's own, and it is about this exact pair.**
tests/test_agents_research_process_appendix.py:1051 asserts

    stamped == store.load("s").observers, "the stamp and the record read the same counters"

and the design note at flow.py:419-428 says why the trail is parked in observers at all -- "a
caller reading the session record ... could otherwise not see the trail at all: the turn's ledger
is deleted moments later" -- with flow.py:445-447 naming the stamp as what gives it per-turn
history "beyond this record's latest-turn copy". That existing test stays green only because its
fixture (`FlowConfig.from_slice({"enabled": True})`, no appendix, no ledger) never produces a
string observer. The invariant is written down; nothing exercises it on the value that breaks it.

What would close it, smallest first:

  - keep the seat's tolerance in the composition -- carry a non-Mapping value through instead of
    dropping it, which is what the line at hook/participant.py:426 already says should happen. Then
    the two sites agree again and the `else` stops being dead code;
  - or, if the kernel's answer really must be `Mapping[str, Mapping[str, Any]]`, then the drop is a
    contract change and wants saying out loud: the producer has to be told (it is a bundled flow in
    this repository, not a third party), and the repo's stamp-equals-record invariant has to be
    restated, because it is false as written.

Either way the fixture gap is worth closing in the same move: the invariant test with an appendix
turned on and a ledger written would have caught this, and is the test that pins the outcome rather
than the shape.

For the record, what I checked on this head and found sound: the importlinter contract is NOT
relaxed -- the thirteen lines this PR adds to pyproject.toml are pure comment, no `ignore_imports`
appears, and `exclude_type_checking_imports = true` sits at line 566 in both the base and this head,
so the type-only edge is documented rather than newly permitted, which is the right direction; the
three genuinely new protocol methods (`ask_select_tools`, `ask_archive`, `ask_system_addendum`) each
have one production caller and each reddens a test when that call is neutered; the rename is swept
-- the only surviving `AgentConduct`/`ConductFactory`/`agent_conduct` mentions are three lines of
dated history in the budget docstring, which this PR itself labels "a dated record of one bump, not
a description of the file today" -- and the new term is defined at CONTEXT.md:170 as the repo
requires; and the budget's three coordinates measure 3,427 -> 3,542 against a ceiling moved
3,440 -> 3,570, i.e. headroom 13 -> 28, with the growth being +17 executable and +79 docstring.

ypflll: Fixed in the latest head. `compose_record` now preserves non-Mapping observer values while continuing to merge Mapping values. The participant and Memory contracts now describe `Mapping[str, Any]`, and the Research Flow stamp invariant test enables the process appendix, writes a real ledger, and asserts the rendered `research_trail` survives identically in the message stamp and session record. Local verification: 110 focused tests passed; Ruff, formatting, import contracts, commit checks, source-language checks, and large-file checks passed after rebasing onto the latest `main`.

gloryfromca (commented):

Reviewed the substantive eighth commit, `fix(agent): preserve scalar observer values`, as well as the rebased series. `compose_record` now retains non-mapping observer values and still merges successive mapping values; the widened participant and Memory contracts match every consumer, while the two other bundled participants remain valid narrower implementations. The Research Flow regression enables the real process appendix, writes a ledger, and proves the rendered scalar trail reaches both the message stamp and session record. I found no compatibility or merge-order regression.

Verification: the prior seven commits remain patch-identical under `git range-diff`; `git diff --check github/main...HEAD` passed; `uv run pytest tests/test_participant_hook.py tests/test_agents_research_process_appendix.py tests/test_kernel_budget.py -x` passed (110 tests); Ruff and format checks passed on all six files in the fix. Every completed GitHub check passes, with four unit jobs pending, and GitHub reports zero unresolved review threads. I covered the repository rules and context, diff and callers, history, backward compatibility, test strength, and architecture boundaries.

LivXue (approved):

Issue #465, opened 2026-09-17
docs: host Raven-Oncall benchmark charts


LivXue:
## AI4AI (Nanochat 50M Pretraining)

![AI4AI (Nanochat 50M Pretraining): Bits Per Byte (BPB), Runtime, Tokens, and Cost](https://github.com/user-attachments/assets/1a1b161e-946b-4049-ad4b-8ed8f22ac658)

## AI4S Internal Benchmark

![AI4S Internal Benchmark: Success Rate, Average Total Runtime, Average Total Tokens, and Average Cost](https://github.com/user-attachments/assets/358fc781-7524-4c1e-997b-fc31943d29d7)

LivXue:
README benchmark figures optimized for full-width, single-column display. All 89 plotted values are unchanged.

![Coding Benchmarks](https://github.com/user-attachments/assets/a516ebc2-f0b4-47d4-b970-7dfd492adc0e)

![DataAgentBench (2026-08-24 Live)](https://github.com/user-attachments/assets/f4ee009a-d24f-4f11-85db-21e358b29104)

![PresentBench](https://github.com/user-attachments/assets/1c8a4339-77f3-4579-85cc-359bc171739d)

![Visual Design](https://github.com/user-attachments/assets/061b5818-595b-4392-b9b6-53a1effb8347)

![AI4AI (Nanochat 50M Pretraining)](https://github.com/user-attachments/assets/8e2b9721-421c-485c-b1dc-1acf901bcb2c)

![AI4S Internal Benchmark](https://github.com/user-attachments/assets/aafe30ca-6693-4d30-956e-73debf3ef2c9)

LivXue:
DataAgentBench and PresentBench re-rendered as vertical columns to reduce figure height in the README. All plotted values are unchanged from the previous horizontal versions.

DataAgentBench: 1320 -> 876 px tall (-33.6%). PresentBench: 1420 -> 944 px tall (-33.5%).

![DataAgentBench (2026-08-24 Live): Raven-Code with Opus-5 achieves 0.8762 Pass@1](https://github.com/user-attachments/assets/75f60aa2-3e2f-42fb-aa51-aa111bf078d0)

![PresentBench: Raven-Design, Claude Code, and public leaderboard scores](https://github.com/user-attachments/assets/c42ff722-8aab-40ec-82e3-0655a859a9d9)

LivXue:
[...]

PR #498, opened 2026-09-18
docs: refresh AI4AI benchmark chart

LivXue:
## Summary

Update the hosted AI4AI (Nanochat 50M Pretraining) chart reference in both bilingual READMEs.

- Replace the old image attachment URL in README.md and README.zh-CN.md.
- Use the updated chart with the Bits Per Byte (BPB) label and the Lower Is Better annotation above the BPB values.
- Keep the chart hosted as a GitHub attachment; no image binary is added to the repository.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Commands:

- `git diff --check`
- `UV_NO_SYNC=1 make check-large-files check-source-language PYTHON_VERSION=.venv/bin/python COMMIT_RANGE=HEAD`
- `UV_NO_SYNC=1 uv run --no-sync pytest tests/test_readme_scope_canon.py tests/test_cli_onboard_commands.py -k readme_quickstart -q`
- Downloaded attachment SHA-256 matches the local PNG.

## Risk

This is a documentation-only URL update. Reverting the commit restores the previous chart.

## Related Issues

#465


ZuyiZhou (approved):
Approving. Two lines, one asset id, the same swap in both files.

I fetched both PNGs and compared them: every number is identical -- BPB 1.053 / 1.038 / 1.037, runtime 4.99 / 4.90 / 3.74, tokens 13.35 / 9.38 / 3.79, cost $50.00 / $39.40 / $0.97, the same three series and the same 2x A800 note. So this is a cosmetic refresh, not a new measurement, and nothing in the surrounding prose needed to move with it. The BPB panel now carries its unit in the heading (`Bits Per Byte (BPB)`) instead of on a line below, which is what the alt text already said, and it gains a "lower is better" arrow.

One nit, not worth a round trip: "Lower Is Better" now appears twice, top right for the figure and again inside the BPB panel. It is true of all four panels, so the per-panel one reads as if it were not.

The red check is not this PR. `unit (py3.12 / 1/4)` reports `5489 passed, 9 skipped`; what failed the job is `--idle-ceiling-strict` tripping on `tests/test_ppt_engine_assets_script_helpers.py::test_a_formula_that_stacks_is_typeset_as_a_picture_in_the_decks_ink` at 5.76s idle -- a runner-speed flake in the deck engine that a two-line README edit cannot reach. main's last two runs are green. Worth a re-run before merging rather than a change here.

PR #501, opened 2026-09-18
feat(ui-web): take the page to the new design, and retire three of its pages

xfng-sd:
## Summary

The page the design draws, and the product it draws for. Everything here was measured off the rendered prototype (`ui-web/docs/design.html`) rather than read out of its stylesheet -- that file carries duplicated rule blocks where the later one wins, so reading it gives the wrong values. Both themes are its own palette now, dark included.

**The rail.** White ground, a hairline seam, one flat step of colour for hover and for selected, and the wordmark back at the head of the row. Three destinations: a new task, a playbook, and the agents that run one. Schedules, channels and memory are set up once and then left alone, so they moved behind the fold that already held two of them. The session list drops its counts and its hairlines -- the rows under a heading are the count, and a rule to the edge drew a box around a list already bounded by whitespace. The one heading with a verb keeps it out of sight until the pointer is on the row.

**The conversation.** One turn, two shapes: a bubble that ends where the sentence ends, and a card of constant width, because a column of cards whose widths follow their word counts reads as a ragged edge. The card holds the whole turn -- the fold, its steps and the answer -- so the process is part of the answer rather than loose text above it. The ground is `--wash` rather than flat paper, the composer is a solid pane, and the landing page is that same page with nothing in it: the composer does not move or resize when the first message lands, which it used to do every time.

**The task pane.** Lifted rather than drawn, and a step on the board is a white card on the dotted ground instead of a tinted cell cut out of it. The flow tab ends in a one-line field addressed to that step's sub-agent by name -- static, since no `tasks.*` method answers it yet. A step nobody started and one that was skipped never got an instance, and neither draws a field.

**Three pages leave the product.** Skills and plugins were two tabs of one `capsPage`, so both going takes the page with them, plus `state/caps.ts` and `features/hosts.ts` -- which existed only because those tabs cleared their box with `innerHTML` and needed nodes React did not own. Knowledge goes with them. Two things survived and were repaired rather than deleted: `features/installed` is the one `ext.list` read that filled three pages, so it keeps the tools the settings dialog draws and drops the rows nothing renders; and the websearch notice keeps its words and loses the button that opened the plugins page. The bundle drops from 966 kB to 880 kB.

Every ratchet in `scripts/gates` shrinks with the tree rather than being widened for it. Several per-domain class rows RISE without a class being added anywhere: the deleted domains were readers of the page's shared vocabulary, so a class they shared with one other domain is that domain's alone now. The page's total debt is unchanged; where it is counted is not, and the comment on that table says so.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Rebased onto `8dd9d24a` (#491) and re-run from there.

- `npm test --prefix ui-web` -- 2358 passed, 0 failed (157 files). Three `KnowledgePage` failures that had been failing before this branch went with the domain.
- `npm run lint --prefix ui-web` -- 0 errors, 5 warnings, all pre-existing `react-hooks/exhaustive-deps`.
- `npm run type-check` for both `ui-web` and `ui-tui` -- clean.
- `npm run lint:i18n --prefix ui-tui` -- generated catalogue up to date.
- `python ui-web/build.py` -- production build and both boot snapshots pass against regenerated goldens.
- `python scripts/check_source_language.py` and `check_large_files.py` over the branch range -- both exit 0.
- Driven in a browser against the offline fixtures, light and dark, at 1440x900: the rail, a conversation, the landing page, and a task pane's node detail. Measured `getComputedStyle` against the prototype element by element rather than comparing screenshots.

One thing worth a reviewer's eye: #491's `first-frame-literals` gate could not run on Windows at all. It takes its root from a file URL's `pathname`, so the walk lands on `D:\D:\...` and dies with ENOENT, and the relative paths it compares against its own pinned list come back with the platform's separator. That is the pair of bugs #488 fixed in seven gates; this one is new since, and it now reads them through the module those seven share.

## Risk

User-visible throughout: this is what the page looks like. The removals are the part to weigh -- skills, plugins and knowledge are gone from the product, not hidden, and their seams and i18n keys go with them. The gateway still answers those methods; nothing in this tree calls them.

Behaviour changes beyond appearance are small and named: the rail's collapse toggle moves to the far end of its row, so collapsing the rail moves the button a rail's width (the prototype never implements a collapsed rail, and this trade was made deliberately); the websearch notice loses its action; and the landing page no longer floats the composer.

Rollback is the branch. Nothing outside `ui-web/`, `i18n/messages.json` and the generated TUI catalogue is touched.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the full target diff and the surrounding page assembly, navigation, transcript, task, installed-extension, language, and lifecycle callers. I also checked the commit history and PR rationale, the Web UI terms and architecture constraints, backward compatibility for the intentionally retired pages, and the adjusted tests/gates for weakening rather than merely accepting their green result.

I specifically tried to refute the task-pane send affordance as a defect: the PR declares it static, the live tasks source currently returns no rows, and it is therefore limited to the offline design fixture rather than ordinary product use. I found no other concrete failure introduced by this revision.

Verification:
- `npm test` in `ui-web`: 157 files, 2,358 tests passed
- `npm run type-check` in `ui-web`: passed
- `npm run lint` in `ui-web`: passed with five pre-existing hook warnings and no errors
- `npm run build` in `ui-web`: passed
- `npm run gen:check --prefix ui-web`: passed
- `npm run type-check --prefix ui-tui`: passed
- `npm run lint:i18n --prefix ui-tui`: passed
- `uv run python ui-web/build.py`: production build passed; both boot snapshots matched
- source-language and large-file checks over `github/refactor/ui_web_architecture...HEAD`: passed

gloryfromca (commented):

I reviewed the new revision against the full target diff and the previously reviewed head. The only delta is removal of two trailing spaces in `ui-web/scripts/check-class-namespace.mjs`; `git diff -w 5259bb273e1e..HEAD` is empty. That leaves the previously checked rules, callers, history, backward compatibility, architecture constraints, and test coverage unchanged, while fixing `git diff --check`.

Verification on this head:
- `npm test` in `ui-web`: 157 files, 2,358 tests passed
- affected class-namespace gate: 1 test passed

ui-web/src/features/transcript/TranscriptPage.tsx
```diff
@@ -1029,7 +1029,7 @@ const AskView = memo(function AskView({ lane, seg }: { lane: Lane; seg: AskData
   }
   const showClip = seg.clipped && !seg.clipOpen
   return (
-    <div className="ask in">
+    <div className="turn me in">
```
gloryfromca, line 1032: No blockers; suggestions only, and they are marked inline.

The class rename leaves the streaming identity test at `TranscriptPage.test.tsx:2088-2108` querying `.ask`. There is no longer any element with that class, so both `askEl` and the later lookup are `null`, and `expect($('.ask')).toBe(askEl)` passes without checking the invariant named by the test (that streaming does not remount the user's turn). Please update those selectors to `.turn.me` so this regression check remains effective.

gloryfromca (commented):

I reviewed the full target diff plus the rebase delta from the previously reviewed head. I covered the repository rules and UI architecture context, changed callers and lifecycle wiring, commit history, compatibility of the intentionally retired pages, and whether the adjusted tests still exercise their stated invariants.

The rebase-specific integration is sound: the target branch's live task chat replaces the branch's old static field, settings retains the raw skill/MCP inventory needed after the old islands are removed, plugin credential metadata now has a settings-owned reader, and MCP status pushes still refresh an open settings dialog. I found one nonblocking vacuous test assertion inline; no product defect is holding this revision up.

Verification:
- `npm test -- --run` in `ui-web`: 181 files, 2,513 tests passed
- `npm run lint`: passed with five pre-existing warnings in untouched files and no errors
- `npm run gen:check`: passed
- `npm run build`: passed
- assembled-page build, boot snapshots, page/CSS/class checks: passed
- `npm run lint:i18n` in `ui-tui`: passed
- source-language and large-file checks over the PR range: passed
- `git diff --check`: passed

0xKT (approved):

PR #502, opened 2026-09-18
fix(agent): keep one cause one streak class in the media envelopes

ZuyiZhou:
## Summary

`failure_class` keys the loop's tool-failure streak on an envelope's `error`
string alone. That works only while the string comes from the small fixed
vocabulary its producers are meant to keep, and it is a convention the module
cannot enforce - the docstring added in #410 names exactly this failure mode:
one cause counting as many.

Every media tool broke it. `_format_http_error` interpolated up to 400
characters of vendor body (which carries a request id) into `error`; the
unreadable-reference handlers interpolated the path; the four catch-alls put
`str(e)` there, which spells the host. So a dead endpoint produced a class per
call, `(tool, failure_class)` never repeated, the streak never reached
`_LOOP_BREAK_THRESHOLD`, and the stop-repeating nudge was unreachable for
`image_generate`, `text_to_speech` and `video_generate` alike.

The URL gate in front of `web_fetch` had it too, in both copies of the tool:
most reasons `validate_url_target` composes name the address they refused, so
a model walking an internal range got one class per address. #410 fixed the
handlers behind that gate and left the gate itself.

`error` now carries only what a reader could enumerate. The variable part moves
to `detail`, which keeps it in front of the model and inside
`is_hard_tool_failure`'s transient-marker scan. Where the split gave the single
call a further key, the batch reducer forwards that too: `hint`, which names the
proxy a 403 can be routed through, and `url`, which names the reference a
guarded fetch refused. A batched picture is told what the single call is told,
which is the property the split has to preserve. Distinct causes stay distinct,
which is the failure mode this trades against: two HTTP statuses, two exception
types and the four vendor refusals each keep their own class, and the tests
assert that direction too.

Found by sweeping for the shape rather than by reading the code: every
`"error":` whose value is not a literal is a candidate, and that grep is the
cheapest guard against this recurring.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/ -q -p no:randomly` -> 23443 passed, 109 skipped
- `make lint-python` -> ruff check and format clean
- `make lint-types` -> all checks passed
- `make lint-imports` -> 10 contracts kept, 0 broken
- `make check-source-language`, `make check-large-files` -> clean
- Revert-to-red, per fix, restoring the source from a backup afterwards:
  reverting `media_gen.py` alone turns the four new class tests red and leaves
  `test_two_different_statuses_stay_two_classes` green; reverting either copy
  of `web.py` turns only that copy's gate test red and leaves the two #410
  streak tests green. For the batch fold, each key is pinned on its own:
  reverting it to `detail` alone turns both new batch tests red, restoring
  `hint` alone leaves the reference test red, restoring `url` alone leaves the
  403 test red.
- After rebase onto 45203a4a6: `uv run pytest tests/test_media_gen_tool.py
  tests/test_security_web_ssrf.py -q -p no:randomly` -> 100 passed;
  `make lint-python`, `make lint-types`, `make check-source-language`,
  `make check-large-files` -> clean.

## Risk

Tool output shape changes: a caller reading the reason out of `error` now
finds it in `detail`. The readers in this repo are the model itself and the
tests updated here; `tests/test_security_web_ssrf.py` is the one behavioural
assertion that moved, and it still pins that a private address is refused for
being private. Rollback is a plain revert of this commit - no state, no
migration, no config.

## Related Issues

N/A. Finishes the envelope work started in #410, which fixed the reader's
handlers and left the gate in front of them and every media tool.


raven/agent/tools/media_gen.py
```diff
@@ -598,7 +604,9 @@ def _batch_answer(self, wanted: list[dict[str, Any]], answers: list[str | BaseEx
                 paths.extend(reply["paths"])
             else:
                 item["error"] = reply.get("error") or "no image returned"
-                for key in ("note", "finish_reason", "retryable"):
+                # ``detail`` travels with ``error``: the reason a picture failed is
+                # there now rather than in ``error``, which carries only the class.
+                for key in ("detail", "note", "finish_reason", "retryable"):
```
gloryfromca: Blocking: batched HTTP 403 failures lose their proxy recovery hint.

`_format_http_error` now moves the 403 recovery advice from `error` into `hint`, but this batch reducer forwards only `detail`, `note`, `finish_reason`, and `retryable`. A normal `image_generate(prompts=[...])` call whose items receive 403 therefore returns only `HTTP 403` plus the vendor body, while the equivalent single call still tells the model to configure `tools.media.proxy` or `HTTPS_PROXY`. I reproduced this directly through `_batch_answer`; neither result retained `hint`. Please preserve `hint` in batch results and cover the batched 403 path.

gloryfromca (commented):
Blocking: batched media HTTP 403 failures lose their proxy recovery hint.

I found one inline blocker. The stable error/detail split otherwise preserves the intended streak behavior and keeps diagnostics model-visible.

Covered: the full diff; surrounding envelope consumers and the agent-loop caller; relevant history; backward compatibility of the JSON envelopes; whether tests were weakened; and the AGENTS.md, CLAUDE.md, CONTEXT-MAP.md, and Runtime architecture constraints.

Verification: the affected suite passed with 307 tests and 17 warnings. The direct batch-reducer reproduction confirmed the missing hint. The source-language gate passed. The full suite did not pass in this checkout: 18,312 passed, 126 skipped, 67 failed, and 235 errored; the output includes missing optional/dev components such as raven_ppt, raven_design, raven_everos, and websockets.

gloryfromca (commented):
Blocking: the batch reducer still drops the HTTP 403 recovery hint.

The new revision adds useful coverage for the images, speech, and video routes, but it does not change `raven/agent/tools/media_gen.py:609`: batch results still omit the `hint` field introduced by `_format_http_error`. The existing inline finding remains current, so I have not duplicated it or resolved its thread.

I reviewed the new delta and rechecked the full branch diff, affected callers, relevant history, backward compatibility, test strength, and the repository and Runtime architecture constraints. I found no additional issues.

Verification: `uv run pytest tests/test_media_gen_tool.py -q` passed with 95 tests. `uv run python scripts/check_source_language.py github/main...HEAD` passed. Current GitHub checks are otherwise green or still in progress.

raven/agent/tools/media_gen.py
ZuyiZhou: Fixed in b24eb7570. The reducer now forwards `hint`, so a batched 403 names `tools.media.proxy` exactly as the single call does, and `test_a_batched_403_names_the_proxy_the_single_call_names` covers that path.

The same hunk dropped a second key this PR created: `url`, where the guarded-fetch refusal names the address, since `error` no longer interpolates it. Batched, that reached the model as a refusal with no address at all. Both keys travel together now, and each has its own test: reverting the fold to `detail` alone turns both red, restoring `hint` alone leaves the reference test red, restoring `url` alone leaves the 403 test red.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at b24eb7570a76: the batch fold now preserves both `hint` and `url`, and the new tests exercise the real batched 403 and guarded-fetch refusal paths. The affected suite passes, so this settles my finding.

gloryfromca (commented):

The revised batch fold preserves every field split out of `error` by this change: `detail`, the 403 recovery `hint`, and the refused-reference `url`. The new tests cover the real batched call paths, and my prior blocker is fixed and resolved. I found no new issues.

Covered: the current full diff and revision delta; media-envelope producers, the batch consumer, and failure-streak caller; relevant history; backward compatibility; test strength; and the AGENTS.md, CLAUDE.md, CONTEXT-MAP.md, and Runtime architecture constraints.

Verification: `uv run pytest tests/test_agent_loop_tool_loop_break.py tests/test_agent_tools_web_providers.py tests/test_agents_research_tools.py tests/test_media_gen_tool.py tests/test_security_web_ssrf.py -q` passed with 312 tests and 17 warnings. `uv run python scripts/check_source_language.py github/main...HEAD` passed. GitHub checks are green or still in progress.

0xKT:
**Not a blocker** -- the model-facing envelope is unaffected. This is about the row that lands on disk.

`_fetch_record` builds the fetch ledger row by copying a fixed key list off the tool's own
envelope (`agents/raven-research/plugins/research-flow/research_flow/tools/web.py:1745-1757`).
That list carries `("error", "error")` at 1756 and no `detail`. This PR moves the refusal
reason out of `error` and into `detail` at web.py:1828, so the reason no longer reaches the row.

Measured at the seam, handing the real `_fetch_record` one envelope each:

    HEAD envelope: {"error": "URL validation failed",
                    "detail": "blocked: private/internal address 127.0.0.1", "url": ...}
      row -> error='URL validation failed', no detail key, nothing in the row names the address

    BASE envelope: {"error": "URL validation failed: blocked: private/internal address 127.0.0.1",
                    "url": ...}
      row -> error='URL validation failed: blocked: private/internal address 127.0.0.1'

Same function, same call; only the envelope differs.

The row is written on the production path, not a window built for the probe:
`_ledger_append(self._fetch_record(url, out))` at web.py:1698-1699.

Why this is more than cosmetic: `_error_shaping` in this same file (web.py:374-381) records what
the ledger's error column is for -- "transport_err: true alone made a drained API key and a
network blip the same row - during the one mid-run outage this project has already eaten, the
ledger could not say which it was. The error string and status make the row diagnosable." After
this change every refused target writes the same undiagnosable row.

Scope, in fairness to the change: the sibling handlers at 1852 and 1865 split the same way before
this PR, so the copy list was already behind them; this PR widens the gap to the validation gate.
And the split itself is right -- the reason names the host, which is exactly the "one cause, many
classes" shape this PR exists to remove. What is missing is the reader update, which this PR did
perform for the media batch fold (`raven/agent/tools/media_gen.py:604-613`) and not here.

One entry in that tuple closes it: `("detail", "detail")`.

gloryfromca:

Confirmed at the production seam: `_fetch_record` keeps `error` but drops the new `detail`, so refused targets now persist as identical `URL validation failed` rows with no address-specific reason. This does not affect the model-facing envelope or mergeability, but it makes my earlier clean stance nonblocking. I am not duplicating the verifier note or its one-entry fix.

Verification: the direct `_fetch_record` reproduction omitted `detail`, and `uv run pytest tests/test_agents_research_tools.py tests/test_agents_research_verbatim_sink.py -q` passed with 98 tests.

0xKT (approved):

PR #486, opened 2026-09-18
docs: sync bilingual README overview

LivXue:
## Summary

Synchronize the Chinese README with the updated English overview. The change aligns the Raven introduction and built-in agent section, including the Host Agent positioning, Raven Evolver description, built-in agent wording, and orchestration readiness statement.

The existing benchmark charts, links, captions, and the remaining bilingual sections are preserved.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- [x] Relevant tests pass locally
- [x] Relevant lint / type checks pass locally
- [x] User-facing docs or screenshots are updated when needed

Commands run:

- `uv run --no-sync pytest tests/test_readme_scope_canon.py tests/test_cli_onboard_commands.py -k readme_quickstart -q` (1 passed)
- `uv run --no-sync pytest tests/test_docker_runtime.py -k container_exposes_no_engine_selector -q` (1 passed)
- `UV_NO_SYNC=1 make check-large-files check-source-language PYTHON_VERSION=.venv/bin/python COMMIT_RANGE=HEAD` (passed)
- `git diff --check` (passed)

## Risk

- [x] Security impact considered
- [x] Backward compatibility considered
- [x] Rollback path is clear for risky changes

This is a documentation-only change. Reverting the commit restores the previous README wording.

## Related Issues

N/A

PR #503, opened 2026-09-18
docs: state the evolver relationship the architecture section already states

ZuyiZhou:
## Summary

The overview says the four built-in agents are "Powered by the **Raven
Evolver** engine". The architecture section of the same file says `evolver/`
is a separate tool that consumes Raven as a library and that the runtime does
not import Evolver, and `pyproject.toml` enforces that with an import-linter
contract literally named "the runtime does not import the evolver". A reader
gets two incompatible answers to whether Evolver is a runtime engine or an
external development tool, and the contract settles which one is true.

This was gloryfromca's blocking review finding on #486. The thread was
resolved when that PR merged, but the text was never changed, so the
contradiction is live on `main` today in both language mirrors.

Both mirrors now put Evolver where it actually runs: refining the shared
harness from outside, against benchmarks, developing the agents rather than
running inside them. That is the reading the review offered for the case where
the intended claim was that Evolver was used to develop the agents. The
overview also takes the canonical spelling of harness self-evolution from the
architecture section, as the review asked.

Scope note: only the Evolver relationship changes. The SOTA sentence in the
same paragraph is untouched - it was not part of the finding, and what it
should say is a product call rather than a consistency fix.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Prose only: two lines, one per language mirror, no code and no new files.

- `git diff --stat` -> `README.md | 2 +-`, `README.zh-CN.md | 2 +-`
- The two gates that apply here are satisfied by construction rather than by a
  command: `check-source-language` exempts `*.md` by suffix (AGENTS.md 1.3),
  and `check-large-files` sees no added or resized file.
- Read back against the two places the claim has to agree with: the Evolver
  row of the agent table, and the harness self-evolution bullet in the
  architecture section. Both mirrors now match both.

## Risk

Documentation wording only. No runtime behaviour, no interface, no build.
Rollback is a plain revert. The claim removed was the one the import-linter
contract forbids being true, so nothing downstream depends on it.

## Related Issues

N/A. Lands the finding from #486 that was resolved without a change.


gloryfromca (commented):

The revised English and Chinese overview text now agrees with the repository layout, the Architecture section, the canonical Layer Seats definition, and the enforced dependency direction: Evolver is an external development tool that consumes Raven and evaluates harness candidates, not an engine imported by the agents at runtime. The bilingual change preserves the existing scope and meaning.

I covered the complete diff, surrounding README architecture text, relevant history, AGENTS.md and CONTEXT terminology, import-linter and pytest architecture contracts, backward compatibility, and whether tests were weakened. There are no changed runtime callers and no test changes; this is documentation-only. Branch, commit, and PR-title naming are compliant.

Verification: `git diff --check` passed; `scripts/check_large_files.py github/main..HEAD` and `scripts/check_source_language.py github/main..HEAD` passed through `uv`; `uv run pytest tests/test_evolver_outside_the_runtime.py tests/test_readme_scope_canon.py -q` passed with 5 tests; the commit-message and PR-title checks passed. The `make` wrapper itself was unavailable in this environment, so I ran its exact underlying gate commands directly.

0xKT (approved):

PR #504, opened 2026-09-18
chore: add codeowners entry for raven/agent

LivXue:
## Summary

Add `.github/CODEOWNERS` with a single entry mapping `/raven/agent/` to `@LivXue`, so a
pull request that changes a file under that directory automatically gets a review request
from the directory's owner.

This only requests a review. The branch ruleset keeps `require_code_owner_review` off, so
no approval requirement changes and no contributor's merge path is affected. Making it a
merge gate would be a separate ruleset change, deliberately not made here.

Two limits are inherent to the mechanism rather than gaps to close later:

- GitHub never requests a review from the author of a pull request, so one opened by the
  owner does not assign them. The API answers that case with HTTP 422.
- A draft pull request gets no automatic request. It is sent when the draft is marked
  ready for review.

CODEOWNERS is read from the base branch, so this governs pull requests opened after it
lands and does not apply retroactively. Ten of the sixteen pull requests open at the time
of writing touch `raven/agent/`. Those were assigned manually, apart from the one the
owner authored, which cannot take the assignment for the reason given above.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other

## Verification

Gates run in a clean worktree branched from `origin/main`, each exit code read rather than
inferred from empty output:

```
make check-large-files      exit 0
make check-source-language  exit 0
make check-commits          exit 0
```

Behaviour checked against GitHub rather than by reading the file. The CODEOWNERS validator
on the pushed branch returns no errors, which is what proves both that the pattern parses
and that the named owner resolves to an account holding write access:

```
gh api "repos/EverMind-AI/Raven/codeowners/errors?ref=..."
{"errors":[]}
```

Not run, and why: `make lint` and `make test-python` analyse Python, TypeScript and
JavaScript, and this diff adds one ten-line text file containing no code, so neither
measures anything about the change. No test in the repository exercises CODEOWNERS
resolution, which is why the validator above is cited instead of a test. Separately
confirmed that nothing reads the directory this file lands in as a set: of the three tests
mentioning `.github/`, one validates backticked documentation paths carrying a known file
extension, one names a single workflow file, and one globs `.github/workflows` only.

## Risk

Nothing changes for any contributor: an added review request does not gate a merge, and
CODEOWNERS confers no permission of its own. The owner handle named here is already public
on every commit and pull request in this repository, so the file discloses nothing new.

Rollback is deleting the file, or the one entry in it. Nothing depends on it.

## Related Issues

N/A


gloryfromca (commented):
Blocking: The commit's co-author trailer must identify the actual Claude model instead of the malformed `claude-opus-5[1m]` value.

AGENTS.md section 3.3 requires the actual current-session model ID in `Co-authored-by`, but the only commit literally contains `Claude (claude-opus-5[1m])`. The repository commit gates pass because they do not validate this trailer field; if merged as-is, GitHub's squash co-author collection would preserve malformed contribution attribution.

The CODEOWNERS change itself is sound: I reviewed `git diff github/main...HEAD`, the surrounding `.github` layout, the target directory, commit and file history, repository rules, PR description, and GitHub-side owner resolution. The new file is in a recognized location, `/raven/agent/` covers the intended subtree, GitHub reports no CODEOWNERS errors for this head, and `@LivXue` has write/admin permission. There are no callers or runtime compatibility concerns, no tests were changed or weakened, and no architecture or domain-term constraints are implicated.

Verification: GitHub CODEOWNERS validation returned `errors: []`. The equivalent commands behind `make check-commits`, `make check-large-files`, and `make check-source-language` all passed against `github/main..HEAD` (`make` itself is unavailable in this environment). No runtime test suite is relevant to this metadata-only change.

userName20260323 (approved):

PR #475, opened 2026-09-17
refactor(ui-web): rebuild the frontend on a typed transport, stores and a react tree

0xKT:
## Summary

Rebuilds the `ui-web/` frontend's architecture without changing the page: the served page, its stylesheet and its behaviour are the same, the code underneath is not. The wire protocol and the Python backend are untouched.

What changed, in three stages plus a layout pass (37 commits, one per step, each landed with every gate green, plus the layout pass squashed in from #478):

- **A. Module graph and types.** The three concatenated script layers (`seam` / `demo` / `live`, ~7,000 lines reached through `window.DS`, `window.RavenShell`, `window.RavenIslands`) became ES modules with an explicit install order; a typed `WsTransport` implements the `RpcTransport` interface and carries the page's reconnect behaviour; the data-source seam and the island registry are imports instead of window objects; a Vite dev server with hot reload proxies to the local gateway.
- **B. Session runtime and sources.** Every RPC domain has a typed `features/<domain>/source.ts`; turn state lives in one `SessionRuntime` per conversation with an ordered pipeline of stages; the offline `?stub=1` page runs on wire-shaped fixtures through a fixture transport instead of a second source layer; the live layer's boot, connection and update-notice code moved into `state/`.
- **C. Chrome components.** `page.html` shrank from 524 lines to a 59-line shell (`<head>`, splash, no-JS notice, two script blocks); every region is rendered by `App.tsx` from a detached React root that portals into `document.body`; page, language, detail drawer, confirm and settings dialogs, menu, toasts, rail, workspace pane, dock, capabilities page, sheet rack, drafts, tooltip, overlays and the Escape order are stores plus components under `src/state/` and `src/chrome/`; document-level listeners are registered from one place in a pinned order; language is a subscription (no more `redrawAll`, no more `querySelectorAll` passes). `src/legacy/` and the RavenShell bridge are gone.

Gates added to prove the page did not move: a sha256 pin on `page.css`, two booted-DOM goldens (stub and live-without-server), 19 per-region DOM goldens, 31 island DOM snapshots, the body-level portal order, the 14-item Escape priority table (91 pairwise cases), the boot listener order, fixture shape checks against the OpenRPC contract, and per-step before/after dumps of the whole body, the boot walk and the language flip (zh / en / unset), all byte-identical.

[...]

PR #506, opened 2026-09-18
fix(*): let the web settings timezone control write, and drop the retired one

LivXue:
## Summary

The timezone box on the web settings page never wrote anything. `settings.set`
receives its cron keys under the spelling the config file uses
(`cron.defaultTimezone`) and hands that straight to `update_cron_config`, which
validates against `CronConfig.model_fields` and so expects the Python field name
(`default_timezone`). Nothing converted between the two, so the writer raised
`KeyError`, the dispatcher's catch-all turned it into `-32603 internal_error`,
and the value never reached disk however many times the box was used.

The fix converts at the RPC boundary with `to_snake`, which that module already
imports. Deriving the field name mechanically instead of hardcoding one string
means the next cron key added to the branch cannot reintroduce the mismatch, and
`update_cron_config` keeps the contract it shares with the CLI.

`cron.forwardChannels` sat in the same whitelist and could not have succeeded
under any spelling: the field left `CronConfig` when delivery became
fire-at-origin, and the loader strips both of its spellings out of a config file
on the way in. The settings page still rendered a row for it. That row, the
whitelist entry, and the two catalogue strings nothing else read are removed.
The key now returns a typed not-writable refusal rather than an internal error,
which is what the CLI has always done for the same input, so the two surfaces
agree.

- `update_cron_config` always writes the camelCase spelling rather than
  following the one a config file already uses. Measured: `CronConfig` inherits
  a base with `extra="ignore"` and the camelCase alias wins regardless of key
  order, so the save takes effect and only a dead key is left behind. This is
  pre-existing behaviour shared with the CLI writer and is not what made the
  control ineffective.
- The settings page reads `cron.defaultTimezone` directly, so a hand-written
  snake_case config shows the schema default until the first save. That display
  gap predates this change, and closing it would touch how the page reads every
  key.
- No defensive `KeyError` to `ConfigValidationError` conversion was added in
  `settings.set`. The whitelist is now exactly the set of writable keys, which
  closes that path structurally.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Written test-first: the new cases were red before the change and green after.
`pytest tests/test_rpc_settings.py -k "cron or forward_channels or timezone"`
reported `3 failed, 2 passed` with
`KeyError: "Unknown cron config key: 'defaultTimezone'"` beforehand and
`5 passed` after. The new ui-web case was `1 failed | 123 passed` before the row
was removed and `124 passed` after.

At the pushed head:

- `pytest tests/test_rpc_settings.py tests/test_config_update.py tests/test_rpc_console.py tests/test_i18n_boundary.py tests/test_cli_cron_commands.py tests/test_config_loader.py`
  -> 331 passed, 1 skipped. The skip is `chmod 0o500 does not block root`, an
  artifact of running the suite as root, unrelated to this change.
- `npx vitest run --no-file-parallelism src/features/settings/SettingsPage.test.tsx`
  -> 124 passed
- the whole `ui-web` suite -> 1835 passed, 2 failed, 0 unhandled errors. The two
  are `deskStore.test.ts` and `palette.test.ts` "when storage refuses", which
  are unevaluable on this machine: node's experimental `localStorage` shadows
  the one the test environment installs, so storage works where those two need
  it to throw. They fail identically without this change.
- `npx vitest run src/__tests__/i18n.test.ts` in `ui-tui` -> 8 passed
- `make lint-python` -> ruff check and ruff format both clean
- `make lint-ui` -> `gen:check` matches the contract, `tsc --noEmit` clean
- `make lint-tui` -> lint, `lint:rpc`, `lint:i18n`, `tsc --noEmit` all clean
- the page gate as CI runs it, minus `npm ci`: `npm run build`, `build.py`,
  `check-page.mjs`, `count-shared-globals.mjs`, `check-css.mjs`,
  `check-class-namespace.mjs` -> all OK
- `check_commit_messages.py`, `check_source_language.py` and
  `check_large_files.py` over the branch range -> all exit 0, against a range
  confirmed non-empty

End to end, `settings.set` with `cron.defaultTimezone` now returns
`applied: true`, leaves `{"cron": {"defaultTimezone": "Asia/Tokyo"}}` on disk,
and `load_config(path).cron.default_timezone` reads it back.

No user-facing document described either control, so none needed updating. The
retirement of fire-time channel selection is already recorded in `CONTEXT.md`
under the Fire-at-origin entry, whose avoid list this change enforces rather
than contradicts.

## Risk

Two user-visible changes, both on the settings page. The timezone box starts
persisting what it is given, where before it failed with an internal error and
wrote nothing. What that value then reaches is narrower than the field's name
suggests: `CronConfig.default_timezone` is read only by the `raven cron config
get` display path, and `_compute_next_run` takes a per-job `tz` or the machine's
local zone rather than this field, so no firing time changes either way. That
gap predates this change, lives in the scheduler, and is not touched here. The
row that delivered schedules to named channels disappears; it wrote to a retired
field, so nothing set there was ever honoured, and no stored value is lost
because the loader already strips that key when reading.

A config file that still carries the retired key keeps loading exactly as
before. That stripping predates this change.

Rollback is a revert of the single squashed commit. No migration, no stored
state, no schema change.

## Related Issues

N/A


gloryfromca (commented):

The RPC boundary now translates the web key to the field name the shared cron writer accepts, and the resulting value round-trips through the loader used by the scheduler. Removing `cron.forwardChannels` is consistent with the earlier Fire-at-origin retirement, its loader migration, the CLI refusal, and the canonical domain guidance. I found no weakened tests or compatibility regression in the touched behavior.

Coverage: I reviewed the full diff; AGENTS.md/CLAUDE.md and CONTEXT terminology; the settings caller, cron writer, schema, loader, and RPC error path; relevant history; backward compatibility; generated i18n consistency; and mergeability against `github/main`. I also checked the documented pre-existing snake_case display/write edge and confirmed it does not keep this fix from taking effect.

Verification: `uv run pytest tests/test_rpc_settings.py tests/test_config_update.py tests/test_config_loader.py tests/test_cli_cron_commands.py -x` (235 passed); `npm test --prefix ui-web -- --run src/features/settings/SettingsPage.test.tsx` (124 passed); `npm run type-check --prefix ui-web` (passed). All completed GitHub checks are green; the coverage gate was still running when this review was posted.

0xKT:
**Not a blocker** -- the write fix is right, and I verified it. This is about what the new test
says it proves.

tests/test_rpc_settings.py:596-598 asserts `load_config(cfg).cron.default_timezone == "Asia/Tokyo"`
under the comment "what decides whether the box did anything is the value the cron scheduler goes
on to read". That assertion re-reads the config the write just produced. No scheduler reads that
field.

Checked at HEAD:

  - raven/proactive_engine/schedulers/cron/service.py:65
        tz = ZoneInfo(schedule.tz) if schedule.tz else datetime.now().astimezone().tzinfo
    The tz-less branch is the machine's local zone, not the configured default.
  - The whole scheduler package consults no configuration: grep for `load_config` and
    `from raven.config` under raven/proactive_engine/schedulers/cron/ returns nothing.
  - All three job-creation sites pass `tz` straight through with no substitution:
    raven/proactive_engine/schedulers/cron/tool.py:226, raven/rpc/methods/console.py:365,
    raven/cli/cron_commands.py:754.
  - The only consumer of the written field anywhere in the tree is the display path
    `raven cron config get` (raven/cli/cron_commands.py:849-860).

The field's own contract says otherwise: raven/config/schema.py:304-305 is
`default_timezone: str = "Asia/Shanghai"` with the docstring "Default IANA timezone for cron
expressions without explicit ``--tz``." That is the promise service.py:65 does not keep.

The missing reader predates this PR -- service.py:65 is unchanged at the base and this PR touches
nothing under raven/proactive_engine/ -- so this is not a regression, and I am not asking this PR
to fix the scheduler. What is new is the sentence, and it is the sentence a later reader will
trust. Worth noting where the user lands either way: before this PR the control failed loudly;
after it, it saves and reports success while no firing time changes.

Two ways to close it, either is fine: narrow the comment to what the assertion actually
establishes (the value round-trips through the writer and back through the loader), or leave the
comment and open an issue for the scheduler to honour the default, referenced from here.

Things I checked and found sound, for the record:

  - `to_snake` is imported at raven/rpc/methods/console.py:22 (already used at :654), and
    `to_snake("defaultTimezone")` is `default_timezone`, a real `CronConfig` field
    (fields: default_timezone, notify_missed). No NameError, no mismapping.
  - The conversion belongs at this boundary: the only other caller of `update_cron_config`
    (raven/cli/cron_commands.py:897) already passes snake_case, so no sibling caller is left
    broken by putting it here rather than in the shared writer.
  - `forward_channels` really is retired -- tests/test_config_loader.py:48 asserts an old config
    carrying it comes back without the attribute -- so dropping the control and the endpoint
    branch strands no live setting.
  - The generated ui-tui catalogue is in sync with i18n/messages.json by the repo's own gate
    (`npm run lint:i18n`, .github/workflows/ci.yml:348), which is green on this head.

gloryfromca:

@0xKT is right about the scheduler-reader claim. I traced the consumers again: `CronConfig.default_timezone` is currently read only by the CLI display path, while `_compute_next_run` uses a per-job `schedule.tz` or the machine local timezone. I therefore withdraw the sentence in my earlier review that described this round-trip as reaching a loader value used by the scheduler.

That correction does not change mergeability. The RPC conversion still fixes the write it claims to fix, the retired control is still correctly removed, and the missing scheduler consumer predates this PR. The remaining suggestion is the one already named above: narrow the new test comment or track the scheduler contract separately.

Verification on the unchanged head: `uv run pytest tests/test_rpc_settings.py -k "cron or forward_channels or timezone" -x` passed all 5 selected tests.

LivXue:
Fixed in `6834bbdc3`. You are right, and the claim was wider than you framed it: the test *name* said it too.

Verified your three checks before touching anything:

- `service.py:65` is `tz = ZoneInfo(schedule.tz) if schedule.tz else datetime.now().astimezone().tzinfo` -- the tz-less branch is the machine's local zone.
- `grep -rn "load_config\|from raven.config" raven/proactive_engine/schedulers/cron/` -> no hits.
- The only readers of the cron block anywhere are `cron_commands.py:855` and `:860` (`getattr(config.cron, k)`, the `raven cron config get` display path) and `gateway_commands.py:603`, which reads `notify_missed` -- a different field. So `default_timezone` has exactly the one consumer you named.

Took your first option, narrowing, but kept the finding where the next reader lands rather than deleting the sentence:

- `test_the_timezone_reaches_the_config_the_scheduler_reads` -> `test_the_timezone_round_trips_through_the_writer_and_the_loader`
- the comment now states what the assertion pins (the writer's spelling against the loader's) and records that the field's one consumer is the display path and that `_compute_next_run` falls back to the local zone.

The PR description needed the same correction and got it: its Risk section said "the timezone box starts taking effect", which would have landed in the squash commit body on `main`. It now says the box starts persisting, that no firing time changes either way, and that the gap predates this PR and lives in the scheduler. Your sentence about where the user lands is the one I had missing.

The description's `-k "cron or forward_channels or timezone"` selector still matches the renamed case: 5 passed; full file 61 passed; 331 passed across the six files at the rebased head.

Not fixing the scheduler here. `schema.py:304-305` still promises "Default IANA timezone for cron expressions without explicit --tz", which `service.py:65` does not keep -- noting it so the gap is on record beyond this thread.

gloryfromca (commented):

The new commit accurately narrows the timezone test to the writer/loader round trip it establishes and explicitly separates the pre-existing scheduler-consumer gap. That resolves the prior nonblocking suggestion without weakening the assertion. A range-diff confirms the implementation reviewed previously is otherwise unchanged after rebasing onto the newer `main`; the apparent CODEOWNERS and README delta is base movement, not part of this PR.

Coverage for this revision: I checked the authored delta and full PR diff, AGENTS.md/CLAUDE.md comment and test rules, the canonical Fire-at-origin guidance, the timezone callers and history, backward compatibility, test strength, architecture boundaries, and mergeability against `github/main`.

Verification: `uv run pytest tests/test_rpc_settings.py -k "cron or forward_channels or timezone" -x` (5 passed); `npm test --prefix ui-web -- --run src/features/settings/SettingsPage.test.tsx` (124 passed); `npm run type-check --prefix ui-web` (passed).

0xKT:
**Not a blocker** -- the control this repaired write now reaches has a spelling rule of its own,
and the file next door states it. On a cron block spelled snake_case the save does not update the
key, it grows a second one beside it, and the reply says there was no previous value.

Measured on this head merged onto 32e612d5, three runs in one file so they share a harness, all
through `settings_set` rather than an internal helper:

    CAMEL    cron={"defaultTimezone": "UTC"}    -> previous='UTC'  block={'defaultTimezone': 'Asia/Tokyo'}
    SNAKE    cron={"default_timezone": "UTC"}   -> previous=None   block={'default_timezone': 'UTC',
                                                                          'defaultTimezone': 'Asia/Tokyo'}
    CONTROL  channels={"send_progress": false}  -> previous=False  block={'send_progress': True}

The third row is the control, and it is what makes the second one mean something: the same
endpoint, the same snake-spelled file, one key routed through `_write_raw_key`/`_as_written`
instead. That path updates in place and reports `previous` correctly. Same door, same input shape,
two answers -- so the difference is the writer, not my setup.

The rule is written down twice in this tree, and neither copy is far away.
`_as_written`'s docstring (raven/rpc/methods/console.py:645-650), about this very endpoint: "a
hand-written config may hold either. A write that always used the camelCase one did not update the
file, it grew a second key beside the first ... The write has to follow the file rather than the
writer's spelling." And again at raven/config/update.py:147-149 for
`set_sentinel_nudge_quota`: "Respects whichever key casing the file already uses ... writing a
second casing for a field already present would duplicate it."

`update_cron_config` does not: raven/config/update.py:52 is `camel_key = to_camel(key)`,
unconditionally.

**Why this is a note and not a block.** The behaviour is the writer's and predates this PR -- it is
there in the first commit of the file's history, and `raven cron config set --default-timezone`
has always had it. What this PR changes is that the RPC path now REACHES the writer at all: on the
base `sub` was `defaultTimezone`, which `CronConfig.model_fields` does not contain, so the call
raised before writing anything. So this is a hole in a feature that starts working today, not a
regression, and nothing that worked stops working: `CronConfig` ignores extras, the camelCase
alias wins, and the value the user typed is the effective one either way
(`load_config().cron.default_timezone == 'Asia/Tokyo'` on both rows above).

What it costs is a stale key the user can keep editing with no effect, and a `previous` the wire
schema declares as required answered as null. Nothing in ui-web reads `previous` today, so the
wrong value stays on the wire rather than reaching a screen.

Smallest thing that closes it: have `update_cron_config` follow the file the way `_as_written`
already does, which fixes the CLI in the same move. The PR's comment argues the conversion belongs
at the call site rather than "in a writer the CLI shares" -- that is a fair call for the
camel/snake translation, and it is separable from this: the writer still has to decide which
spelling to WRITE, and that decision is the one the two docstrings above already made.

For the record, what I checked and found sound on the merged tree: `GATES OK head=6834bbdc
tip=32e612d5 base=c2c5f279 ruff/lint-imports/commit/large/lang rc=0`; 91 passed across
tests/test_rpc_settings.py and tests/test_config_update.py; and the fix is pinned -- reverting
`to_snake` at console.py:577 turns exactly two tests red,
`TestTheCronTimezoneControl::test_the_timezone_round_trips_through_the_writer_and_the_loader` and
`::test_the_previous_timezone_comes_back`.

gloryfromca (approved):

gloryfromca:
#525 takes this change back off `main`. It is open and waiting on a review, and the reason is where the change landed, not what it does.

The frontend half edits `ui-web/src/features/settings/SettingsPage.tsx`, and that file does not exist on `refactor/ui_web_architecture` (#475), which rebuilds the settings surface as `SettingsApp.tsx` plus `ui-web/src/features/settings/pages/`. Rather than carry a change written against the old page across that rewrite, it goes back to you to resubmit against the refactor branch.

What is left to redo there, checked at `bc74bf735`:

- **The backend fix, unchanged.** `raven/rpc/methods/console.py:564` on that branch still holds the joint `("cron.defaultTimezone", "cron.forwardChannels")` whitelist branch, so the `to_snake` conversion and the split into a timezone-only branch apply exactly as you wrote them, and `tests/test_rpc_settings.py` comes with them.
- **Nothing else.** The rebuilt settings pages carry no cron section at all: no `cron.defaultTimezone` reader, no forwardChannels row, and the whole `gui.set.chn.*` group (11 keys on `main`, `cron_to` and `cron_ph` among them) is already absent from `i18n/messages.json`. The row removal and the catalogue cleanup are already true there.

One thing to settle with @0xKT before writing it: the timezone box itself is not on the new settings page either, so with the backend fixed the control still has no UI on that branch. Whether porting it belongs in your PR or in #475 is their call.

PR #507, opened 2026-09-18
feat(docs): add a bilingual documentation site and its deploy pipeline

LivXue:
## Summary

Stands up a bilingual MkDocs site under `docs-site/`, published to GitHub Pages
by a workflow that builds on every pull request and deploys only from `main`.
Fourteen topics ship in English and Chinese: the quick start, self-hosting,
Docker, the WebUI, the sandbox manual, the runtime architecture, the command
reference, the repository layout, the developer workflow, the proactivity design
and reference pair, the self-evolution map and the tracing API.

This pull request is additive. Both READMEs keep their prose and gain one link
to the site in the header row; trimming them is deliberately a second pull
request, so the site can be confirmed live before anything depends on it.

Key decisions:

- **`docs-site/`, not `docs/`.** `docs/README.md` defines `docs/` as design
  notes and dated records, and `docs/plans/` is an archive explicitly not held
  to today's layout. A user manual has the opposite lifecycle.
- **Full bilingual parity**, via the `mkdocs-static-i18n` suffix structure
  (`page.md` beside `page.zh.md`). Parity decays immediately without a guard, so
  a test fails when a page has no twin.
- **Anchors are pinned on the Chinese side.** A CJK heading slugifies to
  nothing, so Markdown numbers it `_1`, `_2`, and inserting one heading
  renumbers every anchor below it, silently breaking links readers already hold.
  Each Chinese heading therefore states the anchor its English twin gets for
  free: 133 heading anchors, identical across the two languages, none
  positional.
- **A left rail, not a top bar.** The header carries the brand, the search box
  and the navigation tree down the left; the table of contents on the right is
  one continuous polyline clipped to the headings currently on screen.
- **No template overrides.** Material customises through `theme.custom_dir` with
  Jinja partials, those partials are `.html`, and `scripts/check_large_files.py`
  holds `.html` in `BLOCKED_ASSET_EXTENSIONS`. The skin is therefore one
  stylesheet remapping Material's own custom properties, plus one script for the
  table-of-contents rail. The brand mark is a `data:` URI for the same reason.
- **An owner-signed `docs-site/` exemption zone** in AGENTS.md section 1.3. The
  i18n config carries the Chinese navigation labels, and `mkdocs.yml` is not
  `*.md`, so the blanket suffix exemption does not reach it. A contract test
  holds the AGENTS.md list and `EXEMPT_PREFIXES` equal, so both moved together.
- **The site never links out to a file it does not host.** Documents that stay
  in the repository are named as plain paths rather than linked, and a guard
  fails a page that links to an external webpage or a repository-relative path.

Two defects found and closed here rather than after publication. The Chinese
site rendered an English sidebar: navigation labels live in `mkdocs.yml`, not in
any page, so comparing the two languages' Markdown finds nothing and only the
rendered output shows it. And a control could be present, sized, animated and
answer `querySelector` while being unclickable, because a popup Material sized
for a top bar landed outside the viewport once the header became a rail.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

All commands run at the pushed head, rebased onto `origin/main`.

```
make check-large-files                                         exit 0
make check-source-language                                     exit 0
make docs-build                                                exit 0, 0 errors
  docs-site/site/index.html and docs-site/site/zh/index.html   both present
  mkdocs_static_i18n: Translated 17 navigation elements to zh
uv run pytest tests/test_docs_site.py tests/test_source_language_check.py \
  tests/test_readme_scope_canon.py tests/test_docker_runtime.py \
  tests/test_living_docs.py                                    44 passed
uv run pytest tests/integration/test_docs_site_controls_e2e.py 13 passed
pre-commit run --files <changed files>                         all hooks pass
git diff --check                                               clean
```

Anchor parity was measured on the built HTML rather than assumed: 14 page pairs,
133 heading anchors compared at h2 through h6, every one identical between the
two languages and none positional.

The controls are hit-tested rather than queried. `document.elementFromPoint` at
the exact point a reader clicks must return the control itself, which is the
only check that catches an off-viewport popup; the language switcher, the search
panel, the repository link and the table-of-contents rail each pass it.

The brand lockup is measured, not eyeballed. A failing test was written first
and reported the wordmark sitting 4.00px below the mark's centre line; it is
0.00px after the fix, and a 4x screenshot puts the residual optical offset at
0.25px.

Every guard this branch adds was confirmed to fail first. The brand-class guard
was mutated from both sides in turn: renaming `.em-eyebrow` in the stylesheet
turned it red, renaming `class="em-standfirst"` in `index.md` turned it red, and
it is green again with both restored. The anchor guard was shown to catch both
failure modes, a missing anchor and a wrong one.

## Risk

No behaviour change to any shipped code path. Nothing imports `docs-site/`, the
new Makefile targets are additive, and the only change outside `docs-site/` and
its tests is one link added to each README's header row.

Rollback is deleting the branch: the site does not exist until Pages is enabled
with `Source: GitHub Actions`, and every other change is inert without it.

Disclosed residue, checked and deliberately not changed:

- The README header now links to `https://evermind-ai.github.io/Raven/`, which
  404s until Pages is enabled. It is the same URL the Documentation section
  already carries.
- The `deploy` job is skipped on pull requests by design, so a green check here
  does not prove deployment works. The first real evidence is the `deploy` job's
  own conclusion on the `main` run after merge.
- GitHub Pages is not enabled on this repository. Measured: `GET /repos/.../pages`
  returns 404. Under `Deploy from a branch` the uploaded artifact is silently
  ignored while the workflow still reports success, so the source must be set to
  `GitHub Actions`.
- `tests/integration/` is excluded by `norecursedirs`, so the browser suite is a
  local guard and never runs in CI.
- The self-hosting page's Makefile target, docker path and port claims are
  unguarded. They were equally unguarded in the README they came from, and the
  owner ruled against adding a guard for them.
- The self-evolution SOP and the benchmark contract stay in the repository and
  are named as plain paths from the pages that reference them. The SOP is a
  translation of an upstream project's internal document and carries redaction
  placeholders, which is a publication decision rather than a formatting one.
- `docs/plans/` still shows the pre-restyle `mkdocs.yml`. CONTEXT.md defines that
  directory as an archive of the tree as it stood, with an explicit rule against
  updating an old plan to match a later change.
- The Chinese pages drop the brand's italic serif accent word. Cormorant Garamond
  carries no CJK glyphs, so it would fall back to another face without any sign.
- The code font is Material's default Roboto Mono. The brand names four
  typefaces and no monospace, so none was invented for it.

## Related Issues

N/A


docs-site/docs/stylesheets/evermind.css
```diff
@@ -0,0 +1,314 @@
+/*
+ * EverMind brand skin for Material for MkDocs.
+ *
+ * The repository forbids committing HTML (AGENTS.md section 7, machine-checked
+ * by scripts/check_large_files.py), so the theme cannot be customised through
+ * Material's usual overrides/*.html partials. Everything here is therefore done
+ * by remapping Material's own CSS custom properties onto the EverMind token set
+ * in .claude/skills/evermind-web-design/references/tokens.css, then correcting
```
gloryfromca: Blocking: remove or replace the stylesheet's untracked local-only source reference.

`git ls-files '.claude/**'` returns no files, so the cited `.claude/skills/evermind-web-design/references/tokens.css` cannot be inspected by maintainers or reviewers. AGENTS.md section 1.1 explicitly forbids source comments that reference locally-visible-only paths; this new header therefore violates a stated hard constraint. The public design spec now records the CSS-only rationale, so this reference can be removed or replaced with a tracked source.

gloryfromca (commented):
Blocking: remove or replace the stylesheet's untracked local-only source reference.

I found one blocking repository-rule violation, marked inline.

I reviewed the full `github/main...7f9aa66a3b97` diff, including the commit added while this review was in progress; the relevant site/config callers and generated link behavior; commit history and action tag availability; AGENTS.md, CLAUDE.md, and CONTEXT routing; backward compatibility; test changes for weakening; and the documented two-stage Pages deployment architecture.

Verification at the latest head: the strict bilingual MkDocs build succeeded and generated both locales with seven translated navigation elements; 40 relevant pytest cases passed; the source-language and large-file gates passed against `github/main...HEAD`; and `git diff --check` passed. GNU make is unavailable in this environment, so I ran the exact `uv` commands behind the Makefile targets. The PR's `docs build` GitHub check is also green. I found no other blockers.

docs-site/docs/stylesheets/evermind.css
LivXue: Fixed in 33d8d6198.

Reproduced before changing anything, rather than taking the finding on trust: `git ls-files '.claude/**'` returns 0 files here too, `.gitignore:57` is `.claude/`, and `AGENTS.md:36` is the rule. You are right, and the pre-submit sweep I ran did not check my own new comment against section 1.1 -- it checked the domain-term registry and stopped there.

The header now names `docs/specs/2026-09-18-docs-site-design.md`, whose Presentation section carries the CSS-only constraint and its rationale, and says the tokens are declared in the file's own `:root` block. Every path it still cites was confirmed tracked with `git ls-files --error-unmatch`.

I checked the class rather than the instance: over `origin/main...HEAD` the branch had exactly one added line citing a gitignored path, the one you named. The remaining `.claude/` hits in the tree are pre-existing and fall into two groups that are not this defect -- `docs/plans/` archive entries, which CONTEXT.md holds to their own date, and references to Claude Code's own `~/.claude/projects/` user directory, a different referent from this repository's ignored `.claude/`.

Re-verified at the new head: `mkdocs build --strict` exit 0 with both locale trees, `tests/test_docs_site.py` 3 passed, `make check-large-files` and `make check-source-language` both exit 0, and the full pre-commit hook set over `origin/main...HEAD` clean including check-yaml.

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed at `33d8d6198534`: the untracked `.claude/` citation is gone, the replacement design-spec path is tracked, and the header now accurately points to the tokens declared in its own `:root` block. The strict bilingual build, 40 focused tests, source-language and large-file gates, and `git diff --check` all pass. This resolves my finding.

gloryfromca (commented):

The only delta from the previously reviewed head replaces the inaccessible source citation with tracked, accurate references, so the prior blocker is resolved; I replied on and resolved that thread. I reviewed the new commit in the context of the full diff, repository rules and context documents, surrounding site/build configuration, history, backward compatibility, unchanged test strength, and the two-stage deployment architecture. No new issue emerged.

Verification at `33d8d6198534`: strict bilingual MkDocs build passed with both locale trees and seven translated navigation elements; 40 focused pytest cases passed; source-language and large-file checks passed against `github/main...HEAD`; the replacement spec path is tracked; and `git diff --check` passed.

tests/test_docs_site.py
```diff
@@ -0,0 +1,127 @@
...
+stylesheet no build step connects it to, so renaming either side errors
+nowhere, reddens nothing, and is visible only to someone looking at the page.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+from markdown.extensions.toc import slugify
```
gloryfromca, line 17: Blocking: make the docs-site unit test importable in the standard unit environment.

`markdown` is only installed by the new `docs` dependency group, but the ordinary unit workflow runs `uv sync --all-extras --dev` and `make coverage-shard` without that group. The earlier focused run passed only because `mkdocs build` had already added the docs packages to the same `.venv`. Reproducing the CI boundary with `uv run --isolated --frozen --python 3.12 --all-extras --dev --no-group docs pytest tests/test_docs_site.py -x -o addopts=''` aborts collection here with `ModuleNotFoundError: No module named 'markdown'`, so the standard unit shard cannot run. This revision introduces the import, normal CI reaches it, and there is no working test path past collection, satisfying all three late-round blocker criteria.

gloryfromca (commented):
Blocking: make the docs-site unit test importable in the standard unit environment.

I found one new blocker in the expanded revision, marked inline.

I reviewed the full `github/main...5887e91b0a64` diff and the four post-review commits, including repository rules and context documents, the generated site and build/deploy callers, relevant history, backward compatibility, test strength and discovery, and the rail/browser architecture. The previously resolved stylesheet thread remains resolved.

Verification: the strict bilingual MkDocs build passed and translated 17 navigation elements; the source-language and large-file gates plus `git diff --check` passed; and 44 focused tests passed after the docs group had been installed. That last environment masks the blocker: an isolated reproduction matching the unit job (`--all-extras --dev --no-group docs`) aborts collection of `tests/test_docs_site.py` with `ModuleNotFoundError: markdown`. The browser suite collected 13 tests locally but all 13 skipped because this runner has no Chrome binary; the PR records a separate 13-pass run.

This is a round-five blocker on all three required grounds: this revision adds the import, the ordinary unit shard reaches it, and the shard cannot run past collection without a code or dependency correction. I found no other new blocker.

tests/test_docs_site.py
LivXue: Fixed in 4a2c0f75e.

Reproduced under your exact command before changing anything, rather than taking the finding on trust: `uv run --isolated --frozen --python 3.12 --all-extras --dev --no-group docs pytest tests/test_docs_site.py -x -o addopts=''` aborts collection here too with `ModuleNotFoundError: No module named 'markdown'`. `grep -n markdown pyproject.toml` returned only `slackify-markdown`, and `ci.yml:129` is `uv sync --locked --all-extras --dev`, so the module had no way into the unit job. You are right, and my local runs passed only because `make docs-build` had already put the docs packages in the same `.venv` - the same masking you named.

The fix declares `markdown` as a dev dependency (`uv add --dev "markdown>=3.10"`). I kept the real import rather than skipping it: the guard's whole value is calling the slugify mkdocs itself calls, and an `importorskip` would have turned the shard green while leaving the guard as absent from CI as it is today.

That the guard now runs is not enough, so I proved it also bites, under the job's own sync flags: mutating one pinned anchor in `docs-site/docs/development.zh.md` from `{ #run-tests }` to `{ #run-testz }` fails it with `development.zh.md: '运行测试' lands on #run-testz, but 'Run tests' lands on #run-tests`, and it passes again with the anchor restored.

Verification at `4a2c0f75e`:

```
uv run --isolated --frozen --python 3.12 --all-extras --dev \
  --no-group docs pytest tests/test_docs_site.py -x -o addopts=''   5 passed
uv run pytest tests/test_docs_site.py tests/test_readme_scope_canon.py \
  tests/test_docker_runtime.py tests/test_source_language_check.py \
  tests/test_living_docs.py                                        44 passed
make docs-build                                                    exit 0, 0 errors
scripts/check_source_language.py / check_large_files.py            exit 0 / exit 0
```

On the browser suite: your runner has no Chrome binary, so all 13 skip there. The PR's 13-pass figure comes from a runner that has one, and `tests/integration/` is excluded by `norecursedirs`, so that suite is a local guard and never runs in CI either way. The anchor guard is the one that had to work in the unit job, and now does.

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed at `4a2c0f75e7c8`: the exact isolated standard-unit command that previously failed collection now passes all 5 docs-site tests without the `docs` group. Declaring `markdown` in the dev group matches the test's direct import while preserving the real MkDocs slugifier check. The 44-test focused set, strict bilingual build, source-language and large-file gates, and `git diff --check` also pass. This resolves my finding.

gloryfromca (commented):

The dependency fix at `4a2c0f75e7c8` is correct: `markdown` is now a direct dev dependency, so the docs-site guard imports and runs in the ordinary unit environment without relying on a prior docs build. The exact isolated reproduction passes 5 tests; the 44-test focused set, strict bilingual MkDocs build, source-language gate, large-file gate, and `git diff --check` all pass. I replied to and resolved the settled unit-environment finding.

This round covered the revision delta and lock metadata, the standard CI unit caller, dependency/backward-compatibility impact, and whether test coverage was weakened; it was not. The prior full-diff review covered the repository rules and context files, docs/deploy callers and history, compatibility, and architectural constraints. I found no new plain error or blocker.

ZuyiZhou (approved):
Approving. Reviewed independently rather than taking the description at its word.

Verified locally at the pushed head:

- `make docs-build` (`--strict`) builds clean; both `site/index.html` and `site/zh/index.html` are produced.
- Anchor parity measured on the built HTML, not the Markdown: 14 page pairs, 133 h2-h6 anchors, every one identical across the two languages, and none of the positional `_1` / `_2` form the pinning exists to prevent.
- The Chinese sidebar renders Chinese and the document carries `lang="zh"`, so the navigation-label defect is actually closed.
- 31 unit tests pass (`test_docs_site.py`, `test_source_language_check.py`), and `make check-large-files` / `make check-source-language` both exit 0.
- Every `raven <command>` in the command reference exists in the real CLI, and the subcommands were checked against `--help` too: `gateway status|reload|stop`, `agents new`, `plugin auth`, `mcp bridge`, `ops connection`, and `tracing compact` (an action argument, correctly documented as one).
- Section 7 is clean: no new file over 100 KiB, no `.html` or `.svg` committed, the brand mark is a 35-character `data:` URI.
- The AGENTS.md zone and `EXEMPT_PREFIXES` moved together and the contract test holds them equal. Action versions match the other 18 workflows.

Three non-blocking notes, none worth another round:

1. The `docs-site/` exemption prefix is wider than the AGENTS.md entry describes. The stated need is the navigation labels in `mkdocs.yml`; the prefix also quietly exempts `stylesheets/` and `javascripts/`. `docs-site/mkdocs.yml` would match the text exactly.
2. Two guards in `tests/test_docs_site.py` disagree in their remediation advice: the relative-link test tells you to make the target an absolute github.com URL, and the external-link test then rejects exactly that. Both pass today because no page has a link of either kind, so it is only the message that misleads.
3. `CHROME = "/usr/bin/google-chrome"` means the browser suite silently skips on macOS rather than running. Combined with `norecursedirs` excluding `tests/integration/`, the rail's coupling to Material's internal class names has no guard that runs anywhere by default.

One thing to know at merge time, already disclosed: the workflow-level `cancel-in-progress: true` outranks the deploy job's own `cancel-in-progress: false`, so two merges landing close together can cancel a deploy in flight. The following push redeploys, so it self-corrects.

Pages still needs `Source: GitHub Actions` before the README link resolves.

PR #382, opened 2026-09-09
feat(cli): full conversation preview in the trajectory browser

forrestjgq:
## Summary

Space on an attempt row in the `raven trajectory` browser previously
showed two truncated per-turn preview lines. This PR replaces it with an
interactive full-conversation viewer:

- Data layer `raven/trajectory/conversation.py`: rebuilds an attempt's
  span snapshot into a causally ordered stream of labeled records (User
  input, LLM input/output, Tool input/output, skill, memory, and
  subagent events). Event-time semantics (inputs at span start, outputs
  at span end) plus nesting depth from parentSpanId keep the causal
  order. Full content comes from artifact files (paths restricted to
  the trace store, streamed 512 KiB cap per file); span previews are
  only a visible fallback. ERROR spans and expected-but-unreadable
  payloads always yield a placeholder record. LLM inputs omit only the
  item-by-item verified common message prefix against the previous call
  (omissions and rewrites are announced); tool_calls of unexecuted
  tools keep their full arguments.
- Rendering: kind-colored labels aligned to one column, hanging-indent
  bodies with word-aware CJK-safe wrapping, degraded/error/meta note
  lines, single-pass turn separators (interleaved traces are never
  regrouped), and a stacked layout on narrow terminals that never drops
  characters.
- Interactive viewer (prompt_toolkit full-screen, entered for every
  preview on a terminal): scrolling, a muted bottom bar with an
  'h for help' hint, filter/collapsed state and the scroll range (laid
  out by display width, never overflowing any terminal), a scrollable
  help page reachable in full on small screens, a global collapse (s:
  bodies over five display lines fold to five plus an ellipsis; error
  and degradation notes never fold), and a label filter (%: real input
  line, case-insensitive, empty clears; filtered-out turn groups lose
  their separators). Resizes re-wrap and re-judge folding live; the
  content caret is hidden (only the filter input shows one). q/Esc
  returns to the list; Ctrl+C cancels the whole browser from any viewer
  state. A rebuild failure falls back to the legacy per-turn preview;
  without a terminal the preview degrades to direct printing.
- Every dynamic field passes the sanitization gates (control
  characters, rich markup, and ANSI never reach the terminal). Adds the
  Conversation Record term to CONTEXT.md.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_trajectory_conversation.py -q` -> 45 passed
[...]

PR #510, opened 2026-09-19
feat(cli): full conversation preview in the trajectory browser

forrestjgq:
## Summary

Space on an attempt row in the `raven trajectory` browser previously
showed two truncated per-turn preview lines. This PR replaces it with an
interactive full-conversation viewer. It re-lands the content of PR #382,
which was auto-closed by the 2026-09-12 history linearization without
being merged; the branch is rebased onto the linearized main, with one
extra commit expressing the CJK display-width test fixtures as unicode
escapes to satisfy the source-language gate that main gained since.

- Data layer `raven/trajectory/conversation.py`: rebuilds an attempt's
  span snapshot into a causally ordered stream of labeled records (User
  input, LLM input/output, Tool input/output, skill, memory, and
  subagent events). Event-time semantics (inputs at span start, outputs
  at span end) plus nesting depth from parentSpanId keep the causal
  order. Full content comes from artifact files (paths restricted to
  the trace store, streamed 512 KiB cap per file); span previews are
  only a visible fallback. ERROR spans and expected-but-unreadable
  payloads always yield a placeholder record. LLM inputs omit only the
  item-by-item verified common message prefix against the previous call
  (omissions and rewrites are announced); tool_calls of unexecuted
  tools keep their full arguments.
- Rendering: kind-colored labels aligned to one column, hanging-indent
  bodies with word-aware CJK-safe wrapping, degraded/error/meta note
  lines, single-pass turn separators (interleaved traces are never
  regrouped), and a stacked layout on narrow terminals that never drops
  characters.
- Interactive viewer (prompt_toolkit full-screen, entered for every
  preview on a terminal): scrolling, a muted bottom bar with an
  'h for help' hint, filter/collapsed state and the scroll range (laid
  out by display width, never overflowing any terminal), a scrollable
  help page reachable in full on small screens, a global collapse (s:
  bodies over five display lines fold to five plus an ellipsis; error
  and degradation notes never fold), and a label filter (%: real input
  line, case-insensitive, empty clears; filtered-out turn groups lose
  their separators). Resizes re-wrap and re-judge folding live; the
  content caret is hidden (only the filter input shows one). q/Esc
  returns to the list; Ctrl+C cancels the whole browser from any viewer
  state. A rebuild failure falls back to the legacy per-turn preview;
  without a terminal the preview degrades to direct printing.
- Every dynamic field passes the sanitization gates (control
  characters, rich markup, and ANSI never reach the terminal). Adds the
  Conversation Record term to CONTEXT.md.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_trajectory_conversation.py
  tests/test_cli_preview_viewer.py tests/test_cli_trajectory_browse.py
  tests/test_trajectory_store.py -q` -> 298 passed (viewer assertions
  run against really rendered screens)
- `uv run pytest -q` (full, on the rebased branch) -> 23463 passed,
  125 skipped; 33 failed + 20 errors reproduce with identical node ids
  on a clean origin/main worktree - environment baseline, not
  introduced here (passed-count delta of 96 equals the tests this
  branch adds)
- `make check-source-language`, `make check-large-files`,
  `make lint-types`, `make lint-python`, `make lint-imports` -> all pass
- Manual: rendered 6 real local attempts at widths 72/120/30 with zero
  over-width lines (largest attempt 7060 lines in 0.35s); drove the
  real TUI in a pty at normal and 44x12 sizes (viewer, help paging,
  collapse, label filtering, Ctrl+C cancel)

## Risk

- Read/display side only; no persisted format changes. The untrusted
  input surface (artifact content, typed filter input) is contained by
  the store-prefix path check, the 512 KiB streamed cap, and per-field
  sanitization.
- User-visible change: the Space preview is now a full-screen viewer;
  the legacy rendering remains as the automatic fallback when the
  rebuild fails, and non-terminal environments print directly.
- Rollback: revert this PR, or restore the previous `_preview_screen`
  call site to get the old preview back.

## Related Issues

N/A


gloryfromca (commented):
Blocking: current LLM input artifacts must be resolved before the full preview can merge.

I found one blocker in the live trace compatibility path. I covered the full diff, AGENTS.md and the canonical trajectory terms in CONTEXT-MAP.md/CONTEXT.md, the preview callers, the current trace writer and artifact resolver, relevant history, backward compatibility, and whether the tests weakened or bypassed behavior; I found no separate architecture issue.

Verification: `uv run pytest tests/test_trajectory_conversation.py tests/test_cli_preview_viewer.py tests/test_cli_trajectory_browse.py -q` passed (221 tests); `uv run pytest tests/test_tracing_api.py tests/test_trajectory_bundle.py -q` passed (89 tests); the source-language and large-file check scripts passed. A direct reproduction using `TraceStore.address_items` and `audit.artifact.v2` showed `contains full message: False`, with the record displaying a `$msg` hash instead.

raven/trajectory/conversation.py
```diff
@@ -0,0 +1,734 @@
...
+        _emit(info, records, "Turn", _PHASE_INPUT, "")
+
+
+def _emit_llm(info: _SpanInfo, records: list[dict[str, Any]], state: Path, chains: _ChainState) -> None:
+    meta = _llm_meta(info.attrs)
+    chain = (info.trace_id, info.parent_id)
+    payload = _slot_payload(state, info.attrs, "llm.input")
+    if payload is not None:
+        obj, text, degraded = payload
+        messages = obj.get("messages") if isinstance(obj, dict) else None
```
gloryfromca, line 464: Blocking: resolve current `audit.artifact.v2` LLM inputs before rendering them.

The active trace writer stores each `llm.input` as a v2 shell whose `messages` entries are `$msg` references. This code parses that shell as ordinary JSON and sends those reference dicts into `render_input`; with a real `TraceStore` artifact I get a role of `?` followed by a `$msg` hash, and the original message text is absent. Consequently the new full-conversation preview hides the entire LLM input for newly recorded calls even though the message blobs are present. The current format needs resolving, including missing-blob degradation, before prefix comparison and rendering; the new tests only construct legacy inline-message payloads, so they do not exercise this ordinary path.

forrestjgq: Fixed in 80fe873e. The llm.input payload now resolves through raven.tracing.artifact_v2.resolve_payload before prefix comparison and rendering: v2 shells render their real message text, a missing blob renders the deterministic placeholder and is announced in the record's degradation note (placeholders keep list positions, so the verified-prefix omission of later calls survives a missing blob instead of degrading to a full dump). The new tests publish shells through the real TraceStore writer (address_items) rather than hand-built inline payloads, and cover the intact-shell, missing-blob, and mixed v1/v2 paths; an end-to-end check built a shell via observability.semconv.llm_input_payload and rendered the actual message text.

gloryfromca: Blocking: v2 message-blob reads still need the preview size bound.

Confirmed: 80fe873e fixes the defect raised in this thread. Real TraceStore v2 shells now render message text, missing blobs produce deterministic placeholders plus a degradation note, and mixed v1/v2 prefix handling remains intact. I am resolving this thread; the remaining size-bound issue is separate and is being posted on the new resolution path.

gloryfromca (commented):
Blocking: v2 message-blob reads must honor the preview size bound.

The original `$msg` rendering defect is fixed and its thread is resolved. I found one new blocker in the resolver path. I rechecked the full PR diff and the 80fe873e delta, AGENTS.md, CONTEXT-MAP.md/CONTEXT.md terminology, the preview caller, the live v2 writer/resolver and image-bearing request path, relevant history, backward compatibility, and whether tests were weakened; no other actionable issue survived refutation.

Verification: `uv run pytest tests/test_trajectory_conversation.py tests/test_cli_preview_viewer.py tests/test_cli_trajectory_browse.py tests/test_tracing_api.py tests/test_trajectory_bundle.py -q` passed (313 tests). The source-language and large-file scripts passed. A real TraceStore reproduction with one 700 KiB image-bearing message produced a 716,900-byte blob, a 716,878-character preview record, `degraded=None`, and the complete payload in memory despite `_ARTIFACT_LIMIT` being 524,288 bytes.

raven/trajectory/conversation.py
```diff
@@ -0,0 +1,765 @@
...
+    if not artifact_v2.is_v2(obj):
+        return obj, None
+    missing: list[str] = []
+
+    def _record(sha1: str) -> dict[str, str]:
+        if sha1 not in missing:
+            missing.append(sha1)
+        return artifact_v2.placeholder(sha1)
+
+    resolved = artifact_v2.resolve_payload(obj, state / "logs" / "audit-artifacts", on_missing=_record)
```
gloryfromca: Blocking: keep v2 message-blob resolution within the preview artifact read cap.

`artifact_v2.resolve_payload` reads each referenced message blob with an unbounded `Path.read_text`, bypassing `_read_artifact` and its 512 KiB I/O/memory limit. This is reachable in normal use because LLM messages can contain inline base64 images: using the real `TraceStore` writer with one 700 KiB image message made `attempt_conversation` materialize all 716,878 rendered characters with no degradation note, even though the configured cap is 524,288 bytes. Production traces already record image-bearing requests around many megabytes, and repeated LLM histories resolve those blobs again per call, so opening a preview can allocate and lay out far beyond the bound or become unusable. Please enforce a bounded read/degradation path for referenced blobs as well as the shell.

forrestjgq: Fixed in f49c291e. Blob resolution no longer goes through resolve_payload's unbounded read_text: each referenced message blob is read through the preview's own bounded reader (streamed 512 KiB + 1 cap, store-prefix check), with reference validation and addressing still owned by raven.tracing.artifact_v2. A blob over the cap renders as a deterministic placeholder ('[message blob over the 512 KiB cap: <sha1>]') instead of its content and is announced in the record's degradation note alongside missing blobs; placeholders keep list positions, so prefix omission still works. Blobs are cached per attempt so repeated LLM histories read each blob once, while every payload still reports its own unavailable blobs (the cache carries the outcome kind). Tests: a real TraceStore-written 516 KiB image-bearing message renders the placeholder with the note and never materializes the content (a spy asserts the read goes through the bounded reader), plus cache-note and mixed missing/oversize cases.

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed with the original real-TraceStore reproduction: the 716,900-byte message blob is now read with a single 524,289-byte request, the payload is absent from the record, and the deterministic oversize placeholder plus degradation note are present. The attempt-wide cache and mixed missing/oversize behavior also pass their focused tests. This resolves my size-bound finding.

gloryfromca (commented):

The bounded v2 blob fix resolves the remaining blocker, and both threads I opened are now resolved. I rechecked the full PR and the f49c291e delta, AGENTS.md and trajectory terminology, the preview caller, current v1/v2 writers and resolver behavior, image-bearing and damaged-store compatibility, relevant history, architecture boundaries, and whether tests were weakened; I found no new actionable issue.

Verification: `uv run pytest tests/test_trajectory_conversation.py tests/test_cli_preview_viewer.py tests/test_cli_trajectory_browse.py tests/test_tracing_api.py tests/test_trajectory_bundle.py -q` passed (316 tests). The source-language and large-file scripts passed. The original 700 KiB reproduction now performs one bounded 524,289-byte read, emits an 87-character placeholder with the degradation note, and does not materialize the payload.

0xKT:
**Not a blocker** -- two things, neither of which changes a rendered pixel today.

**1. `_resolve_v2` is a second decider of the v2 per-field rule, and it already disagrees with the
owner** (raven/trajectory/conversation.py:415)

`raven/tracing/artifact_v2.py:6-7` says of that module: "The canonical serialization, the envelope
key, the reference validation and the per-field resolution rules all live here and nowhere else" --
and gives the reason, that a JavaScript mirror exists and a cross-language round-trip test compares
the two.

`_resolve_v2` delegates the parts its own docstring names -- `is_v2`, `ref_sha1` -- but re-decides
the rest: it strips `artifactFormat` itself and resolves `messages` itself, and it does not resolve
`TEXT_FIELDS` (`systemPrompt`, `prompt`), which the owner does at artifact_v2.py:145-148.

Measured on one shell, both resolvers pointed at the same blobs:

    OWNER artifact_v2.resolve_payload   systemPrompt = 'You are Raven.'
    NEW   conversation._resolve_v2      systemPrompt = {'$msg': 'e7b04557...'}
    messages equal                      True      <- control: both read the same blobs
    whole payload equal                 False
    fields that differ                  ['prompt', 'systemPrompt']

The `messages` line is the control: it is identical, so the divergence is the TEXT_FIELDS policy
and not my harness. (My first attempt put the blobs under the owner's default layout rather than
`state/logs/audit-artifacts`, which made every message read as missing -- worth saying because that
version would have looked like a much larger defect and was my setup, not your code.)

Nothing renders differently today: this module never reads `systemPrompt` or `prompt` -- grep over
conversation.py finds neither outside the resolver -- so I am filing this as what it is, a second
live decider of one rule rather than a visible bug.

The reason for the fork is legitimate and worth keeping: `resolve_payload`'s inner `load()` reads
the blob with an unbounded `read_text` (artifact_v2.py:133), which is exactly the bound this PR set
out to impose. That argues for giving the owner a bounded-read hook -- a `read=` callable, say --
and calling it from here, rather than keeping a partial copy that starts out already differing on
two fields. As it stands the tree has three implementations of one wire rule (this one, the owner,
and server.js:1088) and a round-trip test that compares only two of them.

Smaller, same family: `_oversize_placeholder` (conversation.py:386) is a second placeholder shape
beside `artifact_v2.placeholder`, which this module uses correctly two lines away at :405.

**2. Every preview open renders the whole conversation twice** (raven/cli/trajectory_browse.py:535)

    if not _conversation_lines(records, width):      # rendered, tested for emptiness, discarded
        ...
    if _viewer_available():
        def _make_lines(...): return _conversation_lines(records, view_width, ...)
        if pviewer.view_lines(_make_lines, ...)      # rendered again

The first call exists to answer "is there anything to show", and pays for a full render to do it.
For a long attempt that is the whole cost twice on every open. Either keep the first result and
hand it to the viewer as the initial width's lines, or answer the emptiness question from
`records`.

For the record, what I checked hardest and found sound: the 512 KiB bound is applied where the
docstring says it is -- `_load_message_blob` routes every blob through `_read_artifact`
(conversation.py:402-403) rather than reading it directly, so a content-addressed inline image gets
the same bound as any other artifact; and the four test files this PR touches, plus the owner
module's own suite, are 248 passed on this head.

0xKT (approved):

PR #511, opened 2026-09-19
fix(rpc): let the first model choice land before a loop exists

gloryfromca:
## Summary

A new install cannot be finished from the web. Onboarding and the settings
page both write a provider key and then ask for a model, and that second call
is refused: `config.set key=model` asks the agent-loop factory for a loop
first, the factory builds from the config on disk, and a config with no model
has no provider to resolve, so it raises `missing_credentials` about the very
state this call was about to end. Neither surface says so -- the onboarding
step does not advance, and the settings role stays "Not set".

Reproduced against a gateway on an empty `RAVEN_HOME`, with the key already
written by `model.save_key`:

```
config.set {key: "model", value: "deepseek-chat", provider: "deepseek"}
  -> -32603 internal_error {reason: "missing_credentials", provider: ""}
setup.status -> {provider_configured: false}
```

A refusal for missing credentials is now read as "there is no loop yet",
which the rest of `_set_model` already handles: nothing is validated, nothing
is re-pointed, and the pair is persisted. Because nothing validated it, the
provider named in the request is put to the credential gate every other
surface asks -- through `ProvidersConfig`, since the file holds camelCase
(`apiKey`), a section can carry its credential in more than one shape, and a
key alone is not always enough -- so a keyless provider is still refused on a
fresh install, exactly as anywhere else.

Not introduced by the settings-page work: `refactor/ui_web_architecture` and
PR #494 do not touch `raven/rpc/methods/config.py`, `bootstrap.py` or
`setup.py`. The error text has been pointing at the CLI for the web's own
first run all along ("run `raven onboard` for guided setup").

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_rpc_config.py tests/test_rpc_settings.py -q` --
  115 passed. The two new cases fail without the change: the first-run write
  raises `InternalError`, and the keyless-provider case persists a pair it
  should refuse.
- The keyless case also failed while its fixture wrote `api_key`; the product
  writes `apiKey`, so the fixture now carries the shape `model.save_key`
  leaves behind and the gate is asked the real question.
- `uv run ruff check` and `ruff format --check` over the two files -- clean.
- `scripts/check_source_language.py github/main...HEAD` -- clean.
- Real host, empty `RAVEN_HOME`, `raven serve` on the branch: `model.save_key`
  then `config.set key=model` returns `applied: true` and writes
  `agents.defaults` = `deepseek/deepseek-chat` + `deepseek`; `setup.status`
  flips to `provider_configured: true`; the same call for a provider with no
  key is still refused with `model_not_available`.
- Same host through the browser: the onboarding flow (`?onboard=1`) completes
  welcome, provider, key, model, Finish, and the config carries the pair.

## Risk

- The change only widens what `config.set key=model` accepts, and only when
  the loop cannot be built for want of credentials. With a loop present every
  path is what it was, including the validation that builds the binding.
- A provider with no usable credential is still refused; the gate is the one
  `setup.status` uses, so the two cannot disagree.
- Rollback: revert the commit. The behaviour returns to refusing the first
  model choice, and the CLI (`raven onboard`) remains the way through.

## Related Issues

N/A


raven/rpc/methods/config.py
```diff
@@ -507,7 +569,7 @@ def _set_model(
     session_id, session_scoped = _session_scope(params, "model")
     has_session = session_id is not None
 
-    loop = agent_loop_factory() if agent_loop_factory is not None else None
+    loop = _loop_or_none_on_first_run(agent_loop_factory, raw_value, new_provider)
```
gloryfromca, line 573: Blocking: rebuild or restart the RPC stack after the first-run model is persisted.

This converts the latched startup failure into `None` only inside this handler. `build_rpc_stack` has already captured the original `build_error`, left `agent_loop` and `scheduler` unset, and its factory continues to re-raise that captured error. The handler therefore returns `applied: true` and the onboarding UI closes, but the next `turn.send` still takes the no-scheduler branch and emits the original startup error. I reproduced the sequence with a DeepSeek key: the set succeeded while `stack.agent_loop` and `stack.turn_scheduler` remained absent and `stack.build_error` remained `internal_error`. A fresh standalone `raven serve` still cannot chat until the process is restarted.

raven/rpc/methods/config.py
```diff
@@ -454,6 +455,67 @@ async def config_unset(params: dict) -> dict:
...
+    camelCase (``apiKey``), a section can carry its credential in more than one
+    shape (Gemini's ``apiKeyList``), and a key alone is not always enough
+    (Azure needs an address too). ``setup.py`` learned all three the hard way
+    and says so; this is the same question, so it is the same call.
+    """
+    from raven.config.schema import ProvidersConfig
+    from raven.providers.auth import credential_status
+
+    providers = _load_config().get("providers")
+    if not isinstance(providers, dict):
```
gloryfromca: Blocking: allow externally authenticated OAuth providers through the first-run gate.

The supported OAuth login flows store credentials in token files and do not create a raw `providers` object. `model.options` consequently reports providers such as `openai_codex` as authenticated through `credential_status(..., include_external=True)`, but this early return rejects the same state before `ProvidersConfig` or the external credential check is reached. With an empty raw config and a token-present stub, `credential_status` returned true while `_provider_has_credentials` returned false. Thus an ordinary OAuth first run still cannot persist its first model choice.

gloryfromca (commented):
Blocking: the successful write must leave the current RPC process able to run, and OAuth first runs must pass the external credential gate.

I reviewed the full diff plus the bootstrap/turn path, web onboarding and settings callers, provider credential handling, relevant history, backward compatibility, the repository rules and runtime vocabulary, and whether tests were weakened. The focused config/bootstrap/turn suites pass (122 passed); ruff, format, and diff checks are clean. The broader run including tests/test_rpc_settings.py could not complete in this checkout because the optional raven_everos package is unavailable (56 setup errors).

raven/rpc/methods/config.py
gloryfromca: Fixed the half that belongs here, and left the rebuild out on purpose.

You are right that the write landing is not the process being able to use it: the stack latched `build_error`, the scheduler was never created, and `turn.send` still takes the no-scheduler branch. The reply now carries `needs_restart` when the pair was persisted without a loop, so the caller learns it here instead of on its next send.

I did not rebuild the stack in place, because there is nothing to rebuild with on the path this bug is reached from. `request_swap` -- the guarded "build N+1, then swap" -- exists only in `raven gateway` (`cli/gateway_commands.py`); `raven serve`, which is what the web page runs against, builds its stack once in `cli/serve_commands.py` and holds no swap. And a late loop alone would not be enough: the wiring a turn needs (scheduler, cron, sinks, plugin services) runs inside `build_rpc_stack` under `if agent_loop is not None`, so making the factory retry would hand back a loop the stack is still not wired for.

Making a stack assemble its loop late is a change to how one is assembled, with its own blast radius across the three entry points that build one. I would rather that be a change of its own than a rider on a config handler. Happy to open it if you want it now.

raven/rpc/methods/config.py
gloryfromca: Fixed in 923a980f1. The gate is asked whether or not a section is on file, so a provider whose credential lives in a token file reaches `credential_status(..., include_external=True)` instead of being refused before it. Covered by a case that stubs the gate and asserts both the provider name and that the external half was asked, with the config carrying an empty `providers` block.

raven/rpc/methods/config.py
```diff
@@ -571,13 +634,23 @@ def _set_model(
...
     }
+    if loop is None and agent_loop_factory is not None:
+        # The write landed; the process serving it did not gain a loop. This
+        # gateway built without one -- that is why nothing validated the pair --
+        # and the wiring a turn needs (the scheduler above all) is put together
+        # once, at stack build. So the config is right and this process still
+        # cannot run a turn on it. Said in the reply rather than left for the
+        # caller to discover on the next send, which is where a first run used
+        # to end up without a word.
+        out["needs_restart"] = True
```
gloryfromca, line 664: Blocking: declare and consume the new restart signal before returning it.

`ConfigSetResult`, `rpc-schema/openrpc.json`, and both generated TypeScript clients do not contain `needs_restart`; the contract forbids extra properties. Validating the actual first-run result now raises `extra_forbidden` on this key. More importantly, the web onboarding path awaits `setModel` and discards its result, the settings path only reads `applies_to_session`, and the TUI response type cannot read this field, so no operator is told to restart and the next turn still reaches the latched startup error. The in-place rebuild can remain follow-up work, but this chosen alternative needs to be carried through the wire contract and surfaced by the model-setting callers.

gloryfromca (commented):
Blocking: `needs_restart` must be part of the RPC contract and reach a caller.

I reviewed the full branch and the delta from the prior revision, including the two prior findings, bootstrap and turn wiring, web/TUI model-setting callers, provider authentication, OpenRPC/Pydantic/generated response types, relevant history, backward compatibility, repository rules, runtime vocabulary, and whether tests were weakened. The OAuth gate is fixed, and the author has explicitly deferred live stack rebuilding; that architecture work is therefore a nonblocking follow-up rather than a standing objection here.

Verification: the focused config and schema suites pass (428 passed), and ruff, format, and diff checks are clean. An expanded run reached 534 passed and 6 failures unrelated to this diff: one missing optional `raven_everos` import and five memory fixture failures in `test_rpc_contract_shapes.py`.

raven/rpc/methods/config.py
gloryfromca: Blocking: the restart response still needs to be declared and consumed, but the in-place rebuild is accepted as follow-up work.

That scope split is reasonable: late stack assembly spans bootstrap wiring rather than this config handler. The original rebuild objection is closed here; the current-revision finding is limited to making the chosen `needs_restart` alternative a valid, user-visible RPC result.

raven/rpc/methods/config.py
gloryfromca: Blocking: the separate restart response still needs to be declared and consumed, but this OAuth finding is fixed.

The revised gate now reaches `credential_status(..., include_external=True)` without requiring an on-disk provider section, and the focused test passes. This thread is settled.

raven/rpc/methods/config.py
gloryfromca: Carried through in c00de04a1.

Declared where every other field is: `ConfigSetResult`, `rpc-schema/openrpc.json`, and both generated clients (regenerated, not hand-edited -- `gen:check` and the TUI's `lint:rpc` both report in sync). `tests/test_rpc_schema_match.py` and the contract-shape tests pass, so a real first-run result validates instead of raising extra_forbidden.

Consumed where the operator is: the onboarding step that awaited `setModel` and discarded it now reads the field, and on a first run it shows `gui.onb.restart` ("Saved. Restart the gateway to start chatting: it had no model when it started.", with the zh copy) instead of advancing to the done screen. A case pins it, and it fails on the surfacing being removed.

The settings page's roles card is on the frontend refactor branch rather than here; it reads the same field when that lands.

gloryfromca (commented):

The contract finding is fixed: the real first-run result validates through `ConfigSetResult`, OpenRPC and both generated clients are in sync, and onboarding consumes `needs_restart` and stays off the done screen while showing the operator what to do. No new finding survived review.

The explicitly deferred late stack rebuild and the legacy settings-page consumer remain named follow-up work; they do not block this change under the agreed scope.

I reviewed the full branch and revision delta, the prior discussion, Python and TypeScript callers, wire-schema/codegen compatibility, relevant history, repository rules, runtime vocabulary, backward compatibility, and whether tests were weakened. Verification: 428 Python config/schema tests passed; the web codegen and type checks passed; the onboarding suite passed 13/13; TUI RPC/i18n drift and type checks passed; ruff, format, and diff checks passed.

raven/rpc/methods/config.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

Verified on the current head: the actual first-run result now passes `ConfigSetResult.model_validate`, both codegen drift checks pass, the web and TUI type checks pass, and the onboarding test pins the restart message instead of the done state. This finding is fixed; the separately scoped settings-page work remains the named nonblocking follow-up.

0xKT:
**Not a blocker** -- two things. The first is the one write of the three that your commit message
does not account for; the second is a disagreement the new gate inherits.

**1. The staged pick was a toast on the base and is silence on this head**
(ui-web/src/live/080-overrides.js:124-134)

The commit body names "the two callers that set a model" and hands the settings roles card to
another branch. There is a third: a model picked while the composer is still a draft is staged
(120-settings.js:429-430), announced as staged -- `gui.model.pick_staged`, "Picked {name}; it
applies after you send the first message" -- and written by `applyStagedModel` when the first send
mints the session. That write is session-scoped, and session scope does not take the new field:

    staged pick  (session_id, no scope) -> {"applied": false, "previous": null, "scope": "session", ...}
    settings     (scope=default)        -> {"applied": true,  "needs_restart": true, "scope": "default", ...}

Both measured on this head in one process whose factory raises the stored `missing_credentials`
error the way bootstrap.py:284 does; the second line is the control -- the same harness, one
parameter apart, does produce the new field, so the first line is the scope and not my setup.

`applied: false` resolves. It does not reject. `applyStagedModel` has only `catch (e)` (:129-134),
so neither the toast at :132 nor the `loadProviders` reconcile at :133 runs, and the chip keeps a
model the session does not have. On the base the identical call raised out of
`agent_loop_factory()` -- the expression :573 replaces -- so that same click produced "Could not
switch - no provider is configured yet". The `{"applied": False}` branch at :601 is older than
this PR; what is new is that a live factory now reaches it.

To be fair about the size of it: the turn that follows still fails out loud (turn.py:393-400 emits
the build error when there is no scheduler), so the reader is not left with nothing -- what is lost
is the message about the model and the chip reconcile, on a path whose own comment at :130-131 says
"the pick was announced as staged, so a silent chip flip back would be an unexplained
contradiction".

Worth saying because of where a first run actually lands: the island that reads `needs_restart`
opens only under `?onboard=1` (200-boot.js:65-77, "The page itself stays available on first run"),
so the default route through a keyless gateway is the page and the Settings providers card -- whose
two writes are the deferred one and this one.

**2. The credential gate answers for a provider with nothing on file, and the same file gets the
opposite answer one method away** (raven/rpc/methods/config.py:481)

Measured on this head, `providers: {}`, no AWS variables in the environment:

    bedrock   -> applied=True needs_restart=True  on disk: agents.defaults.provider = 'bedrock'
                 setup.status on the same file    -> provider_configured=False
    deepseek  -> refused, ModelNotAvailableError  (control: the gate does refuse in this state)

`bedrock` declares `AuthMethod(KIND_AMBIENT, (), label="AWS credentials")` (auth.py:249) and an
empty requirement tuple makes `missing()` return `[]` whatever the section holds, so the gate
cannot answer anything but yes -- by design, since the AWS chain is not visible to it. The
divergence is on the other side: setup.py:95 asks `credential_status` only about sections that
exist (`section is not None`), so a credential held outside the config -- the ambient chain, and
the OAuth token file this PR's `include_external=True` was added for -- is invisible to it.
auth.py's own opening says why that matters: "an unstated shape is answered independently wherever
it is needed, and the answers diverge". The new call site obeys that; this is the first write path
on which the older reader's disagreement lands on disk.

For the record, what I checked hardest and found sound:

  - **An unvalidated model cannot brick the next start.** Nothing validates the pair when there is
    no loop, so I wrote `anthropic/no-such-model-anywhere-9999` through this path; it lands in
    `agents.defaults.model`. Building the loop on that config then succeeds, so the re-raise at
    :510 is never reached and `config.set` still works after the restart. Control, to show the
    probe can see a failure at all: the same build on a config whose model is not a string gives
    `InternalError reason='uncaught'`, which is the shape that would re-raise.
  - **The returns-None factory is not on this path.** `serve_commands.py:333` and
    `_gateway_page.py:142` hand `lambda: stack.agent_loop` to `build_app`, which stores it on the
    gateway (ws.py:512) where only `_workspace_root` reads it (ws.py:252-255). `config.set` takes
    the dispatcher's factory from bootstrap.py:284, which raises -- so the new gate is not bypassed
    there, and `needs_restart` is honest, because that factory re-raises a cached error for the
    life of the process.
  - **The tests bite.** Four reverts, four reds: the call site at :573 (5 failed), the credential
    gate at :512-516 (4 failed), `out["needs_restart"] = True` (1 failed), and the ui-web branch at
    OnboardPage.tsx:364 (1 failed). 440 passed across test_rpc_config / rpc_schema_match /
    rpc_registration / rpc_instances_declare_their_errors, 40 across the i18n and install-script
    files, 13/13 on the onboarding suite, and `npm run gen:check` ("generated.ts matches the
    contract (178 methods)") plus `type-check` clean -- so both generated clients are regeneration
    output rather than hand edits.

gloryfromca:
Point 1 is fixed in 2e7bd371d. You are right that it is the one write of the three the commit body did not account for, and right about the mechanism: the refusal resolves, `applyStagedModel` has only a catch, so neither the toast nor the `loadProviders` reconcile ran.

The refused session answer now carries `needs_restart` -- the same reason the default scope already reports, for the same cause: no loop to bind a session to. The recovery reads `applied === false` as the failure it is, says the restart line when the server sent a reason and a named refusal otherwise, and reconciles the chip either way. Two tests over the real `applyStagedModel` through the existing harness, plus one on the RPC; mutation-verified, each reverted separately: dropping the client branch reds both live cases on the missing toast, dropping `needs_restart` from the refused answer reds the RPC case.

Point 2 I am not fixing here, and I want to be explicit rather than quiet about it. You are right that `setup.py:95` asks `credential_status` only about sections that exist, so an ambient chain and an OAuth token file are invisible to it while the new call site can see both -- and that bedrock, with an empty requirement tuple, can only answer yes. That is a real divergence between two readers of the same file, and this PR is the first write path where it lands on disk. But the fix belongs in `_detect_provider_configured`, which every surface reads, not in this handler: making this one call site stricter to compensate would add a third answer to a question that already has two. Filed as follow-up rather than carried here, unless you would rather it block.

Verification on this head: `uv run --all-packages pytest tests/test_rpc_config.py tests/test_rpc_schema_match.py tests/test_rpc_contract_shapes.py -q` 478 passed; ui-web `npx vitest run` 108 files, 1839 tests passed; ruff check and format clean; TUI i18n catalogue regenerated and in sync.

raven/rpc/methods/config.py
```diff
@@ -534,8 +597,16 @@ def _set_model(
...
-            return {"applied": False, "previous": None, "value": raw_value, "scope": "session"}
+            # switch that did not happen. The reason is the one the default
+            # scope reports too, and it belongs here for the same cause: a
+            # refusal resolves, so a caller watching only for a raise hears
+            # nothing at all. A pick staged on a draft goes through this branch
+            # on a first run, and used to end as a chip showing a model the
+            # session does not have.
+            refused: dict[str, Any] = {"applied": False, "previous": None, "value": raw_value, "scope": "session"}
+            if agent_loop_factory is not None:
+                refused["needs_restart"] = True
```
gloryfromca: No blockers; suggestions only, and they are marked inline.

`needs_restart` is documented and rendered as meaning the write landed and a restart will apply it, but this session-scoped branch returns `applied: false` and persists nothing. A direct first-run probe returned this flag while `config.json` remained unchanged with no `agents.defaults`; after restart the gateway therefore has no selected model to load, despite the UI saying `Saved. Restart the gateway to start chatting`. Preserve the choice somewhere startup reads before returning this flag, or leave the flag absent and report an actionable refusal. This stays nonblocking because choosing the default through onboarding/settings is a working way out.

gloryfromca (commented):

The resolved-refusal handling now catches the staged session write and reconciles the chip, but the new `needs_restart` value overstates what happened: this branch persisted no model for a restarted process to consume. That is marked inline as a nonblocking correction because onboarding/default-model setup remains a working route. The separately filed `_detect_provider_configured` credential-status divergence remains a named follow-up and does not block this revision.

Covered the delta and full diff, the staged caller and restart contract, history, backward compatibility, test strength, and the repository rules/context. Verification passed: `uv run --all-packages pytest tests/test_rpc_config.py tests/test_rpc_schema_match.py tests/test_rpc_contract_shapes.py -q` (478 passed), focused web tests (29 passed), Ruff check/format, web codegen and type-check, TUI i18n sync, and `git diff --check`.

raven/rpc/methods/config.py
gloryfromca: Fixed in 84dd03957, and you are right about the contradiction: the flag says the write landed and a restart applies it, and on that branch nothing landed, so the page promised a restart that comes back to a gateway with no model still.

Of the two ways out you named I took the second. Persisting the pick would have made the flag honest, but it would also widen a choice the reader made for one conversation into the default every new conversation starts on -- the staged pick is session-scoped by construction, and that is the scope the draft was picked under.

So the refusal travels on its own and the recovery names the way out: "set a model in Settings first - this gateway started without one". The `needs_restart` absence is now pinned rather than incidental (`assert "needs_restart" not in result`), and the live case asserts the toast and the chip reconcile; mutation-verified by removing the client branch, which reds it on the missing toast.

Verification: `uv run --all-packages pytest tests/test_rpc_config.py tests/test_rpc_schema_match.py tests/test_rpc_contract_shapes.py -q` 478 passed; ui-web `npx vitest run` 108 files, 1838 tests passed; ruff clean; TUI i18n catalogue in sync.

gloryfromca: No blockers; suggestions only, and they are marked inline.

Confirmed on this head: the first-run session result is now `applied: false` without `needs_restart`, the config remains unchanged, and the live recovery reports the refusal and reloads the chip. That settles the false restart promise. The guidance still hands the next default write to the already named Settings restart-consumer follow-up; that remains nonblocking under the recorded scope decision.

gloryfromca (commented):

The false restart signal is fixed: the session-scoped refusal no longer carries `needs_restart`, its caller reports the refusal and reconciles the model chip, and a direct first-run probe confirms the config remains unchanged. No new findings on this revision. The already recorded Settings restart consumer, late stack assembly, and shared provider-status divergence remain nonblocking follow-ups under the scope decisions on the change.

Covered the full diff and current delta, the RPC contract and callers, history, backward compatibility, test strength, and repository rules/context. Verification passed: 478 RPC/schema tests, 28 focused live-model tests, Ruff check/format, both TypeScript checks, web codegen, TUI i18n sync, and `git diff --check`.

gloryfromca (commented):

This head has the same tree object as the reviewed `84dd03957552`; only the co-author trailer was normalized. The code diff, generated outputs, callers, compatibility, tests, and repository-rule coverage are therefore unchanged, and `git diff --check github/main...HEAD` remains clean. No new findings. The previously named Settings restart consumer, late stack assembly, and shared provider-status divergence remain nonblocking follow-ups.

raven/rpc/methods/config.py
0xKT, line 1: **Blocking:** the same defect the delta fixes on the draft path is still live on the ordinary one.
`persistModel`'s live-session branch discards the reply, so a refused switch leaves the chip moved
and the page saying "Switched to X".

First, my own share of this: it was live at c00de04a too, and I passed that head. The note I left
there named the default-scope surface and the staged pick; I did not follow the third caller. So
this is not new to this revision -- it is the reader the fix did not reach, and I should have
found it a revision ago.

**The two branches, side by side.** The taught one, from this delta:

    ui-web/src/live/080-overrides.js:128-137   const r = await rpc.call('config.set', {... session_id})
                                               if (r && r.applied === false) { toast(...); loadProviders(...) }

and the one three lines further up the same file's sibling, untouched:

    ui-web/src/live/120-settings.js:425-428    if (sid) {
                                                 await rpc.call('config.set', {... session_id: sid});
                                                 return;                      <- the reply is dropped
                                               }

Both send a session-scoped write. Only the first reads the answer.

**What the reader sees.** `choose()` moves the chip before the round trip and rolls it back only in
`catch` (ui-web/src/features/model/store.ts:145, :157-161). `persistModel` returns `undefined` on
that branch, which is not `'staged'`, so :154 toasts `gui.model.pick_switched` -- "Switched to
{name}". A refusal that RESOLVES never reaches the catch. The tree already pins this outcome: the
picker's own test installs a `persist` that resolves undefined
(ui-web/src/features/model/ModelPicker.test.tsx:52) and asserts `store.current()` is the new model
with `toasts == ['gui.model.pick_switched ...']` (:291-305). That test is green, and the shipped
session branch resolves exactly that way whatever `applied` says.

**It is this PR's doing.** At base the same call raised out of `agent_loop_factory()` -- the
expression :573 replaces -- so the catch fired, the chip rolled back, and the reader got "Could not
switch". On this head the server answers `{"applied": false, ...}` and resolves
(raven/rpc/methods/config.py:613). Measured on this head, one process, two scopes:

    session_id, no scope  -> {"applied": false, "previous": null, "scope": "session"}
    scope=default         -> {"applied": true,  "needs_restart": true, "scope": "default"}

**And the rule is the one you wrote into this delta**, at config.py:610-612: "The refusal is what
the caller gets, and it is the caller's job to say it -- a refusal resolves, so watching only for a
raise leaves a chip on a model the session does not have." 120-settings.js:425-428 is a caller that
does not say it.

**Reachable, though not on a bare first run.** The picker only opens for a provider whose section
lists models (`offered = p.configured ?? p.models`, types.ts:41, fed by 120-settings.js:147 where
`configured` is always an array, so `??` never falls through to `p.models`), and a fresh install has
none. The state that satisfies everything at once is an existing install whose default provider's
credential is gone -- key rotated out, say -- while another provider still has a key and models
added. The gateway then starts loop-less, `session.create` and `session.resume` still work because
they go through `_safe_invoke_factory`, which swallows the raise
(raven/rpc/methods/session.py:66-80), and the reader is in a live conversation with a working
picker.

What closes it: the three lines you already wrote, on the other branch. The reply is in hand at
120-settings.js:426; `r.applied === false` there wants the same toast and the same
`loadProviders(sid, gen)` reconcile -- and `choose()` needs the chip put back, which today only its
catch does, so either that branch throws or `persistModel` reports the refusal to it.

For the record, what I verified and found sound on this head: the draft-path fix bites (removing
`if (r && r.applied === false)` fails "speaks and reconciles when the refusal RESOLVES"); dropping
`needs_restart` from the refusal is right and pinned
(`test_a_first_run_session_pick_says_why_it_was_refused` fails if it is put back); there is exactly
one `applied: false` return and it is reachable only when the cached build error was
`missing_credentials`, so the reason `gui.model.refused` asserts is sound for a first run; both
locales render as one sentence; the ui-tui i18n and RPC generators reproduce their committed output
byte for byte and `tsc` is clean on both trees; ui-web `gen:check` matches the contract at 178
methods; 65 passed in tests/test_rpc_config.py and 41 in the two ui-web suites.

gloryfromca: Fixed in ccc3f601c. The finding holds and the shape of my mistake is the one worth naming: I taught one caller to read `applied: false` and did not enumerate the others. Three callers send a session-scoped model write; the enumeration belonged before the first fix, not after you found the second reader. That is the rule I had just written into the server comment, applied to two of three places.

`persistModel`'s live-session branch now reads the reply and throws `{ data: { detail: T('gui.model.refused') } }` when the write was refused. Thrown rather than reported, for the reason you gave: the rollback it needs is the one `choose`'s catch already does, and `data.detail` is the shape that catch renders -- `ModelPicker.test.tsx:322` already pins "forward then back" plus `gui.op.switch_failed {"detail":...}` for exactly that shape, so the two layers meet without a new path.

Pinned at the layer that was wrong: `the live-session write > rejects a refusal, because only the caller's catch puts the chip back` drives the real `persistModel` through the existing harness, settles `{applied: false}`, and asserts the rejection carries `gui.model.refused` and that `modelSet` was never called. A second case pins that an applied write still resolves undefined. Mutation-verified: dropping the throw reds the first with "expected null to be truthy". The harness needed `T` to build the function, which it now supplies the way `stagedHarness` does.

On your own share -- noted, but it does not move much of the weight. The reader that got the wrong answer is a caller of the code I changed, and finding my callers is my job, not the reviewer's.

Verification on this head: ui-web `npx vitest run` 108 files, 1840 tests passed; `npm run type-check` clean.

gloryfromca: Addendum -- the enumeration I owed you, done now rather than after a third round. Every caller that writes a model through `config.set`, and what each does with the answer:

    080-overrides.js:128   session (staged draft)   reads applied -> toast + loadProviders reconcile
    120-settings.js:426    session (live pick)      reads applied -> throws, choose's catch rolls back
    200-boot.js:9          default (onboarding)     reads needs_restart -> shows the restart line

There is no fourth. `persistModel` itself has exactly two references -- `store.ts:149` inside `choose`, which is the try/catch, and the wiring at `120-settings.js:480` that hands it over as `persist` -- so the throw has one consumer and that consumer handles it.

The two other `config.set` calls in the live layer write `permissions.mode`, not a model, and are outside this change.

0xKT: **Re-verified fixed** on ccc3f601, merged onto the current tip 1d80cd4f.

The live-session branch now reads the reply and throws the shape `choose()` already knows how to
handle (ui-web/src/live/120-settings.js:426-433):

    const r = await rpc.call('config.set', {... session_id: sid});
    if (r && r.applied === false) throw { data: { detail: T('gui.model.refused') } };

Throwing rather than reporting is the right call and better than what I asked for: the rollback
this needs is the one the catch already does, so the chip goes back by the path that was always
responsible for it instead of a second one. The tree pins the whole chain end to end, and both
halves were already there:

  - the new test rejects with `data.detail == 'gui.model.refused'` and asserts no `modelSet`
    happened (ui-web/scripts/model-refresh-live.test.mjs:408-419). It drives the SHIPPED
    `persistModel` -- the harness builds it out of the real file -- so it cannot pass while
    120-settings.js differs;
  - `choose()`'s side was already pinned by a test that rejects with exactly that shape:
    ui-web/src/features/model/ModelPicker.test.tsx:322-334 asserts the chip goes forward then back
    (`['minimax-m2', 'minimax-m3']`) and the toast is `gui.op.switch_failed {"detail": ...}`.

So the reader now sees "Could not switch - set a model in Settings first - this gateway started
without one", with the chip back on the model the session actually holds.

Measured on the merged tree: removing the one new line turns "rejects a refusal, because only the
caller's catch puts the chip back" red, and nothing else. 77 passed across
scripts/model-refresh-live.test.mjs, src/features/model and src/features/onboard; 439 passed across
tests/test_rpc_config.py, test_rpc_schema_match.py and test_rpc_registration.py;
`GATES OK head=ccc3f601 tip=1d80cd4f base=3dd06f6e ruff/lint-imports/commit/large/lang rc=0`.

The tip moved while I was reviewing (3dd06f6e -> 1d80cd4f, which is #503 landing). Its two files
are README.md and README.zh-CN.md and neither is in this PR's fourteen, so every measurement above
is about the tree you would get.

gloryfromca (commented):

The live-session caller now reads `applied: false` and rejects into the model picker’s existing catch, which restores the optimistic chip and reports the refusal. The staged-session and onboarding callers remain correct, and no new finding survived review. The separately scoped default-Settings restart consumer, late stack assembly, and shared provider-status divergence remain the previously named nonblocking follow-ups.

Covered the current delta and full diff, all model `config.set` callers, the picker rollback contract, history, compatibility, test strength, and repository rules/context. Verification passed: 478 RPC/schema tests, the complete web suite (108 files, 1,840 tests), web type-check and codegen, plus `git diff --check`. The web suite emitted known happy-dom aborted-fetch teardown diagnostics but had zero failures.

gloryfromca:
Superseded by #524, on `refactor/ui_web_architecture`, and closed for a reason that is this PR's own fault line rather than a preference.

Two of the three frontend files this touches -- `ui-web/src/live/120-settings.js` and `ui-web/src/live/080-overrides.js` -- do not exist on the refactor branch, which deletes `ui-web/src/live/` whole. Landing the fix here would have meant a modify/delete conflict when that branch merges, and the frontend half being deleted rather than carried: the same two defects live on in the files that replace them (`features/model/source.ts:169`, `state/session/runtime.ts:452`), so the deadlock's visible consequence would have come back with the merge. The refactor branch merges today, which settles the timing question that made a main-side fix worth considering at all.

Everything here is on #524: the RPC half applied cleanly (`config.py` and `test_rpc_config.py` are byte-identical on both branches), the contract change was made by hand because `config.set` declares its result inline there rather than through a named component, the onboarding change applied cleanly, and both generated clients were regenerated rather than ported. The reply-reading half is written against that branch's own files.

Everything the reviews here established stands, and the moved code is the code they were run against. Thanks for the four rounds on it -- the staged-pick regression, the false restart promise, and the third caller were all found here.

PR #512, opened 2026-09-19
fix(config): let a borrowed address follow the lender it came from

gloryfromca:
## Summary

An EverOS role points a section at a provider: the settings card writes the
model and asks `lend_provider_credentials` for that provider's key and
address to go beside it. The address was copied only when there was one to
copy -- the reasoning being that a section with no lender address may be
holding one the reader typed by hand.

A section being pointed at a new lender is holding the OLD lender's address,
though, and leaving the field out of the answer keeps it. Moving the memory
role from an OpenRouter model to a DeepSeek one left DeepSeek's key at
OpenRouter's address, which the far end refuses. DeepSeek is not a corner
case here: every vendor LiteLLM routes carries no `default_api_base`, so the
lend answers with no address for all of them.

The lender answers about the address either way now, empty included. An empty
value reads where it lands the way an unset one does: the EverOS writer sets
the field, and the section stops carrying somebody else's endpoint.

Found while running the settings page's acceptance case for the EverOS roles
against a real gateway. The lend rule is from #414, not from the settings
work that surfaced it.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_config_update_providers.py tests/test_rpc_settings.py
  tests/test_cli_onboard_commands.py tests/test_rpc_console.py -q` -- 641 passed
  on the branch this was written on; 191 on the two closest suites here.
- The new case fails without the change: the lender answers `{"api_key": ...}`
  where the section needs `{"api_key": ..., "base_url": ""}`.
- `uv run ruff check` and `ruff format --check` over both files -- clean.
- Real gateway, temp `RAVEN_HOME`, EverOS plugin present: `settings.everosSet`
  with `borrow_from: "deepseek"` writes the model and DeepSeek's key and
  leaves no `base_url` behind; the same call with `borrow_from: "openrouter"`
  writes OpenRouter's address back. Before the change the first of those left
  `base_url = "https://openrouter.ai/api/v1"` under DeepSeek's key.

## Risk

- Two callers, both of which point a section at a provider: the EverOS card
  (`settings.everosSet`) and the CLI onboarding reuse step. For a lender that
  has an address, nothing changes. For one that has none, the borrowing
  section now stops carrying the previous lender's -- which is the fix.
- A reader who typed an address by hand into an EverOS section and then
  borrowed from an addressless provider loses that typed value. That is the
  same event as the bug, read the other way; the section is being pointed
  somewhere else, and the card shows the resolved address.
- Rollback: revert the commit.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the full diff and traced both callers: the settings RPC passes the explicit empty `base_url` into the EverOS writer, where it deletes the stale field, while CLI onboarding deliberately falls back to its curated provider URL. I also checked the lending and writer history, backward compatibility for lenders with configured/default addresses, test changes for weakening, and the applicable AGENTS.md/CLAUDE.md and CONTEXT-MAP.md/CONTEXT.md constraints.

Verification: `uv run --all-packages pytest tests/test_config_update_providers.py tests/test_rpc_settings.py tests/test_cli_onboard_commands.py -x` passed (523 tests; one non-failing asyncio cleanup warning). Ruff check/format, commit-message checks, source-language checks, and `git diff --check` passed. The first plain `uv run pytest` attempt could not collect the plugin-dependent suites because `raven_everos` was not installed; rerunning with all workspace packages supplied the plugin and completed successfully.

raven/config/update_providers.py
0xKT, line 1: **Blocking:** the empty address is not read as "unset" by the consumer -- it is read as "delete" --
and for the two vendors this file already knows an address for, including the one in the incident,
the borrow now leaves the role with no address at all.

The fix is right about the bug. Keeping the lender's old address with a new vendor's key is worse
than what follows, and I could reproduce that incident on the base. What this changes is which
wrong state you land in.

**The empty value is not equivalent to leaving the field out.** Three links, all in production
code:

    raven/config/update_providers.py:1878   return {"api_key": ..., "base_url": base_url}   # "" possible
    raven/rpc/methods/console.py:1227       clean.update(lent)
    plugins-dist/everos-memory/raven_everos/config.py:554-556
                                            if value == "": merged.pop(key, None)

`console.py:1227` is the load-bearing step: the same function filters empty values out of the
user's own fields thirty lines earlier (`:1199-1200`, `if v: clean[k] = v`), and the borrowed
mapping goes in through `update` without that filter. So `""` reaches the writer, and the writer
deletes the key.

**The file already holds the address this now drops.** `lend_provider_credentials` resolves
`lent.api_base or spec.default_api_base` and stops. Measured on this head:

    deepseek     registry default_api_base = ''    _PROVIDER_BASE_URL_FALLBACK = 'https://api.deepseek.com/v1'
    openai       registry default_api_base = ''    _PROVIDER_BASE_URL_FALLBACK = 'https://api.openai.com/v1'
    openrouter   registry default_api_base set     (so the borrow works for this one)
    groq, zai    neither

`_PROVIDER_BASE_URL_FALLBACK` is at raven/config/update_providers.py:1107, 770 lines above this
change, under the comment that says why it exists: "they rely on the SDK's built-in default, which
a bare OpenAI client doesn't know. A bare client needs an explicit base_url." It has two readers
already -- `:1182` and `:1957` -- and this function is not one of them.

The consumer of the section really is a bare OpenAI client: plugins-dist/everos-memory/
raven_everos/onboard.py:118-119 answers `"no base_url configured"` and stops when the field is
empty.

So the incident this PR was written for -- a role moved from an OpenRouter model to a DeepSeek one
-- now ends with DeepSeek's key and no address, in a tree that knows DeepSeek's address.

**And one case loses with no upside.** A role already pointing at a hand-typed gateway, whose key
is refreshed by borrowing from a lender that has no address of its own: on the base the address
survived, because the field was left out; on this head it is deleted. Nothing about that action
concerned the address. That is the case the comment you removed was protecting, and it is the
reason I am filing this as blocking rather than as a note: it turns a working configuration into a
broken one through an unrelated edit.

**The new test does not reach any of this.** `test_borrow_hands_back_an_empty_address_when_the_
lender_has_none` asserts `lend_provider_credentials("openai") == {"api_key": "sk-lend", "base_url":
""}` -- at the function boundary, on one of the two vendors the fallback table has an address for,
and it never follows the value to `merged.pop`. It pins the shape, not the outcome.

What would close it, smallest first:

  - add the fallback the file already has: `lent.api_base or spec.default_api_base or
    _PROVIDER_BASE_URL_FALLBACK.get(provider, "")`, which is what `_bare_openai_settings` at :1957
    already does through the same contract object. Then the answer is the address for
    deepseek/openai/openrouter, and `""` is left for vendors nobody has an address for;
  - and for those, decide deliberately what `""` should mean at console.py:1227 -- deleting the
    section's address is a choice worth making on purpose rather than inheriting from the writer's
    empty-string convention.

A test that follows one borrow through `settings.everosSet` to the written `everos.toml` would pin
whichever answer you choose; the current one cannot.

gloryfromca: Blocking: borrowing DeepSeek credentials deletes the required usable address.

The finding holds. Reproducing the full `settings.everosSet` path on this exact head, starting with an OpenRouter LLM section and borrowing a configured DeepSeek key, returned `{"applied": true}` but wrote only `model = "deepseek/deepseek-chat"` and `api_key = "sk-deepseek"`; `base_url` was absent. EverOS 1.2.3 rejects an LLM configuration without `base_url`, while Raven already has `https://api.deepseek.com/v1` in `_PROVIDER_BASE_URL_FALLBACK`.

So this revision replaces the stale-address failure with a missing-required-address failure for the incident path. `uv run --all-packages pytest tests/test_config_update_providers.py tests/test_rpc_settings.py tests/test_cli_onboard_commands.py -x` still passes (523 tests), which confirms the current unit test does not cover the persisted result. The correction needs to carry Ravens known fallback through the lending path and add an end-to-end RPC regression that asserts the written section remains usable; the semantics for vendors with no known address can be decided separately.

gloryfromca: Fixed in 42ce2b93e. The finding holds and the premise I wrote into the comment was wrong: I checked all three links you named, and `clean.update(lent)` at console.py does bypass the empty-value filter thirty lines above it, so `""` reaches the writer and `merged.pop` deletes the key.

The lending path now resolves the address through the same three steps `resolve_provider_credentials` uses, `_PROVIDER_BASE_URL_FALLBACK` included, keyed through `canonical_provider_name`. DeepSeek, OpenAI and OpenRouter now answer with their own address; `""` is left only for a vendor no table here knows, and the comment says why deleting is the deliberate answer there rather than an inherited convention: what stands in the section was reached with somebody else's key, so a loud "no base_url configured" beats sending this key to that address.

On the test: you were right that it pinned the shape and not the outcome. It is replaced by one that names the address, and the regression you asked for now exists -- `test_the_borrowed_address_replaces_the_lenders_predecessor` writes an `[llm]` section holding OpenRouter's address, borrows a keyed DeepSeek through `settings_everos_set`, and reads the written everos.toml. Mutation-verified: removing the fallback line makes it fail with `Right contains 1 more item: {'base_url': 'https://api.deepseek.com/v1'}`, which is the symptom you reproduced.

Verification: `uv run --all-packages pytest tests/test_config_update_providers.py tests/test_rpc_settings.py tests/test_cli_onboard_commands.py` 525 passed; ruff check and format clean; `git diff --check` clean.

gloryfromca (commented):

The prior blocker is fixed on this revision. The lender now uses the same configured address, registry default, and canonical-name fallback precedence as the existing credential resolver. A direct reproduction of the OpenRouter-to-DeepSeek transition writes DeepSeek's key together with `https://api.deepseek.com/v1`, and the new RPC regression follows the value through `settings.everosSet` into `everos.toml`.

Reviewed the complete diff and the delta from the prior head, both callers, relevant resolver/writer history, backward compatibility, test integrity, and the applicable AGENTS.md/CLAUDE.md and CONTEXT-MAP.md/CONTEXT.md constraints. The remaining empty-address case is explicit for lenders with no usable address known to Raven and fails closed instead of retaining another lender's endpoint.

Verification: `uv run --all-packages pytest tests/test_config_update_providers.py tests/test_rpc_settings.py tests/test_cli_onboard_commands.py -x` passed (525 tests). Ruff check/format, commit-message checks, source-language checks, and `git diff --check` passed. The existing open review thread was started by another reviewer, so it was not resolved here.

raven/config/update_providers.py
0xKT: **Re-verified fixed** on 42ce2b93. Both halves of what I asked for are here, and I measured them
rather than reading them.

The three links I named now end differently. `lend_provider_credentials` resolves
`lent.api_base -> spec.default_api_base -> _PROVIDER_BASE_URL_FALLBACK[canonical_provider_name(...)]`
(update_providers.py:1886-1891), so on this head, with a key on file and nothing else:

    deepseek    {"api_key": "sk-deepseek",   "base_url": "https://api.deepseek.com/v1"}
    openai      {"api_key": "sk-openai",     "base_url": "https://api.openai.com/v1"}
    openrouter  {"api_key": "sk-openrouter", "base_url": "https://openrouter.ai/api/v1"}
    groq / zai  {"api_key": ...,             "base_url": ""}

End to end through `settings_everos_set` into the written everos.toml, both borrows in one run
against a section already holding a hand-typed gateway address:

    borrow_from=deepseek  -> base_url = 'https://api.deepseek.com/v1'
    borrow_from=groq      -> base_url = <DELETED>      <- control: the delete is the lender,
                                                          not the harness

So the incident the PR was written for now lands on DeepSeek's own address, and the case I filed
as blocking -- a role pointed at a hand-typed gateway losing its address to an unrelated key
refresh -- is gone for every vendor a table here knows.

It remains for groq, zai and any vendor no table knows, and that is now a written decision rather
than an inherited convention: "what stands there was reached with somebody else's key, so a loud
'no base_url configured' beats sending this key to that address" (update_providers.py:1882-1885).
I would have weighed the hand-typed-gateway case the other way, but the decision is reasoned, it
fails loudly instead of misrouting, and it is recorded where the next reader meets it. That is the
second bullet of what I asked for, so it closes.

The regression exists and it bites. `test_the_borrowed_address_replaces_the_lenders_predecessor`
(tests/test_rpc_settings.py:520) writes an `[llm]` section holding OpenRouter's address, borrows a
keyed DeepSeek, and asserts the whole written table -- so a dropped key fails it, which the old
boundary assertion could not do. Removing the one fallback line turns two tests red:

    tests/test_config_update_providers.py::test_borrow_answers_with_the_address_raven_knows_for_the_lender
    tests/test_rpc_settings.py::test_the_borrowed_address_replaces_the_lenders_predecessor

193 passed across tests/test_config_update_providers.py and tests/test_rpc_settings.py on this
head. One thing worth saying because it looks alarming and is not this PR: through the board
runner, tests/test_cli_onboard_commands.py reports 34 failures on this tree -- and the identical
34, by name, on an unrelated PR's tree that does not touch this file. They are the runner's
plugin entry point, not your change.

gloryfromca (commented):

This revision has the same source tree as the previously reviewed `42ce2b93eeb4`; the range-diff changes only the latest commit's co-author trailer to the repository-required format. Tree identity and an empty `git diff 42ce2b93eeb4 HEAD` confirm that the previously verified behavior and 525-test result apply unchanged. `scripts/check_commit_messages.py` and `git diff --check` pass. The prior review thread has been resolved by its opener.

raven/config/update_providers.py
0xKT, line 1: **Blocking:** the case I closed as a written decision is the one I can now show is not announced.
For groq and zai the borrow still deletes a working address, and nothing tells the reader -- the
role keeps reporting itself configured.

I closed the earlier thread on two grounds: the table now carries the address for every vendor it
knows, and for the rest the delete is a deliberate choice that "fails loudly". The first holds.
The second is the half I did not check, and it is wrong at the layer the reader actually looks at.

**Measured on this head, end to end through `settings_everos_set`, against a section already
holding a hand-typed address**, both borrows in one run:

    borrow_from=groq      base_url = <DELETED>                       role_configured_in -> True
    borrow_from=deepseek  base_url = 'https://api.deepseek.com/v1'   role_configured_in -> True

The second line is the control: the vendor the table knows keeps its address. The right-hand
column is the point -- the predicate cannot tell the broken section from the working one. It is
`bool(sec.get("model") and sec.get("api_key"))` (raven_everos/config.py:507), and the web LED is
the same question, `const on = !!(cur.model && cur.api_key_set)`
(ui-web/src/features/settings/SettingsPage.tsx:2250). Neither reads `base_url`. A third control,
to show the predicate is live rather than stuck on: with the key removed it answers False.

So "a loud `no base_url configured`" is true only of `onboard.py:118`, which runs when something
actually calls the role. Until then the section reports itself configured, with no address.

**Who this reaches.** The lender list is
`providers.filter(pv => pv.on && pv.kind !== 'oauth' && pv.kind !== 'local')`
(SettingsPage.tsx:2332-2334), so every authenticated key provider is offered, groq and zai
included. The scenario is the one the old comment protected and the one I filed the first time: a
role pointed at an address the reader typed, whose key is refreshed by borrowing. On the base the
address survived because the field was left out. On this head it is deleted, by an action that
concerned only the key, and the page goes on saying the role is set up.

**Why I am not asking you to read `display_api_base`.** I checked that first and it is the wrong
field: registry.py:76-88 says it is the settings pane's value and "read through `display_api_base`,
never directly", and the registry states the hand-out answer separately --
`usable_default_api_base` is `""` for groq and zai on purpose, pinned by
tests/test_provider_registry_data.py:151 with the reason "would now send its own base to the
router". Promoting a display address into operational use is the mistake that comment exists to
prevent.

What closes it, smallest first:

  - two entries in `_PROVIDER_BASE_URL_FALLBACK` (:1105-1108), which is the table for exactly this
    shape -- a vendor LiteLLM reaches by key whose address a bare OpenAI client does not know.
    groq is `https://api.groq.com/openai/v1` and zai is `https://api.z.ai/api/paas/v4`; both are
    already in the tree as `shown_api_base` (registry.py:1163, :805), so this adds no new fact,
    it copies one the file already states into the table the file already reads. That is the same
    move that closed deepseek and openai;
  - or, if the delete is to stand for the remainder, make it visible where the reader looks:
    `role_configured_in` and the LED both answer a question whose answer is "yes" for a section
    that cannot be called.

A test in the shape of `test_the_borrowed_address_replaces_the_lenders_predecessor` but with groq
as the lender would pin whichever you choose; nothing covers that vendor set today.

For the record, the rest of this revision re-verified clean on e2b4a0d9: the head moved from
42ce2b93 while I was reviewing, and `git merge-tree --write-tree` gives the identical merged tree
d25d2e9c for both, with `git range-diff` showing only the `Co-authored-by` trailer changed -- so
the earlier measurements stand. 193 passed across tests/test_config_update_providers.py and
tests/test_rpc_settings.py; removing the fallback line turns two tests red.

One nit while I am here: `test_borrowing_keeps_the_section_url_when_the_lender_has_none`
(tests/test_rpc_settings.py:504) is now named for behaviour this PR removes -- its lender `custom`
carries a registry default, so what it pins is "some truthy address is present", not "the section
kept its own". The assertion is not vacuous (a blank address makes it fail), but the name is
misleading about the new semantics.

gloryfromca: Blocking: Groq and Z.ai borrowing must not silently leave an unusable role reported as configured.

The finding holds. On this exact head, full `settings_everosSet` calls for both Groq and Z.ai returned `{"applied": true}`, deleted the existing hand-typed `base_url`, and left `role_configured_in(...) == True`; the DeepSeek control retained `https://api.deepseek.com/v1`. EverOS rejects the addressless role only when it is later used, so the save and the settings status provide no loud failure.

`uv run --all-packages pytest tests/test_config_update_providers.py tests/test_rpc_settings.py tests/test_cli_onboard_commands.py -x` still passes (525 tests), confirming the current regressions do not cover these offered lenders. The provider fallback or the configured-state/reporting contract needs to cover this path before the revision is clean.

gloryfromca: Fixed in 283bc74a8, by the first of the two ways out.

The half I got wrong is the one you named: I wrote "fails loudly" into the code as the justification and never checked the layer that renders the state. `role_configured_in` is `bool(sec.get("model") and sec.get("api_key"))` and the LED asks the same question -- neither reads `base_url` -- so a section with no address reports itself set up until something calls the role. Asserting a property of the reader without asking the reader is the same mistake twice in one file.

groq and zai now carry `https://api.groq.com/openai/v1` and `https://api.z.ai/api/paas/v4`, copied from the `shown_api_base` the registry already states, and the comment records why it is that field and not `display_api_base` or `usable_default_api_base` -- the latter is empty for both on purpose so neither sends its own base to the router, which is a different question from what a bare client must be told.

Pinned in the shape you asked for: `test_a_vendor_the_table_knows_only_by_its_shown_address_still_lends_one` writes an `[llm]` section holding a hand-typed gateway, borrows a keyed groq through `settings_everos_set` and reads the written everos.toml. Mutation-verified -- removing the two entries reds it with `KeyError: 'base_url'`, which is your `<DELETED>`.

On the remainder: after this, what answers "" is two kinds and neither is an oversight -- a vendor on another protocol (anthropic, gemini), and a deployment whose address only its operator knows (azure_openai, hosted_vllm, gpustack). The unit test moved off groq onto `gpustack` and says which kind it stands for, and the code comment now says a vendor arriving in that branch wants a table entry instead of it. I removed the "fails loudly" claim rather than rewording it.

Your nit is fixed too: `test_borrowing_keeps_the_section_url_when_the_lender_has_none` is now `test_a_borrow_hands_over_an_address_as_well_as_a_key`, named for what it pins.

Verification: `uv run --all-packages pytest tests/test_config_update_providers.py tests/test_rpc_settings.py tests/test_cli_onboard_commands.py tests/test_provider_registry_data.py -q` 550 passed; ruff check and format clean.

gloryfromca: Fixed in 283bc74a8, by the first of the two ways out.

The half I got wrong is the one you named: I wrote "fails loudly" into the code as the justification and never checked the layer that renders the state. `role_configured_in` is `bool(sec.get("model") and sec.get("api_key"))` and the LED asks the same question -- neither reads `base_url` -- so a section with no address reports itself set up until something calls the role. Asserting a property of the reader without asking the reader is the same mistake twice in one file.

groq and zai now carry `https://api.groq.com/openai/v1` and `https://api.z.ai/api/paas/v4`, copied from the `shown_api_base` the registry already states, and the comment records why it is that field and not `display_api_base` or `usable_default_api_base` -- the latter is empty for both on purpose so neither sends its own base to the router, which is a different question from what a bare client must be told.

Pinned in the shape you asked for: `test_a_vendor_the_table_knows_only_by_its_shown_address_still_lends_one` writes an `[llm]` section holding a hand-typed gateway, borrows a keyed groq through `settings_everos_set` and reads the written everos.toml. Mutation-verified -- removing the two entries reds it with `KeyError: 'base_url'`, which is your `<DELETED>`.

On the remainder: after this, what answers "" is two kinds and neither is an oversight -- a vendor on another protocol (anthropic, gemini), and a deployment whose address only its operator knows (azure_openai, hosted_vllm, gpustack). The unit test moved off groq onto `gpustack` and says which kind it stands for, and the code comment now says a vendor arriving in that branch wants a table entry instead of it. I removed the "fails loudly" claim rather than rewording it.

Your nit is fixed too: `test_borrowing_keeps_the_section_url_when_the_lender_has_none` is now `test_a_borrow_hands_over_an_address_as_well_as_a_key`, named for what it pins.

Verification: `uv run --all-packages pytest tests/test_config_update_providers.py tests/test_rpc_settings.py tests/test_cli_onboard_commands.py tests/test_provider_registry_data.py -q` 550 passed; ruff check and format clean.

gloryfromca (commented):

The Groq/Z.ai blocker is fixed on this revision. Both providers now resolve through the existing bare-client fallback table; direct end-to-end checks through `settings.everosSet` persisted `https://api.groq.com/openai/v1` and `https://api.z.ai/api/paas/v4` respectively, while keeping the role configured. The unknown-address fixture now uses GPUStack, whose address is operator-defined and cannot have a global fallback.

Reviewed the full branch diff and the delta from `e2b4a0d9992a`, the RPC and onboarding callers, resolver/registry semantics, backward compatibility, test integrity, and the applicable repository/context constraints. `uv run --all-packages pytest tests/test_config_update_providers.py tests/test_rpc_settings.py tests/test_cli_onboard_commands.py -x` passed (526 tests). Ruff check/format, commit-message checks, source-language checks, and `git diff --check` passed. The remaining open blocker thread was started by another reviewer, so it was not resolved here.

raven/config/update_providers.py
0xKT: **Re-verified fixed** on 283bc74a. Both halves are answered -- the one I asked for and the one I
was wrong to have accepted last time.

**The address.** `_PROVIDER_BASE_URL_FALLBACK` now carries groq and zai (:1117-1118). Measured end
to end through `settings_everos_set` into the written everos.toml, three borrows in one run
against a section already holding a hand-typed gateway address:

    borrow_from=groq       base_url = 'https://api.groq.com/openai/v1'
    borrow_from=zai        base_url = 'https://api.z.ai/api/paas/v4'
    borrow_from=gpustack   base_url = <DELETED>        <- control: the delete still happens
                                                          where no table can hold an address

At the function boundary the same three plus an unspecced vendor answer identically, so the change
is in the table and not in the caller.

**The justification.** The comment that said the delete "fails loudly" now says what I measured
instead (:1892-1897): the section's own readers ask `model and api_key`, cannot tell an addressless
section from a working one, and nothing reports the gap until the role is called. That is the
correction, written where the next reader meets it, and it points at the table as the answer for
anything that turns up in that branch.

What is left in the branch is now a set I cannot fault: another protocol, or a deployment whose
address only its operator knows. The new unit test names `gpustack` for exactly that
(tests/test_config_update_providers.py:1836), which is a better example than the groq it used to
use, because groq's address was knowable and gpustack's is not.

**The tests bite and the stale name is gone.** Removing the two table entries turns
`test_a_vendor_the_table_knows_only_by_its_shown_address_still_lends_one`
(tests/test_rpc_settings.py:546) red -- it drives the RPC and reads the written file, so it pins
the outcome and not the boundary. 194 passed across tests/test_config_update_providers.py and
tests/test_rpc_settings.py on this head. And the nit I raised in passing is fixed properly rather
than renamed: `test_a_borrow_hands_over_an_address_as_well_as_a_key` (:504) now says what it pins.

One thing I am not asking you to change, recorded so it is not lost: for a vendor that does reach
the delete branch, `role_configured_in` still answers True and the LED beside it still lights.
That is the same blindness, now with a smaller and better-argued set behind it, and the comment
says so. If that set ever grows, the reader-side fix is the one to reach for.

0xKT (approved):

PR #513, opened 2026-09-19
fix(*): stop a settings write from waiting on the model catalogue

gloryfromca:
## Summary

A tools switch sat busy for two seconds before it moved. Measured on a real gateway, with the page's own socket tapped:

```
settings.set   +0ms
settings.get   +15ms
model.options  +16ms          <- 1.7-2.0s, every write
config.get     +2073ms
switch flipped  2124ms
```

`source.set` reloaded the whole settings load after every write, and that load ends in `model.options` -- a live read of every configured vendor. No settings key changes what it answers. The catalogue moves into `loadSettingsWithProviders`, which boot, the page load and the provider writes call; a plain write reloads the config alone. The same switch now flips in 21ms, and `model.options` is gone from the write path.

Two further things the page got wrong about built-in tools:

- The fixed switch was drawn like every other one -- full opacity, pointer cursor, same colour. `aria-disabled` is true, which says so to a screen reader and to nobody else, so the row read as a control that ignores clicks. It is greyed now, with a default cursor.
- Which tools carry no switch was inferred from the card a tool sat in. That forced `cancel_dag`, `dag_status` and `resolve_dag_node` into the meta-tool card to come out fixed, away from the tool they control. `ext.list` now reports the fact itself -- the union `tool_search.py` already applies, the two meta tools plus every schema-hidden one -- and the three controls sit beside `run_subagent_dag` while still reading as built in. The remaining card is `tool_search` and `tool_call`, named for what it holds again.

`tool_call` is in the schema by necessity: the model has to see the doorway to reach a hidden tool. Asserting the hidden set alone let it through as a live switch, which a run on a real gateway caught and the test now pins.

Not in this PR: `model.options` itself is slow because litellm resolves a huggingface context window over the network on every call, at a URL carrying a doubled path segment that 404s six times per call. That fix is separate.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `npx vitest run` in ui-web: 179 files, 2563 tests passed.
- `uv run --all-packages pytest tests/test_rpc_console.py tests/test_settings_tool_groups.py tests/test_rpc_schema_match.py tests/test_rpc_contract_shapes.py -q`: 524 passed.
- `npm run type-check`, `npm run gen:check` (184 methods), ui-tui `npm run lint:rpc` and `npm run lint:i18n`: all in sync.
- `uv run ruff check` and `ruff format --check`, `make check-source-language`, `make check-large-files`: clean.
- Real gateway, temp RAVEN_HOME, page driven through playwright: switch round trip 2124ms before, 21ms after; the three DAG controls render greyed under the collaboration card; `tool_call` renders greyed under tool search.
- Mutation-verified, each reverted one at a time: putting the catalogue back into every write reds the new source test on `+ "model.options"`; dropping the greyed class reds the tools page test; dropping a permission push reds the chip gate; reporting the hidden set without the meta tools reds the ext.list test on `tool_call`.

## Risk

User-visible: a settings write no longer refreshes the provider rows. Nothing a settings key writes feeds `model.options`, and every write that does -- the provider ops, `add_models`, `set_fields` -- still reloads it. The built-in rows change appearance and three of them change which card they sit in; no tool gains or loses a switch, and the counter still counts switchable rows only.

Contract: `ExtToolRow` gains an optional `builtin`. Both generated clients are regenerated rather than hand-edited. An older client ignoring the field draws what it drew before.

Rollback: revert the commit; nothing is persisted by this change.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the full diff against `refactor/ui_web_architecture`, including the surrounding settings/model sources, `ext.list`, the tool registry and progressive-disclosure path, production call sites, and relevant history. Also checked the repository rules in AGENTS.md/CLAUDE.md, the Runtime and Web UI context documents, wire compatibility, architecture boundaries, and whether the changed tests weakened coverage.

The settings split keeps provider catalogue reads on boot, dialog load, and provider mutations, while ordinary `settings.set` writes refresh only configuration. The tool metadata remains an optional wire addition, older clients ignore it, and current switchability/counting follows the registry-provided row rather than card placement. The source regression test and permission-chip gate still exercise the behavior they claim.

Verification:
- `uv run --all-packages pytest tests/test_rpc_console.py tests/test_settings_tool_groups.py tests/test_rpc_schema_match.py tests/test_rpc_contract_shapes.py -q`: 524 passed.
- Changed-area Vitest run (`source.test.ts`, `Tools.test.tsx`, and `permission-chip.test.mjs`): 19 passed.
- Web type-check and codegen check, plus TUI RPC and i18n generation checks: passed.
- The first full web run reported 2,560 passed and 3 failures in the untouched KnowledgePage suite under unsupported Node 23; rerunning that exact file passed all 111 tests. GitHub's page check and complete CI matrix are green.
- `git diff --check`: clean.

gloryfromca (commented):

The rewritten commit has the same tree object as the previously reviewed revision (`13f657305466ef2073879b193b8b367627df2203`); only commit metadata changed. The new `Co-authored-by: Claude (claude-opus-5) <noreply@anthropic.com>` trailer parses correctly, is ASCII-only, and now follows the AGENTS.md/CLAUDE.md format.

Because the source and test trees are byte-for-byte identical, the prior review's diff, callers/history, backward-compatibility, test-integrity, and architecture conclusions remain applicable; no code tests were rerun for this metadata-only rewrite. `git diff --check` remains clean.

gloryfromca (commented):

This revision rebases the already-reviewed patch from `c418e0bb` onto `bc74bf735`. `git range-diff` reports the PR commit as unchanged; the apparent large delta is the advanced target branch, while the PR still changes the same 20 files.

This round checked the rebase interactions in the touched settings/installed sources and generated RPC contracts, along with repository rules, callers/history, backward compatibility, test integrity, and the Runtime/Web UI architecture constraints. The new base's settings credential-field work and installed-source simplification do not change the settings-load split or built-in-tool metadata behavior. No tests were weakened.

Verification on the rebased tree:
- Focused Python RPC/schema suite: 526 passed.
- Changed-area web tests: 19 passed.
- Web type-check and RPC codegen check: passed.
- TUI RPC and i18n generation checks: passed.
- `git diff --check`: clean.

0xKT (approved):

PR #514, opened 2026-09-19
perf(providers): stop resolving a window over the network for nothing

gloryfromca:
## Summary

`resolve_context_window` went to huggingface.co even when the caller asked it not to fetch. That flag governs this module's own OpenRouter read; it does not reach the one LiteLLM does inside `get_model_info`, which GETs a repo's `config.json` on every call and caches nothing.

`model.options` resolves a window per configured vendor, so it paid six sequential round trips -- about 1.5s -- and the settings page reloaded it after every write:

```
 235 ms  https://huggingface.co/huggingface/deepseek-ai/DeepSeek-V4-Pro/raw/main/config.json
 309 ms  https://huggingface.co/huggingface/moonshotai/Kimi-K3/...
 257 ms  https://huggingface.co/huggingface/zai-org/GLM-5.3/...
 ... six in all, 1513 ms
```

They bought nothing. The id handed over still carries its routing prefix, so the address asked for is `huggingface.co/huggingface/<repo>` -- not a repo, 404 every time, window None either way. The corrected address answers 200, which is how the doubled segment was confirmed rather than guessed.

`may_fetch` sits beside `may_prompt` and is asked at the same two places, for the same reason that one exists: a lookup meant to read a table instead does something slow, and every entry point that resolves a model reaches it.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- Real gateway, temp RAVEN_HOME, over the wire: `model.options` 1965 / 1709 / 1725 ms before, 349 / 243 / 241 ms after, the same 55 providers.
- One window in process: `resolve_context_window("huggingface/deepseek-ai/DeepSeek-V4-Pro", allow_fetch=False)` 559 ms and one HTTP call before, 0.0 ms and none after, returning None both times.
- `uv run --all-packages pytest` over every test file that names `resolve_context_window`, `may_prompt` or `metadata_candidates`: 577 passed.
- Mutation-verified: removing the guard reds the new test with `handed over: ['huggingface/deepseek-ai/DeepSeek-V4-Pro']`.

## Risk

A huggingface id no longer gets a window from LiteLLM. It was not getting one: the lookup 404s. The table and the OpenRouter catalogue, which this module consults first, are unaffected, and every other vendor takes the same path it took before. If a real huggingface window is wanted later it needs a cached lookup with the prefix stripped, which is a separate change and is why the guard is named for what it prevents rather than for the vendor.

Rollback: revert the commit.

## Related Issues

N/A


gloryfromca (commented):
Blocking: the branch name and Claude co-author trailer must satisfy AGENTS.md before merge.

The implementation itself is clean in this review. I found no functional defect after checking the diff, all three guarded LiteLLM paths (context window, max output, and rates), the installed LiteLLM behavior, metadata candidates, RPC/provider/agent callers, and relevant history.

Two repository hard constraints still need correction:

- The head branch is `fix/model_options_skips_the_hf_round_trip`. Its short description has seven words, while AGENTS.md requires 3-5.
- Commit `d59e1ae6b282` uses `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>`. AGENTS.md requires `Co-authored-by: Claude (<actual-model-id>) <noreply@anthropic.com>`; `1M context` is not the model ID.

AGENTS.md says its hard-constraint violations are rejected, so these remain blockers even though the runtime change is sound.

Coverage: I checked AGENTS.md and CLAUDE.md (the same policy), CONTEXT-MAP.md and the Runtime terminology, the complete target diff, surrounding implementation, callers, history, backward compatibility, test-strength changes, and the relevant Runtime/RPC architecture boundary. The new test was not weakened and directly traps the prohibited handoff.

Verification:
- `uv run pytest tests/test_provider_rates.py -q`: 77 passed.
- The first broader affected-file run without workspace packages: 490 passed, 87 setup errors, all from missing `raven_everos` in `test_cli_doctor_commands.py`.
- Re-running the same affected files with the repository workspace environment, `uv run --all-packages pytest ... -q`: 577 passed.

gloryfromca:
Both hard-constraint findings hold, and both are fixed -- but the branch one cannot be fixed in place, so this PR is superseded by a new one from the renamed branch.

- Branch name: `fix/model_options_skips_the_hf_round_trip` is seven words. The replacement is `perf/model_options_hf_round_trip` -- five, and the type now matches the commit's `perf(providers)` rather than contradicting it.
- Trailer: corrected to `Co-authored-by: Claude (claude-opus-5) <noreply@anthropic.com>`. The same wrong trailer was on four other commits of mine tonight; all are rewritten and force-pushed with a lease, content byte-identical (`git diff <old> <new> --stat` empty in each case).

The commit is otherwise unchanged: same tree, same tests, same measurements.

Two branch names already under review have the same problem and I did not rename them, deliberately: `fix/first_run_model_without_a_loop` (#511) and `fix/borrowed_base_url_follows_the_lender` (#512) are both six words. Renaming means closing and reopening, which would throw away the review threads on them -- including an open question I put to a reviewer. Say the word and I will do it; otherwise I would rather carry the violation visibly than destroy the record that the work depends on.

gloryfromca:
Superseded by #515, from the renamed branch `perf/model_options_hf_round_trip`. Same commit, same tree; both findings from the review here are fixed.

PR #515, opened 2026-09-19
perf(providers): stop resolving metadata over the network for nothing

gloryfromca:
## Summary

`resolve_context_window` went to huggingface.co even when the caller asked it not to fetch. That flag governs this module's own OpenRouter read; it does not reach the one LiteLLM does inside `get_model_info`, which GETs a repo's `config.json` on every call and caches nothing.

`model.options` resolves a window per configured vendor, so it paid six sequential round trips -- about 1.5s -- and the settings page reloaded it after every write:

```
 235 ms  https://huggingface.co/huggingface/deepseek-ai/DeepSeek-V4-Pro/raw/main/config.json
 309 ms  https://huggingface.co/huggingface/moonshotai/Kimi-K3/...
 257 ms  https://huggingface.co/huggingface/zai-org/GLM-5.3/...
 ... six in all, 1513 ms
```

They buy nothing, and no id we can pass would change that. LiteLLM computes the stripped repo name and then does not use it:

```
_get_model_info_helper   split_model = potential_model_names["split_model"]        # 'deepseek-ai/DeepSeek-V4-Pro'
                         if custom_llm_provider == "huggingface":
                             _get_max_position_embeddings(model_name=model)        # the caller's original string
```

`split_model` holds the bare repo id and is not used; `model` still carries the routing prefix, so the address asked for is `huggingface.co/huggingface/<repo>`, which is not a repo. (`stripped_model_name`, despite the name, still carries it -- measured.) `get_max_tokens` one function away does the same branch correctly, by reassigning `model` from `get_llm_provider` first.

Still present on upstream `main` today (`utils.py:5817`; `:5678` in the pinned 1.85.0). Passing the bare repo id from here does not help either: `get_llm_provider` then raises for want of a provider. Measured both ways -- with the prefix, 404 and `{}` after 533ms; without it, `BadRequestError` in 1ms. So no argument from this side reaches a working lookup, and restoring the capability means an upstream fix or our own cached read.

**Guarded at three sites, not two** -- an earlier draft of this description said two, and it was wrong. The window and max-output lookups read the static table *before* the guard, so a model the table knows still answers and only the dynamic call is skipped. The rates lookup has no table read of its own (its own comment says why), and `cost_per_token` on a huggingface id was measured going to the network for 799ms to return `(0, 0)`, which this module already treats as unknown. So nothing that answers today stops answering.

`may_fetch` sits beside `may_prompt` for the same reason that one exists: a lookup meant to read a table instead does something slow, and every entry point that resolves a model reaches it.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other

## Verification

- Real gateway, temp RAVEN_HOME, over the wire: `model.options` 1965 / 1709 / 1725 ms before, 349 / 243 / 241 ms after, the same 55 providers.
- One window in process: `resolve_context_window("huggingface/deepseek-ai/DeepSeek-V4-Pro", allow_fetch=False)` 559 ms and one HTTP call before, 0.0 ms and none after, returning None both times.
- `uv run --all-packages pytest` over every test file that names `resolve_context_window`, `may_prompt` or `metadata_candidates`: 577 passed.
- Mutation-verified: removing the guard reds the new test with `handed over: ['huggingface/deepseek-ai/DeepSeek-V4-Pro']`.

## Risk

A huggingface id no longer gets metadata from LiteLLM's dynamic lookup. It was not getting any: the window lookup 404s and the rates lookup answers `(0, 0)`. The static table and the OpenRouter catalogue, which this module consults first, are unaffected, and every other vendor takes the same path it took before.

Restoring the capability is a separate change and cannot be done by choosing a better id: it needs either an upstream fix at `utils.py:5678` or our own cached read of `huggingface.co/<repo>/raw/main/config.json`. That is why the guard is named for what it prevents rather than for the vendor -- the vendor is incidental, the network call in a hot path is the problem.

Rollback: revert the commit.

## Related Issues

Supersedes #514, which was opened from a branch whose name broke the 3-5 word rule. Same commit, same tree; the review on #514 found no functional defect.


gloryfromca (commented):
Blocking: the PR Type checkbox has to match the `perf` commit type.

The implementation itself is clean in this review. The static table is still consulted before the new guard, Hugging Face's dynamic LiteLLM result remains the same unknown/zero result the callers already handled, and the OpenRouter and fallback tiers are unchanged. I found no functional or backward-compatibility defect.

The remaining blocker is in the PR description. AGENTS.md says the Type checkbox must mirror the commit type and maps unlisted types to Other. This PR and commit use `perf`, but the description checks Fix; please check Other instead.

Coverage: I checked AGENTS.md and CLAUDE.md, CONTEXT-MAP.md and the Runtime Token Rates terminology, the complete target diff, the three guarded LiteLLM paths, metadata-candidate generation, affected RPC/provider/agent callers, relevant history, backward compatibility, test-strength changes, and the Runtime/UI architecture boundary. I also inspected the pinned LiteLLM implementation directly.

Verification:
- `uv run --all-packages pytest <the ten affected test files> -q`: 577 passed in 29.74s.
- `uv run ruff check raven/providers/rates.py tests/test_provider_rates.py`: passed.
- `uv run ruff format --check raven/providers/rates.py tests/test_provider_rates.py`: 2 files already formatted.
- `git diff --check github/refactor/ui_web_architecture...HEAD`: passed.

An initial shell invocation passed the selected filenames as one zsh scalar and collected no tests (exit 5); the array-based rerun above is the valid result.

gloryfromca:
Fixed: the Type box now reads Other, alone. The mapping in AGENTS.md section 3.7 names five types and sends everything else to Other; `perf` is one of the everything else, and I had checked Fix by habit rather than by the rule -- which is exactly what that section says the checkbox is not, "not a fresh judgement call".

Nothing else changed: same head `84019d308`, same tree, same commit message.

Incidentally, the zsh trap you hit passing the filenames as one scalar is in this repository's own operating notes -- `pytest $FILES` reaches pytest as a single nonexistent path and exits 5 with "no tests ran". `${=FILES}` or an array is the fix, which is what your rerun did.

gloryfromca (commented):

The PR description now checks Other, matching the `perf` commit type and clearing the sole blocker from the prior review. The code revision is unchanged, so the previously reported 577 passing affected tests remain applicable; no tests were rerun for this metadata-only correction. There are no review threads from this reviewer to resolve.

gloryfromca (commented):

This head is a commit-message-only rewrite of the reviewed revision: its tree and parent are identical to `84019d3089f0`. The revised commit and PR wording now describe all three guarded metadata paths, and the title, ASCII-only message, co-author trailer, branch name, and Other type selection satisfy the repository rules.

The latest target was fetched at `bc74bf7355ee`; `git merge-tree --write-tree HEAD github/refactor/ui_web_architecture` completed cleanly, and the full target diff remains the same 63-line change previously reviewed for callers, history, backward compatibility, test strength, and architecture boundaries. No review threads are open.

Verification:
- Source and test trees are byte-identical to the revision whose affected suite passed 577 tests, so pytest was not repeated for this metadata-only rewrite.
- `uv run ruff check raven/providers/rates.py tests/test_provider_rates.py`: passed.
- `uv run ruff format --check raven/providers/rates.py tests/test_provider_rates.py`: 2 files already formatted.
- `git diff --check github/refactor/ui_web_architecture...HEAD`: passed.

gloryfromca (commented):

The new commit changes only the `may_fetch` explanation, and its account matches the pinned LiteLLM 1.85.0 implementation: `_get_model_info_helper` computes the bare `split_model` but sends the original qualified `model`, while `get_max_tokens` first replaces `model` with the provider-split value. The delta therefore documents the existing workaround without changing runtime behavior, compatibility, tests, callers, or architecture.

Repository rules, the incremental and full target diffs, commit metadata, test-strength implications, current PR metadata, and the latest target merge were checked. The merge-tree check completed cleanly and no review threads are open.

Verification:
- `uv run pytest tests/test_provider_rates.py -q`: 77 passed.
- `uv run ruff check raven/providers/rates.py tests/test_provider_rates.py`: passed.
- `uv run ruff format --check raven/providers/rates.py tests/test_provider_rates.py`: 2 files already formatted.
- `git diff --check github/refactor/ui_web_architecture...HEAD`: passed.

gloryfromca (commented):

This revision rebases the two previously reviewed commits onto the current target. `git range-diff` reports both commits as patch-identical, while the PR diff remains confined to `raven/providers/rates.py` and `tests/test_provider_rates.py`. The rebase therefore adds no new change to callers, behavior, compatibility, test strength, or architecture.

Repository rules, both commit messages, current PR metadata, the full target diff, and the target merge were checked. The merge-tree check completed cleanly and no review threads are open.

Verification on the rebased tree:
- `uv run pytest tests/test_provider_rates.py -q`: 77 passed.
- `uv run ruff check raven/providers/rates.py tests/test_provider_rates.py`: passed.
- `uv run ruff format --check raven/providers/rates.py tests/test_provider_rates.py`: 2 files already formatted.
- `git diff --check github/refactor/ui_web_architecture...HEAD`: passed.

0xKT (approved):

PR #516, opened 2026-09-19
fix(docs): give each language its own search index and settle the site shell

LivXue:
## Summary

The documentation site built one search index for both languages, so a reader on
the English pages was offered Chinese ones: results they cannot read, behind
links that leave the language they chose. A build hook now splits the merged
index in two and points the Chinese pages at their own copy, which the theme
also resolves every result link against.

Where that split runs is the whole fix. The merge lands in a plugin's own
`on_post_build`, so the split has to be the last thing that event does. Running
it on `on_shutdown` looks equivalent and is not: `mkdocs serve` never shuts
down, so the preview went on serving the merged index however often it rebuilt.
The new guard makes the same assertion against `mkdocs build` and against
`mkdocs serve`, because passing one says nothing about the other.

The rest settles the shell against the reference layout:

- the rail's foot is its own box the navigation cannot show through, and the tab
  icon is the EverMind mark;
- the footer keeps to the content column instead of covering the fixed columns
  once the reader reaches the page foot;
- the contents column reaches the screen foot with its scrollbar hidden behind a
  fade, and follows the reader by centring the first lit entry;
- search opens as a centred dialog over the page. The form itself travels into
  that dialog, so what stays in the rail is drawn from the field's own metrics
  and its shape does not change on a click. Neither state animates: the theme
  transitions the two shapes into each other, and leaving search played that out
  in the corner of the page after the reader had moved on.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/integration/test_docs_site_search_e2e.py` -> 2 passed.
  Without this change the preview case fails: the Chinese index is HTTP 404 and
  the English index carries 146 Chinese pages.
- `uv run pytest tests/integration/test_docs_site_controls_e2e.py` -> 25 passed
- `uv run pytest tests/test_docs_site.py tests/test_source_language_check.py` ->
  31 passed
- `make check-commits`, `make check-source-language`, `make check-large-files` ->
  exit 0; `ruff check` on the added files -> no findings
- `make docs-build` -> the root index holds 149 entries and no Chinese page; the
  Chinese index holds 146, none of them addressed from the site root

Driven in a real browser at 1560x820 in both languages: a word that appears in
both trees returns 10 results and 0 from the other language, and the corner of
the page is pixel-identical to its never-opened state at every frame after
leaving search. Every page of the built site was checked at the page foot (28
pages, 0 faults) for a footer covering a column, for a column whose last entry
cannot be reached, and for the rail's foot icons being clickable.

No page content changed, so nothing else needed updating with it.

## Risk

User-visible: search no longer returns pages from the other language, and the
search control, the footer and the two fixed columns change shape. The field
left in the rail while the dialog is open is a drawn replica, and every rule
behind it sits above the theme's desktop breakpoint, so narrow viewports keep
the theme's own full-screen search untouched.

Rollback: revert the commit. The hook is one file, and dropping it from the
`hooks:` list in `docs-site/mkdocs.yml` restores the previous merged index on
its own.

## Related Issues

N/A


gloryfromca (commented):
Blocking: replace the malformed `claude-opus-5[1m]` model ID in both required co-author trailers.

The `[1m]` suffix is an ANSI-formatting fragment, not an actual model ID. AGENTS.md section 3.3 requires the current-session model ID so model contributions remain distinguishable, and GitHub carries these trailers into the squash commit.

I otherwise found no code defect. I covered the full `github/main...975806563ce9` diff, the generated MkDocs site, the Material search/base callers, `mkdocs-static-i18n` post-build ordering, the desktop shell CSS/TOC JavaScript, relevant history, backward compatibility at the desktop breakpoint, and the new tests for weakening or false-green assertions. I also checked AGENTS.md/CLAUDE.md and CONTEXT-MAP.md; this change introduces no Runtime/TUI domain term or architecture-boundary issue.

Verification:
- `uv run --frozen --python 3.12 --group docs pytest tests/integration/test_docs_site_search_e2e.py -q -o addopts=`: 2 passed.
- `uv run pytest tests/test_docs_site.py tests/test_source_language_check.py -q`: 31 passed.
- `uv run --group docs mkdocs build --strict`: passed; English produced 149 entries with no `zh/` locations and Chinese produced 146 entries rooted to its locale.
- Source-language, large-file, commit-message, `git diff --check`, and Ruff rule checks passed.
- `tests/integration/test_docs_site_controls_e2e.py`: 25 skipped because `/usr/bin/google-chrome` is absent. Redirecting the unchanged suite to Playwright Chromium reached browser startup but failed because this image lacks `libatk-1.0.so.0`, so I could not independently execute the 25 browser cases here.
- `ruff format --check` still says `docs-site/hooks/search_by_language.py` would be reformatted; that is an automated formatting-gate result, not a separate review finding.

gloryfromca (commented):
Blocking: the previously reported malformed co-author model ID still has to be corrected.

The code delta from `975806563ce9` is clean: `docs-site` is now included in the shared Python lint roster, the hook matches Ruff formatting, and the expanded roster also passes the existing `ty` gate. No new code finding emerged. The earlier trailer defect remains in the existing commits and is also present on the new commit; I am not repeating its rationale here.

Coverage this round: the full diff and new delta, Makefile callers (`lint-python` and `lint-types`), repository rules and commit history, backward compatibility, test integrity, and the previously reviewed docs-site architecture.

Verification:
- Ruff check and format over the expanded `PYTHON_LINT_TARGETS`: passed (`1977 files already formatted`).
- `ty check` over the expanded non-test roster: passed.
- `uv run --frozen --python 3.12 --group docs pytest tests/integration/test_docs_site_search_e2e.py -q -o addopts=`: 2 passed.
- Source-language, large-file, commit-message, and `git diff --check` checks: passed.

gloryfromca (commented):

The new delta correctly keeps `docs-site` in the Ruff roster while excluding it from `ty`, whose normal lint environment does not install the MkDocs dependency group. I found no new code issue. Under the fourth-round convergence rule, I no longer hold the previously reported metadata objection as a merge blocker. There are no review threads from me to resolve.

Coverage this round: the full diff and Makefile delta, lint target callers and dependency environment, project rules and history, backward compatibility, test integrity, and the previously reviewed docs-site architecture.

Verification:
- Ruff check and format over the expanded roster: passed; 1977 files already formatted.
- `ty check` over the corrected non-test, non-docs-site roster: passed.
- `uv run --frozen --python 3.12 --group docs pytest tests/integration/test_docs_site_search_e2e.py -q -o addopts=`: 2 passed.
- Source-language, large-file, commit-message, and `git diff --check` checks: passed.

ZuyiZhou (approved):
Approving. The search split is the load-bearing part, so I traced it end to end rather than trusting the description.

Verified locally:

- `make docs-build` is clean and the split is real: 149 entries in the root index, 146 in the Chinese one, zero leakage in either direction, and every `location` in both resolves to a file that actually exists in its own tree.
- The `base` rewrite is correctly scoped, which was my main worry. Material builds the index URL as `new URL("search/search_index.json", config.base)`, so shifting `base` is the right lever. The search worker is a *separate* `search` key that the template writes literally (`../../assets/javascripts/workers/search.*.min.js`), so it is untouched and Chinese pages still load it from the true root. The two other things that resolve against `base` are `../versions.json` and `sitemap.xml`, and they sit behind `version: null` and `navigation.instant`, neither of which this site enables. So the rewrite reaches the index URL and the result hrefs and nothing else.
- `_shallower` holds at every depth: `zh/index.html` gets `.` and `zh/quick-start/index.html` gets `..`, both landing on `site/zh/`. The `".."` case does not start with `"../"` and correctly falls through to the `"."` branch.
- Both new tests fail with the hook removed, in exactly the shape described: the preview returns HTTP 404 for the Chinese index. The guard is not one that passes either way.
- `@event_priority(-100)` is load-bearing and proven so. I flipped it to `+100`, which puts the hook ahead of the i18n merge, and the build test caught it. The `if not english or not chinese` short-circuit doubles as re-run protection.
- Mobile is genuinely untouched: 202 of the 212 added CSS lines sit inside `@media screen and (min-width: 76.25em)`, and the 10 that do not are custom-property declarations that only desktop rules consume.
- `ruff check docs-site` passes now that it is in `PYTHON_LINT_TARGETS`, `lint-types` filters both `tests` and `docs-site` as intended, 31 unit tests pass, and `check-large-files` / `check-source-language` exit 0.

One thing worth doing, not here:

`tests/integration/` is excluded by `norecursedirs`, so the guard this fix depends on never runs in CI. The controls suite has a reason to stay local, since it needs a browser at a hardcoded `/usr/bin/google-chrome`. This new file does not: it needs `uv` and the `docs` group and nothing else, both of which the `docs build` job already has. Promoting just `test_docs_site_search_e2e.py` into that job would turn this from a regression guarded on your laptop into one guarded on every pull request. If it moves, the fixtures' `pytest.skip` on a failed build should become a hard failure, or CI will go green on a site that did not build.

That matters because the hook's two early returns are deliberately silent. If a future `mkdocs-static-i18n` moves the merge, renames the file, or stops merging, the split quietly stops happening and the only thing that would notice is a test that does not run.

Two minor notes, neither worth acting on: both indexes inherit `config.lang: ["en", "zh"]`, so the English index has lunr pull a Chinese tokenizer it will never use; and `follow()` writes `scrollTop` from the window scroll handler, where the 1px tolerance is what keeps it from fighting the reader.

0xKT:
**Not a blocker** -- the search half is real and I could break it on purpose. The other 383 lines
of test cannot run anywhere in this repo, so the shell they describe ships unverified by anything
that executes.

**What I verified by building the site, not by reading it.** `uv run --group docs mkdocs build
--strict` on this head merged onto 32e612d5, then reading the output:

    search/search_index.json       149 entries, 0 beginning with zh/
    zh/search/search_index.json    146 entries, 0 beginning with zh/   (addressed from their own root)

and the two e2e cases really execute rather than skipping -- `2 passed in 12.72s`, which is a real
build plus a real preview server. Unregistering the hook in mkdocs.yml turns both of them red, the
built-site one and the preview one. That is the fix pinned at both ends.

Two things I suspected and could not make true, recorded so nobody re-checks them:

  - `_shallower` returning `""` for a page whose base is `"../"`. mkdocs writes the base without a
    trailing slash, so the string is `".."` at depth 1 and `"../.."` at depth 2. Measured on the
    built tree: `zh/index.html` gets `base='.'` (the `else` branch) and `zh/architecture/index.html`
    gets `base='..'`. Both correct, at both depths.
  - the data: URI favicon being mangled by the theme's `url` filter. It comes out byte-identical in
    `<link rel="icon">` on the root page and on a deep `zh/` page. And `navigation.instant` is not
    among the enabled features, so re-pointing `base` reaches the search index and nothing else.

**The finding: `test_docs_site_controls_e2e.py` has no runner.** Three separate gates stand between
it and ever executing, and all three hold today:

    pyproject.toml:472   norecursedirs = ["tests/integration", ...]   -- outside default collection
    .github/workflows/ci.yml:297                                      -- CI names ONE integration
                                                                         file, and it is not this one
    pyproject.toml:124-129  playwright is the opt-in `browser` extra, documented for raven.browser;
                            no job installs it, so line 29 `pytest.importorskip` fires
    tests/integration/test_docs_site_controls_e2e.py:35,88
                            CHROME = "/usr/bin/google-chrome", and a missing binary skips

Run here through the board runner it reports `1 skipped` on the import alone. The control that
makes that mean something is its neighbour: `test_docs_site_search_e2e.py`, same directory, same
runner, same invocation -- it runs and passes. So the difference is the dependency, not the way I
called it.

I am filing this as a note rather than a block because the integration directory being out of
default scope is this repo's deliberate arrangement, not something this PR changed, and the search
suite shows the arrangement works when the tool is present. What is new is a suite whose tool is
never present and whose file says nothing about how to run it. A line in its docstring naming the
command -- the extra, and the chrome path it wants -- would cost nothing and would tell the next
reader whether the hit tests have ever been green.

Smaller, same family: the rail's replica search field carries its label in CSS
(`docs-site/docs/stylesheets/evermind.css`, `content: "Search"` with an
`html[lang^="zh"]` override carrying the Chinese word as a CSS string literal). That is complete for the two locales the site has, and
a third one would silently read "Search" -- the only string on the page that does not come from the
catalogue.

For the record, what else I checked and found sound: all seven files read line by line;
`GATES OK head=6d781e90 tip=32e612d5 base=3dd06f6e ruff/lint-imports/commit/large/lang rc=0`; the
`follow()` added to toc-progress.js is defined at :159 and first reachable through `start()` on
DOMContentLoaded, so there is no temporal-dead-zone hazard in calling it from `sync` at :150; and
the Makefile's stated reason for filtering docs-site out of `lint-types` holds -- `docs` is a
dependency-GROUP (pyproject.toml:535,557) and `lint-types` runs `--all-extras`, which does not pull
groups, so ty really would read mkdocs as missing while ruff still covers the tree.

PR #517, opened 2026-09-19
docs: add a showcase of six real multi-agent runs

ZuyiZhou:
## Summary

Adds a Showcase section to the README, between the built-in agents and the
third-party agent presets. The agent sections above it each carry benchmark
charts; this one shows what the agents produce together on real work.

Six runs, laid out as two tables of cells. Each cell holds two screenshots:
the task graph exactly as Raven ran it, and what the run produced. The first
table is the four runs that end in a slide deck, and all four graphs show the
same orchestration shape, three research legs in parallel, a node that reduces
them to a single argument, and one design node, so they read as one family
rather than four unrelated demos. The second table is two runs that end in
something else: a comparison board across six orchestration frameworks, and a
parameter sweep that Raven-Code wrote, Raven-Oncall ran and Raven-Design
plotted.

Everything on display is an output of the run itself, not an illustration.
The comparison board was re-rendered from its own source in English so the
section does not mix languages; the one remaining Chinese artifact is the
Song-dynasty deck, which was asked for in Chinese.

Images are hosted as GitHub user attachments, matching how the existing README
images are served, so no assets land in the repository.

Two things left out on purpose, each worth its own change: per-deck download
links, because both larger decks exceed the 25 MiB attachment limit and need
release assets; and a run whose task graph is not yet captured in English.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `make check-large-files` and `make check-source-language` both pass; the
  change adds no files, only README markup.
- Rebased onto f10ac9f11 and re-ran both gates afterwards.
- Rendered the section locally at GitHub's 860px README column width to check
  cell alignment, image scaling and the two-column layout.
- Confirmed all twelve attachments return HTTP 200 to an anonymous request,
  and that each one matches its source file by checksum.
- Paired every graph with its own artifact by content rather than by file
  name, because the source screenshots use two different numbering schemes.

No tests or lint were run: this change touches README markup only.

## Risk

README-only, with no runtime behaviour change. Rollback is reverting the
commits; the twelve images are user attachments, so nothing is left behind in
the repository.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the full `github/main...HEAD` diff and its three-commit history, the surrounding README structure, AGENTS.md/CLAUDE.md and the context map. I also checked backward compatibility, test integrity, and architecture impact: this is an additive README-only change, touches no callers or runtime boundaries, and changes or weakens no tests.

All twelve new attachments load anonymously and visually match their titles and alt text. GitHub's Markdown renderer preserves both tables and all twelve images. `git diff --check`, the large-file gate, source-language gate, and commitlint all pass. The documented `make` wrapper could not run because `make` is not installed here, so I ran its exact `uv`-based gate commands directly. The complete GitHub suite, including docs build, repository-file checks, and unit tests, is green.

0xKT:
**Not a blocker** -- the section is sound and every image in it is real. One thing it leaves
behind: the two READMEs stopped matching.

**What I measured rather than read.** All twelve asset URLs resolve to actual images -- 200,
`image/png`, 56,770 to 2,730,070 bytes, none missing. Worth saying how I got there, because my
first pass was wrong: a `HEAD` request returns 403 for every one of them, which reads as twelve
dead links. The control is an image that was already in this README before your change
(`d1376ca8...`): it answers 403 to `HEAD` and 200 to `GET` exactly the same way. So the 403 is
`user-attachments` refusing the method, not the links, and there is nothing wrong with any of them.

**The parity.** README.md and README.zh-CN.md have matched section for section -- thirteen `##`
headings each, same order, same subjects. Measured on the base and on this head:

    base   README.md 13 headings   README.zh-CN.md 13
    head   README.md 14 headings   README.zh-CN.md 13     <- Showcase, English only

and the diff touches one file. So a reader who arrives at the Chinese README -- which the English
one links to at the top -- now gets a repository tour with no showcase in it, and no sign that one
exists. Every other section in that file has a counterpart; this is the first that does not.

I am filing it as a note rather than a block because nothing is wrong with what landed and nothing
that worked stops working: a missing translation is absence, not breakage, and the section is pure
addition. But the parity here is not incidental -- it holds for all thirteen existing sections,
which is what makes the gap visible rather than unremarkable.

The cheapest close is the same block with the six titles and the two lines of prose translated;
the markup, the table structure and all twelve URLs carry over unchanged.

For the record, what else I checked: the whole diff read line by line (49 added lines, one file,
nothing removed -- I looked for deletions specifically, because a README edit that rewrites while
adding is where a line goes missing quietly); every image is linked rather than committed, which
is what section 7 of AGENTS.md asks for ("Store public-report assets outside git and link to
them"), and the machine gates agree --
`GATES OK head=c8549ec9 tip=0eb8b41b base=f10ac9f1 ruff/lint-imports/commit/large/lang rc=0`;
the added prose is English, as that file's language requires; and each `` carries alt text
that describes its own picture rather than repeating the title.

One thing I could not check and am not claiming either way: whether each task-graph image shows the
shape its alt text describes. Four of them say "three Raven-Research nodes running in parallel into
a synthesis node, then one Raven-Design node" in identical words, which is plausible if those four
runs really did take the same shape, and I have no way to read a PNG's contents to confirm it.

0xKT (approved):

PR #521, opened 2026-09-19
fix(docs): seat the search dialog's clear button on its own centre line

LivXue:
## Summary

The search dialog stacked two controls in one corner and seated neither on
the line the query is typed on.

The theme places its clear button at a fixed distance from the form's top,
an offset that centres the button only in a form of the theme's own height.
This dialog's form is taller, with two consequences from the one cause: the
button sat 10px above the magnifier and the text beside it, and it landed in
the same corner as the chip naming the key that opens search, overlapping it
by 21px.

The button is centred against the form box and against its own container
now, so it follows whatever height the form takes rather than assuming one.
The chip is confined to the closed state, where it is still true: a reader
already inside the dialog has nothing to press that key for, and the field
left in the rail keeps its chip and is unchanged in both states.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Clicking always reached the button, overlap and all, so a guard written
around the click would have passed while the fault was plainly visible in a
screenshot. The guard measures the boxes instead: the chip and the button
must not intersect, and the button's centre must be within a pixel of the
magnifier's. Run against the unchanged site it reports a 21px overlap and a
10px drift, which is how the fault was confirmed before it was fixed.

Measured in a real browser at 1560x820, vertical centres in the open dialog:

| element | before | after |
| --- | --- | --- |
| form box | 195 | 195 |
| magnifier | 195 | 195 |
| query text | 194.5 | 194.5 |
| clear button | 185 | 194.5 |

- `uv run pytest tests/integration/test_docs_site_controls_e2e.py tests/integration/test_docs_site_search_e2e.py` -> 30 passed
  -> exit 0, against an environment synced the way CI installs
- `pre-commit run --from-ref origin/main --to-ref HEAD`, `make check-commits`,
  `make check-source-language`, `make check-large-files` -> exit 0
- `mkdocs build --strict` -> built
- captured the rail at rest, the rail while the dialog is open, and the
  dialog itself: the rail is identical in both states and the dialog's
  corner holds one control

## Risk

User-visible and confined to the open search dialog above the desktop
breakpoint: the clear button moves down onto the centre line, and the key
chip no longer appears while the dialog is open. Narrow viewports keep the
theme's own full-screen search untouched.

Rollback: revert the commit. The change is two rules in one stylesheet, and
either can be dropped on its own.

## Related Issues

N/A


gloryfromca (commented):

I reviewed the full `github/main...HEAD` diff, the surrounding custom search rules, Material 9.7.7's search markup and base CSS, and the history that introduced the desktop dialog. The checked-state selectors confine both changes to the open desktop dialog, closing search restores the existing rail field, and viewports below the project's desktop breakpoint remain on Material's own layout. The flex-centred options box tracks the form's scaled height, while hiding the shortcut chip removes the competing control from that corner.

I also checked the repository rules, compatibility, test integrity, and architecture impact. The comments explain theme constraints that are not evident from the declarations, the integration-test filename remains canonical, the source-language and large-file gates pass, and no domain term or module boundary is introduced. The regression test was not weakened to simulate success: it opens real search results and measures the rendered chip, clear-button, and magnifier boxes.

Verification: `uv run pytest tests/test_docs_site.py` passed (5 tests); Ruff passed on the changed test; the strict MkDocs build passed; and the source-language and large-file scripts passed for `github/main..HEAD`. The browser-suite command initially produced 3 passes and 27 skips because this runner has no `/usr/bin/google-chrome`. After installing the declared Playwright extra and pointing the unchanged fixture assertions at its Chromium, collection ran but all 27 browser cases errored at setup because the runner lacks `libatk-1.0.so.0`; none reached an assertion. I therefore did not count the browser suite as passing.

ZuyiZhou (approved):
Approving. The browser guard needs Linux Chrome, so I built the site, served it and measured the boxes myself at 1560x820 rather than taking the table on trust. Both columns reproduce exactly:

| | before | after |
| --- | --- | --- |
| chip vs clear button | overlap 21px | chip suppressed, no overlap |
| clear button centre vs magnifier | -10px | -0.5px |

Form box and magnifier sit at 195 in both states, so the button moving from 185 to 194.5 is the whole fix and nothing else shifted.

Both rules are inside `@media screen and (min-width: 76.25em)` and both are gated on `[data-md-toggle="search"]:checked`, so the closed field and every narrow viewport are untouched -- I confirmed the closed state separately, where the button already centres correctly because the form is the theme's own height.

Suppressing the chip rather than moving it is the right call: it names the key that opens search, which is not a thing a reader already inside the dialog can act on.

The note that a click-based guard would have passed is worth keeping. That is the second time in this series a guard measured the wrong property -- see #518, where the query carried the boundary marks -- and measuring geometry is the correct answer both times.

Same standing note as #516 and #518: this guard lives in `tests/integration/`, which `norecursedirs` excludes from CI, so it protects nothing on a pull request.

Issue #126, opened 2026-07-14
docs: replace the README banner asset


cyfyifanchen:
## Problem

The current README hero banner needs to be replaced. Both `README.md` and `README.zh-CN.md` currently reference the same GitHub user-attachment asset as the Raven banner.

Current asset:

- `https://github.com/user-attachments/assets/5a99d736-49ee-49c9-8b51-890f14078e78`

## Scope

- Replace the Raven banner in `README.md`.
- Replace the matching Raven banner in `README.zh-CN.md`.
- Keep the English and Chinese READMEs visually aligned.
- Preserve the existing badge and language-link layout below the banner.

## Acceptance criteria

- The banner uses the approved new Raven asset.
- The banner renders correctly on GitHub desktop and mobile widths.
- The old banner asset is no longer referenced by either README.
- The README top section still loads without broken images.

LivXue:
Updated Raven README banner: One raven, a whole flock of specialists.

![Raven banner: One raven, a whole flock of specialists](https://github.com/user-attachments/assets/d1376ca8-14ce-4d03-a912-c3c2fdefe555)

LivXue:
Updated Raven README banner: italic ampersands in the tagline rule.

![Raven banner](https://github.com/user-attachments/assets/ff05474a-03f5-4ec2-b42b-55f1508ede06)

LivXue:
Recompressed copies of the README assets hosted in this thread, for README.md and README.zh-CN.md. Every still is now under 400 KB. The old comments stay as they are; each item names the asset it replaces and how it was re-encoded.

- replaces `ff05474a-03f5-4ec2-b42b-55f1508ede06`: 1.59 MB -> 379 KB, JPEG quality 92, 4:4:4 chroma, same size

  ![Raven banner](https://github.com/user-attachments/assets/ae944083-9c61-4218-8372-2c3c0d483d4b)

PR #522, opened 2026-09-19
docs: refresh the README banner

LivXue:
## Summary

Both READMEs point at a re-rendered hero banner on line 3. The composition and the
1536x656 geometry are unchanged; the tagline rule now sets its ampersands in italic so
they match the italic words on either side.

The image is not committed. `scripts/check_large_files.py` rejects an added `.png` by
extension at any size, so the asset is hosted as a GitHub attachment and the README
references it by absolute URL. It was uploaded as a new comment on the existing banner
issue rather than by editing the old one, whose URL is still live in the published
README and in every fork and cache of it.

The banner UUID is shared between the English and the Chinese README, so one replaced
image is two edits carrying the same new URL. The workflow figure further down is the
language-specific one, and it is untouched.

Checked and deliberately not changed: line 3 uses plain Markdown image syntax rather
than the anchor-wrapped `` shape the benchmark figures use. That is
the shape the banner already had, and converting it is a separate decision from
swapping the asset.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `git grep` for the previous asset UUID across all tracked files returns 0 hits after
  the change; before it, exactly two, both README line 3. Nothing survives in
  `docs-site/`, which carries no copy of the README and no `user-attachments` URL at
  all.
- Byte check of the published asset against the local render: a GET of the exact URL
  as written in the README returns HTTP 200 `image/png` with sha256
  `c62a38ab3bf0c8fe82dc194bd8250ee99e324bfdd13242a271f18a8f5c69e107`, identical to the
  file on disk.
- A HEAD probe of that URL returns 403, but so does a HEAD probe of the previously
  published banner, which is live and rendering. The signed storage URL behind the
  redirect refuses HEAD, so GET is the only probe that carries information here. Noted
  because the 403 reads like a failed upload and is not one.
- `make check-large-files check-source-language` -> exit 0.
- `scripts/check_commit_messages.py origin/main..HEAD` -> exit 0.
- No automated test covers this and none was added: no file under `tests/` references
  `user-attachments`, and `tests/test_living_docs.py` globs `raven/**/README.md` and
  `ui-tui/README.md`, not the repository root README. The byte comparison above is the
  verification.

The two boxes above are left blank on purpose. No test suite exercises a README image
URL, and `make lint` covers Python that this diff does not touch. The checks that do
apply to a Markdown-only diff are the three gates listed above, and all three pass.

## Risk

One rendered image at the top of the README landing view. No code path, no
configuration, no dependency, no behaviour change. The artwork carries only the product
tagline, so there is nothing disclosed by publishing it.

Rollback is a single revert of this commit, which restores the previous URL. The
previous asset is untouched and still served, so the rollback needs no re-upload.

## Related Issues

#126 hosts the asset. This PR does not close it.


gloryfromca (commented):

The new hosted asset resolves as a valid 1536x656 PNG, preserves the prior composition and wording, and makes the stated italic ampersand correction. Both localized READMEs use the same URL, and no prohibited asset is added to the repository.

Coverage: I checked AGENTS.md constraints, the complete diff, surrounding README content, the prior banner-refresh history, backward compatibility and rollback, test scope and absence of test weakening, and the repository asset/architecture rules. There are no runtime callers for this documentation-only URL change.

Verification: the large-file, source-language, and commit-message scripts passed directly through `uv`; pre-commit passed for both READMEs; and `git diff --check` passed. No automated test targets root README image URLs. All completed GitHub checks were green when reviewed; one unit-test shard was still pending. (`make` itself is unavailable in this environment, so I invoked the exact underlying scripts.)

ZuyiZhou (approved):
Approving. Every claim here is checkable and all of them check out:

- The new URL returns HTTP 200 `image/png`, sha256 `c62a38ab3bf0c8fe82dc194bd8250ee99e324bfdd13242a271f18a8f5c69e107` -- the exact digest stated -- and the PNG header reads 1536x656, so the geometry really is unchanged.
- The old UUID has zero hits across tracked files; the new one appears in exactly two, both README line 3, same URL in each.
- The old asset still returns 200, so the rollback claim holds: reverting restores a URL that is still served and needs no re-upload.

Hosting it as an attachment rather than committing it is the right reading of section 7, which says to keep public-report assets outside git and link to them. At 1.55 MB a committed copy would have failed the extension block and the size limit both.

Leaving the two verification boxes blank and explaining why is the correct call, not a gap: no suite covers a README image URL, and `test_living_docs.py` globs `raven/**/README.md` and `ui-tui/README.md`, not the root. The byte comparison is the verification, and it is a better one than a test would have been.

The HEAD-returns-403 note was worth writing down. I would have read that as a failed upload too.

PR #524, opened 2026-09-20
fix(*): let a first run set its model, and report a refused write

gloryfromca:
## Summary

Two halves of one first run, moved onto this branch because the other half of each lives in files this branch deletes.

**A fresh install could not finish setup from the page.** `config.set key=model` built a provider before persisting anything; the build resolves against the CURRENT config, and with no model set that raises `MissingCredentialsError`. The model could not be set because no model was set. `setup.status` asks for a provider key AND `agents.defaults.model`, so both surfaces read the install as unconfigured and offered the same dead end again.

The gate now asks the credential question directly, through the same `credential_status(..., include_external=True)` every other surface uses -- asked even when no section is on file, so a provider whose credential lives in an OAuth token file passes it. A build that fails for want of credentials on a first run reads as "no loop yet" rather than as a refusal, the write lands, and the reply carries `needs_restart`: the config is right and this process still cannot run a turn on it, because the wiring a turn needs is assembled once at stack build. Onboarding shows that instead of advancing to a done screen it cannot back up.

**That change makes a refused session write RESOLVE, and two callers here dropped the answer:**

```
features/model/source.ts:169    await gateway().call('config.set', {... session_id: sid})   <- reply dropped
                                return
state/session/runtime.ts:452    try { await gateway().call('config.set', {... session_id}) }
                                catch (e) { toast(...); loadProviders(...) }                 <- a refusal never reaches it
```

`choose` moves the chip before the round trip and rolls it back only in its catch, so the reader picked a model, the chip moved, the page said "Switched to X", and the session was on none of it. The live pick now throws the shape that catch already renders (`data.detail`), so the rollback goes through the path that was always responsible for it. The staged apply says it and reconciles, which is what its own comment already promised for the raising case.

**Why here and not against main.** The RPC half is identical on both branches, but its frontend reaches `live/120-settings.js` and `live/080-overrides.js` on main -- files this branch deletes. Landing it there would have meant a modify/delete conflict at merge, and the fix being deleted rather than carried. Supersedes #511, which is closed.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `npx vitest run` in ui-web: 189 files, 2758 tests passed. `npm run type-check` clean.
- `uv run --all-packages pytest tests/test_rpc_config.py tests/test_rpc_schema_match.py tests/test_rpc_contract_shapes.py tests/test_rpc_registration.py -q`: 502 passed.
- `npm run gen:check` matches the contract at 185 methods; ui-tui `lint:rpc` and `lint:i18n` in sync. Both generated clients are regeneration output, not hand edits.
- Mutation-verified, each reverted on its own: dropping the throw reds "rejects a refused session write"; dropping the staged branch reds "speaks and reconciles when the refusal RESOLVES rather than raises"; putting `needs_restart` back on the refused session answer reds `test_a_first_run_session_pick_says_why_it_was_refused`.
- The refusal assertion compares against `t('gui.model.refused')` rather than the English copy, so a wording change cannot silently pass it.

## Risk

User-visible: a first run can now set a model and is told the gateway must come back; a refused session write says so and puts the chip back, where it previously claimed success. An applied write is unchanged.

Contract: `config.set` gains an optional `needs_restart`. An older client ignoring it behaves as before. The offline fixture does not model it -- its gateway always has a loop -- and it is pinned in the fixture-shape gate's `config.set` group with that reason.

Rollback: revert either commit independently; the RPC half and the reply-reading half do not depend on each other to compile.

## Related Issues

Supersedes #511.


gloryfromca (commented):

The refactored live-session caller now turns `applied: false` into the rejection shape the picker already uses to restore the optimistic chip, while the staged-session caller reports the refusal and reloads the session model under the captured generation ticket. The catalogue addition and generated TUI output are in sync, and no finding survived review.

Covered the full target diff, both callers and their rollback/reconciliation paths, the RPC reply contract, the corresponding pre-refactor fix and review history, backward compatibility, test strength, AGENTS.md/CLAUDE.md and Web UI context rules, and the source/seam/SessionRuntime architecture constraints.

Verification: the focused model-source suite passed (31/31); web type-check, codegen drift, source-language check, and `git diff --check` passed; ESLint completed with 0 errors and 7 pre-existing warnings in untouched files. The full web run reached 2,755 passing tests and 2 failures in untouched `KnowledgePage.test.tsx`; rerunning that file alone passed 111/111, so those failures are not attributed to this change.

gloryfromca (commented):

The added first-run path is sound: it narrows the startup exception to `missing_credentials`, verifies the selected provider through the shared credential gate including external OAuth state, persists only default-scoped writes, leaves session-scoped writes as explicit refusals, and carries `needs_restart` through the RPC contract and onboarding UI. The prior live and staged session-write fixes remain intact.

Two previously deferred items remain named follow-ups rather than blockers: the refactored Settings default-model path still discards `needs_restart`, so that entry route does not proactively ask for the required restart; and `setup.status` still cannot recognize a sectionless external OAuth credential, so onboarding's recheck can disagree with the new credential gate. Late in-process stack assembly is likewise outside this change; the implemented onboarding route correctly reports that a restart is required.

Covered the revision delta and full target diff, bootstrap and turn wiring, every web model-write caller, RPC/Pydantic/OpenRPC/generated-client compatibility, the source change's review history and scope decisions, backward compatibility, test strength, AGENTS.md/CLAUDE.md and Runtime/Web UI context rules, and the source/seam/SessionRuntime architecture constraints.

Verification passed: 492 RPC/config/schema tests; 88 focused web tests; a complete web rerun with 189 files and 2,758 tests; web and TUI type checks; both RPC codegen checks; TUI i18n drift; web ESLint with 0 errors and 7 pre-existing warnings in untouched files; Ruff check and format; source-language, commit-message, and `git diff --check` gates. The first full web attempt hit one unrelated `KnowledgePage.test.tsx` race in untouched code; the complete rerun passed.

gloryfromca (commented):

This head is a clean rebase of the two reviewed patches onto the updated target: `git range-diff` pairs both old and new commits exactly, so no conflict resolution changed their substance. Re-reading the full target diff and the callers on the new base found no new defect.

The previously named Settings `needs_restart`, `setup.status` external-OAuth, and late in-process stack-assembly follow-ups remain deferred and nonblocking; this revision neither resolves nor worsens them. There are no new inline suggestions on this head.

Covered the revision delta and full target diff, callers and history, backward compatibility, test strength, AGENTS.md/CLAUDE.md and Runtime/Web UI context rules, and the source/seam/SessionRuntime architecture constraints.

Verification passed: 492 RPC/config/schema tests; 88 focused web tests; the complete web suite with 181 files and 2,517 tests; web and TUI type checks; web and TUI RPC codegen checks; TUI i18n drift; Ruff check and format; source-language, commit-message, and `git diff --check` gates. Web ESLint completed with 0 errors and 5 pre-existing warnings in untouched files. An initial concurrent test batch produced timeout failures under resource contention; the affected focused suites and the complete web suite passed when rerun serially.

0xKT (approved):

PR #525, opened 2026-09-20
revert(*): back out the web settings timezone write fix

gloryfromca:
## Summary

Reverts #506. The fix it carried edits `ui-web/src/features/settings/SettingsPage.tsx`,
a file the ui-web rebuild on `refactor/ui_web_architecture` (#475) replaces with
`SettingsApp.tsx` and its `pages/` tree, so the frontend half cannot travel across
that rewrite. The backend half goes with it so the change stays one reviewable unit:
it is redone on the refactor branch instead, where the settings page it has to edit
actually exists. #506 carries a note asking its author to resubmit there.

Nothing else landed on `main` between #506 and this revert, so the result is the
pre-#506 tree exactly.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other (revert)

## Verification

`git diff --stat 04eb8a31d 9f1fd7c73` is empty: this head's tree
(`00f113b27b7443562f83e89f27adcf8874c1dc91`) is byte-identical to the tree of
`04eb8a31d`, the commit `main` sat on before #506 merged. The unit suite,
lint, and the contract checks ran on this head in CI rather than locally.

`coverage gates` is red and stays red: the diff-coverage gate scores the 11
restored lines in the `cron.forwardChannels` validation branch, which had no test
covering them before #506 deleted them. A revert cannot raise that number without
writing tests for code it is removing from the writable set.

## Risk

The web settings timezone control goes back to failing its write with
`-32603 internal_error`, which is the behaviour every release before #506 shipped.
Rollback is re-applying #506, or the resubmission on `refactor/ui_web_architecture`.

## Related Issues

Reverts #506.


raven/rpc/methods/console.py
```diff
@@ -559,22 +559,21 @@ async def settings_set(params: dict, *, agent_loop_factory=None) -> dict:
...
-        # field names. Nothing else in this endpoint crosses the two, so the
-        # conversion belongs here rather than in a writer the CLI shares.
-        sub = to_snake(key.split(".", 1)[1])
+            try:
+                ZoneInfo(value)
+            except Exception:
+                raise ConfigValidationError(f"unknown timezone: {value}") from None
+        elif not isinstance(value, list) or not all(isinstance(x, str) for x in value):
+            raise ConfigValidationError("forwardChannels must be a list of strings")
         prev = update_cron_config(sub, value)
```
gloryfromca, line 577: Blocking: the web timezone save must continue translating its key before calling the cron writer.

`ChannelPage` sends `cron.defaultTimezone`, so `sub` is `defaultTimezone`. `update_cron_config` checks membership in `CronConfig.model_fields`, which contains `default_timezone`; at this line an ordinary edit therefore raises `KeyError` and leaves the config unchanged. This reproduces through `settings_set({"key": "cron.defaultTimezone", "value": "UTC"})` at this head. Retain the `to_snake(...)` boundary conversion or provide an equivalent supported mapping.

ui-web/src/features/settings/SettingsPage.tsx
```diff
@@ -2425,11 +2425,24 @@ function ExecPage({ s }: { s: SettingsState }): JSX.Element {
...
   const raw = s.snap.raw
+  const fwd = (V(raw, 'cron.forwardChannels', []) as string[]) || []
   return (
     <>
       <Scard title={t('gui.set.chn.card')}>
         <SwiRow label={t('gui.set.chn.progress')} k="channels.sendProgress" on={V(raw, 'channels.sendProgress', true) === true} />
         <SwiRow label={t('gui.set.chn.hints')} k="channels.sendToolHints" on={V(raw, 'channels.sendToolHints', false) === true} />
+        <TextRow
+          label={t('gui.set.chn.cron_to')}
+          k="cron.forwardChannels"
```
gloryfromca, line 2436: Blocking: the retired forward-channel control must not be restored without a live configuration contract.

`CronConfig` has no `forward_channels` field, and the loader strips both spellings because delivery is fire-at-origin. Editing this restored row sends `cron.forwardChannels` to the cron writer, which raises `KeyError` and writes nothing. Keep the row and RPC key removed; restoring the feature would require real routing semantics and schema support, not only this control.

gloryfromca (commented):
Blocking: preserve the working timezone save and keep the retired forward-channel setting removed.

This exact revert reintroduces both failures corrected by #506. Calling `settings.set` at this head with ordinary values produced `KeyError` for both keys and left the config unchanged: the cron writer accepts Python field names while the restored RPC path passes `defaultTimezone`, and `forwardChannels` has no schema field at all. The latter also conflicts with the fire-at-origin architecture and the loader migration that strips this retired key. The focused suites remain green because this revert deletes the five regression cases that covered these paths.

Coverage: I reviewed the full `github/main...HEAD` diff; AGENTS.md and CLAUDE.md; CONTEXT-MAP.md and the canonical fire-at-origin guidance; the settings UI caller, RPC boundary, cron writer, schema, loader, scheduling consumers, relevant history and #506 record; backward compatibility; test strength; architecture boundaries; and current mergeability.

Verification: `uv run pytest tests/test_rpc_settings.py tests/test_config_update.py tests/test_config_loader.py tests/test_cli_cron_commands.py -x` (232 passed); `npm test --prefix ui-web -- --run src/features/settings/SettingsPage.test.tsx` (123 passed); `npm run type-check --prefix ui-web` (passed). A direct isolated-config invocation of `settings_set` reproduced `KeyError` for both `cron.defaultTimezone = "UTC"` and `cron.forwardChannels = ["telegram"]`, with `{}` still on disk.

raven/rpc/methods/console.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

The amended PR record explicitly accepts the temporary `-32603 internal_error` regression and defers the backend and replacement-UI repair to #475. That scope decision settles this as a blocker for this revert. The fix is not yet present on #475's current head (`bc74bf7355ee` still passes `defaultTimezone` to the cron writer), so the timezone write remains a named follow-up there.

ui-web/src/features/settings/SettingsPage.tsx
gloryfromca: No blockers; suggestions only, and they are marked inline.

The amended PR record defers the whole #506 repair to the replacement settings implementation on #475. That scope decision settles this as a blocker for this revert. The retired control cleanup is not yet visible on #475's current head, so keeping `cron.forwardChannels` out of the replacement UI and RPC whitelist remains a named follow-up there.

gloryfromca (commented):

The tree is unchanged from the previously reviewed revision, so both measured regressions remain real. The amended PR record now explicitly accepts the temporary timezone `internal_error` and defers the whole #506 repair to the replacement settings implementation on #475. That author-level scope decision means neither item blocks this revert; both earlier threads have been replied to and resolved.

Named follow-up: #475's current head already omits the retired `forwardChannels` row from the replacement UI, but its RPC backend still passes `defaultTimezone` without conversion and still whitelists `cron.forwardChannels`. The resubmission there still needs the backend conversion/refusal and regression coverage before that branch merges.

Coverage for this revision: I reviewed the full `github/main...HEAD` diff and the delta from `51482dacc338`; rechecked the AGENTS.md/CLAUDE.md constraints, canonical fire-at-origin guidance, callers, schema, loader, history, backward-compatibility impact, removed tests, architecture boundaries, amended PR record, and the stated #475 follow-up branch. The new commit changes attribution/message metadata, not the source tree.

Verification: `uv run pytest tests/test_rpc_settings.py tests/test_config_update.py tests/test_config_loader.py tests/test_cli_cron_commands.py -x` (232 passed); `npm test --prefix ui-web -- --run src/features/settings/SettingsPage.test.tsx` (123 passed); `npm run type-check --prefix ui-web` (passed).

PR #527, opened 2026-09-20
fix(agent): expand shell variables before the workspace fence scans

LivXue:
## Summary

`restrict_to_workspace` is an operator's promise that the agent cannot reach outside
its workspace. It kept that promise only for paths spelled literally.

`ExecTool._check_workspace_restriction` extracted candidate paths from the command as
typed, then called `os.path.expandvars` on what it had extracted. Extraction wants a
`/` sitting on a word boundary, and `$HOME/` puts a letter there, so a variable
spelling produced no candidate at all and the expansion never saw the text that hid
the path. The expansion sat downstream of the extraction it needed to feed.

An unset name is the sharpest spelling: the shell drops it, so `$NOPE/etc/shadow` is
`/etc/shadow`, and there is no literal form to fall back on. Nothing else was looking
either, because `cat` is read-only and `default_tier` answers `allow` for a read-only
command without asking; for a read the fence was the only thing in the way.

Measured with the fence on, before this branch and after it:

| command | before | after |
|---|---|---|
| `cat /etc/shadow` | refused | refused |
| `cat $HOME/.ssh/id_rsa` | ran | refused |
| `cat "$HOME/.ssh/id_rsa"` | ran | refused |
| `cat $NOPE/etc/shadow` | ran | refused |
| `cat ${NOPE:-/etc/shadow}` | ran | refused |
| `cat ${UNSET:-$HOME/secret}` | ran | refused |
| `cat ${PWD%/*}/outside.txt` | ran | refused |
| `cat ${PWD%$PWD}/etc/passwd` | ran | refused |
| `cd /; cat etc/shadow` | ran | refused |
| `cd -- /; cat etc/passwd` | ran | refused |
| `sh -c 'cat $HOME/secret'` | ran | refused |
| `env sh -c 'cat $HOME/secret'` | ran | refused |
| `command cd /; cat etc/shadow` | ran | refused |
| `cat $PWD/notes.txt` | ran | ran |
| `echo 'keys go in $HOME/.ssh'` | ran | ran |
| `cd subdir; cd ..; ls` | ran | refused |
| `cd missing; cd ..; ls` | ran | refused |
| `cd subdir \|\| cd ..; ls` | ran | refused |
| `cd subdir \| cat; cd ..; ls` | ran | refused |
| `cd subdir & cd ..; ls` | ran | refused |
| `pushd ..; ls` | ran | refused |
| `test -d subdir && cd subdir && echo r; cd ..; ls` | ran | refused |
| `cd subdir && cd .. && ls` | ran | ran |
| `cd subdir && cd ..; cat notes.txt` | ran | ran |

The fix expands first, with the shell's own rules, then scans. Each rule is taken
from the shell rather than invented:

- an unknown name expands to nothing, as the shell does;
- the environment that decides is the child's allowlisted baseline, not `os.environ`,
  so a name the command will never be given reads as unset;
- single quotes expand nothing, so `echo 'keys go in $HOME/.ssh'` stays a sentence;
- an escaped `$` yields the bare character, which this level does not expand but a
  nested shell handed it does;
- a brace body goes back through the same pass, so a parameter inside a fallback word
  or a trim pattern is read on the same terms as one anywhere else;
- `%NAME%` is read only where cmd.exe is the shell, and there without regard to case,
  because that is what cmd does and `%VAR%` means nothing to `sh`;
- `PWD` resolves to the directory the command runs in, because a shell sets it from
  there whatever this process inherited.

A nested shell's payload is scanned in the shell that runs it, recursively, through
the policy's own embedded-shell helper and depth bound. Both readers of a segment
unwrap `env`, `sudo`, `command` and leading assignments first, through the policy's
own helper, so the fence and the deny list agree on what a segment runs rather than
keeping two lists.

Stepping out is the same escape as reaching out, so a `cd` out of the workspace is
refused as well. Only the destinations the scan cannot see needed this: `/` has
nothing after it, `..` is not absolute, and a `cd` with no argument names `$HOME` by
saying nothing. `cd /etc` was already refused, because `/etc` is a path like any
other. Each `cd` starts from where the last one landed, but only where a separator
proves it got there: `&&` runs its right side because the left one returned zero.
After anything else the move may not have happened, so both readings are held and a
later step that leaves from either is refused.

### The posture

Bypasses of one family kept appearing: everything the fence did not model was
allowed, so every gap was silent and the next one would be too. The contract is now
stated on `_check_workspace_restriction`:

**A parameter expansion is resolved faithfully or the command is refused.** The set
of spellings is finite, so the residue is closed. Substitution, case folding,
indirection, offsets and a brace this cannot parse each refuse with the construct
named. A refusal is visible and can be argued with; the failure it replaces could
not be.

**Command substitution is a declared limit, not a gap.** It holds an arbitrary
program, so no textual guard can resolve it, and refusing it would refuse
`echo "built at $(date)"` along with everything else. Paths written literally inside
one are still scanned. Two constructs are exempt from the refusal for a reason that
does not depend on modelling them: `${#NAME}` and `$((...))` yield numbers, and a
number cannot name an absolute path.

The same rule decides what the directory walk may carry. A `cd` the shell may not
have run cannot be carried, and the separator is what says whether it ran: a bracket
or a pipe puts it in a subshell, `||` runs it only when the one before failed, `;`
continues whether it failed or not. Only an `&&` chain is knowable, and the chain is
the unit rather than the step -- it stops at its first failure, so what follows one
inherits the position after any prefix of it. The walk carries those prefixes and
unions them where the chain ends.

The directory walk took three rounds, and the second and third findings were both in
code written to close the one before. Recorded because the shape repeats: the first
version advanced the walk on every `cd`, so a `cd` into a directory that does not
exist left the walk a level deeper than the shell and the `cd ..` after it read as a
return. The second read the separator after each `cd`, which closed that and four
more spellings of it (`||`, a pipe, a background `&`, a bracket) but treated `&&` as
proof the `cd` ran -- and a condition earlier in the same chain can skip it. The third
makes the chain the unit: it stops at its first failure, so what follows inherits the
position after any prefix of it.

Every one of these was measured against bash rather than reasoned about, in a
workspace with and without the directory the command names, because which branch
leaks depends on that. `pushd` was found the same way and is the one finding here
that came from neither review round.

Widening the fence from "no outside path is named" to "and the command does not walk
out" is the one deliberate contract change beyond the posture. Without it the rest is
bypassed by `cd /etc; cat shadow`.

- **The permission tier.** A read-only command is tiered `allow` and asks nobody. That
  is why this defect mattered, but it is a separate control that applies to every
  install including the ones with the fence off, so changing it is not in scope for
  making the fence keep its own promise.
- **A symlink inside the workspace pointing outside.** `cat link` names no absolute
  path, so there is nothing for a textual scan to refuse. Both sides of the comparison
  already resolve physically, so a named path cannot be laundered through a symlink.
- **A path computed at run time**, such as one decoded from base64 in a substitution.
  The sandbox executor is the boundary for both of these, which is what the code says
  about itself.

Blast radius: `_check_workspace_restriction` returns before any of this when
`restrict_to_workspace` is off, which is the default, so an install that did not raise
the fence runs none of the new code.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Branch scoped with three dots against the rebased base:
`git rev-list --left-right --count origin/main...HEAD` reports `0 14`.

- `COLUMNS=200 uv run --frozen --python 3.12 --all-extras pytest -q`
  2 failed, 23826 passed, 112 skipped in 636.80s under coverage. A second run of
  the same head gave 1 failed, 23803 passed.
- The one failure is `test_install_script.py::test_resolve_node_dir_answers_each_case_it_exists_for`
  ("node without npm provisions a runtime"). It is not introduced here: it fails
  identically with this change removed, and it fails the same way when its file is run
  alone, so it is neither this branch's nor an ordering artifact.
- A second failure is intermittent and its attribution is left open rather than
  claimed: `test_agents_research_tools.py::test_a_thin_dataset_card_falls_back_to_the_metadata_the_table_needs`
  (`KeyError: 'served_url'`). It appeared in two of three full-suite runs of this
  branch and not in the one full-suite run taken with this change removed, which is a
  single baseline sample and not enough to rule the branch out. Against that: the test
  passes when its file runs alone both with and without this change, and it passes
  when its file and the new one run together in a single process, so the new tests do
  not reach it directly. The mechanism that does fit is in that file: it selects a
  process-wide `ContextVar` 29 times and never restores the token, so its state leaks
  to whichever tests share its xdist worker, and the default `--dist load` decides
  that per run rather than by a fixed split. Adding tests changes that distribution
  without changing any code the test touches. Not changed here, since the isolation
  defect is that file's own.
- `uv run --frozen --python 3.12 --all-extras pytest tests/test_shell_workspace_fence.py -q`
  43 tests, 116 cases, all passing. Every case added for a reported bypass was watched
  failing before its fix; the rest are over-blocking guards, which pass on both sides
  by design and are there to fail if a fix reaches too far.
- Reverting only the first commit's hunk in `raven/agent/tools/shell.py`, with the new
  tests left in place, turned 13 of the then 38 cases red: every case written for the
  defect, and none of the over-blocking guards. The tree was restored afterwards and
  `git diff HEAD` confirmed empty, since a three-way reverse-apply stages what it
  applies.
- `make coverage-diff` (the gate that failed on the first revision):
  `Diff coverage: 97.89% (186/190 executable changed lines)`.
- Recursion through brace bodies terminates because a body is strictly shorter than
  the text holding it and a value is substituted rather than re-read; checked at 200
  levels of nesting and 500 parameters in one body, both answer without a
  `RecursionError`.
- Fence verdict against a real bash, fourteen shapes, each run in a workspace with and
  without `subdir`: `holes: 0   over-blocks: 0`. A hole is allowed-and-leaks; an
  over-block is refused-while-neither-branch-leaks, which is why both fixture states
  are needed to claim either.
- The walk's separator rule is held from both sides. Never ending an `&&` chain turns
  10 cases red, every escape class; ending it at every separator turns 6 red, the
  proven walks that must keep running. Dropping `pushd`/`popd` recognition turns 3
  red; recognising `popd` without its worst-case reset turns 1.
- `pytest tests/test_*shell*.py tests/test_permissions*.py tests/test_*exec*.py
  tests/test_sandbox*.py` -- 958 passed.
- `ruff check` and `ruff format --check` clean on both changed files.
- `scripts/check_commit_messages.py`, `scripts/check_large_files.py` and
  `scripts/check_source_language.py` over `origin/main..HEAD`: all pass.

The docs box is unchecked because nothing documents the fence's path semantics. The
only mention outside the code lists `restrictToWorkspace` as a config field, which
this does not change; there was no published claim describing the old behaviour.

Not reproduced here: the end-to-end refusal on Windows. A Windows path is not absolute
to a POSIX `Path`, so a verdict test would turn on the host running the suite rather
than on the expansion. The percent-expansion tests assert the expanded text and drive
the Windows branch through a class attribute instead.

## Risk

With the fence on, commands that name an outside path through a variable, that `cd`
out of the workspace, or that use a parameter expansion the fence cannot resolve are
now refused where they previously ran. The first two are the control doing what its
name says. The third is the posture, and it is the one that can refuse a command that
would have been harmless: `echo ${NAME^^}` names no path and is refused because
nothing here can prove it does not.

That trade is deliberate. An over-block is visible to the operator and can be argued
with or reported; an under-block is silent, and eight of them were found across two
review rounds on this branch before it was ready.

One shape that was refused mid-branch now runs: a `cd` sequence chained with `&&`
that returns to the workspace, such as `cd subdir && cd .. && ls`. The `;` spelling
of the same walk is refused, and correctly: when the directory does not exist the
`cd` fails, the `;` carries on from where the shell stood, and the `cd ..` leaves the
workspace. Measured against bash, it reads the parent.

Nothing changes for a default install, where the fence is off and this code does not
run.

Over-blocking was the failure mode guarded against throughout, and one instance of it
was introduced and then removed within this branch: reading `%VAR%` on POSIX refused
`echo '%HOME%/notes'`, which `sh` prints literally. Workspace-relative `$PWD` use,
single-quoted text containing a `$`, `cd` inside the workspace and back, `cd` into an
operator-added extra directory, `cd -- subdir`, a wrapper in front of inside work, a
nested payload that stays inside, a fallback that resolves inside, a trim that
resolves to a basename or matches nothing, `${#NAME}`, `$((...))`, `date +%Y%m%d`,
`printf '%s%s'` and `cut -d/` style delimiter arguments all still run, each with a
test.

Rollback is reverting these commits; the fence returns to its previous behaviour with
no migration and no persisted state involved.

## Related Issues

N/A


raven/agent/tools/shell.py
```diff
@@ -328,6 +333,147 @@ def _check_workspace_restriction(self, command: str, cwd: str) -> str | None:
...
+        A sandboxed executor runs somewhere with its own environment, which
+        nothing here can read. The values are a host-side approximation there;
+        the rule that carries is the first one, which needs no value to hold.
+        """
+        out: list[str] = []
+        quote = ""
+        index = 0
+        while index < len(command):
+            char = command[index]
+            if quote == "'":
```
gloryfromca, line 701: Blocking: nested shell command strings must be scanned in the shell that executes them.

`sh -c 'cat $HOME/secret'` still returns `None` from `_check_workspace_restriction`: the single quotes correctly suppress expansion in the outer shell, but their contents are the inner shell's program and `$HOME` expands there. The permission policy already recognizes `sh -c`/`bash -c` payloads, and a nested `cat` remains an allowed read, so with DirectExecutor this reads outside the workspace without approval. Please apply the fence to recognized nested-shell payloads as well.

raven/agent/tools/shell.py
```diff
@@ -328,6 +333,147 @@ def _check_workspace_restriction(self, command: str, cwd: str) -> str | None:
...
+                name_match = cls._VARIABLE.match(command, index)
+                if name_match is not None:
+                    value = env.get(name_match.group(1) or name_match.group(3), "")
+                    # A brace body stands beside the value rather than being
+                    # applied to it: both are candidates, because which one the
+                    # shell uses depends on a value this cannot read. The word
+                    # operator becomes a space so its word starts on a
+                    # boundary -- `${NOPE:-/etc/shadow}` is that path whenever
+                    # the name is unset, and nothing else would put it in view.
+                    remainder = cls._BRACE_WORD_OPERATOR.sub(" ", name_match.group(2) or "")
```
gloryfromca: Blocking: brace modifiers that can derive an outside path must not be treated as inert text.

With a workspace `/tmp/review-workspace`, `cat "${PWD%/*}/outside.txt"` is allowed here, but the shell removes the final component from `PWD` and reads `/tmp/outside.txt`. This implementation instead scans the in-root value `/tmp/review-workspace` and leaves `%/*` beside it, so it never sees the path the shell will use. The value-removal forms need either faithful handling or a conservative refusal when they can move a path beyond the allowed roots.

raven/agent/tools/shell.py
```diff
@@ -328,6 +333,147 @@ def _check_workspace_restriction(self, command: str, cwd: str) -> str | None:
...
+
+        for segment in segments:
+            tokens = list(segment)
+            while tokens and tokens[0] in ("(", "{"):
+                tokens.pop(0)
+            if tokens and tokens[0].startswith("("):
+                tokens[0] = tokens[0].lstrip("(")
+            if not tokens or tokens[0] != "cd":
+                continue
+            arguments = [token for token in tokens[1:] if token not in ("-L", "-P")]
```
gloryfromca: Blocking: cd parsing must skip the standard -- option separator before choosing its target.

`cd -- /; cat etc/passwd` is allowed: this code treats `--` as the destination (inside the workspace), while the bare `/` is also invisible to `_extract_absolute_paths`; the shell changes to root and reads `/etc/passwd`. Please consume `--` as an option terminator before selecting the destination (including the no-destination/home case).

raven/agent/tools/shell.py
```diff
@@ -328,6 +333,147 @@ def _check_workspace_restriction(self, command: str, cwd: str) -> str | None:
 
         return None
 
+    # `$NAME` and `${NAME}`, the two spellings that carry a path. `$(`, `$$`
+    # and `$?` are deliberately not names, so they fall through untouched.
+    _VARIABLE = re.compile(r"\$(?:\{([A-Za-z_]\w*)([^}]*)\}|([A-Za-z_]\w*))")
```
gloryfromca, line 408: Blocking: Windows percent-style variables must be expanded before the Windows path scan.

DirectExecutor uses the platform shell, so on Windows the ordinary spelling is `%USERPROFILE%\.ssh\id_rsa`, not `$HOME/...`. This matcher leaves `%USERPROFILE%` untouched and the Windows path regex sees no drive prefix, allowing the command outside the workspace. The existing fence and environment allowlist explicitly support Windows, so the expansion pass needs to cover the syntax that `cmd.exe` actually executes there.

gloryfromca (commented):
Blocking: the remaining variable and cd forms that bypass the workspace fence must be handled.

I reviewed the full diff plus the ExecTool callers, the permission policy's shell parsing, DirectExecutor and Boxlite execution, and the relevant history. I also checked AGENTS.md, CLAUDE.md, and CONTEXT-MAP.md; the branch/commit/source/test naming rules are satisfied and no new domain term needed definition. The tests are additive rather than weakened, and I checked backward compatibility on both direct and sandboxed call paths.

Local verification: `uv run --frozen pytest -q tests/test_shell_workspace_fence.py tests/test_shell_comments.py tests/test_sandbox_unit.py` passed with 166 tests. Direct probes against `_check_workspace_restriction` reproduced each inline bypass while confirming the advertised direct `$HOME` case is blocked. Because these commands still reach outside the configured roots through the exact shell features this change now interprets, this revision is not ready to merge.

raven/agent/tools/shell.py
LivXue: Reproduced and fixed in 5b4701da5.

    ALLOWED  sh -c 'cat $HOME/secret'
    BLOCKED  bash -c "cat $HOME/secret"
    BLOCKED  sh -c 'cat /etc/shadow'

The last two show the edge of the hole: double quotes expand in the outer shell,
and a literal path inside single quotes still sits on a boundary the scan takes,
so only the single-quoted name was invisible.

The payload is now scanned with the same rules, recursively, through
`_embedded_shell_command` and `_MAX_EMBEDDED_SHELL_DEPTH` rather than a second
parser, so the fence and the deny list agree on what counts as a nested shell.

One thing the fix needed beyond that: Python's shlex keeps `\$` inside double
quotes where a real shell drops the backslash, so two-deep nesting handed the
inner shell `\$HOME` and neither level expanded it. An escaped `$` now yields
the bare character. This level does not expand it, but it travels on, and the
shell handed it does.

    BLOCKED  sh -c 'cat $HOME/.ssh/id_rsa'
    BLOCKED  /bin/sh -c 'cat $HOME/.ssh/id_rsa'
    BLOCKED  sh -c 'cd /; cat etc/shadow'
    BLOCKED  sh -c "sh -c 'cat \$HOME/.ssh/id_rsa'"
    ALLOWED  sh -c 'cat notes.txt'
    ALLOWED  sh -c 'cat $PWD/notes.txt'

raven/agent/tools/shell.py
LivXue: Reproduced and fixed in 42ebd258a. With the workspace at /tmp/review-workspace:

    ALLOWED  cat "${PWD%/*}/outside.txt"
    the fence scanned:  cat "/tmp/review-workspace %/*/outside.txt"

That second line is the defect as you described it: the in-root value scanned,
the operator left standing beside it. The reasoning behind it was wrong in a
way worth naming -- `%` was filed as "reshapes a value, introduces no path",
and a trim does not introduce a path, it moves one, upward and out of the root.

You offered either faithful handling or a conservative refusal. This takes both,
and the split is the point. `%`, `%%`, `#` and `##` are now applied, the doubled
forms taking the longest match and the single ones the shortest. Every other
brace body is refused with the construct named.

Both rather than one, because they are not alike: a trim is common and
resolvable, so refusing it would cost ordinary commands, while substitution,
case folding, indirection and offsets are neither, and allowing them leaves the
fence's promise depending on which spellings happen to be implemented. That is
now stated as a contract on `_check_workspace_restriction`: a parameter
expansion is resolved faithfully or the command is refused, which closes the
residue rather than shrinking it. Command substitution stays outside that rule
and is documented as a declared limit -- it holds an arbitrary program, no
textual guard can resolve it, and refusing it would take `echo "$(date)"` too.

    BLOCKED  cat "${PWD%/*}/outside.txt"   now scans: cat "/tmp/outside.txt"
    BLOCKED  cat ${PWD%%/*}/etc/shadow
    BLOCKED  cat ${HOME#/home}/outside.txt
    BLOCKED  cat ${PWD/x/y}/z              unsupported shell expansion
    BLOCKED  cat ${PWD^^}/x                unsupported shell expansion
    BLOCKED  cat ${!HOME}/x                unsupported shell expansion
    BLOCKED  cat ${@}/x                    unsupported shell expansion
    ALLOWED  cat ${HOME##*/}/notes.txt     a basename, so it stays inside
    ALLOWED  cat ${PWD%/}/notes.txt        no match, value unchanged
    ALLOWED  echo ${#HOME}                 a count is a number

raven/agent/tools/shell.py
LivXue: Reproduced and fixed in cec183e10.

Before, with the workspace at /tmp/review-workspace:

    ALLOWED  cd -- /; cat etc/passwd
    BLOCKED  cd /; cat etc/passwd

The argument walk dropped `-L` and `-P` and then took the first remaining word,
so `--` was read as the destination, resolved inside the workspace, and passed.

`--` is now consumed as the option terminator. Past it every word is an operand,
so a `-` there names a directory rather than $OLDPWD, and no word at all means
what a bare `cd` means. The flag skip also became a leading-only walk, so
`cd -- -L` names its directory instead of losing it.

    BLOCKED  cd -- /; cat etc/passwd
    BLOCKED  cd -- ..; cat outside.txt
    BLOCKED  cd -L -- /; cat etc/passwd
    BLOCKED  cd --; cat .ssh/id_rsa
    ALLOWED  cd -- subdir && ls

raven/agent/tools/shell.py
LivXue: Reproduced and fixed in bab9ac456.

    ALLOWED  type %USERPROFILE%\.ssh\id_rsa
    the expansion pass returned it unchanged

Checked the premises before fixing rather than taking them: `direct_executor.py`
calls `create_subprocess_shell`, which is cmd.exe on Windows, and the
environment allowlist carries USERPROFILE, APPDATA, HOMEDRIVE and HOMEPATH.

`%NAME%` is now substituted, but only for a name the child is actually given.
That is cmd.exe's own rule for an undefined name, which it leaves standing as
written, and it is also what keeps a POSIX host out of the way: those names do
not exist there, so nothing is rewritten and `date +%Y%m%d` survives. The check
sits ahead of the quote branches because cmd.exe has no quoting that suppresses
it, and hiding a run it would expand behind POSIX quoting would under-block on
the platform the rule exists for.

Disclosed rather than claimed: I could not reproduce the end-to-end refusal on
Windows. A Windows path is not absolute to a POSIX `Path`, so a verdict test
would turn on the host running the suite instead of on the expansion. The test
asserts the expanded text:

    type %USERPROFILE%\.ssh\id_rsa   ->   type C:\Users\victim\.ssh\id_rsa

The verdict half is the pre-existing Windows path scan, which this does not
change.

raven/agent/tools/shell.py
gloryfromca: Blocking: standard command wrappers still hide nested shell payloads from this scan.

The bare `sh -c` reproducer is fixed, including the two-level escape case. However, `command sh -c 'cat $HOME/secret'` and `env sh -c 'cat $HOME/secret'` still return `None` from `_check_workspace_restriction`. This path calls `_embedded_shell_command` on the wrapped segment directly, while the permission policy first applies `_unwrap_command_wrappers`; DirectExecutor therefore still runs the inner read outside the workspace. This thread remains open for the wrapper forms.

raven/agent/tools/shell.py
gloryfromca: Blocking: separate shell-expansion paths still need correction before this revision can merge.

This specific finding is fixed. I verified that the original `${PWD%/*}` reproducer is now resolved to `/tmp/outside.txt` and refused, while the no-match and basename cases remain allowed. I am closing this thread.

raven/agent/tools/shell.py
gloryfromca: Blocking: separate shell-expansion paths still need correction before this revision can merge.

This specific finding is fixed. I verified `cd -- /`, `cd -- ..`, the flag-plus-terminator form, and the terminator-only home case are refused, while `cd -- subdir` remains allowed. I am closing this thread.

raven/agent/tools/shell.py
gloryfromca: Blocking: percent expansion must be restricted to Windows and use Windows case-insensitive lookup.

The exact uppercase reproducer now expands, but the implementation differs from both platform shells. On POSIX, where `HOME=/tmp/outside-home`, `echo '%HOME%/notes'` is now refused even though `sh` prints that text literally, because the percent pass runs unconditionally. On Windows, `cmd.exe` resolves environment names case-insensitively, but `%UserProfile%\\.ssh\\id_rsa` is left unchanged by the exact-key dictionary lookup and still bypasses the path scan. This thread remains open for those two directions.

gloryfromca (commented):
Blocking: wrapped shell payloads, nested fallback expansions, and platform-inaccurate percent expansion still need correction.

I re-reviewed the complete diff and the six follow-up commits, including the permission-policy helpers, DirectExecutor/Boxlite callers, relevant history, backward compatibility, and the repository rules in AGENTS.md, CLAUDE.md, and CONTEXT-MAP.md. The existing tests were not weakened; the new tests cover the four reported reproducers, and I independently confirmed the brace-trim and `cd --` fixes, replied to their threads, and resolved them.

The nested-shell and Windows threads remain open with concrete wrapper/platform reproducers, and one new nested-parameter finding is inline. Nonblocking follow-up: `_steps_outside` resolves every `cd` from the initial cwd, so `cd subdir; cd ..; ls` is refused even though it returns to the workspace; this has a straightforward compound-command workaround and does not add another merge blocker.

Local verification passed: `uv run --frozen pytest -q tests/test_shell_workspace_fence.py tests/test_shell_comments.py tests/test_sandbox_unit.py` (202 passed) and `uv run --frozen pytest -q tests/test_permissions_gate.py tests/test_shell_approval.py` (467 passed). `git diff --check github/main...HEAD` also passed.

raven/agent/tools/shell.py
```diff
@@ -326,8 +381,243 @@ def _check_workspace_restriction(self, command: str, cwd: str) -> str | None:
...
+
+        Every other body is refused. Substitution, case folding and indirection
+        can each hand the command a path this cannot compute, and allowing them
+        would make the fence's promise depend on which spellings happen to be
+        implemented -- which is the defect, not a smaller version of it.
+        """
+        if not body:
+            return value
+        if cls._BRACE_WORD_OPERATOR.match(body):
+            return f"{value} {cls._BRACE_WORD_OPERATOR.sub(' ', body)}"
```
gloryfromca: Blocking: fallback words containing parameter expansions must be expanded before scanning.

With `UNSET` absent and `HOME=/tmp/outside-home`, `cat ${UNSET:-$HOME/secret}` still returns `None` from `_check_workspace_restriction`, while the shell expands the fallback's `$HOME` and reads outside. `_resolve_brace` emits the fallback text after the main walker has passed that position, so the inserted parameter is never expanded or refused; trim patterns have the same issue (for example `${PWD%$PWD}/etc/passwd`). This contradicts the new resolved-or-refused contract. Please recursively process nested expansions in brace bodies, or refuse bodies containing constructs this pass has not resolved.

raven/agent/tools/shell.py
LivXue: Reproduced and fixed in d7e1eda3d.

    ALLOWED  command sh -c 'cat $HOME/secret'
    ALLOWED  env sh -c 'cat $HOME/secret'
    BLOCKED  sh -c 'cat $HOME/secret'

You are right about the cause, and it is sharper than the reproducers show: the
previous commit reused `_embedded_shell_command` but not the `_unwrap_command_wrappers`
that precedes it in the policy, so it reused half of the agreement it claimed to
be keeping. Both readers of a segment now unwrap first.

Writing the wrapper cases out found one more of the same shape that is not in
your list: `command cd /; cat etc/shadow` named the same builtin behind the same
wrapper and was refused only when spelled bare. The directory check unwraps too.

    BLOCKED  command sh -c 'cat $HOME/.ssh/id_rsa'
    BLOCKED  env sh -c 'cat $HOME/.ssh/id_rsa'
    BLOCKED  env FOO=bar sh -c 'cat $HOME/.ssh/id_rsa'
    BLOCKED  sudo sh -c 'cat $HOME/.ssh/id_rsa'
    BLOCKED  FOO=bar sh -c 'cat $HOME/.ssh/id_rsa'
    BLOCKED  command cd /; cat etc/shadow
    ALLOWED  env sh -c 'cat notes.txt'

The branch was rebased, so the sha in my earlier reply here has moved:
5b4701da5 is now 10e9fb3da.

raven/agent/tools/shell.py
LivXue: Both directions reproduced and fixed in c271415a5.

    BLOCKED  echo '%HOME%/notes'            <- refused, and `sh` prints it literally
    ALLOWED  type %UserProfile%/.ssh/id_rsa <- exact-key lookup missed the name

The first is mine and the reasoning I gave for it was wrong, not just the code.
The commit claimed the Windows names do not exist on a POSIX host; HOME, PWD,
USER and TMPDIR are on the executor's allowlist on both. The test offered as
proof used `date +%Y%m%d`, which passes only because `Y` and `m` happen to be
unset, so it never exercised the direction it was cited for.

`%VAR%` is cmd.exe syntax and is now read only where cmd.exe is the shell. The
lookup folds case, as cmd resolves names, and a name the child does not have is
still left standing, which is also what cmd does.

    ALLOWED  echo '%HOME%/notes'
    ALLOWED  echo %HOME%/notes
    ALLOWED  date +%Y%m%d
    ALLOWED  printf '%s%s' a b
    and with the Windows branch driven from a POSIX host:
    %UserProfile%\.ssh\id_rsa  ->  C:\Users\victim\.ssh\id_rsa
    %NOT_A_REAL_VARIABLE%/x    ->  unchanged

The platform answer is a class attribute rather than a call to `os.name`, which
is what lets the Windows branch be driven from this host. The end-to-end refusal
on Windows is still not reproduced here, for the reason given before.

Rebased since my earlier reply here: bab9ac456 is now a2f224bab.

raven/agent/tools/shell.py
LivXue: Reproduced and fixed in 73cf99394.

    ALLOWED  cat ${UNSET:-$HOME/secret}
    ALLOWED  cat ${PWD%$PWD}/etc/passwd

You have found the one spelling the contract claimed and did not cover, and the
contradiction is the right way to put it: the body was neither expanded nor
refused, just handed through. A body is emitted after the walk has passed the
position it lands in, so nothing read it again.

The body now goes back through the same pass, so a parameter in a fallback word
or a trim pattern is expanded or refused on the same terms as one anywhere else.
Taking the recursion rather than a blanket refusal keeps the resolvable cases
working, which the second group below is there to hold.

The brace pattern stops at the first `}`, so nesting is something this cannot
parse rather than something it resolves. Under the same contract that is now a
refusal instead of a fall-through to text.

    BLOCKED  cat ${UNSET:-$HOME/secret}
    BLOCKED  cat ${UNSET:-${HOME}/secret}
    BLOCKED  cat ${PWD%$PWD}/etc/passwd
    BLOCKED  cat ${UNSET:-$NOPE/etc/shadow}
    BLOCKED  cat ${UNSET:-${X:-/etc/shadow}}   unsupported shell expansion
    ALLOWED  cat ${UNSET:-notes.txt}
    ALLOWED  cat ${UNSET:-$PWD/notes.txt}

Recursion terminates because a body is strictly shorter than the text holding it
and a value is substituted rather than re-read. Checked at 200 levels of nesting
and 500 parameters in one body: both answer without a RecursionError.

LivXue:
All three blocking findings are fixed, each in its own commit, and answered in
their threads. The nonblocking one is fixed too.

`cd subdir; cd ..; ls` was refused because every `cd` resolved from the directory
the command started in. Each now starts from where the last one landed. Carrying
forward is off wherever a bracket appears: a `cd` inside a subshell does not
outlive it, and the splitter has already dropped the bracket that said so, so
carrying past one would read `(cd subdir); cd ..` as returning to the workspace
while the shell sits a level above it. Where the structure cannot be seen the
stricter reading holds, which is the same rule the expansion contract follows.
The bracket test is deliberately crude; a literal bracket in an argument costs
only strictness. Fixed in 59dd71589, with the sequences spelled as repeated
`cd ..` because a literal `../` is refused by the traversal rule before any of
this runs.

The branch was rebased onto the current base after your review, so the shas in my
earlier replies have moved:

    5b4701da5 -> 10e9fb3da   nested shell payload
    42ebd258a -> 5e1e1063b   brace trim
    cec183e10 -> b11f4fe78   cd option terminator
    bab9ac456 -> a2f224bab   percent expansion, first revision

Verification at the new head: 39 tests and 98 cases in the fence file, all
passing, with every case added for a reported bypass watched failing first;
632 passing across the shell, rpc and permission-gate files; `make coverage-diff`
reports 100.00% over 156 changed lines, which was the gate that was red on the
first revision.

One disclosure that is not a finding of yours. A second full-suite failure,
`test_agents_research_tools.py::test_a_thin_dataset_card_falls_back_to_the_metadata_the_table_needs`
(`KeyError: 'served_url'`), is intermittent: two of three full-suite runs of this
branch, and absent from the one run taken with this change removed. That is a
single baseline sample, so I am not claiming it away. Against ownership: the test
passes when its file runs alone with and without this change, and passes when its
file and the new one run together in one process. The mechanism that fits is in
that file, which selects a process-wide `ContextVar` 29 times and never restores
the token; `--dist load` then decides per run which tests share its worker. I
have not changed it, since the isolation defect is that file's own. It is written
up in the description.

raven/agent/tools/shell.py
gloryfromca: Blocking: the new cd state tracking still allows an escape after a failed or skipped cd.

This wrapper finding is fixed. I verified that `command sh -c`, `env sh -c`, assignment-wrapped shells, and `command cd /` are now refused while wrapped inside-work remains allowed. I am closing this thread.

raven/agent/tools/shell.py
gloryfromca: Blocking: the new cd state tracking still allows an escape after a failed or skipped cd.

This percent-expansion finding is fixed. I verified POSIX `%HOME%` text is left alone and the Windows branch resolves `%UserProfile%` case-insensitively while preserving an undefined name. I am closing this thread.

raven/agent/tools/shell.py
gloryfromca: Blocking: the new cd state tracking still allows an escape after a failed or skipped cd.

This nested-parameter finding is fixed. I verified fallback words and trim patterns now expand their own parameters, unresolved nested braces are refused, and inside-work fallback forms remain allowed. I am closing this thread.

raven/agent/tools/shell.py
```diff
@@ -326,8 +389,297 @@ def _check_workspace_restriction(self, command: str, cwd: str) -> str | None:
...
+            target = arguments[0] if arguments else env.get("HOME", "")
+            if not target:
+                continue
+            try:
+                destination = (here / Path(target).expanduser()).resolve()
+            except Exception:
+                return True
+            if not any(root == destination or root in destination.parents for root in roots):
+                return True
+            if carry_forward:
```
gloryfromca: Blocking: directory state must not advance past a cd that the real shell may skip or fail.

In a fresh workspace with `outside.txt` in its parent, `cd missing; cd ..; cat outside.txt` returns `None` from `_check_workspace_restriction`, then DirectExecutor prints the outside file. This code advances `here` to `workspace/missing` even though the real `cd` fails; because `;` continues from the original workspace, the next `cd ..` leaves the fence. `false && cd subdir; cd ..; cat outside.txt` has the same failure because operator control flow was discarded. The new `cd subdir; cd ..` test does not create `subdir`, so its semicolon case actually takes this escape path when executed. Please model or conservatively handle whether each directory change can run and succeed before carrying its destination forward.

gloryfromca (commented):
Blocking: cd state tracking must not assume skipped or failed directory changes succeeded.

I re-reviewed the complete diff and the four new fix commits, including the shell-policy helpers, DirectExecutor and sandbox callers, relevant history, backward compatibility, and the repository rules in AGENTS.md, CLAUDE.md, and CONTEXT-MAP.md. I verified the wrapper, platform-percent, and nested-brace fixes against their reproductions, replied to the author's explanations, and resolved all three prior threads.

The remaining blocker is new in the latest directory-state change and is marked inline. I reproduced it end to end with a temporary workspace: the fence returned allowed and DirectExecutor printed the contents of a sibling file outside the workspace after the preceding `cd` failed.

Local verification otherwise passed: `uv run --frozen pytest -q tests/test_shell_workspace_fence.py tests/test_shell_comments.py tests/test_sandbox_unit.py tests/test_permissions_gate.py tests/test_shell_approval.py` (693 passed), and `git diff --check github/main...HEAD` passed. The tests are additive, but the new semicolon `cd` case does not create its target directory, so it currently asserts allowed for the same failed-`cd` sequence that escapes in a real shell.

raven/agent/tools/shell.py
LivXue: Reproduced, and the class is wider than the two spellings you named. Against a real
bash, with `outside.txt` in the workspace's parent:

    fence    bash        command
    ALLOWED  LEAKS       cd missing; cd ..; cat outside.txt
    ALLOWED  LEAKS       false && cd subdir; cd ..; cat outside.txt
    ALLOWED  LEAKS       cd subdir | cat; cd ..; cat outside.txt
    ALLOWED  LEAKS       cd subdir & cd ..; cat outside.txt
    ALLOWED  LEAKS       pushd ..; cat outside.txt

`||` is a sixth: it runs its right side only when the `cd` failed, which is exactly
when the shell is still standing where it began. And you are right about the test --
the fixture's workspace is empty, so `subdir` never exists and `cd subdir; cd ..; ls`
was asserting that the escape is allowed.

I did not model whether a `cd` can succeed. The fence runs before execution, and
`mkdir sub && cd sub` creates its own destination, so any such model is an
approximation; one that errs toward allowing is another hole. What is knowable is the
separator: **`&&` is the only one that proves the command before it succeeded**,
because it runs its right side *because* the left returned zero. After anything else
the move may not have happened.

One position could not carry that. My first attempt kept `here` unchanged after an
unproven `cd`, and `cd a && cd b; cd ..; cd ..` still walks out: the shell may be at
`ws/a`, and two `cd ..` from there leave, while a single tracked position sees only
one of the two readings. The walk now carries the set of places the shell can be
standing in -- converging on `&&`, holding both readings otherwise -- and refuses
when any of them leaves.

That needed the separator, which `_split_on_operators` consumed and dropped. It grew
a sibling yielding each piece with the operator that terminated it, and
`_command_segments` the same; both originals are thin wrappers over the new ones, so
no existing caller reads anything new (467 permission tests unchanged). This is the
one part that reaches outside `raven/agent/`.

Two rounds, because the first version was not good enough. A whole-command predicate
(`carry unless any separator other than && appears anywhere`) closes every escape
above and refuses `cd sub && cd ..; cat notes.txt`, which cannot leave the workspace
in either branch. That was rejected and replaced with the per-separator reading, which
refuses none of these:

    ALLOWED  cd sub && cd .. && cat notes.txt
    ALLOWED  cd sub && cd ..; cat notes.txt
    ALLOWED  cd sub && cd .. && ls | head
    ALLOWED  mkdir d && cd d && cd .. && ls

Re-run against bash at the new head: every leaking spelling refused, every allowed one
confirmed not to leak, `holes: 0`.

Proved load-bearing in both directions rather than one. Making the walk always
converge (the old behaviour) turns 6 cases red -- the five escape classes and the
subshell guard. Making it never converge (the over-blocking reading) turns 6 red --
the two legitimate walks and the four cases above. Dropping `pushd`/`popd` recognition
turns 3 red; recognising `popd` without its worst-case reset turns 1.

`pushd` is a separate commit. `popd` and a bare `pushd` take their destination off a
stack this cannot follow, so the walk adds the shallowest place either can land, which
is where the command began.

Flagging rather than leaving it in the diff: `test_a_cd_that_returns_to_the_workspace_still_runs`
changed meaning. Its `;` spellings were the escape, so they move to the refusal test;
the `&&` spellings stay, and the refusal test gains `&&` cases of its own, since those
are the only ones that reach the converging branch at all.

gloryfromca: No blockers; suggestions only, and they are marked inline.

The reported failure modes are fixed. I verified the failed `cd`, skipped `cd`, `||`, pipe, background, and `pushd` cases, and the focused suite passes.

One narrower control-flow form remains for follow-up:

    test -d subdir && cd subdir && echo ready; cd ..; cat outside.txt

The fence returns `None`; when the test fails, bash skips the `cd` and `echo`, then the later `cd ..` reads the parent file. A non-`cd` segment does not carry conditional state, so the `cd`'s following `&&` collapses the position set even when a preceding `&&` skipped that `cd`.

I am not keeping this as a blocker in round six: it is not introduced relative to the target branch—`main` has no directory-state check and already allows it—and this revision materially narrows that existing class. The original objection in this thread is settled, so I am closing it.

gloryfromca (commented):

The standing directory-state blocker is fixed: the separator-aware walk now refuses failed/skipped `cd`, `||`, pipeline, background, subshell, and `pushd` escapes while retaining proven `&&` walks. The original thread is resolved.

Named follow-up: conditional state before a `cd` is still not represented. `test -d subdir && cd subdir && echo ready; cd ..; cat outside.txt` is allowed and reads the parent when the test fails. This is nonblocking at this round because target `main` already has no directory-state scan; the revision substantially reduces that pre-existing class rather than introducing it.

I covered the repository rules, the complete diff against `github/main`, this revision's delta and history, the separator parser's existing callers, backward-compatibility of the token-only wrappers, and whether tests were weakened. The changed semicolon expectation corrects an unsafe test; the proven `&&` cases remain covered. Verification: `uv run pytest -q tests/test_shell_workspace_fence.py tests/test_shell_comments.py tests/test_sandbox_unit.py tests/test_permissions_gate.py tests/test_shell_approval.py` passed (707 tests), manual fence-plus-bash reproductions confirmed both the fixes and the named follow-up, and `git diff --check github/main...HEAD` passed.

0xKT:
**Not a blocker** -- an A stands beside this, and the change is a clear improvement. But the one test
that names the recursion cap cannot fail, and it stands directly in front of a hole that is still
open.

**The fence stops looking past four nested shells, and allows.** Measured on this head, wrapping
`cat $HOME/.ssh/id_rsa` in N layers of `sh -c`:

    nest 0 .. 4  ->  refused: "path outside working dir"
    nest 5, 6    ->  ALLOWED

`_MAX_EMBEDDED_SHELL_DEPTH = 4` (raven/permissions/shell_policy.py:104), and past it
`_check_workspace_restriction` returns None rather than refusing. The boundary is its own control:
one layer either side of the cap answers differently, so this is the cap and not something else.

**This is NOT a regression, and that is why it is a note.** The same probe on the merge-base
0cb8b799 answers ALLOWED at every depth including zero -- before this change the fence never saw
`$HOME` at all. The cap and its permissive edge predate the branch. The PR strictly improves the
fence; it just leaves this one way around it, and closing it is a one-word change (refuse at the
bound instead of returning None) that is arguably out of scope here.

**What is this PR's is the test.** tests/test_shell_workspace_fence.py:551-559,
`test_a_nesting_deeper_than_the_cap_still_answers`, asserts

    assert refusal(fenced, command) in (None, *_REFUSALS)

which is the complete range of `refusal()`. No implementation can fail it. Checked by mutation, all
through the board runner:

    control: never recurse (`if _depth < _MAX...` -> `if False:`)  ->  9 failed, 103 passed
             ...and `-k nesting_deeper` alone                      ->  1 passed
    cap 4 -> 3                                                     ->  112 passed
    cap 4 -> 2                                                     ->  112 passed
    cap removed entirely (`if True:`)                              ->  112 passed
    cap 4 -> 1                                                     ->  1 failed  (a different test)

So the control proves the suite reaches the recursion -- nine tests die when it is removed -- and the
test named after the cap survives every one of those mutants, including deleting the recursion it
is named for. Nothing else in the 768-line file pins the cap's value either.

The rest of the file is strong: reverting both production files to base leaves 60 of its 112 failing,
so it does pin the change it was written for. It is this one assertion that reports coverage it does
not have, about the one part of the fence that is still passable.

Smallest thing that would close it: assert the exact answer for that input, whichever the intended
one is. If the intended answer is a refusal, the cap needs the one-word change too and the test then
pins both.

LivXue:
The named follow-up is fixed rather than carried, at 741716b4.

Reproduced first, against bash, in a workspace with and without `subdir`. Four
spellings of the guard reach the parent and the fence allowed all four:

    test -d subdir && cd subdir && echo ready; cd ..; cat outside.txt
    false && cd subdir && echo ready; cd ..; cat outside.txt
    [ -d subdir ] && cd subdir && echo ready; cd ..; cat outside.txt
    grep -q x notes.txt && cd subdir && echo ready; cd ..; cat outside.txt

You located it exactly. The `&&` after the `cd` proves it succeeded, which it
does -- but only if the `cd` ran, and a condition earlier in the same chain can
skip it. Converging there dropped the reading where the shell never moved.

The correction is that the chain, not the step, is the unit the walk carries. A
chain stops at its first failure, so what follows one inherits the position
after any prefix of it: none of it, some of it, all of it. The walk accumulates
those prefixes and unions them where the chain ends.

The part I had wrong twice is where a chain ends. It ends at the separator of
any segment, not of a `cd` -- your reproducer leaves its chain on `echo ready;`,
and the previous rule only consulted separators on `cd` segments, so it never
saw the exit. The per-segment move moved into `_moved` for that reason: the
branch that closes a chain has to run for every segment, including the ones
that used to return early.

Re-measured at the new head over fourteen shapes, each in both fixture states:
`holes: 0   over-blocks: 0`. That includes the four above, the five separator
classes from the previous round, `pushd`, and the proven walks
(`cd sub && cd ..; cat notes.txt`, `cd a && cd b && cd .. && cd ..`,
`mkdir -p d && cd d && cd ..`), every one of which still runs.

Held from both sides rather than one. Never ending a chain turns 10 cases red,
every escape class including these four; ending it at every separator turns 6
red, the proven walks. Suites: 958 passed across the shell, permissions, exec
and sandbox files; `lint-python`, `lint-imports`, `lint-types`,
`check-commits`, `check-large-files`, `check-source-language` all exit 0.

One thing I am not claiming to have closed: a directory that appears between the
check and the run. The fence is static, `mkdir sub && cd sub` creates its own
destination, and nothing here models the filesystem changing under the command.
That case is allowed and is safe for the walk's purpose, since a `cd` into a
directory created in the same chain is proven by the same `&&`.

gloryfromca (commented):

The named guard-chain follow-up is fixed. I reproduced all four forms from the author’s note: each is now refused, while the proven `&&` walks remain allowed. Tracking the chain’s prefix positions and ending it on separators from every segment addresses the specific state loss without changing the token-only policy callers. The separately recorded recursion-cap note remains nonblocking and is not duplicated here.

I covered the repository rules, the revision delta and full PR file set, the separator walker and its callers, commit history, backward compatibility, test expectations, and the existing architecture boundary. The new tests add coverage for the reported leak and do not weaken prior assertions. Verification: `uv run pytest -q tests/test_shell_workspace_fence.py tests/test_shell_comments.py tests/test_sandbox_unit.py tests/test_permissions_gate.py tests/test_shell_approval.py` passed (711 tests); manual fence checks passed for the four blocked guards and three allowed walks; `git diff --check github/main...HEAD` passed. All review threads I opened are resolved.

0xKT:
**Not a blocker** -- but it is a behaviour regression against both `main` and the
head I graded A, and I want to be explicit that I considered blocking and chose
not to. Re-review of head 741716b4; the thirteen earlier commits are `=` in
`git range-diff ecf397ef...741716b4`, so this is entirely about the new commit.

The escape it closes is real and the fix is the right shape. What follows is the
cost it charges on the other side.

## The chain collapse fires at `|`, where the and-chain has not ended

`shell.py:509` is `if separator != "&&":` and `:514` is
`here = chain = _either(chain, here)`. So a `|` ends the chain and re-admits the
pre-chain position.

But `|` binds tighter than `&&`. `A && B | C && D` is three pipelines joined by
`&&`, and the `|` sits inside the second one -- it is not a boundary between
and-chain members. Reaching `D` still requires `A` to have returned zero, so the
starting directory is not a place the shell can be, and admitting it refuses a
command that cannot escape.

Reproduced through `ExecTool(working_dir=ws, restrict_to_workspace=True).execute()`,
with the two controls that isolate the cause:

    cd build && ls | wc -l && cd ..      REFUSED
    cd build && ls && cd ..              allowed     (same shape, no pipe)
    cd build && ls | wc -l               allowed     (pipe, no later cd)

Neither control alone refuses; it takes a pipe inside the chain *and* a `cd`
after it. bash never leaves the workspace in any of the three. The same command
is allowed at `main` and at ecf397ef, so this is new with this commit.

It is an over-block and not a hole -- `_either` is a union, so the collapse can
only grow the position set, and a 167,310-command differential grid found 576
verdict changes, every one REFUSED-to-allowed and none the other way. Of those,
144 are commands bash can never put any shell outside the workspace for, and
every one of the 144 contains a `|`. The classification is exact in both
directions.

What makes it worth a comment rather than a footnote is the refusal's shape:
`_boundary_error` returns `retryable=False, blocks_call=True` and appends
`STOP_RETRY_INSTRUCTION`, so the model is told not to find another way. And this
PR's own test file states the law in its docstring: "a workspace-relative command
starts getting refused and the operator turns the fence off, which costs more
than the hole did."

## Do not take the one-line fix

`if separator not in ("&&", "|", "|&")` turns the fence suite green -- 116
passed -- and opens a hole: `cd subdir | { cd ..; cat outside.txt; }` flips from
REFUSED to allowed while real bash prints the outside file. Each pipeline element
is its own subshell starting where the *pipeline* started, so exempting `|`
also drops the pre-pipeline position. The correct fix tracks where the pipeline
began rather than skipping the collapse.

That the suite stays green under a change which demonstrably opens an escape is
its own finding: nothing in it pins `|` on the escape side except
`cd subdir | cat; cd ..; ls`, which the one-liner happens to still catch.

## Why this is not blocking

The only way to remove the over-block today is to revert the collapse, and that
reopens the escape the commit exists to close -- which I would rather have shut.
The fence is opt-in (`tools.restrict_to_workspace`, default False), and I found
no instance of the affected shape in the repo. So: merge it, and treat the
pipeline-start tracking as the follow-up.

One smaller thing while you are in there: the commit makes `_MAX_WALK_POSITIONS`
reachable by a shape that never reached it before -- a 64-long `&&` chain of
`cd` builtins now refuses at n=64, where ecf397ef did not. The comment above the
bound calls that shape unreachable for a command a person or a model writes, and
I agree it is; the margin just moved.

0xKT (approved):

PR #528, opened 2026-09-20
fix(*): keep a sub-agent's result when it lands during shutdown

ZuyiZhou:
## Summary

A sub-agent that finished while the host was shutting down lost its result. The announce path submits a SUBAGENT turn to the spine, and once the scheduler is draining that submit raises SchedulerDrainingError; nothing in the announce caught it, so the announcing task died with the exception unretrieved (asyncio's "Task exception was never retrieved" was the only trace) and the result text went nowhere. The failure path was no better: `_run_subagent`'s own `except Exception` announces the error through the same call and raised the same way out of the handler.

Two layers, both needed, then three follow-ups review asked for.

1. `SubagentManager._inject` now catches the draining refusal, logs the undelivered text in full at error level with the conversation it was for, and returns False; the four announce sites (spawn result, DAG result, DAG exception, unprompted report) skip the `subagent.delivered` marker on False, because nothing re-entered the conversation and there is no seam for a client to draw. A completed run's announce text already carries its `Record:` path, so the result stays recoverable from disk as well. Only SchedulerDrainingError is caught: the wrong-loop RuntimeError next to it is a programming error and still surfaces.

2. Three hosts sealed the spine before cancelling the sub-agents, which made the window certain rather than a race:
   - `raven serve` (`raven/rpc/bootstrap.py`) ran `turn_teardown()` before `subagents.cancel_all()`;
   - the TUI (`raven/cli/tui_commands.py`) did the same;
   - the gateway (`raven/cli/gateway_commands.py`) tore the page mount down before `cancel_all()`, so a sub-agent whose conversation lives on the page announced into an already sealed page spine.
   All three now cancel first. That is the order the gateway's own spine teardown and the generation swap (`_unbind_generation`) already kept, so no host is on a new sequence. The one-shot CLI (`agent_commands.py`) is left alone: `_wait_for_background_work` loops until no sub-agent is running before its teardown, so it has no window.

3. Cancelling first left `cancel_all()`'s snapshot no longer final, which review caught. The sweep reads `_running_tasks` once and then yields for up to five seconds draining them, and the scheduler is still accepting work in that window, so a main turn still in flight could dispatch a new sub-agent after the snapshot was taken and nothing would sweep it: the host seals the spine next, and a CLI child runs with `start_new_session`, so it would outlive the process. `cancel_all()` now closes dispatch admission before it takes the snapshot, so the snapshot is final by construction. `spawn()` refuses at the top (cheap, before the MCP preflight and before a registry row exists for a run that will never start) and again after `_write_spawn_status`, its last await, since the close can land in that window; from the second check to `_track` nothing awaits, so on a single-threaded loop whatever passes it is indexed. The gate sits inside `cancel_all()` rather than in each host because the generation swap has the same window (`_unbind_generation` cancels, then tears the page mount down), and a sweep wired per host is wiring the next host can forget.

   The admission flag is one-way, which is what every caller of `cancel_all()` means: the three host shutdowns, and the swap, which builds a new `SubagentManager` with the generation that replaces this one. A caller that means to keep serving wants `cancel_by_session` or `set_paused`, both unchanged; the docstring now says so.

4. The gate's DAG door was wrong in a way review caught and measured: `adopt_background_run` cancelled the task it was handed, but `SubAgentDagTool._dispatch` creates that task and adopts it in the same synchronous step, so the cancel landed before the task's first tick, and a task cancelled then never enters its body. `_run_detached`'s own `CancelledError` handler, the one that stops the outbox, never ran: a blocking `run_subagent_dag` dispatched after the sweep parked on `outbox.take()` for good, and a backgrounded one told the model its run had started when nothing was running and no announce was coming. Reachable on the reload path too, since the generation swap closes admission and then gives in-flight turns their grace. `adopt_background_run` now returns the refusal text once admission has closed and `None` once the task is indexed, the `str | None` contract `charge_dag_run` already has; `_dispatch` treats a refusal as its result in both modes, stopping the outbox, retiring the run now rather than a tick later, and returning the text. `charge_dag_run` refuses first once admission has closed, before the graph is minted and before a task exists, and a refused graph costs zero dispatches; the adopt door stays because the tool yields between the two. Both doors return one constant in the quota refusal's `Error:` shape.

5. Review asked, as a question rather than a block, why `spawn` read the pause gate before the shutdown gate: a host that is both paused and shutting down answered with the pause text, which tells the model to ask the user to resume, and during a shutdown nobody can. The shutdown gate reads first now, with a two-line comment on why the order matters. The same commit adds the Unreleased changelog entry this fix owed.

Not a race but a guaranteed window: on serve and the TUI any sub-agent still running at shutdown that finished during the grace period lost its result every time.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Head d526a3f6, five commits (the announce fix and the host ordering; two tests pinning the step-over of a failing cancel; the admission gate the first blocking finding asked for; the DAG refusal the second blocking finding asked for; the gate order the question asked for, with the changelog entry), rebased onto origin/main at cf3c95fd. `git range-diff` against the previous head shows the four earlier commits replayed unchanged (`=`) and the fifth differing only in where its changelog entry sits, under two entries main gained at the same place. Main's own changes to the two shutdown files since the previous base are the memory-drain report helper's rename and the control-plane port fallback, neither inside the reordered block, and the suites were rerun on the rebased head rather than carried over.

Reproduction before the fix (a submit that raises SchedulerDrainingError, then `_announce_result`): the exception escaped the announce. After the fix the same call returns, and the error log carries the full announce text.

```
COLORTERM=truecolor uv run --frozen --python 3.12 --all-extras pytest -q -p no:randomly tests/test_subagent_*.py tests/test_playbook_tool.py tests/test_rpc_bootstrap.py tests/test_cli_tui_commands.py tests/test_cli_gateway_commands.py
  2062 passed in 46.57s

COLORTERM=truecolor uv run --frozen --python 3.12 --all-extras pytest -q    (full suite, in its own worktree at head d526a3f6)
  23971 passed, 110 skipped in 327.77s

make lint-types
  All checks passed!
uv run --frozen --python 3.12 --extra dev ruff check / ruff format --check on the changed files
  All checks passed!
make check-commits; PYTHONPATH=. uv run ... scripts/check_source_language.py origin/main...HEAD; PR_TITLE=... make check-pr-title; COMMIT_RANGE=origin/main...HEAD make check-large-files
  exit 0, all four
```

Diff coverage is read from this PR's `coverage gates` job on this head. The previous head measured `Diff coverage: 96.67% (58/60 executable changed lines)` against the 90% threshold; the two uncovered lines were the moved `page_mount.teardown()` pair inside the gateway's shutdown closure, which no unit harness executes and whose order the source-order test pins instead. The last commit moves no executable line out of coverage: the swapped gate is exercised by its new test.

Revert-to-red for the first two commits, one production file at a time (`git checkout origin/main -- <file>`, run, `git checkout HEAD -- <file>`), with the commits already landed so the checkout is a real revert:

```
raven/agent/subagent/manager.py reverted   -> 7 failed  (the 4 announce paths, the 2 _run_subagent shapes, the record-path test)
raven/rpc/bootstrap.py reverted            -> 1 failed  (test_the_owning_teardown_cancels_subagents_before_the_spine_seals), 2 passed
raven/cli/tui_commands.py reverted         -> 1 failed  (test_rpc_runner_cancels_subagents_before_the_spine_seals), 2 passed
raven/cli/gateway_commands.py reverted     -> 1 failed  (test_the_gateway_shutdown_cancels_subagents_before_the_page_spine_seals), 2 passed
```

Each host's test is red only when its own file is reverted, so the three ordering tests cover three separate fixes. The two failing-cancel tests (`test_a_cancel_that_fails_does_not_keep_the_spine_from_sealing`, `test_rpc_runner_seals_the_spine_even_when_the_cancel_fails`) are not revert-to-red tests: they pass on origin/main too, because they pin that a failing cancel is stepped over rather than the ordering. Mutating each host's `except Exception` branch to re-raise turns each one red (1 failed apiece), which is the guarantee they hold. The serve and TUI tests are behavioural (the teardown steps are recorded and their order asserted); the gateway one is a source-order check, matching the existing shutdown-order test beside it, because that teardown is a closure inside a 1100-line command with no import seam.

The admission gate is verified by mutation rather than revert-to-red, since reverting `manager.py` to origin/main would take the announce fix with it and could not separate the three guards:

```
baseline                                       3 passed
MUTANT  spawn's post-await check removed       1 failed   (test_a_spawn_is_refused_when_the_shutdown_sweep_lands_while_it_is_starting)
MUTANT  spawn's top check removed              1 failed   (test_a_spawn_after_the_sweep_is_turned_away_before_it_writes_anything)
MUTANT  adopt_background_run's guard removed   1 failed   (test_a_dag_run_handed_over_after_the_sweep_is_cancelled_not_adopted)
MUTANT  `_dispatch_closed = True` removed      3 failed
CONTROL restored                               3 passed
```

The first of those tests poses the reported race directly: it drives `cancel_all()` from inside the spawn's own registry write, which is the last await before the task would be created.

The DAG refusal is verified through the real tool, the shape the second finding asked for, plus mutation and a backup-based revert of each production file to the previous head:

```
baseline                                6 passed
MUTANT  outbox.stop() removed           1 failed  (blocking: take() parks past its 1s bound)
MUTANT  adopt's return ignored          2 failed  (blocking hangs past its 5s bound; background says started)
MUTANT  adopt returns None on refusal   4 failed
MUTANT  charge's early door removed     1 failed
REVERT  dag_tool.py to previous head    2 failed  (both tool modes)
REVERT  manager.py to previous head     4 failed
CONTROL restored                        6 passed
```

`test_a_graph_submitted_after_the_shutdown_sweep_is_refused_in_both_modes[blocking|background]` dispatches through `SubAgentDagTool.execute` against a real `SubagentManager` after `cancel_all()`, with the charge door deliberately unwired so the adopt door is the one under test: the call returns the refusal inside a bound, `active_run_ids()` is empty, the manager tracks nothing, nothing is announced, and in blocking mode the outbox (captured by patching `Outbox` in the tool module) answers `take()` with `Stopped`. `test_a_dag_run_handed_over_after_the_sweep_is_refused_and_cancelled_unstarted` replaces the earlier manager test that started its task before adopting, which was the case that already worked: it now creates and adopts with no await between and asserts the body never ran. `test_a_dag_run_charged_after_the_sweep_is_refused_before_it_costs_a_dispatch` pins the early door and that both doors say the same thing. The 6 in the matrix includes two pre-existing tests the selection also matched; they stayed green throughout.

The gate order is verified by mutation, since both orders refuse and only the text differs:

```
baseline                                       3 passed  (the two admission tests above plus the new one)
MUTANT  gates in the old order, pause first    1 failed  (test_a_paused_host_that_is_shutting_down_gives_the_terminal_reason), 2 passed
CONTROL restored                               3 passed
```

The new test carries its own control: with the host paused and not shutting down, the pause text (with its ask to resume) is still what a spawn gets; after `cancel_all()` the same spawn gets the shutdown refusal with neither "paused" nor "resume" in it.

`ui-tui` is untouched (`git diff --stat origin/main...HEAD -- ui-tui` is empty), so its gate was not rerun. Integration tests were not run: nothing here reaches a real resource.

## Risk

User-visible: on serve and the TUI, sub-agents are now cancelled before the scheduler seals rather than after. Both build their spine with `build_rpc_spine`'s default `shutdown_grace` of 0.0, so no in-flight turn was being waited on either way, so the only observable change is that a finishing sub-agent's result now reaches the log instead of vanishing. On the gateway the page mount now tears down after `cancel_all()` instead of before; the generation swap has always done it in that order.

Also user-visible: a spawn or DAG dispatch that arrives once shutdown has begun is now refused rather than started into a process that is about to exit. A spawn reads `Spawn refused: the host is shutting down...`; a `run_subagent_dag` call reads `Error: the host is shutting down and is starting no more sub-agents. No sub-agent was run.` in both modes, so a blocking call returns at once instead of parking through the shutdown grace, and a backgrounded call no longer claims a run that `dag_status` cannot find. A host that is paused and shutting down says shutting down, not paused.

Rollback: revert the commits. The change is self-contained (one manager method's return type plus four call sites, three reorderings inside teardown blocks, and one reordering of two refusals).

## Related Issues

N/A


raven/rpc/bootstrap.py
```diff
@@ -391,6 +391,26 @@ async def teardown() -> None:
...
+        # adapter a run still going can hand another write. A mounted stack
+        # leaves this to its host, which owns the sub-agents and cancels them
+        # before tearing the mount down.
+        if owns_loop:
+            try:
+                from raven.acp_client.client import begin_drain
+
+                begin_drain()
+                if agent_loop is not None:
+                    await agent_loop.subagents.cancel_all()
```
gloryfromca, line 411: Blocking: prevent new dispatches from entering between this sweep and Spine teardown.

`cancel_all()` snapshots `_running_tasks` once and can then yield for up to five seconds while draining those tasks. At this point the scheduler is still accepting work, and `spawn()`/DAG admission has not been closed, so an in-flight main turn can dispatch a new sub-agent after that snapshot. `turn_teardown()` then cancels the main turn, but nothing sweeps the newly detached task before the ACP pool, MCP transports, and backend are closed; a CLI child can even outlive the host because it runs in a new session. The same reordered pattern appears in the TUI and gateway page shutdowns. Please quiesce dispatch admission before the snapshot, or perform a final sub-agent sweep after sealing the Spine.

gloryfromca (commented):
Blocking: shutdown must prevent new sub-agent dispatches from entering after the cancellation snapshot.

I reviewed the full diff and the relevant manager, scheduler, RPC/page mounting, gateway generation, and teardown paths, plus the history that established the prior ordering. I also checked AGENTS.md and the Runtime terms in CONTEXT.md, backward compatibility of `_inject`, and the test changes for weakened coverage; I found no additional issue there.

Verification: `uv run pytest tests/test_subagent_manager.py tests/test_cli_gateway_commands.py tests/test_cli_tui_commands.py tests/test_rpc_bootstrap.py -x` passed (295 tests).

One blocking race is marked inline.

raven/rpc/bootstrap.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in 6595fc283344: `cancel_all()` closes dispatch admission before taking its snapshot, `spawn()` rechecks after its last await, and late DAG adoption cancels the already-created task. The new race tests pass, so this objection no longer stands.

gloryfromca (commented):

The new terminal admission gate fixes the reported shutdown race: it closes before `cancel_all()` snapshots tasks, covers a spawn already inside an await, and cancels a DAG task that reaches adoption after closure. I reviewed the revision delta and full PR diff, all production `cancel_all()` callers, DAG task ownership, scheduler/host teardown ordering, relevant history, backward compatibility, AGENTS.md and CONTEXT.md constraints, and whether tests were weakened; no further issue remains.

Verification: `git diff --check github/main...HEAD` passed. `uv run pytest tests/test_subagent_manager.py tests/test_cli_gateway_commands.py tests/test_cli_tui_commands.py tests/test_rpc_bootstrap.py -x` passed (298 tests). A broader run including `tests/test_subagent_dag_runner.py` reached 341 passes and 2 failures before xdist stopped; both failures reproduce with `-n0` in untouched DAG-memory tests because the optional EverOS memory plugin is not installed, so they are not caused by this revision.

The prior blocking thread is resolved.

raven/rpc/bootstrap.py
ZuyiZhou: You are right, and thank you - the reorder traded the lost result for a leaked run. Fixed in 6595fc283.

I took the first of your two options, quiescing admission, but put the gate inside `cancel_all()` rather than in each host, for two reasons. The generation swap has the same window (`_unbind_generation` cancels and then tears the page mount down), so a final sweep wired into the three host shutdowns would have left that fourth caller open. And a sweep-after-seal is per-host wiring that the next host to be written can forget, which is how the ordering drifted apart in the first place.

`cancel_all()` now sets `_dispatch_closed` before it takes the snapshot, so the snapshot is final by construction.

`spawn()` checks twice, which is the part worth reviewing. The top check (beside the `_paused` one) is the cheap refusal, before the MCP preflight and before a registry row exists for a run that will never start. It is not sufficient on its own: `_preflight_mcp` and `_write_spawn_status` both await, so a close landing in that window would still reach `create_task`. The second check sits after `_write_spawn_status`, which is this method's last await, and from there to `_track` nothing awaits, so on a single-threaded loop whatever passes it is indexed and therefore reachable by the sweep. The `pending` row the spawn already wrote needs no undoing: `reconcile_instance_rows` (raven/agent/subagent/instances.py:368) reads a pending row with no live handle back as `interrupted`.

`adopt_background_run()` is the DAG's admission and its run arrives already started, so refusing it has to cancel it - tracking it would index a run the sweep has already passed, and returning without either would be the leak itself.

The flag is one-way, and `cancel_all`'s docstring now says so, because every caller means to retire the manager: the three host shutdowns, and the swap, which builds a new `SubagentManager` with the new generation (`build_runtime` -> `AgentLoop.__init__`, main.py:506, and `agent` is rebound at gateway_commands.py:972). A caller that means to keep serving wants `cancel_by_session` or `set_paused`, which are unchanged.

Three tests, each pinning one guard:

- `test_a_spawn_is_refused_when_the_shutdown_sweep_lands_while_it_is_starting` drives the sweep from inside the spawn's own registry write, which is exactly the race you described;
- `test_a_spawn_after_the_sweep_is_turned_away_before_it_writes_anything` asserts no registry row is written, which is what the top check buys;
- `test_a_dag_run_handed_over_after_the_sweep_is_cancelled_not_adopted` covers the DAG side.

Mutation matrix, since revert-to-red against origin/main would not separate them:

```
baseline                        3 passed
MUTANT  late check removed      1 failed   (sweep-lands-mid-spawn)
MUTANT  top check removed       1 failed   (turned-away-before-writing)
MUTANT  adopt guard removed     1 failed   (dag-handover)
MUTANT  _dispatch_closed = True removed     3 failed
CONTROL restored                3 passed
```

One thing I got wrong myself and fixed: the DAG test first awaited the adopted task unbounded, so with the guard mutated away it hung the suite rather than failing it - an unguarded adopt leaves the run going on its 3600s sleep. It now waits with a timeout and asserts `task.cancelled()` on the task itself.

raven/agent/subagent/manager.py
0xKT, line 1: **Blocking:** the new cancel in `adopt_background_run` lands before the task's first tick, so the
run's own cancel handler never runs. A blocking `run_subagent_dag` dispatched in that window parks
forever, and a backgrounded one is reported to the model as started.

The gate is right that a run arriving after the sweep must not be left untracked. What it does not
account for is *when* it arrives. `SubAgentDagTool._dispatch` creates the task and adopts it in the
same synchronous step:

    raven/agent/subagent/dag_tool.py:1160   task = asyncio.create_task(self._run_detached(...))
    raven/agent/subagent/dag_tool.py:1189   self._adopt(run_id, task, origin.conversation)
    raven/agent/subagent/dag_tool.py:1195   event = await outbox.take()

Nothing awaits between 1160 and 1189, so `task.cancel()` at manager.py:766 cancels a task that has
never run.

**A task cancelled before its first tick never enters its coroutine body.** That is the whole
mechanism, so I measured it rather than reasoned about it, with the control that distinguishes it:

    ### cancelled-before-first-tick: entered=[] cancel_handler_ran=[]
    ### cancelled-after-it-started:  entered=[1] cancel_handler_ran=[1]

`_run_detached`'s `except asyncio.CancelledError: outbox.stop()` (dag_tool.py:1562-1564) lives
inside that body. `stop()` is the only thing that resolves a parked taker
(dag_adjudication.py:415-425); `_retire` pops the outbox out of the index (dag_tool.py:1185) but
never stops it. So the future created at dag_adjudication.py:276 is never resolved and the
`await outbox.take()` at dag_tool.py:1195 parks for good.

**Measured through the real tool**, with `_adopt` wired to a real manager the way
raven/agent/loop/wiring.py wires it, and no await inserted by the probe:

    ### close_admission=True  -> HUNG (8s timeout)    active_run_ids=[] mgr_running=0
    ### close_admission=False -> RETURNED in ~0.01s   'DAG run ... finished: 1 completed, 0 failed,
                                                       0 cancelled, 0 skipped (of 1).'

**This PR introduces it.** With only the seven added lines at manager.py:761-767 removed and
nothing else changed, the same probe returns instead of hanging:

    ### close_admission=True -> RETURNED: 'DAG run ... finished: 1 completed, 0 failed, 0 cancelled,
                                           0 skipped (of 1).'

On the base `adopt_background_run` is a docstring and `self._track(...)`, so nothing could cancel
the task between `create_task` and `await outbox.take()`.

**The window is not only process exit.** `_unbind_generation` -- the reload path, where the process
keeps serving afterwards -- closes admission first and gives in-flight turns their grace after:

    raven/cli/gateway_commands.py:860   await agent.subagents.cancel_all()   -> _dispatch_closed = True
    raven/cli/gateway_commands.py:869   await gw_teardown()
    raven/gateway/spine.py:216          await scheduler.shutdown(grace=shutdown_grace)
    raven/spine/scheduler.py:596-599    phase 3 waits `grace` for in-flight turns; phase 4 cancels them
    raven/config/schema.py:952          shutdown_grace: float = 5.0

A turn still running during those five seconds that calls `run_subagent_dag` blocking does not
finish inside the grace: it spends all of it parked and is then cascade-cancelled, and the user
loses that turn's reply. The model does not have to choose the blocking form for this to happen --
a replan re-dispatch passes `background=not bound` (dag_tool.py:1265). The three shutdown paths
this PR edits have the same order.

**The backgrounded half tells the model the opposite of what happened.** Same probe,
`background=true`, admission closed:

    ### background-after-close -> 'DAG run ... started in the background (1 nodes). Check its
        progress with tool_call name "dag_status" ... I'll report the result when it finishes --
        keep working, and do not submit this graph again.'
    ###   tool._runs=[] mgr_running=0

`dag_status` has nothing to find and no announce is coming. `spawn`'s two new gates return
`_SHUTDOWN_REFUSAL` (manager.py:82) so the model learns; this third door returns nothing.

**The test this PR adds for the gate does not reach the production shape.**
`test_a_dag_run_handed_over_after_the_sweep_is_cancelled_not_adopted`
(tests/test_subagent_manager.py:481) does `await running.wait()` at :494 before adopting, so it
pins the case where the task has already started -- the case that works. The production caller is
the other one, and the tree already knows that window exists: the `_retire` comment at
dag_tool.py:1166-1169 names "a task cancelled before its first tick never enters `_run`" and covers
the index bookkeeping for it. It does not cover the parked taker, and the invariant three lines
below the adopt call ("the run cannot produce an event into an empty tray", dag_tool.py:1191-1193)
is the one that breaks.

What would close it, smallest first: stop the outbox on the refusal path, so the parked taker is
resolved the way a real cancellation resolves it; and hand the caller something it can see, as the
two spawn gates do. A test that dispatches through `SubAgentDagTool.execute` with admission closed,
rather than adopting a hand-rolled task that has already started, would pin both halves.

gloryfromca: Blocking: late DAG adoption must unblock foreground dispatches and return a truthful refusal.

Confirmed independently through the real `SubAgentDagTool` wired to a closed `SubagentManager`. With `background=False`, `execute()` timed out after 0.5s while `active_run_ids()` was empty and the manager tracked zero tasks. With `background=True`, it returned the "started in the background" result even though adoption had already cancelled the task and the manager tracked zero tasks. This matches the pre-first-tick cancellation analysis: `_run_detached` never reaches `outbox.stop()`, while `_retire` only removes the index entry.

The existing added test starts its task before adoption, so it does not cover the production ordering. Verification: `uv run pytest tests/test_subagent_manager.py tests/test_subagent_dag_runner.py::test_an_unstarted_background_run_retires_every_per_run_entry -x` passed (163 tests), but the real-tool probe above reproduces the missing behavior. My prior clean stance is withdrawn pending this fix.

ZuyiZhou: Confirmed, and thank you for measuring it rather than arguing it - the pre-first-tick cancel is exactly right, and the `entered=[]` control is now an assertion in the manager test. Fixed in e59a6f7ce.

I took your smallest shape, both halves. `adopt_background_run` now returns the refusal text once admission has closed, `None` once the task is indexed - the same `str | None` contract `charge_dag_run` already has, so the DAG tool learns the same way it learns about a quota refusal. `_dispatch` treats a refusal as its result in both modes: it stops the outbox the way `_run_detached`'s handler would have, calls `_retire(task)` now rather than a tick later (the pattern the foreground path already uses for a finished run), and returns the text. So the blocking call no longer reaches `outbox.take()` at all, and the backgrounded one no longer says "started in the background".

Also added the early door you implied: `charge_dag_run` refuses first once admission has closed, before the graph is minted and before a task exists, and a refused graph costs zero dispatches. The adopt door stays because `_execute` yields between charging and `_dispatch`. Both doors return one constant, `_DAG_SHUTDOWN_REFUSAL`, in the quota refusal's `Error:` shape rather than the spawn prefix, since the DAG's refusals already carry that shape.

The `TaskAdopter` alias now reads `-> str | None`. The only production adopter is the manager's; test adopters returning `None` still read as adopted.

Tests, in the shape you asked for:

- `test_a_graph_submitted_after_the_shutdown_sweep_is_refused_in_both_modes[blocking|background]` dispatches through `SubAgentDagTool.execute` with a real `SubagentManager` after `cancel_all()`, charge deliberately unwired so the second door is the one under test. Blocking: returns the refusal inside a 5s bound, `active_run_ids() == []`, manager tracks nothing, no announce, and the outbox (captured by patching `Outbox` in the tool module) answers `take()` with `Stopped` inside 1s. Background: returns the refusal, no "in the background", no tray created.
- `test_a_dag_run_handed_over_after_the_sweep_is_refused_and_cancelled_unstarted` replaces the test you flagged: task created and adopted with no await between, then asserts `task.cancelled()`, `entered == []`, and the refusal.
- `test_a_dag_run_charged_after_the_sweep_is_refused_before_it_costs_a_dispatch` pins the early door and that the two doors say the same thing.

Mutation matrix on those, plus each production file reverted to the previous head (backup-based, since the fix was uncommitted at the time):

```
baseline                                6 passed
MUTANT  outbox.stop() removed           1 failed  (blocking: take() parks past its 1s bound)
MUTANT  adopt's return ignored          2 failed  (blocking hangs past 5s; background says started)
MUTANT  adopt returns None on refusal   4 failed
MUTANT  charge's early door removed     1 failed
REVERT  dag_tool.py to previous head    2 failed  (both tool modes)
REVERT  manager.py to previous head     4 failed
CONTROL restored                        6 passed
```

(The 6 includes two pre-existing tests the `-k` selection also matched; they stayed green throughout, as they should.)

One thing worth your eye: the refusal is returned before `_dispatch` ever parks a taker, so `outbox.stop()` is not what un-hangs the blocking call in this shape - the early return is. `stop()` is there so the tray's state is truthful for anything still holding a reference, and the blocking test pins it separately (the first mutant row). If you would rather it not be called at all on a tray nothing took from, that is a one-line change and I will make it.

gloryfromca (commented):

The revision fixes the pre-first-tick cancellation defect in both directions: late adoption now returns a refusal that stops and retires the foreground outbox, and background dispatch returns that same refusal instead of claiming a nonexistent run started. The earlier charge door also refuses before task creation, while adoption remains the necessary post-await race check.

I reviewed the revision delta and full PR diff, manager/DAG callers and task ownership, scheduler and generation-shutdown ordering, backward compatibility of the adopter callback, relevant history, AGENTS.md and CONTEXT.md constraints, and whether the tests exercise the production ordering rather than weakening coverage. No further issue remains from my side. The remaining open thread belongs to its original reviewer, so I have not resolved it.

Verification: `git diff --check github/main...HEAD` passed. A real-tool foreground/background probe returned the shutdown refusal within 0.5s in both modes, never reported a start, and left no active run. `uv run pytest tests/test_subagent_manager.py tests/test_cli_gateway_commands.py tests/test_cli_tui_commands.py tests/test_rpc_bootstrap.py tests/test_subagent_dag_runner.py::TestBackgroundRun::test_a_graph_submitted_after_the_shutdown_sweep_is_refused_in_both_modes -x` passed (301 tests).

raven/agent/subagent/manager.py
```diff
@@ -819,6 +852,9 @@ async def spawn(
                 f"{SPAWN_REFUSED_PREFIX}delegation is paused. The user paused sub-agent "
                 "spawning; do the work in this turn instead, or ask them to resume."
             )
+        if self._dispatch_closed:
```
LivXue: **Question - consider reading the shutdown gate before the pause gate.** `spawn` tests
`_paused` at 849-854 and `_dispatch_closed` at 855, so a host that is both paused and
shutting down answers a dispatch with the pause refusal: "the user paused sub-agent
spawning; do the work in this turn instead, or ask them to resume." During a shutdown
nobody can resume anything, and the reason the model is given is not the reason the
dispatch was refused. Swapping the two blocks would let the terminal reason win. This is
code-reading only: I did not construct the paused-and-closing combination at runtime.

LivXue:

Scope: I reviewed only the `raven/agent/**` portion of this PR -- 2 of 10 changed files.
The remaining 8 files were read as context but are not covered by this verdict.

I reviewed the in-scope diff at e59a6f7c, both files of the sensitive zone change by
change, the five `cancel_all()` callers, the generation-swap ordering, the DAG tool's two
refusal doors, the scheduler's draining contract, and the import-seat contracts.

Confirmed findings:

- One question, no defect: `spawn` reads `_paused` before `_dispatch_closed`
  (manager.py:855), so a paused host that is also shutting down answers with the pause
  refusal and tells the model to ask the user to resume. See the inline note.

Verification:

- `tests/test_subagent_manager.py`: 163 passed; `tests/test_subagent_dag_runner.py`: 260
  passed. Both at e59a6f7c, with PYTHONPATH pointed at that revision so the run used the
  PR's code rather than the checkout.
- Mutation: `adopt_background_run` returning None instead of the refusal makes both modes
  of `test_a_graph_submitted_after_the_shutdown_sweep_is_refused_in_both_modes` fail, and
  the background mode fails on the exact "started in the background" report that 0xKT
  measured on the previous revision. The guard is load-bearing and the test reaches the
  production shape.
- Mutation: dropping `outbox.stop()` from the refusal path fails the blocking mode on its
  `take()` bound.
- `lint-imports` at this revision: 10 kept, 0 broken.

Checked and deliberately not reported: `_inject` logging the full undelivered text at
ERROR (the author weighed it explicitly, and a failure-path result has no record file to
fall back on); the one-way `_dispatch_closed` flag having no reset (every `cancel_all()`
caller retires the manager, and `_unbind_generation()` runs before `agent` is rebound to
the next generation); the late door spending a dispatch credit on a run that never starts
(reachable only during shutdown).

The pre-first-tick cancellation defect was filed by 0xKT at 05:27 and is fixed in this
revision. I reproduced the pre-fix failure independently to check the fix, but the finding
is his; this is not a second opinion on it. His thread is still open.

raven/agent/subagent/manager.py
0xKT: **Re-verified fixed** at e59a6f7c. Both halves of the finding are closed, measured with the same
probe that found it -- two ParticipantHook-free cases entered through `SubAgentDagTool.execute` with
`_adopt` wired to a real manager whose admission is closed, plus the admission-open control:

    background=false  -> "Error: the host is shutting down and is starting no more sub-agents.
                          No sub-agent was run."            (was: HUNG, 8s timeout)
    background=true   -> the same refusal                    (was: "started in the background",
                                                              for a run already cancelled)
    CONTROL, admission open -> "DAG run ... finished: 1 completed, 0 failed, 0 cancelled, 0 skipped"

so the fix changes the closed-admission path and nothing else. `tool._runs` is empty and the manager
tracks zero in both refusal cases.

The shape is right, not just the symptom. `adopt_background_run` now RETURNS the refusal rather than
cancelling silently, and `_dispatch` settles what the cancelled task owed -- `outbox.stop()` so the
taker below is not left parked on a tray nothing will fill, `_retire(task)` for the index, and the
refusal as the result in both modes. That is the same reasoning the finding rested on, and putting
it at the call site means the next caller of `adopt_background_run` inherits the obligation visibly
rather than by accident.

Nothing further from me on this thread.

raven/agent/subagent/manager.py
ZuyiZhou: Done in a9738df4, the way you suggested: `spawn` reads `_dispatch_closed` before `_paused`, with a two-line comment on why the order matters, since nothing else about the two blocks says it.

Your reading was right, and I constructed the combination you did not: `test_a_paused_host_that_is_shutting_down_gives_the_terminal_reason` sets the pause, spawns once as a control (the pause text, with its ask to resume), then runs `cancel_all()` and spawns again, asserting the shutdown refusal with neither "paused" nor "resume" in it. With the two gates put back in the old order that test is the one that fails (1 failed, 2 passed among the three admission tests); restored, all three pass.

Same push: rebased onto current main (`git range-diff` shows the four earlier commits replayed unchanged), and the Unreleased changelog entry this fix owed. Full suite on the rebased head: 23794 passed, 110 skipped.

gloryfromca (commented):

The tip correctly gives the terminal shutdown refusal precedence over the reversible pause refusal, while the paused-only control preserves the existing resume guidance. The changelog accurately describes the full shutdown-result fix. I reviewed the tip delta and full PR diff, the spawn and DAG admission ordering, all retirement callers and relevant history, backward compatibility, AGENTS.md and CONTEXT.md constraints, and whether the added test preserves prior coverage. No new issue remains from my side; threads opened by other reviewers remain theirs to close.

Verification: `git diff --check github/main...HEAD` passed. `uv run pytest tests/test_subagent_manager.py tests/test_cli_gateway_commands.py tests/test_cli_tui_commands.py tests/test_rpc_bootstrap.py tests/test_subagent_dag_runner.py::TestBackgroundRun::test_a_graph_submitted_after_the_shutdown_sweep_is_refused_in_both_modes -x` passed (302 tests). `make check-large-files` could not start because `make` is unavailable; its underlying command, `PYTHONPATH=. uv run --frozen --python 3.12 --extra dev python scripts/check_large_files.py github/main..HEAD`, passed.

ZuyiZhou:
Rebased onto main at cf3c95fd to clear the changelog conflict: this branch and main had both inserted at the top of Unreleased / Fixed, so the only manual resolution was the order of the entries; nothing else conflicted. `git range-diff` against a9738df4: four commits replayed `=`, the fifth differs only in that changelog hunk. Main's changes to the two shutdown files since the previous base (the memory-drain report helper's rename, the control-plane port fallback) sit outside the reordered block.

Rerun on the rebased head d526a3f6: the focused suites (`tests/test_subagent_*.py tests/test_playbook_tool.py tests/test_rpc_bootstrap.py tests/test_cli_tui_commands.py tests/test_cli_gateway_commands.py`) 2062 passed; the full suite 23971 passed, 110 skipped in its own worktree; lint-types, lint-imports, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. The description's head and numbers are updated to this head.

gloryfromca (commented):

Reviewed the full `github/main...HEAD` diff at d526a3f6ed66 after the rebase. `git range-diff` shows four commits patch-identical to the previously reviewed series and the remaining difference confined to neighboring changelog context. I also rechecked the repository rules and runtime vocabulary, shutdown and generation-swap callers, DAG adoption and announce contracts, backward compatibility, and whether the tests were weakened; I found no new error.

Verification: the focused shutdown suite passed (308 tests); `git diff --check`, the source-language gate, and the large-file gate all passed. The original review thread is resolved, and I did not alter anyone else's threads.

gloryfromca (commented):

This head only merges current `main` into the branch. I read the current full diff and the two incoming main commits: all five feature commits compare `=` under `git range-diff`, and the old and new PR diffs have the same stable patch ID. The incoming routing/attachment work does not alter the shutdown callers or contracts.

Coverage: repository rules and updated runtime vocabulary; the diff, callers, and history; backward compatibility; test strength; and the shutdown/admission architecture. Verification: 363 focused shutdown and routing tests passed; `git diff --check`, the source-language gate, and the large-file gate passed. My review thread remains resolved.

LivXue (approved):

PR #529, opened 2026-09-20
chore(*): catch the ui-web rebuild up to main

gloryfromca:
## Summary

`refactor/ui_web_architecture` last shared a commit with `main` at `5a29ea22`
(#451). This merges `main` (`dbe259f6`) into it so the conflicts get settled
once, now, instead of all at once under #475, and so CI runs on the merged tree
rather than on a branch GitHub already reports as `CONFLICTING`.

Seven paths conflicted. How each was settled:

- `ui-web/src/features/settings/SettingsPage.tsx`, its test, and
  `ui-web/src/live/120-settings.js` -- deleted on this branch, edited on `main`
  (modify/delete, so git leaves `main`'s copy in the tree). The rebuilt settings
  surface replaces all three; the deletions stand.
- `i18n/messages.json` and `ui-tui/src/i18n/messages.generated.ts` -- `main`'s
  only change to either since the merge base is #506 dropping
  `gui.set.chn.cron_to` and `cron_ph` (8 and 4 lines, no additions). This branch
  already drops the whole `gui.set.chn.*` group, so its side is the superset and
  wins.
- `raven/skill_hub/hub.py` -- each side added one import from
  `skill_hub.audit`. `write_install_meta` is called at line 621,
  `INSTALL_META` at line 699, and both exist in the merged `audit.py`, so both
  are imported.
- `ui-web/build.py` -- `main` rewrote the `newline=""` comment, this branch
  added `flush=True` to the print under it. Both kept.

What came in quietly and is worth knowing: #506's cron timezone fix auto-merged
into `raven/rpc/methods/console.py` and brought its five cases in
`tests/test_rpc_settings.py` with it. The rebuilt settings pages have no cron
section at all, so on this branch the backend can now write
`cron.defaultTimezone` correctly and nothing on the page does it. Whether that
control gets ported, and in which PR, is a separate decision.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other (catch-up merge)

## Verification

Run in a worktree on the merged commit:

- `uv run pytest tests/test_rpc_settings.py tests/test_skill_hub_client.py tests/test_skill_hub_policy.py tests/test_skill_hub_source.py tests/test_skill_hub_tools.py tests/test_i18n_boundary.py -q` -- 174 passed.
- `npm test` in `ui-web` -- 181 files, 2513 tests, all passed.
[...]

PR #531, opened 2026-09-20
ci(*): keep commit lint off the default branch's history

gloryfromca:
## Summary

A pull request whose base is not the default branch carries, in its
`BASE..HEAD` range, every commit the default branch has gained since the two
diverged. Both commit linters read that range, so they re-read history that
already passed this gate when it landed and that nobody can rewrite now.

#529 is the case that surfaced it: merging `main` into
`refactor/ui_web_architecture` puts 38 `main` commits in range, one of which
(`b01534d0`, #471) carries a `Co-authored-by` trailer with a non-ASCII name.
The gate reports that PR red for a message it did not write. The same will
happen to every future catch-up merge.

`scripts/check_commit_messages.py` now takes several revision arguments and
passes them to `git log`, so a caret exclusion works. Commitlint has no
equivalent flag, so the workflow iterates the same revision list and lints one
message at a time. On a pull request the workflow excludes the repository's
default branch; a push to the default branch is unchanged and still lints what
it pushed.

This does not touch the offending commit on `main` and does not relax a rule:
a commit that a PR actually introduces is linted exactly as before.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [x] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

`uv run pytest tests/test_check_commit_messages.py tests/test_commit_lint.py -q`
-- 15 passed. Three of those cases are new and cover a repo where the default
branch carries a rejected message and the feature branch merges it in: the
range fails without an exclusion, passes with one, and still fails when the
bad commit is the branch's own.

Mutation check on the new cases: replacing `_commit_messages(args)` with
`_commit_messages(args[:1])`, which is the behaviour before this change, turns
`test_excluding_the_default_branch_skips_the_history_it_carries` red and leaves
the other three green. The pin fails when the fix is removed.

Against the real history behind #529, with the checker run directly:

- `check_commit_messages.py "bc74bf73..a0666244c"` -- exit 1, naming
  `b01534d02` exactly as CI does.
- the same range plus `^github/main` -- exit 0.

The workflow's new loop was run as written against both ranges: 0 commits to
lint on the catch-up range, and a planted `Update stuff.` commit on a feature
branch still sets the failure flag.

`make lint-python` -- clean. `scripts/check_source_language.py` over the branch
-- exit 0; the test's non-ASCII message is built from an escape, so the file
itself stays ASCII.

## Risk

Commitlint now starts once per commit instead of once per range, so a PR with
many commits spends a second or so more in this job. The exclusion only ever
removes commits that are already on the default branch, so nothing a PR
introduces escapes either linter.

Rollback is reverting this PR; the previous behaviour is one range argument and
one commitlint invocation.

## Related Issues

N/A. Unblocks the `commit messages` check on #529.


gloryfromca (commented):

The pull-request-only exclusion gives both commit-message checkers the same effective revision set, while the default-branch push path remains unchanged. Branch-owned commits are still linted and merge commits are still skipped.

Verification:
- `uv run pytest tests/test_check_commit_messages.py tests/test_commit_lint.py -q` - 15 passed.
- After `npm ci`, the workflow's exact per-commit loop against this PR base and `^github/main` linted both branch commits with `failed=0`; the Python checker exited 0 on the same revisions.
- `git diff --check github/main...HEAD` - clean.
- GitHub's `commit messages` and `repository files` jobs pass on this head.

Coverage included the diff, workflow callers and history, push and command-line backward compatibility, test-strength checks, and the applicable AGENTS.md/CLAUDE.md source, comment, commit, and test rules. CONTEXT architecture constraints are not applicable because this change introduces no runtime domain term or layer-boundary change.

0xKT (approved):

PR #532, opened 2026-09-20
refactor(agent): ask a role once with the turn's whole roster

ypflll:
## Summary

- Gather every active `ParticipantHook` seat for a phase and pass the complete registration-ordered roster to the Memory, Planning, Capability, and Action roles in one delegated call.
- Preserve default composition, diagnostic trails, nested product axes, addendum replacement, and outbound transformation behavior.
- Isolate participant factory and axes discovery failures, and always clear phase roster state.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [x] Refactor
- [ ] Other

## Verification

- `uv run ruff check --no-cache raven plugins-dist agents tests` - passed.
- `uv run ruff format --check raven/agent/hook/composite.py raven/agent/hook/participant.py tests/test_participant_hook.py tests/integration/test_participant_roster_e2e.py` - 4 files already formatted.
- `uv run lint-imports` - 10 contracts kept, 0 broken.
- `uv run pytest -q tests/test_participant_hook.py tests/test_agent_hook_contract.py tests/test_core_stacks.py tests/integration/test_participant_roster_e2e.py tests/integration/test_subagent_charter_e2e.py tests/integration/test_agent_playbook_e2e.py` - 124 passed after rebasing onto the latest `main`.
- The full default suite was also compared against the same-HEAD baseline in the same environment. Error counts were identical; one existing research test fluctuated under parallel execution and passed in isolation.
- Documentation and screenshots are not applicable because this refactor has no user-facing UI or configuration change.

## Risk

The grouped role decision runs at the last active participant position. An interleaved non-participant hook before that position therefore runs before the grouped decision; a regression test pins this ordering. A failing participant factory is omitted from the roster without blocking the remaining participants. Rollback is a revert of this PR.

## Related Issues

N/A


raven/agent/hook/participant.py
```diff
@@ -376,46 +440,51 @@ def _splice_addendum(self, ctx: Any, text: str, seat: "_Seat") -> None:
...
     async def before_execute_tools(self, ctx: AgentHookContext) -> HookDecision:
-        seat = self._seat(ctx)
-        return self._with_trail(
-            seat, self._decide(await self._review(self._step(ctx, phase="execute_tools"), seat.participant))
+        seats = self._claim_roster(ctx, "before_execute_tools")
+        if seats is None:
+            return HookDecision()
+        participants = tuple(seat.participant for seat in seats)
+        return self._with_trails(
+            seats, self._decide(await self._review(self._step(ctx, phase="execute_tools"), participants))
```
gloryfromca, line 449: Blocking: a raised participant call must not prevent the rest of the roster from running.

The default composers do not catch exceptions per participant. Now that the whole roster is invoked inside the claimant hook, an exception from the first participant escapes the combined role call; `CompositeHook` catches that claimant hook as a no-op, and every later participant is skipped. I reproduced this with `[Broken, Later]`: only `Broken.review()` ran, so `Later` could not return its `End` verdict. Before this change, each seat was a separate composite child, so the first exception was isolated and the later seat still ran. This also violates the documented participant invariant that an answer which raises counts as silence. Please retain per-participant failure isolation when grouping the roster.

gloryfromca (commented):
Blocking: preserve per-participant exception isolation in grouped role calls.

I found one concrete regression. The focused verification passed: `uv run pytest -q tests/test_participant_hook.py tests/test_agent_hook_contract.py tests/test_core_stacks.py tests/integration/test_participant_roster_e2e.py tests/integration/test_subagent_charter_e2e.py tests/integration/test_agent_playbook_e2e.py` (`124 passed in 10.06s`). I also ran a direct two-seat failure-path reproduction; it confirmed that the later seat is skipped.

I reviewed the complete diff and covered the participant/composite callers, default harness composers, assembly ordering, the pre-change behavior and file history, backward compatibility, test strength, and the applicable AGENTS.md and CONTEXT.md architecture constraints. I found no weakened tests or additional actionable defects after trying to refute the candidates.

gloryfromca (commented):
Blocking: the existing participant-exception isolation defect remains unresolved.

This head only merges the latest `main`; the effective four-file PR patch is unchanged from the prior revision. The focused suite still passes (`124 passed in 6.59s`), but the targeted two-seat reproduction still runs only the raising participant and skips the later participant, so the open inline finding remains valid. I found no new findings.

For this revision I checked the merge delta and full effective diff, the participant/composite callers and default harness composers, relevant history and backward compatibility, test strength, and the applicable AGENTS.md and CONTEXT.md constraints.

raven/agent/hook/participant.py
ypflll: Fixed in the latest head. Every default participant composer now catches failures per participant and treats that answer as silence, so the rest of the roster still runs. The same isolation covers intake, advice, review, salvage, judge, system addendum, outbound, archive, and tool selection. A focused `[Broken, Later]` regression proves that a raising `Broken.review()` does not prevent `Later.review()` from returning `End`. Local verification after rebasing onto current `main`: 130 focused tests passed, Ruff and formatting passed, and all 10 import contracts were kept.

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed at `4b075c5fcbc0`: the original `[Broken, Later]` reproduction now reaches both participants and preserves `Later`'s terminating verdict. I also ran the expanded focused suite, including the synchronous judge composition path: `178 passed in 6.72s`. The failure-isolation finding is fixed.

gloryfromca (commented):

The per-participant failure guards restore the documented silence-on-error behavior across the default composers, including synchronous judge composition and the role-less outbound path. The original two-seat reproduction now reaches the later participant and preserves its terminating verdict. Expanded focused verification passed: `178 passed in 6.72s`. I found no new findings.

I reviewed the fix delta and complete effective diff, participant/composite and harness callers, relevant history, backward compatibility, test strength, and the applicable AGENTS.md and CONTEXT.md architecture constraints. The prior finding has been replied to and its thread is resolved.

gloryfromca (commented):

This revision is a rebase onto current `main`; the five-file participant patch has the same stable patch ID as the previously reviewed clean head. The prior thread remains resolved, and the expanded focused suite passed again: `178 passed in 7.44s`. I found no new findings.

I checked the rebase delta and complete effective diff against the refreshed GitHub base, relevant callers and history, backward compatibility, test strength, and the applicable AGENTS.md and CONTEXT.md constraints.

gloryfromca (commented):

This is another rebase-only revision. Against GitHub's refreshed `main`, the effective five-file participant patch has the same stable patch ID as the previously reviewed clean head. The prior thread remains resolved, and expanded focused verification passed again: `178 passed in 6.76s`. I found no new findings.

I checked the rebase delta and complete effective diff, relevant callers and history, backward compatibility, test strength, and the applicable AGENTS.md and CONTEXT.md constraints.

0xKT:
**Not a blocker** -- an A stands beside this. The isolation this PR establishes has one hole, and it
is in the half the helper does not cover: a participant whose `drain_trail` raises now throws away
the whole roster's decision, where before it cost only its own seat.

`_with_trail` (raven/agent/hook/participant.py:244-258) calls the drain outside any try:

    drain = getattr(seat.participant, "drain_trail", None)
    trail = drain() if callable(drain) else []

`_answer`/`_answer_sync` (raven/agent/harness/participants.py:66-79) guard the nine verbs, and
nothing guards this. At the base each seat was its own composite child, so a raising drain was
caught per hook. At this head `_with_trails` (:260-264) drains every seat onto the ONE decision the
claimant returns, so the first bad drain discards what another seat already produced.

**Measured, with the control that distinguishes it.** Two `ParticipantHook`s in a `CompositeHook`,
entered at `before_execute_tools` -- the same method the turn path calls -- with no await inserted
by the probe. The roster is [bad -> Accept, later -> End("closed")]:

    bad.drain_trail raises  -> short_circuit_result=None     events: review:bad, review:later, drain:bad
    bad.review raises       -> short_circuit_result='closed'  (CONTROL: the nine-verb guard works)
    nobody raises           -> short_circuit_result='closed'  (CONTROL)

Note `drain:later` never runs either, so the second seat's own trail is lost as well.

**This PR introduces it.** The same probe file, unchanged, on the base tree (`git archive` of the
merge-base into a scratch dir, run through the same runner): all three cases answer `'closed'`.
`_with_trails` does not exist at the base.

**Why this is a note and not a blocking finding.** `drain_trail` has exactly one definition in the
repository -- `raven/contracts/participant.py:241`, `return self.__dict__.pop("_trail", [])` -- and
`dict.pop` with a default cannot raise. No in-tree participant overrides it, so nothing that ships
here can reach this today. It needs an out-of-tree participant.

That population is precisely the one the guard beside it was written for, which is the reason to
mention it at all. The docstring at :247-252 says the `getattr` is there because "a participant that
implements the nine verbs without inheriting `AgentParticipant` is a participant, and asking it for
a method it never claimed would raise into the composite's catch -- which would log the plugin as
broken and drop the answer it just gave". The absent method is handled; the raising one is not, and
it now drops more than it used to. CONTEXT.md:179 states the same invariant from the other side:
"an answer that raises or does not parse counts as silence rather than as a refusal."

Smallest thing that would close it: wrap the drain the way `_answer` wraps a verb -- one try around
`drain()`, the trail treated as empty on a raise, and a log line naming the participant. A test in
the shape of the probe above would pin it.

gloryfromca:

I independently reproduced @0xKT's `drain_trail` case at this head: both participants review, then the first drain raises, the later participant's `End` decision is discarded, and its drain is skipped. The expanded focused suite still passes (`178 passed in 6.71s`). Because the repository's only `drain_trail` implementation cannot raise and no in-tree participant overrides it, this remains an out-of-tree compatibility follow-up rather than a reason to hold the merge.

That evidence changes my standing stance from `clean` to `nonblocking`; the revision can merge as it stands. My previously resolved thread remains settled.

0xKT (approved):

Issue #530, opened 2026-09-20
docs: host the EverMind ecosystem banner


LivXue:
Hosting issue for the banner shown at the top of the README's EverMind Ecosystem section. Assets are attached as comments on this issue and referenced from README.md and README.zh-CN.md by absolute URL, since image files cannot be committed to the repository.

LivXue:
EverMind ecosystem banner for the README ecosystem section (2172x724).

![EverMind ecosystem banner](https://github.com/user-attachments/assets/bb13d4d5-e254-4ba0-9ed0-b2c0ccd4bc6a)

LivXue:
Cropped to a banner: 2172x511 (4.25:1), down from 2172x724 (3.00:1). Removes the empty field above and below the orbital composition; the composition itself is untouched.

![EverMind ecosystem banner, cropped](https://github.com/user-attachments/assets/6c4392ea-2f82-42c7-acde-8dfecb2e6c2c)

ZuyiZhou:
<img width="2000" height="1568" alt="Image" src="https://github.com/user-attachments/assets/10679855-e2c5-4422-81c3-17cd784123ce" />

LivXue:
Recompressed copies of the README assets hosted in this thread, for README.md and README.zh-CN.md. Every still is now under 400 KB. The old comments stay as they are; each item names the asset it replaces and how it was re-encoded.

- replaces `6c4392ea-2f82-42c7-acde-8dfecb2e6c2c`: 1.40 MB -> 385 KB, JPEG quality 96, 4:4:4 chroma, same size

  ![The EverMind ecosystem: the EverMind mark and its slogan on an orbital field](https://github.com/user-attachments/assets/65a5e3f1-dccb-496f-94f5-9782070ec8b4)

- replaces `10679855-e2c5-4422-81c3-17cd784123ce`: 539 KB -> 180 KB, 256-color PNG, same size

  ![Retrieval sweep dashboard: recall@k is set by top_k alone and latency stays broadly flat, with a sixteen-cell grid of measured recall and latency and the best cell ringed at top_k 10, chunk_size 1024](https://github.com/user-attachments/assets/9e0ee7b0-4265-4b84-bb50-8034406894b3)

PR #534, opened 2026-09-20
docs: restyle the README agent sections and add an ecosystem banner

LivXue:
## Summary

A presentation pass over both READMEs. No prose was rewritten and no benchmark
number changed; all four changes are about how the Built-in Agents and EverMind
Ecosystem sections read.

1. The four agent subheadings (Raven-Research, Raven-Code, Raven-Design,
   Raven-Oncall) now carry a single chevron, one step lighter than the double
   chevron the section headings already use. That gives depth by weight without
   introducing a second symbol family.
2. The six Raven-Code, Raven-Design and Raven-Oncall benchmark charts render at
   95% of the line instead of full width. Both READMEs reference the same six
   assets, so the width is set in each file or the two language pages disagree
   about the same picture.
3. The EverMind Ecosystem section now opens with a brand banner, cropped to
   2172x511 (4.25:1) from the 3.00:1 original. The empty field above and below
   the orbital composition is gone and the remaining margin is equal on both
   sides, which the original was not. The composition itself is untouched.
4. The two Raven-Oncall captions state the result their chart shows rather than
   naming the comparison.

- The two Raven-Oncall charts label the same model differently: DS-V4.1-Flash
  on AI4AI, DS-V4-Flash on AI4S. Pre-existing, and correcting it means
  regenerating a chart rather than editing text.
- The Raven-Research chart keeps its full width, and is now the only chart in
  that section which does.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Gates, each against a non-empty branch range:

    make check-large-files              exit 0
    make check-source-language          exit 0
    scripts/check_commit_messages.py    exit 0

Tests:

    uv run --frozen --all-extras pytest tests/test_living_docs.py \
      tests/test_scope_canon.py tests/test_commit_lint.py -q
    15 passed

Assets: for each uploaded image the published URL was fetched with GET and its
sha256 compared against the local file. Both matched. HEAD is not usable for
this check, because user-attachments answers 403 to it whether or not the
upload succeeded.

Greps:

- The superseded banner asset id has zero occurrences repo-wide.
- The four agent heading anchors have zero references in either README, in the
  documentation site, or in the workflows, so the slug change the chevron
  prefix causes breaks nothing that links to them.

The first two boxes are blank on purpose. No test reads the two root READMEs:
test_living_docs globs the package READMEs, test_scope_canon reads a
documentation-site page, and test_external_consumer_surface reads a different
README. The three suites above ran because they are the nearest documentation
guards, not because they cover this change. Repository lint does not cover
Markdown.

## Risk

User-visible: the four agent subheadings gain a symbol, so their generated
anchors change; six charts render slightly narrower; the ecosystem section
gains a banner; two captions assert a result. Nothing outside the two README
files changes.

Security: no credential, endpoint or permission is touched. The uploaded image
is a brand banner, public by intent.

Backward compatibility: the heading anchors are the one break. Nothing in the
repository links to them, but an external bookmark to one of the four would
stop resolving.

Rollback: revert the commit. Images are attachments rather than committed
files, and superseded assets stay published, so any link already shared keeps
resolving.

## Related Issues

#530 hosts the ecosystem banner assets and stays open for that purpose.


gloryfromca (commented):

Reviewed the complete `github/main...HEAD` diff and both rendered README variants. The heading hierarchy renders as intended, the documented permalink change has no repository callers, the benchmark claims match the charts, and the banner URL resolves to the stated 2172x511 image.

Coverage included the applicable AGENTS.md rules (Markdown language exemption, externally hosted report assets, branch and commit conventions), CONTEXT-MAP terminology, surrounding README history, backward compatibility, test-integrity checks, and architecture boundaries; this documentation-only change does not cross a runtime or UI boundary, and no tests were weakened.

Verification: `git diff --check github/main...HEAD` passed; `check_large_files.py github/main..HEAD`, `check_source_language.py github/main..HEAD`, and `check_commit_messages.py github/main..HEAD` passed; 17 focused documentation and README-contract tests passed. The environment has no `make` executable, so I ran the two underlying repository gate scripts directly.

No findings.

ZuyiZhou (commented):
Not approving yet -- one of the two new captions is contradicted by the chart directly beneath it.

**The AI4S caption.** It now reads "Raven-Oncall significantly outperforms Claude Code on both success rate and cost for AI4S tasks". The chart's Success Rate panel has three bars:

- Claude Code / Opus-5: 64.71%
- Raven-Oncall / Opus-5: 82.35%
- Raven-Oncall / DS-V4-Flash: **41.18%**

The third bar is a Raven-Oncall configuration scoring 23.53 points *below* Claude Code on the exact axis the caption claims it significantly outperforms on. The caption says "Raven-Oncall" without qualification, the chart shows two Raven-Oncall series, and a reader who looks from the sentence to the picture sees the opposite of what they were just told. Cost is fine -- $13.60 to $5.10 to $0.10 supports the claim on both series.

The old caption named the comparison and was accurate. Suggested fixes, either works: qualify the configuration ("Raven-Oncall with Opus-5 ..."), or keep the assertion to the axis where both series win ("... significantly outperforms Claude Code on cost, and on success rate with Opus-5").

**The AI4AI caption** is defensible but the adverb is doing uneven work. Both Raven series beat Claude Code on BPB, so nothing is contradicted, but the quality margin is 1.053 to 1.037 -- about 1.5% -- while cost runs $50.00 to $0.97. "Significantly ... on both quality and cost" earns the word decisively on one axis and thinly on the other. Worth a second look while you are in there; not a blocker on its own.

Everything else in the pull request checks out and I would have approved on it alone:

- Six charts move 100% to 95% in each README, and all seven assets this touches appear exactly once in both files, so the two languages cannot disagree about the same picture.
- The ecosystem banner returns HTTP 200 and its PNG header reads 2172x511, a ratio of 4.25:1 -- the crop is what the description says it is.
- The four agent heading anchors have zero references anywhere in the tree outside `docs/plans` and `docs/specs`, so the slug change from the chevron prefix breaks no internal link. The external-bookmark caveat in the Risk section is the right thing to have flagged.
- Leaving the first two verification boxes blank and saying why is correct: no suite reads the root READMEs.

Happy to approve as soon as the AI4S caption matches its chart.

Handsome-wzw (approved):

PR #535, opened 2026-09-20
feat(ui-web): redraw the raven mark the browser tab and the processor row share

LivXue:
## Summary

Replaces the drawing in `ui-web/icon/raven.svg`. Nothing else moves: no path,
no stylesheet rule and no caller changes.

That one file has two consumers, because `ui-web/build.py` copies it over
`dist/assets/raven.svg` after the rest of the asset tree is in place:

- the browser tab icon, through `ui-web/src/page.html`;
- the mark on the first row of the file-processing list in the knowledge base
  settings dialog, through `ravenIconPath()` in
  `ui-web/src/shell/provider-mark.tsx` and its one call site in
  `ui-web/src/features/knowledge/KnowledgePage.tsx`.

The second one is easy to miss, so it was checked rather than assumed. The row
sizes the mark at 17px through `.kbprocr .provider-icon` in
`ui-web/src/styles/page.css`, not at the 20px the base `.provider-icon` rule
gives, and it is the first row, so it renders at full opacity. The image
carries no `data-tone` attribute, so none of the six `invert(1)` rules in that
stylesheet matches it and the theme switch leaves it alone. Both were confirmed
by rendering the built asset at 17px on the light and the dark surface.

The new file keeps a `viewBox`, which is what lets one drawing serve a 17px row
and a 32px high-dpi tab.

This does not touch `ui-web/src/assets/agents/raven.svg`, the separate mark the
agent rows draw. Those two files were already different drawings and still are.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Gates, against this branch's own range rather than a stale one:

- `scripts/check_large_files.py origin/main..HEAD` exit 0. `ui-web/` is exempt
  from the extension list by `scripts/check_large_files.py`, and the file is
  1199 bytes, down from 2524.
- `scripts/check_source_language.py origin/main..HEAD` exit 0.
- `scripts/check_commit_messages.py origin/main..HEAD` exit 0.
- `npx commitlint --from origin/main --to HEAD` exit 0.
- `npm run gen:check --prefix ui-web` exit 0, 178 methods matched.
- `npm run type-check --prefix ui-web` exit 0.

Tests. CI does not run `make test-ui`, so the one test file that names this
asset was run directly:

- `npm test --prefix ui-web -- --run src/features/knowledge/KnowledgePage.test.tsx`
  gives 5 failed, 104 passed, 0 unhandled errors, exit 1.
- All 5 are the `the recall test` cases, and node names the cause itself:
  `localStorage is not available because --localstorage-file was not provided`.
- They are inherited, not introduced. `git diff origin/main..HEAD --name-only --
  ui-web` lists this SVG and nothing else, so the JavaScript that run exercised
  is byte-identical to `origin/main`; the run was the baseline. No test imports
  or reads the file, so its bytes cannot reach an assertion.

Rendering. The built asset was rendered in headless Chrome at 16, 32, 48 and 64
pixels on a light and a dark tab strip, and at the row's real 17px box on both
themes.

The first box is left unchecked deliberately. The relevant file reports 5 red,
and while the evidence above shows all 5 are inherited and unreachable from this
change, the suite does not come back green on this machine and saying it does
would be false. On the third box: `grep` over the tracked tree finds no prose
describing this drawing. The favicon sentences in `ui-web/src/styles/page.css`
and `ui-web/src/shell/agent-mark.tsx` describe `miromind.svg` and the agent
mark, both of which this leaves alone.

## Risk

User-visible in two places, both cosmetic. Rollback is reverting the commit and
rebuilding; `dist/` is generated, so nothing downstream holds a copy.

- At 16px the halo ring largely disappears. Its stroke is 1.6 units on a 64
  unit canvas, which is 0.4 of a physical pixel there, and a count of ring
  pixels falls from 245 at 64px to 17 at 16px. A high-dpi tab draws 32px, where
  it holds. Widening the stroke would change a drawing that was chosen as it is,
  so it is reported rather than edited.
- No test pins the `viewBox` of this file, and none is added here. No icon in
  the tree has such a guard, and adding the first one is a wider decision than
  this change.

The security note: a favicon is parsed by the browser directly, so the file was
checked for active content. It has no `script`, `foreignObject`, `image`,
`xlink:href`, `href`, `url(`, `@import` or event-handler attribute; the only
external token is the SVG namespace declaration.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the complete `github/main...HEAD` diff and found no defect worth raising. I covered the repository rules in `AGENTS.md`/`CLAUDE.md` and the UI/runtime boundary in `CONTEXT-MAP.md`; traced both consumers through `ui-web/build.py`, `page.html`, `ravenIconPath()`, the knowledge processor row, and its CSS; inspected the file's history and the unchanged separate agent-row mark; checked backward compatibility of the asset path/cache flow; confirmed no tests were weakened; and checked the passive SVG/content and frontend architecture constraints.

Verification on this checkout:
- `npm test --prefix ui-web -- --run src/features/knowledge/KnowledgePage.test.tsx`: 109 passed
- `npm run build --prefix ui-web`: passed
- `uv run python ui-web/build.py`: passed; the copied asset is byte-identical
- UI type check and RPC generation check: passed
- source-language, large-file, commit-message, XML/passive-content checks: passed

The worktree remained clean. Completed CI checks are green; three unrelated Python unit shards were still running when this review was posted.

ZuyiZhou (approved):
Approving. Small diff, and the claims that matter are the ones about reach rather than the drawing, so those are what I checked.

Both consumers are real and are the only two: `ui-web/build.py:210` copies the file over `dist/assets/raven.svg`, `ui-web/src/page.html:7` points the tab icon at it, and `ravenIconPath()` at `provider-mark.tsx:267` has its single call site at `KnowledgePage.tsx:655`. The 17px sizing is there too -- the base `.provider-icon` rule is 20px and `.kbprocr .provider-icon` overrides it to 17px, with `:first-child` at full opacity, exactly as described. The second consumer really was easy to miss and finding it before shipping is the right instinct.

`viewBox="0 0 64 64"` is present, which is what makes one file serve both sizes. The security scan reproduces: zero matches for script, foreignObject, image, xlink:href, href, url(, @import or an event-handler attribute, so nothing active reaches a browser that parses a favicon directly. 2524 bytes down to 1199, and `ui-web/` is in the exempt prefix list in `check_large_files.py`, so the section 7 SVG block correctly does not apply.

Leaving the test box unchecked and explaining it is the right call. The 5 red cases are `localStorage is not available because --localstorage-file was not provided`, which is a Node 26 behaviour on this machine and not a branch fault -- the reasoning that a diff touching only an SVG cannot reach a JS assertion is sound, and the baseline argument backs it.

The 16px halo note is worth having written down: reporting that the ring thins to 17 pixels rather than quietly widening the stroke is the right way to handle a drawing decision that is not yours to make.

PR #537, opened 2026-09-20
fix(*): assemble a serve stack that a first run came up without

gloryfromca:
## Summary

A fresh install still could not chat without restarting the gateway. #524 made
that honest -- the reply carries `needs_restart` and onboarding says so -- but
the limit itself was untouched. This removes it.

`raven serve` assembles its whole stack once, at startup. With no model
configured `build_agent_loop` refuses, so there is no loop, and everything a
turn needs is built inside `if agent_loop is not None`: the scheduler, cron, the
plugin services, MCP. Writing a model in from the page changed a file and
nothing else -- nothing went back to build what was skipped, and the next send
met the error recorded at startup.

`config.set` now asks its host to assemble that stack, and reports what came
back. Assembling is not the handler's job: it calls a seam the host supplies, so
the wiring stays with whoever owns the stack (both reviewers on #524 asked for
that split). `needs_restart` narrows rather than disappears -- it now means a
build was tried and this process still cannot run a turn.

**Scope: only a process that never built a loop takes this path.** The trigger
is the answer that says so, which #524's credential gate already computes, so an
ordinary model switch on a running gateway re-points the loop exactly as before
and pays nothing. A settings write does not assemble anything (#513 is intact).

Three details are load-bearing rather than stylistic:

- **The loop-less stack is dropped, not torn down.** Its teardown ends by
  closing the browser and the ACP pool, both process-global and both the new
  stack's now. A first-run stack owns nothing else -- no cron, no plugin
  services, no memory backend, no MCP transports -- so there is nothing to
  release and nothing for a second assembly to collide with. This is what makes
  the case a one-shot build rather than the gateway's generation swap
  (`gateway_commands._request_swap`), which has to keep generation N serving
  while N+1 comes up.
- **The replacement stack is handed the emitter it replaces.** A subscription
  lives in the emitter, and the page re-subscribes only when its socket
  reconnects -- which a swap does not do. A fresh emitter would leave every open
  stream unfed, with no error anywhere.
- **The download routes read their store per request.** A first run assembles
  after the app is built, so the store was `None` at `add_files_routes` time and
  that process would have had no download surface for the rest of its life.

Not in this change: repointing a model on a running gateway by swapping the
stack. That needs `build_rpc_stack` split into assembly and start-up, because
today it starts cron, the plugin services and MCP as part of building -- a
second stack beside a live one would collide on all three. Separate change.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Real host, empty `RAVEN_HOME`, driven over the gateway's own RPC socket (the
startup log confirms the first-run state: `agent loop failed to build
(internal_error); chat turns will error`). Save a provider key, set a model,
send a turn, no restart in between:

```
config.set -> {"applied": true, "previous": null, "value": "deepseek/deepseek-v4-pro", "scope": "default"}
turn.send  -> message.start / episode.start / token.delta / message.complete
```

No `needs_restart` in the reply, and the assembly brought up the rest of the
spine: `Cron service started`, the playbook runtime, the sandbox executor and
five ACP sub-agents reaching `auto-verify ready`.

Reverse check, same host, same script, with the seam unwired: `config.set`
answers `needs_restart: true` and the turn ends in `error (-32603
internal_error)` -- today's symptom.

- `uv run pytest tests/test_rpc_config.py tests/test_rpc_bootstrap.py
  tests/test_rpc_transports_deliverables.py tests/test_cli_serve_commands.py
  tests/test_rpc_registration.py tests/test_rpc_transport.py
  tests/test_tracing_api.py tests/test_plugin_services_lifecycle.py -q`: 292
  passed.
- Five new cases, each mutation-verified on its own: dropping the seam call reds
  the first-run assembly case; reporting a failed assembly as success reds the
  restart case; asking the seam on every model write reds the running-process
  case; a fresh emitter per assembly reds the subscription case; reading the
  deliverable store once reds the late-store case.

## Risk

User-visible: a first run can finish setup and start chatting without bringing
the gateway back. A process that already has a loop is unchanged.

Contract: `needs_restart` keeps its type and its meaning narrows, so a client
reading it behaves the same; it is now set only when an assembly was tried and
failed.

The one behaviour a reviewer should weigh is the dispatcher rebind: a socket
already open keeps serving, because the transport reads `gateway.dispatcher` per
frame and the in-flight call holds its own reference. Subscriptions survive by
construction (the emitter is carried); sessions do not exist yet on a first run.

Rollback: revert the commit. Nothing else depends on the seam, and the parameters
it adds are optional with `None` defaults, so every other caller of
`build_rpc_stack` is unaffected.

## Related Issues

N/A


raven/cli/serve_commands.py
```diff
@@ -323,13 +323,51 @@ async def _serve_main(port: int, open_browser: bool) -> None:
...
+
+        The stack left behind is dropped rather than torn down: its teardown
+        ends by closing the browser and the ACP pool, and both are
+        process-global -- the new stack's now.
+        """
+        nonlocal stack
+        if stack.agent_loop is not None:
+            return True
+        async with building:
+            if stack.agent_loop is not None:
```
gloryfromca: Blocking: serialize first-run model writes with late assembly so every successful write reaches the live runtime.

This second check prevents two builds, but it does not reconcile a model write that already ran through the old dispatcher while the first build was yielding. `WsGateway` starts each frame in its own task, so the following ordering is reachable with two tabs: A writes DeepSeek and starts building it; B enters the old `config.set`, writes OpenAI to disk, and waits on this lock; A installs the DeepSeek stack; B acquires the lock, sees `stack.agent_loop`, and returns success without applying OpenAI to that loop. I reproduced that ordering: both calls returned `applied: true` without `needs_restart`, while the live model was `deepseek/deepseek-chat` and the persisted model was `openai/gpt-5`. The next turn therefore uses a different model from the successful selection, and a restart changes it again. The model write and assembly need one serialization boundary, or a waiter must apply its persisted choice to the stack that won the build.

gloryfromca (commented):
Blocking: serialize first-run model writes with late assembly so every successful write reaches the live runtime.

The focused 292-test suite passes, as do 27 architecture/runtime tests and Ruff checks. I covered AGENTS.md/CLAUDE.md and the Runtime/Web UI context rules, the full diff, the onboarding/config.set/WebSocket callers, assembly and teardown lifecycle, commit history, backward compatibility, and whether tests were weakened. The tests are additive, but they do not exercise overlapping writes through the loop-less dispatcher; that is the one blocker identified inline.

gloryfromca (commented):
Blocking: the concurrent first-run model-write race in the open thread still has to be fixed.

The `_ServedStack` extraction preserves the same ordering: each loop-less `config.set` writes its model before entering `ensure()`, and a waiter that finds the first assembled loop still returns success without applying its own later write. The added concurrency case overlaps two `ensure()` calls, so it proves that only one stack is assembled but does not cover two concurrent model writes or the live-versus-persisted default afterward. No new functional findings surfaced.

Verification: the focused RPC/serve suite passed with 299 tests, and the architecture/runtime suite passed with 27 tests. `ruff check` passed. `ruff format --check` reported that `tests/test_rpc_config.py` would be reformatted; that automated-format result is not a review finding. I rechecked the repository rules and Runtime/Web UI context, the complete diff against the current target, callers and lifecycle, history, backward compatibility, and that the tests were additive rather than weakened.

gloryfromca (commented):
Blocking: the open concurrent first-run model-write race still has to be fixed.

This revision only restores two blank lines in `tests/test_rpc_config.py`; it does not change the late-assembly ordering covered by the open thread. No new findings surfaced.

Verification on this head: the focused RPC/serve suite passed with 299 tests, `ruff check` passed, and `ruff format --check` passed for all 10 changed Python files. I rechecked the complete target diff, repository and Runtime/Web UI rules, affected callers and lifecycle, history, backward compatibility, architecture boundaries, and that the tests were not weakened.

raven/cli/serve_commands.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

The underlying race remains as described, but it does not meet the fourth-round blocking bar: it requires overlapping first-run model writes, which is not ordinary single-wizard use, and an affected operator can recover by switching the model once more after assembly or by restarting. Carrying this as a named follow-up rather than continuing to hold the change; this thread is settled from the reviewer side.

gloryfromca (commented):

The tree is identical to the prior revision; this head only rewrites the formatting-fix commit metadata. The concurrent first-run model-write race remains a named follow-up: it requires overlapping setup writes rather than ordinary single-wizard use, and another model switch or restart recovers the live runtime. The earlier thread has been replied to and resolved accordingly. No new findings surfaced.

Verification on this head: the focused RPC/serve suite passed with 299 tests, `ruff check` passed, and `ruff format --check` passed for all 10 changed Python files. I rechecked the complete target diff, repository and Runtime/Web UI rules, affected callers and lifecycle, history, backward compatibility, architecture boundaries, and that the tests were not weakened.

raven/cli/serve_commands.py
gloryfromca: Taking the race as a named follow-up, and recording the shape here so it is not re-derived.

The ordering you reproduced is real and the second check under the lock does not close it: a waiter that persisted its own model while the winner was building acquires the lock, finds a loop, and reports success for a write that never reached that loop. The two-tab first run it needs is outside the wizard's path, and switching the model once more after assembly applies it, which is why I am not holding the change for it either.

What a fix needs, for whoever picks it up: the persisted write and the assembly have to share one serialization boundary, or a waiter has to re-apply its own persisted choice to whichever stack won the build. The second is the smaller change -- `_set_model` already knows how to point a live loop at a binding, and the waiter has the value it wrote in hand when `ensure()` returns True.

0xKT (approved):

PR #538, opened 2026-09-20
fix(ui-web): keep a dag run's row, and let the strip be where a graph is named

xfng-sd:
## Summary

Two changes to how a `run_subagent_dag` graph reaches the reader. The first is
a correctness fix, and it is what makes the second safe.

**A dag run could lose its row in the tasks panel.** `refresh()` replaces the
whole row list with what `tasks.list` answered, and the server answers from the
state it held when it was asked -- so a graph dispatched during that round trip
is already on screen when the answer lands, and the answer does not know about
it. The session guard that was there covers the reader switching conversations
mid-read; nothing covered a live frame landing mid-read. Only two of the five
reducers can insert a row, and `dag.run_started` fires exactly once per run:
`applyNodeUpdated` and `applyRunCompleted` both map over rows that already
exist. So a dag row clobbered there stayed invisible for the whole life of the
run, in the strip above the composer and in the desk pane alike, and no later
frame could bring it back. A spawn recovered on its next status frame, which is
why this only ever showed on graphs. The fix is a tick bumped by every live
frame and captured around the read: unchanged, the answer replaces as before;
moved, the answer speaks for every row it has not been overtaken on, a frame
speaks for what it moved after the read was asked, and a row a frame retired is
not put back. Preferring the frame on a contested row rather than the answer is
the self-correcting direction: the answer is the richer copy, but it may be
describing a node the reader has already watched finish and nothing would fix
that, while every terminal frame already schedules a reconcile that brings the
richer copy back a moment later.

**The dag sheet is gone.** A graph docked in the rack as a `.dsheet` put the
whole of itself between the transcript and the box you type in for as long as
the run lasted, and left a settled one there afterwards. The strip below the
rack already names every running task in a line each, and its chip opens the
same graph in the desk's task pane with room to pan it. The trail's delegation
card is built from the same frames and stays -- that is the conversation's
durable record, and it is not what this removes.

What is kept is the run itself: `openDagRun` reads it to turn a receipt in that
card into the node a pane opens on, and `state/session/resume.ts` puts it back
on reload. Two resume tests said "raises the sheet" and asserted a rack count;
they pin the run now, which is what they were always about. `mount.tsx` holds no
JSX once the root and the rack filing go, so it is `mount.ts`, and the four
gates that pinned the sheet -- the file name, two import edges, the rack's
docking table and three class tallies -- shrink with it rather than being
widened for it.

Net ~800 lines removed. The bundle drops from 929.0 kB to 926.7 kB.

Scope note: this suppresses the sheet for every dag run, not only for a
playbook run. `dag.run_started` carries no playbook marker -- the name is
derived server-side in `tasks.list` from the node-id prefix -- so telling the
two apart on the page would have needed a new wire field. Suppressing both was
the simpler and the wanted behaviour.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run from `ui-web/`, on this branch rebased onto `refactor/ui_web_architecture`
at b73b95887:

- `npm run type-check` -- clean
- `npm run lint` -- 0 errors (5 pre-existing `react-hooks/exhaustive-deps`
  warnings, none in files this branch touches)
- `npx vitest run` -- 184 files, 2530 tests, all pass
- `npm run build && python build.py` -- built dist/index.html, and both boot
  snapshots match their goldens unchanged (the sheet was never in the first
  frame)
- `node scripts/check-class-namespace.mjs` -- OK
- `node scripts/check-css.mjs` -- OK
- `node scripts/check-page.mjs` -- OK
- `npm run gen:check` -- generated.ts matches the contract

The race fix carries four cases in `features/tasks/store.test.ts`: the row a
frame inserted survives the read, the answer does not undo a frame that moved
the same row, the answer still wins for a row the frames left alone, and a
plain read with nothing racing it still drops a row the answer no longer
carries.

Review round one raised two blocking findings, both addressed in 00a2244b2:
the merge preferred the answer on a contested row where the frame was newer
(now stamped and resolved per row), and `ui-web/CONTEXT.md` plus five other
places still described the deleted sheet.

Round two reopened the second of those: the sweep was run with a grep for
contiguous `dagsheet`, and the repository spells the term `dag sheet`, so most
of it was missed. ed226c228 finishes it -- eleven source comments, two gate
descriptions and two tests that pinned a tenant which can no longer dock.
`grep -RIni -E "dagsheet|dag sheet" ui-web/src ui-web/scripts ui-web/*.md` now
returns only the `_Avoid_` line in CONTEXT.md that retires the term.

Not verified in a browser: the offline stub canvas does not replay its scripted
`run_subagent_dag` to these surfaces. Replaying the same conversation against
the pre-change build produced an identical empty rack and identical chips, so
the change causes no observable difference there, but the dag path's evidence
is the suite rather than the page. Why the stub's dag script goes nowhere looks
like a separate fixture gap.

## Risk

User-visible: a running graph is no longer drawn above the composer. It is
named there instead, as a chip in the task strip, and the full graph is one
click away in the desk's task pane. The trail's delegation card in the
transcript is unchanged, so the conversation's own record of the run is intact.
The one thing genuinely lost is the settled graph that used to persist in the
rack after a run finished; that run is still listed in the task pane's settled
group.

No contract, backend or storage change -- the whole diff is page-side. The
stored session note still carries the fold bit it always did; nothing reads it
now, which is noted in the code as follow-up cleanup rather than removed here.

Rollback is a revert of the single commit; nothing else depends on it.

## Related Issues

N/A


ui-web/src/features/tasks/store.ts
```diff
@@ -76,19 +76,45 @@ export const rowKey = (row: TaskRow): string => `${row.kind}:${row.id}`
...
+     drops the row that frame inserted, and only two of the five reducers can
+     ever insert one -- `dag.run_started` fires exactly once per run, and
+     `applyNodeUpdated` / `applyRunCompleted` only ever map over rows that
+     already exist. So a dag run clobbered here is invisible for the whole of
+     its life, which is what this merge is for: the server wins on every row
+     it answered, and a row it has not caught up to yet survives until the
+     next read. */
+  const named = new Set(answered.map(rowKey))
+  const missed = store.get().rows.filter((r) => !named.has(rowKey(r)))
+  patch({ rows: [...missed, ...answered], loaded: true })
```
gloryfromca: Blocking: do not let an older tasks.list row overwrite a same-key live update.

The captured tick tells us a live frame landed after this request began, while the comments and PR description establish that `answered` is the server state from when the request was made. For a row present in both lists, this ordering makes the older `answered` copy win. A concrete sequence is: an existing node is `running`; `refresh()` starts; `onNodeUpdated()` marks it `completed`; then the older list answer returns it as `running`. This line puts the task and node back to `running`, and `dag.node_updated` schedules no reconcile, so the strip/pane stays stale until another frame or read happens. The new same-key test currently asserts this stale precedence instead of covering the newer-live-state case.

ui-web/src/features/dag/DagSheet.tsx
```diff
@@ -1,99 +0,0 @@
-/* The dag sheet: the graph a `run_subagent_dag` call is orchestrating, docked
```
gloryfromca, line 1: Blocking: update the canonical Web UI vocabulary when deleting the DAG sheet.

`ui-web/CONTEXT.md` still defines a Sheet and the Sheet rack as including a DAG graph, and it points directly to this now-deleted `DagSheet.tsx`. The DAG manifest and domain-shape gate comments repeat the same obsolete architecture. Because CONTEXT.md is the repository's canonical vocabulary that contributors are required to consult, this deletion leaves the documented architecture false and links readers to a nonexistent component. Please update those definitions/comments to describe the retained DAG state and the tasks strip/desk ownership.

gloryfromca (commented):
Blocking: protect same-key live updates and update the DAG sheet architecture definitions.

I reviewed the full target diff and traced the changed task-row lifecycle through the session stages, source wiring, composer strip, desk surface, DAG resume path, and session cleanup. I also checked the branch history, backward-compatibility claims, deleted/replacement tests, AGENTS.md/CLAUDE.md, and the canonical Web UI context. Two blocking findings are inline.

Verification on this head:
- `npm test -- --run`: 180 files, 2,479 tests passed
- `npm run lint`: 0 errors and 5 warnings, all in untouched files
- `git diff --check`: passed

The sheet-specific test removal matches the intended UI removal, but the retained DAG state path loses some direct coverage; that is not a separate blocker here.

ui-web/src/features/tasks/store.ts
xfng-sd: Agreed, and fixed in 00a2244b2 -- the ordering was wrong and the test pinned the wrong side of it.

The tick told us a frame landed after the read was asked; it did not tell us which rows that frame was about, and taking `answered` for every row it named discarded the newer copy of exactly the ones it was. Your sequence is the one that hurts: `applyNodeUpdated` is the single consumer that does not schedule a reconcile, so the row stayed at `running` until another frame or read happened.

The stamp is per row now. `liveAt` maps a row key to the tick at which a frame last moved it, and which rows a frame moved is read off object identity rather than asked of the reducers -- each one rebuilds only the row it touches and passes the rest through by reference, so the diff is exact and `live.ts` stays pure. The merge then goes row by row: the answer speaks for everything it has not been overtaken on, a frame speaks for what it moved after the read was asked, and a row a frame retired (a spawn cancelled before it ran) is not put back.

On which side to prefer, since the client cannot actually order the server's snapshot against the frame: preferring the frame is the self-correcting direction. The answer is the richer copy -- the tokens, the output files and the final error text are only ever in it -- but preferring it can leave a node the reader has already watched finish showing as running, with nothing scheduled to fix it. Preferring the frame costs the richer fields only until the reconcile that every terminal frame already schedules lands.

One more thing that fell out of doing it per row: a row the answer dropped that no frame touched is now gone, where the first version kept it.

`prefers the server's copy of a row the answer did name` is replaced by `does not let the answer undo a frame that moved the same row`, which is your sequence verbatim -- it fails against the previous merge, and I checked that it does rather than assuming. A second case pins that the answer still wins for a row the frames left alone, so the richer-copy path stays covered.

ui-web/src/features/dag/DagSheet.tsx
xfng-sd: Agreed, and fixed in 00a2244b2.

`ui-web/CONTEXT.md`: the **Sheet** and **Sheet rack** entries no longer list a dag graph among what docks above the composer, and the link to the deleted component is gone. Added a **Task strip** entry for where a run is named now -- the chip, the pane it opens, and the trail's delegation card as the run's durable record -- with an `_Avoid_: "dag sheet"` line so the retired term is on the page that governs the vocabulary rather than only in a commit message.

Also swept the rest of what still described the old architecture, which was more than the two you named:

- `features/dag/manifest.ts` -- said "a sheet in the rack above the composer, mounted per run"
- `scripts/gates/domain-shape.test.mjs` -- the `dag` entry repeated it
- `features/dag/store.ts` -- header said what it draws "is decided here and in DagSheet.tsx"
- `features/composer/ClarifySheet.tsx` -- header pointed at `features/dag/DagSheet.tsx` as the pattern it follows
- `src/main.tsx` and `state/session/stages.ts` -- the import alias was `dagSheet`, now `dagRun`; it had not been a sheet since the previous commit and the name was the last thing still saying otherwise
- `state/sheetRack.test.ts` -- a fixture used the literal class `dagsheet`, now a neutral label

`grep -ri dagsheet ui-web/src ui-web/scripts ui-web/*.md` comes back empty.

On the coverage note in your summary: agreed that the retained DAG state path lost some direct cover when `DagSheet.test.tsx` went. The two resume cases that asserted a rack count now assert the run is put back, which is the part that still exists, and `mount.runStarted.test.ts` still covers the live layer into that state. Happy to add cases for `advance` / `settle` against the store directly if you would rather not leave that to the transcript's own suites.

ui-web/src/features/tasks/store.ts
gloryfromca: Blocking: the remaining issue is confined to stale DAG-sheet architecture text in the other thread.

The row-ordering fix itself holds. `liveAt` marks only keys whose reducer output changed identity (including renamed or retired rows), the merge preserves only post-request rows, and the replacement test reproduces the exact running -> completed -> stale running sequence. I also checked duplicate/no-op frames, rows untouched by a different frame, retirement, reset, and terminal reconcile paths and found no regression here. Closing this thread.

ui-web/src/features/dag/DagSheet.tsx
gloryfromca: Blocking: sweep the remaining source comments and rack test that still describe a DAG sheet.

The canonical CONTEXT entry and the files listed in your reply are fixed, but the grep used there only searched contiguous `dagsheet`; the repository mostly spells the term `dag sheet`. `grep -RIni -E "dagsheet|dag sheet" ui-web/src ui-web/scripts ui-web/*.md` still finds active descriptions in `state/sheetRack.ts`, `chrome/SheetRack.tsx`, `features/composer/clarify.ts`, transcript/subagent mount and store comments, CSS, and `components/Ico.tsx`. Most concretely, `chrome/SheetRack.test.tsx` still creates a `.dsheet` and pins “stacks a question raised over a running graph on top of it,” even though this revision removes that tenant. These are not historical notes like the retrospective in `dag/mount.ts`; they describe current callers and ordering constraints that no longer exist. The new vocabulary is right, but the source and test still contradict it, so this thread remains open.

gloryfromca (commented):
Blocking: remove or update the remaining DAG-sheet source references and obsolete rack test.

The per-row live ordering fix resolves the runtime blocker and its replacement tests cover the reported same-key regression. The canonical Web UI vocabulary is also corrected, but the source sweep is incomplete: active rack, composer, transcript, subagent, icon, and CSS comments still describe the deleted sheet, and `chrome/SheetRack.test.tsx` still pins a graph as a rack tenant. The row-ordering thread is resolved; the vocabulary thread remains open for this remainder.

I reviewed the refreshed target diff, the fix delta, relevant callers and reducer identity contracts, branch history/range-diff, backward compatibility, test replacement, AGENTS.md/CLAUDE.md, and the Web UI architecture vocabulary. I found no additional runtime blocker and no evidence that tests were weakened to hide a failure.

Verification on this head:
- `npm test -- --run`: 184 files, 2,530 tests passed
- `npm run lint`: 0 errors and 5 warnings, all in untouched files
- `git diff --check`: passed

ui-web/src/features/dag/DagSheet.tsx
xfng-sd: Agreed on all of it, and fixed in ed226c228. The grep behind my last reply was the mistake -- I searched contiguous `dagsheet` and the repository spells it `dag sheet`, so I reported a clean sweep off a pattern that could not have found most of it. Ran yours, plus a wider one for `graph sheet` / `sheet.*graph`, and took the union: seventeen places.

Eleven were comments describing a live mechanism rather than history, and all now say what is true:

- `state/sheetRack.ts` (x2) and `chrome/SheetRack.tsx` -- the "mixed child list" rationale
- `features/transcript/mount.tsx` (x2) and `features/subagents/mount.tsx` -- the flushSync notes that justified themselves with "the dag sheet selects its node next"
- `features/subagents/store.ts` (x2) -- what opens nodes there, and what reads `sel()`
- `components/Ico.tsx` -- the local-copy note waiting on a renderer merge that can no longer happen as written
- `features/composer/clarify.ts` -- the `data-asks` fold rationale
- `state/session/stages.ts` -- the replan note
- `features/dag/types.ts` -- `task_summary` described as what titles the sheet
- `styles/page.css` -- the shared-controls header
- and two gate descriptions you did not list, `i18n-keys` and `island-lang`, which both still called the graph a sheet in the rack

On `chrome/SheetRack.test.tsx`: retargeting it to two real tenants turned out to be impossible, and the reason is worth recording. Both remaining tenants -- the clarify question and the approval -- call `dropClass('csheet')` before they dock, so no two real sheets are ever on screen together to order. The graph was the other half of that pair. So the case now stacks a question over an already-docked sheet whose second tenant is a bare host that says in the comment that it is a stand-in and why. I would rather pin the rack's own contract with an honest placeholder than delete the only cover for the portal-per-host arrangement, since that is what breaks if the portals ever move to `#sheetRack` -- but if you would rather see the case deleted outright now that no caller can exercise it, say so and I will.

`state/sheetRack.test.ts` was the same shape of problem: it asserted that a sheet handed over with no interior is absent from the view list and called that "the graph's kind". The contract is still `add`'s, but no caller takes it now, and the comment says exactly that. Related: I noted on `RackSheet` that the optional view on `add` is a capability with no caller -- the graph was its only one. Removing the branch is a real simplification but a wider change to a state module than a review fix should carry, so I left it documented rather than ripped out. Happy to do it in a follow-up if you want it gone.

`grep -RIni -E "dagsheet|dag sheet" ui-web/src ui-web/scripts ui-web/*.md` now returns one line: the `_Avoid_` entry in CONTEXT.md that retires the term.

Verification on ed226c228, rebased on b73b95887:
- `npx vitest run`: 184 files, 2,530 tests passed
- `npm run type-check`: clean
- `npm run lint`: 0 errors, 5 warnings, all in untouched files
- `npm run build && python build.py`: both boot goldens match unchanged
- `check-class-namespace` / `check-css` / `check-page`: OK

gloryfromca: No blockers; this can merge as far as I am concerned.

The sweep holds. The remaining matches are either the canonical retired-term entry or explicitly historical context, while current rack, renderer, session, gate, and node-opening descriptions now match the code. Keeping the rack ordering case with an honestly labeled bare-host stand-in is appropriate: it tests a supported `add` contract and portal-per-host ordering without pretending the removed DAG tenant still exists. I do not need that case deleted. Closing this thread.

gloryfromca (commented):

The per-row race fix remains sound, and the final sweep now makes the source comments, architecture gates, canonical vocabulary, and rack tests agree with the removal of the DAG sheet. Both threads I opened are resolved; I found no new finding.

I reviewed the complete refreshed target diff, the cleanup delta, affected callers and reducer contracts, branch history, backward compatibility, test changes, AGENTS.md/CLAUDE.md, and CONTEXT vocabulary. The removed sheet tests were replaced where retained behavior still needs coverage; I found no test weakening used to mask a failure.

Verification on this head:
- `npm test -- --run`: 184 files, 2,530 tests passed
- `npm run lint`: 0 errors and 5 warnings, all in untouched files
- `npm run build`: passed
- `uv run python build.py`: passed; both boot snapshots matched
- class namespace, CSS, and page checks: passed
- `git diff --check`: passed

0xKT (approved):

PR #540, opened 2026-09-20
fix(*): three settings findings from the post-merge read

gloryfromca:
## Summary

Three findings from reading the merged settings dialog on a real host, one per
area. Each is small and independent; they travel together because they are one
reading pass.

**Provider marks in the shared model picker.** `components/ModelPicker` drew the
provider column as plain text, so a settings role row offered a name where every
other provider list in the page offers the brand mark. It now takes
`ProviderIcon` from `components/ProviderMark`, the same component the provider
cards and the composer's own list use. No CSS change: `.model-picker-prov` was
already a flex row with a gap.

**What `ext.list` calls built in.** The flag was computed from the registry's
schema-hidden set. Hidden from the schema and withheld from the model are
different mechanisms: a schema-hidden tool is still reachable through
`tool_call`, and the DAG controls (`run_subagent_dag`, `cancel_dag`,
`dag_status`, `resolve_dag_node`) sit in that set while still honouring the
`tools.disabledTools` switch the page draws. Drawing them as built in told the
reader their switch does nothing, which is false. The flag now answers the
question the switch asks: the two tool-search meta-tools, plus the MCP resource
and prompt meta-tools the loop registers and withdraws on its own. Those are the
only names whose switch the loop would not honour.

**About's update check restored to two steps.** The check started the upgrade
itself, so the one button was really two actions and A47's second step had no
entry point. The check now returns the newer version and the row offers the
upgrade beside it; the reader takes the second step. The seam gained
`checkUpdate(): Promise<string | null>` and `upgrade()`, so the page chrome keeps
owning the flow and the card only draws it.

Two more landed after review, plus a gap the first round left behind.

**The upgrade button never appeared.** The check calls `redrawSettings()`
before it resolves; `SettingsApp` keys the panel by that epoch, so the About
row that asked for the check is replaced before the setter it left behind
runs. Neither order worked -- had the setter won the race, the replacement
would still have mounted with no version. The version is now read from where
`showUpNote` already retains it, through a new `upgradeLatest()` export, so
the redraw that broke the state is what delivers the value. A version found
by the background poll offers the upgrade here too, which is what A47 asks
for: the row offers it once a newer version is known, not only when this
button was the one that found it.

**The meta-tool card was named after a feature it did not contain.**
`tool_search` is registered only when progressive tool disclosure is on
(`tools.toolSearch.enabled`, off by default), so a default install saw a card
called Tool search holding only `tool_call`, which is not tool search. The
card is now Meta tools, `ext.list` reports the meta-tool the loop skipped the
way `_gated_tools` already reports a key-gated one -- absent from the list
reads as deleted rather than switched off somewhere else -- and the row says
where its switch actually is.

**The offline fixtures still carried the old predicate.** `rpc/fixtures/ext.ts`
and `test/settingsHarness.ts` called `cancel_dag` built in, so the offline page
drew the defect this PR fixes, and the harness pinned it well enough that the
page test asserted the old behaviour and passed. Correcting the fixtures reds
that test, which is the proof it was pinning the wrong fact. The switchable
tool counter moves with them: a tool drawn as built in was left out of the
total, so the page under-reported what the reader can switch off.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Automated:

```
cd ui-web && npm test              -> 182 files, 2523 passed
cd ui-web && npm run type-check    -> clean
cd ui-web && npm run lint          -> 0 errors, 5 pre-existing warnings
uv run --all-extras pytest tests/test_rpc_console.py tests/test_rpc_schema_match.py \
    tests/test_rpc_contract_shapes.py tests/test_rpc_registration.py -q  -> 534 passed
```

Each new test was mutation-verified:

- revert the picker to plain text -> `ModelPicker.test.tsx` reds with
  `Anthropic2: expected null to be truthy`
- compute `builtin` from the hidden set again -> `test_rpc_console.py` reds with
  `cancel_dag answers to the switch, so it carries one`
- let the check call `askUpgrade()` again -> `wire.test.ts` reds with
  `expected null to be '0.3.0'`

Real host, before and after, on a host that can actually report an update (a
uv-tool install with a seeded update cache): before, the rail said a new
version was available and the About row offered nothing; after, the row offers
"Upgrade to 9.9.9" and clicking it asks rather than starts. The other half of
that path, where the button's own live check finds the newer version, is not
reachable on a development host and stays covered by the unit tests above.

Real host (`raven serve`, temporary RAVEN_HOME, driven through the page):

- the role picker's provider column draws `IMG:openrouter`, `IMG:deepseek`,
  `IMG:gemini`
- the tools page lists the four DAG controls with their switches and marks only
  `tool_call` as built in
- About's check reports up to date and no longer opens the upgrade dialog by
  itself. The newer-version branch is not reachable on a development host:
  `_upgrade_command_works()` gates the whole update notice on a uv-tool install,
  so that half is covered by the unit tests above.

## Risk

User-visible: provider marks appear in the settings role picker; the four DAG
controls lose the "built in" tag and read as switchable, which they are; the
About row grows a second button after a check that finds a newer version.

The `builtin` field is advisory -- the page draws it, nothing gates on it -- so a
wrong value misinforms rather than breaks. Rollback is per commit; the three
areas share no code.

## Related Issues

N/A


ui-web/src/features/settings/pages/About.tsx
```diff
@@ -21,15 +24,28 @@ export function workspacePath(raw: Record<string, unknown>, configPath: string):
...
     <Card>
       <Row label={t('gui.settings.about.version')}>
         <Rov>{version || t('gui.settings.about.version_unknown')}</Rov>
       </Row>
       <Row label={t('gui.settings.about.update')}>
-        <button type="button" className="mini ghost" onClick={(e) => void store.source().checkUpdate(e.currentTarget)}>
+        <button
+          type="button"
+          className="mini ghost"
+          onClick={(e) => { void store.source().checkUpdate(e.currentTarget).then(setNewer) }}
```
gloryfromca: Blocking: the successful update check remounts About before it can retain the discovered version.

`checkUpdate()` invokes `redrawSettings()` before resolving. That increments `epoch`; `SettingsApp` keys the panel by that epoch, and the store's `flushSync` contract unmounts this `About` instance before `.then(setNewer)` runs. The setter therefore targets the old instance, while the replacement mounts with `newer === null`, so a live successful check never renders the new Upgrade button. The current tests divide the path: `About.test.tsx` substitutes a check without the redraw, while `wire.test.ts` never mounts the page. Keep the discovered version somewhere that survives the epoch remount, or avoid that remount before returning, and cover the integrated path.

gloryfromca (commented):
Blocking: the successful update check remounts About before it can retain the discovered version.

I found one blocking behavior defect, marked inline. I covered the repository rules (AGENTS.md / CLAUDE.md / CONTEXT-MAP.md / ui-web/CONTEXT.md / ui-web/CONTRIBUTING.md), the full diff, affected callers and relevant history, backward compatibility and generated RPC schema, whether tests were weakened, and the documented UI layering and store contracts. The provider-mark and tool-switchability corrections otherwise match their callers and contracts.

Verification:
- `uv run pytest tests/test_rpc_console.py -q`: 97 passed
- `npm test` in `ui-web`: 182 files and 2523 tests passed
- `npm run type-check` in `ui-web`: passed
- `npm run gen:check` in `ui-web`: passed

gloryfromca:
Confirmed and fixed in 99ab1084d.

I checked the mechanism before changing anything: `redraw()` sets `epoch: get().epoch + 1` (store.ts:191) and `SettingsApp.tsx:87` keys the panel `key={s.epoch}`, so the subtree is replaced, not re-rendered. Worth adding that neither order saved it -- had the setter won the race, the replacement would still have mounted with `newer === null`. The bug was holding the value in the row at all, not the timing.

The version was already being retained, so nothing new holds it: `showUpNote()` keeps it for the rail's notice and `askUpgrade()` has always read it from there, it just had no export. `app/updates.ts` now exports `upgradeLatest()`, the settings seam carries it as `newerVersion()`, and About reads it on every draw. The redraw that broke the state is what now delivers the value.

Two consequences, both deliberate:

- `checkUpdate()` returns `Promise<void>`. Nothing reads the answer any more, and leaving it was what let the two tests drift apart.
- A version found by the background poll now offers the upgrade here as well. A47 asks the row to offer the upgrade once a newer version is known, not only when this button was the one that found it, so I took that as the correct reading rather than filtering it out.

On the split you named: `About.test.tsx` now redraws inside its fake check, the way the real one does, so the remount is in the test rather than outside it; and both files assert on the retained version instead of a return value, which is the fact the page reads. Reverting About to a held `useState` reds both of the new cases -- `offers the upgrade after the check, across the redraw that check does` and `offers the upgrade for a version the reader never asked about` -- which is the reproduction of what you reported.

Also in this push: `ruff format` on `tests/test_rpc_console.py`, which CI caught separately. Full web suite 2524 passed, python lint clean.

ui-web/src/features/settings/pages/About.tsx
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed on 99ab1084d236: `About` now reads the retained version through `newerVersion()` after the epoch remount, and the regression case makes its fake check call `store.redraw()` before asserting the upgrade control. That exercises the lifecycle boundary the prior tests missed, and the original failure no longer holds.

Verification on this head: `uv run pytest tests/test_rpc_console.py -q` passed 97 tests; the full `ui-web` suite passed 2,524 tests across 182 files; type-check and RPC generation checks also passed.

gloryfromca (commented):

The repair resolves the prior lifecycle defect: the discovered version now outlives the epoch-keyed settings subtree, and the regression test exercises the redraw/remount before checking the upgrade control. The prior thread has been replied to and resolved.

This round covered the repository rules (AGENTS.md / CLAUDE.md / CONTEXT-MAP.md / ui-web/CONTEXT.md / ui-web/CONTRIBUTING.md), the full target diff and repair delta, affected callers and update-state history, backward compatibility and generated contracts, test-strength changes, and the documented seam/store/layering constraints. No new plain errors or merge blockers remained.

Verification:
- `uv run pytest tests/test_rpc_console.py -q`: 97 passed
- `npm test` in `ui-web`: 182 files and 2524 tests passed
- `npm run type-check` in `ui-web`: passed
- `npm run gen:check` in `ui-web`: passed

raven/rpc/methods/console.py
```diff
@@ -307,6 +318,37 @@ def _needs_of(name: str) -> dict | None:
...
+    """
+    rows: list[dict] = []
+    from raven.agent.tools.tool_search import META_TOOL_NAMES
+
+    for name in sorted(META_TOOL_NAMES - registered):
+        rows.append(
+            {
+                "name": name,
+                "description": "",
+                "enabled": False,
```
gloryfromca, line 343: No blockers; suggestions only, and they are marked inline.

Follow-up: this synthesized `tool_search` row says `enabled: false`, but `builtin: true` routes it through the fixed-control branch in `Tools.tsx`, which hard-codes `aria-checked=true`; the CSS also places the thumb on the checked side. With the default-off tool-search setting, the page therefore visually and accessibly reports the tool as on while the adjacent note says it is controlled elsewhere. The regression test checks that the control is fixed and disabled but does not check its reported state.

This is introduced and default-reachable, but the adjacent note provides a working route to the real setting, so it does not meet the fourth-round blocking bar. No change is requested in this PR; please carry the state mismatch as the named follow-up in the review summary.

gloryfromca (commented):

The previous About lifecycle blocker remains fixed. The new head only adds spacing to the meta-tool note and does not change that result.

One named follow-up remains: a synthesized, default-off `tool_search` row is displayed and announced as checked. This revision introduces it and it is reachable under the default configuration, but it does not meet the fourth-round blocking bar because the adjacent note gives the operator a working route to the setting.

Coverage included the repository instructions, the full target diff and latest delta, affected callers and history, backward compatibility, test strength, and the settings/RPC architecture. No weakened tests were found.

Verification: on 8d256c16590a, `uv run pytest tests/test_rpc_console.py -q` passed 99 tests, the full `ui-web` suite passed 2,524 tests across 182 files, `npm run type-check` passed, and `npm run gen:check` matched 185 RPC methods. The move to this head is a one-line CSS spacing change; `git diff --check github/refactor/ui_web_architecture...80a0a6f9b376` passes.

raven/rpc/methods/console.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

This revision changes the fixed control to `aria-checked={!!known_.on}`, but the live row does not preserve the wire state used to synthesize it. `mkToolRow()` only forces `on = false` when `needs` is present; otherwise it replaces `on` with a getter over `tools.disabledTools`. The synthesized `tool_search` row has `enabled: false`, `needs: null`, and is not in the default disabled list, so production still gives this component `known_.on === true` and the control remains checked.

The new assertion passes because `settingsHarness.snap()` hand-builds `tool_search` with `on: false`, bypassing `mkToolRow()`. The original state-mismatch follow-up therefore remains. It is still nonblocking for the same reason: the adjacent note gives the operator a working route to the real setting.

gloryfromca (commented):

The About lifecycle fix remains sound, and this revision introduces no new blocker. The named `tool_search` follow-up is not actually fixed: `Tools.tsx` now reads `known_.on`, but the live `mkToolRow()` adapter replaces the synthesized row's wire-level `enabled: false` with a getter over `tools.disabledTools`, which is true on the default configuration. The new page assertion bypasses that adapter by supplying `on: false` directly.

This remains nonblocking under the late-round bar. The mismatch is introduced and default-reachable, but the adjacent note still gives the operator a working route to the actual tool-search setting. The original thread remains resolved and the item stays named here as a follow-up rather than an open objection.

Coverage included AGENTS/CLAUDE and the relevant CONTEXT and ui-web contribution rules; the full target diff and latest delta; callers and history; backward compatibility; test strength; and the settings/RPC architecture. No existing test was weakened, though the new regression assertion does not traverse the production adapter where the state is lost.

Verification on this head: `uv run pytest tests/test_rpc_console.py -q` passed 99 tests; `npm test -- --run` passed 2,524 tests across 182 files; `npm run type-check` passed; and `git diff --check github/refactor/ui_web_architecture...HEAD` passed.

gloryfromca:
Fixed in 9f4e4ed4f, and you caught the half I had just missed.

I had found the same thing a few minutes earlier from the other end -- the real host still drew `tool_search` as on after the `aria-checked` change -- so both of us landed on `mkToolRow`. `on` is an accessor over `tools.disabledTools`, which is the right question for a switchable tool and a meaningless one for a tool no switch on this page can move: `tool_search` is reported precisely because the loop did not register it, and it is not in that list either, so the accessor answered on and the previous commit changed nothing a reader sees.

A builtin row now keeps `enabled` as it came off the wire.

Verified on a real host rather than from the test, because the test was the thing that missed it:

```
tool_search: aria-checked=false, fixed=true,  thumb left  (off)
tool_call:   aria-checked=true,  fixed=true,  thumb right (on)
cancel_dag:  aria-checked=true,  fixed=false, thumb right (on)
```

On why the page test could not see it: the settings harness builds its rows directly instead of through `mkToolRow`, so that mapping had no test at all -- the fixture asserted a shape the real path never produces. There is one now (`features/installed/source.test.ts`), covering both sides: a builtin row keeps the wire state, a switchable row asks the disabled list. Dropping the branch reds the first.

Full web suite 2532 passed, type-check clean, lint 0 errors.

gloryfromca:

Confirmed on 9f4e4ed4fec6: `mkToolRow()` now returns fixed rows with the `enabled` state reported by `ext.list` before installing the `tools.disabledTools` accessor used by switchable rows. The new adapter-level test covers both the default-off `tool_search` row and the on `tool_call` row, so it exercises the production layer the previous page test skipped. The prior state-mismatch follow-up no longer holds.

Verification on this head: the targeted RPC suite passed 99 tests; the full web suite passed 2,526 tests across 183 files; type-check and RPC generation checks passed; and the full target diff passes `git diff --check`.

gloryfromca (commented):

The latest delta fixes the remaining `tool_search` display mismatch at the live adapter boundary: fixed rows preserve the loop-reported `enabled` state, while switchable rows retain the existing `tools.disabledTools` accessor. The new test covers that distinction in `mkToolRow()` itself. The earlier About lifecycle issue remains fixed, and no new findings survived review.

Coverage included the repository rules (AGENTS/CLAUDE, relevant CONTEXT, and ui-web contribution constraints), the full target diff and latest delta, callers and history, backward compatibility, test strength, and the settings/RPC architecture. Tests were strengthened rather than weakened. Both review threads opened here are resolved.

Verification: `uv run pytest tests/test_rpc_console.py -q` passed 99 tests; `npm test -- --run` passed 2,526 tests across 183 files; `npm run type-check` passed; `npm run gen:check` matched 185 methods; and `git diff --check github/refactor/ui_web_architecture...HEAD` passed.

gloryfromca (commented):

The new revision adds the stacked #541 role-picker fix and then rebases the seven-patch stack onto the advanced target. The picker now escapes the settings panel scroller, remains bounded by the dialog, follows resize geometry, and closes only when its anchor actually scrolls away. Only `RolePill` supplies the new anchor; unanchored callers retain their existing layout. No concrete failure survived review.

A range comparison against the pre-rebase stack found six identical patches and only conflict-context movement around setup translations in the seventh; the settings behavior previously reviewed remains intact. Both review threads opened here remain resolved.

Coverage included AGENTS/CLAUDE, relevant CONTEXT and ui-web contribution rules, the full target diff, the stacked picker delta, rebase history, callers, backward compatibility, test strength, and the settings/component architecture. No tests were weakened.

Verification on the actual rebased head in a temporary detached worktree: `uv run pytest tests/test_rpc_console.py -q` passed 99 tests; `npm test -- --run` passed 2,544 tests across 186 files; type-check passed; lint completed with 0 errors and 5 pre-existing warnings; RPC generation matched 189 methods; the production build passed with both boot snapshots matching; and the full target diff passed `git diff --check`.

0xKT (approved):

PR #541, opened 2026-09-20
fix(ui-web): float the role picker over the dialog

gloryfromca:
## Summary

Stacked on #540, which this branch's parent commit is; review that one first.
The two touch the same component.

The settings role picker was absolutely positioned inside the row that opened
it, and those rows live inside the settings panel's own scroller (`.spanels`,
`overflow-y: auto`), which clips an absolutely positioned child. Measured on a
1512x900 window, ten of the eleven role rows lost part of the 400px panel:

| row | cut off | visible |
|---|---|---|
| main chat | 69px | 331 of 400 |
| session naming | 144px | 256 of 400 |
| skill filter | 207px | 193 of 400 |
| embedding | 270px | 130 of 400 |
| rerank | 333px | 67 of 400 |
| the five media rows | 69-322px | - |
| the last row | 0 | 400 of 400 |

Opening one also scrolled the panel, because the search field takes focus and
the browser brings it into view, so the rows above moved at the moment of
opening.

**What changed.** `lib/popover.ts` gained `anchorRow`, beside the `place` the
composer's two popovers already share. It hangs the panel below the row where
the dialog has room under it, above the row where it does not, and clamps to the
dialog either way.

`position: fixed` rather than absolute, so the scroller has nothing to clip: a
fixed box resolves against `.smodal`, which is above the scroller and is already
the containing block `page.css` gives the add-model drawer, for the same reason
and in the same words -- "clipped by this dialog rather than floating over the
whole window". No portal and no new body-level layer: the dialog is the bound
the prototype draws, and it is already a containing block.

The origin is measured rather than assumed. Which ancestor a fixed box resolves
against depends on one having a transform, and `.smodal`'s comes from its entry
animation; reading back where `top: 0` lands keeps the arithmetic right whether
that stays true or not.

**Scroll.** An open panel closes when its row scrolls away rather than following
it. Following means re-placing on every scroll frame, which is geometry to
maintain for a gesture nobody makes while choosing; a panel that stays put while
its row slides off is naming the wrong row. The test is that the anchor moved,
not that a scroll fired -- focusing the search field can scroll the panel by
itself, and closing on the event would close the panel on the frame it opened.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
cd ui-web && npm test           -> 182 files, 2527 passed
cd ui-web && npm run type-check -> clean
cd ui-web && npm run lint       -> 0 errors, 5 pre-existing warnings
python ui-web/build.py          -> boot-snapshot OK, 250 and 251 nodes match golden
```

Four cases were added and each was mutation-verified:

- always hang below -> `flips above a row the panel would overrun` reds
- close on any scroll -> `closes once the row it names has scrolled away, and not
  before` reds
- never place at all -> both placement cases red

Real host (`raven serve`, temporary RAVEN_HOME, 1512x900, driven through the
page): every one of the eleven role rows opened with zero pixels clipped at
either edge of the dialog, against the table above taken the same way before the
change.

## Risk

User-visible: the panel floats over the dialog instead of pushing the rows
apart, and closes if the list scrolls under it.

`anchorRow` is new and has one caller, so nothing that works today changes
route; `place` and the composer's own picker are untouched. The component
without an `anchor` behaves exactly as before, which is what keeps the change
off every other caller. Rollback is the one commit.

A follow-up worth considering separately: the page now has two model pickers,
this one and the composer's portalled `features/model/ModelPicker`, and
`lib/popover.ts` says in its own comment that the second was meant to serve the
settings row too. Merging them would delete this component, but it would put the
change on the most used control on the page, so it does not belong in a fix.

## Related Issues

N/A


ui-web/src/components/ModelPicker.tsx
```diff
@@ -47,8 +53,33 @@ export function ModelPicker({ title, providers, current, onPick, onClose, emptyN
...
+     search field can itself scroll the panel a little, and a listener that
+     closed on the event would close the panel on the frame it opened. */
+  useEffect(() => {
+    if (!anchor) return
+    const was = anchor.getBoundingClientRect().top
+    const off = (): void => {
+      if (Math.abs(anchor.getBoundingClientRect().top - was) > 1) onClose()
+    }
+    document.addEventListener('scroll', off, true)
+    window.addEventListener('resize', off)
```
gloryfromca: No blockers; suggestions only, and they are marked inline.

A width-only resize can move the role pill horizontally while leaving its `top` unchanged (for example, as the 94vw desktop dialog narrows above the 900px breakpoint). This handler then does nothing, so the fixed picker keeps its old horizontal coordinate and can be clipped by `.smodal`. Please either compare both axes and close, or rerun `anchorRow` on resize. Closing and reopening recovers immediately, so I do not consider this merge-blocking.

gloryfromca (commented):

Reviewed `github/fix/settings_three_review_findings...HEAD` at `94b092253db4`. I covered the repository rules in `AGENTS.md`, `CONTEXT-MAP.md`, `ui-web/CONTEXT.md`, and `ui-web/CONTRIBUTING.md`; the complete diff; the settings dialog DOM and CSS containing blocks; the sole production `ModelPicker` caller; relevant picker/popover history; backward compatibility of the optional unanchored path; architecture boundaries; and whether tests were weakened. The fixed-position approach escapes the `.spanels` scroller, the measured origin accounts for the dialog's transformed containing block, the existing unanchored behavior remains available, and no assertions were removed or relaxed. One recoverable resize issue is marked inline.

Verification: `npm test --prefix ui-web` passed 182 files and 2,527 tests; the two directly affected suites passed 19 tests; `npm run type-check --prefix ui-web` passed; `npm run lint --prefix ui-web` passed with five pre-existing warnings in untouched files; and `git diff --check github/fix/settings_three_review_findings...HEAD` passed.

ui-web/src/components/ModelPicker.tsx
gloryfromca: No blockers; suggestions only, and they are marked inline.

The rebase leaves this width-only resize edge case unchanged. It remains recoverable by closing and reopening the picker, so it is being carried as a named nonblocking follow-up in the current revision summary rather than left as an open review objection.

gloryfromca:
Fixed in de641aade -- the premise holds. `.smodal` is `min(1000px, 94vw)`, so narrowing the window above the breakpoint moves the row on one axis only, and a guard that reads `top` sees nothing.

I took the second of your two options rather than the first. Comparing both axes would close the panel on a resize, and a resize does not do to the row what a scroll does: a scroll slides the row out from under the panel, so a panel left pointing at it is worse than no panel; a resize leaves the row exactly where it is in the list and only moves the dialog around it. Taking the panel away there is a loss for a gesture that never touched the list. So the two listeners now do different things -- scroll closes on a moved row, resize calls `anchorRow` again and the panel follows.

`top` stays the only axis the scroll guard compares, and the comment now says why: a vertical scroller is the only thing that moves the row under a scroll, and the horizontal case is the resize, which re-places.

New case: `re-places rather than closes when a resize slides the row sideways`. Mutation-verified by pointing the resize listener back at the scroll guard, which reds it on the stale coordinate (`expected '632px' to be '420px'`).

Full web suite 2529 passed, type-check clean, lint 0 errors.

One correction to the earlier CI noise on this PR: the three jobs my watcher reported as failures on 2db1fbd2d were `cancelled`, not `failed` -- editing the PR title started a fresh run that superseded the in-flight one. The real failure before that was the commit scope, and `settings` is now `ui-web` in both the commit and the title.

gloryfromca (commented):

This head is a content-equivalent rebase of the reviewed picker change onto the updated target at `99ab1084d`; the five-file PR diff is unchanged from the prior revision, and no new findings appeared. Rechecked the current full diff against `AGENTS.md`, `CONTEXT-MAP.md`, `ui-web/CONTEXT.md`, and `ui-web/CONTRIBUTING.md`, along with callers, relevant history, backward compatibility, test integrity, and architecture boundaries.

Named follow-up: the previously noted width-only resize positioning edge case remains recoverable and nonblocking. Its old inline thread is resolved so it does not read as an outstanding merge objection.

Verification on `2db1fbd2d219`: `npm test --prefix ui-web` passed 182 files and 2,528 tests; `npm run type-check --prefix ui-web` passed; `npm run lint --prefix ui-web` passed with the same five warnings on untouched files; and `git diff --check github/fix/settings_three_review_findings...HEAD` passed.

gloryfromca:
One ordering note, since the summary and the head have drifted apart.

The latest revision summary verifies `2db1fbd2d219`, which is the rebase commit. The resize fix landed after it as `de641aade`, and that is the current head. The test count in the summary is the tell: it reports 2,528, and `de641aade` is 2,529 -- the extra case is `re-places rather than closes when a resize slides the row sideways`.

So the named follow-up is not outstanding. `window.addEventListener('resize', ...)` now calls `anchorRow` again instead of the close guard, which is the second of the two options from the original comment; the reasoning for choosing it over comparing both axes is in the comment above.

Nothing to do on my side -- flagging it only so the thread is not read as a carried edge case when the head no longer has it.

gloryfromca:

Confirmed on `de641aade`: routing resize to `anchorRow` fixes the reported width-only stale coordinate while preserving the picker, and the new test covers that behavior. One related stale-baseline edge remains after a height resize and is marked inline in the current review; it is recoverable and nonblocking.

ui-web/src/components/ModelPicker.tsx
```diff
@@ -47,8 +53,43 @@ export function ModelPicker({ title, providers, current, onPick, onClose, emptyN
...
+     opened. Only `top` is compared because only `top` is what a vertical
+     scroller moves; the horizontal case is the resize, which re-places. */
+  useEffect(() => {
+    if (!anchor) return
+    const was = anchor.getBoundingClientRect().top
+    const off = (): void => {
+      if (Math.abs(anchor.getBoundingClientRect().top - was) > 1) onClose()
+    }
+    const again = (): void => {
+      if (box.current) anchorRow(box.current, anchor)
```
gloryfromca: No blockers; suggestions only, and they are marked inline.

After a height resize, the centered `.smodal` moves the anchor vertically and `again` re-places the picker correctly, but `was` still holds the opening-time top. Because the document listener captures descendant scrolls, the next scroll inside `.model-picker-right` makes `off` see that old resize delta and close the picker even though the anchor has not moved since re-placement. Refreshing the baseline in `again` would preserve the intended guard. This needs a height resize followed by a scroll and reopening recovers, so it is a nonblocking follow-up rather than a merge objection.

gloryfromca (commented):

The reported width-only resize defect is fixed on `de641aade`: resize now re-runs `anchorRow`, preserves the open picker, and the added test covers the stale horizontal coordinate. One adjacent, recoverable height-resize baseline case is marked inline and carried as a named follow-up; it does not hold up this change.

This round covered the current target diff and `2db1fbd..de641aad` delta, repository and Web UI context rules, the production caller and dialog CSS, relevant history, backward compatibility, architecture boundaries, and test integrity. No assertions were removed or weakened.

Verification: `npm test --prefix ui-web` passed 182 files and 2,529 tests; the two directly affected suites passed 20 tests; `npm run type-check --prefix ui-web` passed; `npm run lint --prefix ui-web` passed with five warnings on untouched files; and `git diff --check github/fix/settings_three_review_findings...HEAD` passed.

gloryfromca:
Fixed in 66c10cfc0. The premise holds and the detail about the capture-phase listener is the part that makes it matter: `.model-picker-right` is its own scroller, so choosing a model reaches that document listener -- the stale baseline would have been read on the most ordinary gesture in the panel, not a rare one.

`again` now refreshes `was` after re-placing, so the baseline moves with the panel.

New case: `does not read a height resize as a row that scrolled away`. It resizes, asserts the panel survived, scrolls without moving the row and asserts it is still open, then moves the row and asserts it does close -- so the guard is still doing its job rather than being disabled. Mutation-verified by dropping the refresh line, which reds it with `the row has not moved since the panel was re-placed: expected [ 1 ] to deeply equal []`.

Full web suite 2530 passed, type-check clean, lint 0 errors.

This one was mine to introduce: the resize handler in de641aade created the window between the panel's position and the guard's baseline, and I did not look at what else held that top.

gloryfromca:

Confirmed on `66c10cfc0`: the resize path now updates the scroll guard baseline after re-placement, and the regression test proves both that the first unrelated captured scroll stays open and that a later real row movement still closes. The prior finding is settled.

gloryfromca (commented):

The post-resize scroll baseline now moves with the re-placed picker, and the added regression test covers both the false-close path and the later genuine-scroll close. The prior finding is resolved, and no new findings remain.

This round covered the complete current target diff and `de641aad..66c10cfc` delta, repository and Web UI context rules, the production caller and dialog geometry, relevant history, backward compatibility, architecture boundaries, and whether tests were weakened. The assertions were extended rather than relaxed.

Verification on `66c10cfc0b25`: `npm test --prefix ui-web` passed 182 files and 2,530 tests; the directly affected suites passed 21 tests; `npm run type-check --prefix ui-web` passed; `npm run lint --prefix ui-web` passed with five warnings on untouched files; and `git diff --check github/fix/settings_three_review_findings...HEAD` passed.

gloryfromca (commented):

This revision is a content-equivalent rebase onto target `8d256c165`; the old and new target-relative diffs have the same stable patch ID, and the PR remains the same five picker files previously reviewed clean. Rechecked the current diff, repository and Web UI rules, callers and relevant history, backward compatibility, test integrity, and architecture boundaries; the updated base introduces no interaction with the picker change. All previously raised threads remain resolved.

Verification on `d49784f03b33`: `npm test --prefix ui-web` passed 182 files and 2,530 tests; `npm run type-check --prefix ui-web` passed; `npm run lint --prefix ui-web` passed with five warnings on untouched files; and `git diff --check github/fix/settings_three_review_findings...HEAD` passed.

gloryfromca (commented):

This head is another content-equivalent rebase, now onto target `80a0a6f9b`; the target-relative stable patch ID is unchanged from the previously clean revision. Rechecked the current diff, repository and Web UI rules, callers and history, backward compatibility, test integrity, and architecture boundaries. The target's one-line meta-tool spacing change does not interact with the picker patch, and all prior threads remain resolved.

Verification on `140401936610`: `npm test --prefix ui-web` passed (182 files, 2,530 tests); `npm run type-check --prefix ui-web` passed; `npm run lint --prefix ui-web` passed with five warnings on untouched files; and `git diff --check github/fix/settings_three_review_findings...HEAD` passed.

gloryfromca (commented):

This head is a content-equivalent rebase onto target `1201d03dc`; its target-relative stable patch ID remains unchanged from the previously clean revision. Rechecked the complete current diff, repository and Web UI rules, callers and history, backward compatibility, test integrity, and architecture boundaries. The target's new Tools-page switch change is outside the picker files and introduces no interaction with this patch; all prior review threads remain resolved.

Verification on `0330480037f0`: `npm test --prefix ui-web` passed (182 files, 2,530 tests); `npm run type-check --prefix ui-web` passed; `npm run lint --prefix ui-web` passed with five warnings on untouched files; and `git diff --check github/fix/settings_three_review_findings...HEAD` passed.

gloryfromca (commented):

This head is a content-equivalent rebase onto target `9f4e4ed4f`; its target-relative stable patch ID is unchanged from the previously clean revision. Rechecked the complete current diff, repository and Web UI rules, callers and history, backward compatibility, test integrity, and architecture boundaries. The target's installed-tool state fix has no shared files or call path with the picker change, and all prior review threads remain resolved.

Verification on `4434b401dc29`: `npm test --prefix ui-web` passed (183 files, 2,532 tests); `npm run type-check --prefix ui-web` passed; `npm run lint --prefix ui-web` passed with five warnings on untouched files; and `git diff --check github/fix/settings_three_review_findings...HEAD` passed.

PR #542, opened 2026-09-20
feat(*): a conversation starts in a folder of your choosing, and the rail groups by it

Handsome-wzw:
## Summary

A conversation can now be started in a folder of the person's choosing, from the page, the way a code editor's folder switcher works: on the new-task screen the composer carries a folder chip that opens No folder / Recent / Open folder..., and the folder picked rides on `session.create` as its `workdir`. The engine already pinned a session's turns to that directory (shell, filesystem, deliver and media tools follow the binding); the page had no way to set it.

The choice is made once. After the first message the chip only reports the conversation's folder and cannot be pressed, which is exactly the engine's semantics: the pin is taken at creation and every later turn reads it.

The rail splits the rest of the list on the same fact: conversations pinned to a folder sit under their own collapsible heading, each row wearing the folder's name (full path on hover); conversations on the policy default keep the old heading, renamed to say what it is once the other group exists. Pinned and cron rows are untouched.

Two small backend additions carry it:

- `session.list` items gain an optional `workdir`, the stored override or absent, which is what the rail groups on.
- `fs.dirs` lists the subdirectories of one absolute directory (home when omitted), directories only, dotfiles omitted, capped at 500, and marks each entry -- and the listed directory itself -- with whether `session.create` would accept it (`validate_override`: not the agent home, not its ancestors, not its memory/skills/sessions trees). The picker greys those out instead of offering a folder the create would refuse.

"Recent" is derived from the rail's own rows, newest activity first, so nothing new is stored and every browser sees the same list. The offline demo page installs neither hook; the chip works and "Open folder..." says it needs a gateway.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_rpc_console.py tests/test_rpc_session.py tests/test_rpc_contract_shapes.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py tests/test_ui_language_repaint.py`: 661 passed. New: six `fs.dirs` tests (subdirectories only, dotfiles omitted, sort, `ok` on entries and on the listing, home default, root has no parent, relative/file/missing/unreadable refused, a child vanishing mid-listing), one `session.list` test (the override or null), and the `fs.dirs` round trip against its declared result model.
- `scripts/coverage_gate.py diff --base-ref upstream/main`: 97.83% of changed executable lines (45/46).
- `ruff check` and `ruff format --check`: clean.
- ui-web: `npm run gen:check` (179 methods, in sync), `npm run type-check`, `npx vitest run`: 109 files, 1849 tests passed. New: `shell/workdir.test.ts` (draft/locked states, recents deduped, stage and unstage, the browse walk with greyed entries and a disabled use button, the offline and refused-path notes), three rail tests (the two groups, the tag on either separator, folding), two `open-conversation` tests (the staged folder rides the create and lands on the row; none staged creates with none).
- `python ui-web/build.py`, `check-page.mjs`, `check-css.mjs`, `check-class-namespace.mjs`: OK.
- ui-tui: `npm run gen:rpc`, `npm run gen:i18n`, `npm run lint:rpc`, `npm run lint:i18n`, `npm run type-check`: in sync and clean.
- On a real gateway, in the served page: opened the picker on the new-task screen, browsed home into a demo folder, picked it; the chip named the folder; promoted the draft; the chip locked with the fixed-once-started note and no longer opened; `fs.list` for the new session rooted at the demo folder; a turn asking the agent to run `pwd` answered with that folder; `session.list` carried it as `workdir`; the rail showed the row under the folder group with its tag, and the other rows under the no-folder heading.

## Risk

`fs.dirs` lists directory names on the gateway's machine to an authenticated page. The page already lists and reads files under a session's working directory, and the agent's shell tool reaches the whole machine, so this widens what the page can see by directory names only; dotfiles stay hidden and the agent's own data is marked unusable, matching `session.create`'s refusal. A page that never presses the chip behaves exactly as before: `session.create` is called with no `workdir`, and a `session.list` consumer that ignores the new optional field sees the old shape. Reverting the commit restores the previous page and contract; sessions already pinned keep working, since the pin was the engine's feature already.

## Related Issues

N/A


ui-web/src/shell/workdir.test.ts
```diff
@@ -0,0 +1,205 @@
+// @vitest-environment happy-dom
```
gloryfromca, line 1: Blocking: add the required file-level purpose documentation to this new test file.

AGENTS.md section 1.1 requires every new code file to document its purpose in English with a module docstring or equivalent file-level documentation. The Vitest environment directive configures the runner, and the comment at line 12 only explains the markup helper; neither documents what this test module covers. Please add a short file-level purpose comment before the imports.

gloryfromca (commented):
Blocking: add the required file-level purpose documentation to the new workdir test file.

One repository-rule blocker is marked inline.

Covered: the full `github/main...HEAD` diff; AGENTS.md, CLAUDE.md, and the context map; changed callers and relevant workdir/session history; RPC models, wire schema, and generated clients; draft promotion, session reopening, rail grouping, and filesystem browsing; the authenticated loopback transport and data-exposure boundary; backward compatibility; and whether tests were weakened. I found no additional functional, security, compatibility, or architecture defect after attempting to refute each candidate.

Verification:
- `uv run pytest tests/test_rpc_console.py tests/test_rpc_session.py tests/test_rpc_contract_shapes.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py`: 650 passed after syncing the declared `dev` extra. The initial environment-only run was 644 passed and 6 failed because `raven_everos` was not installed.
- ui-web `gen:check`, `type-check`, and `vitest run`: 109 files, 1849 tests passed.
- ui-tui RPC/i18n generation checks and type-check: passed.
- Ruff check/format, source-language and large-file gates, production web build, page/CSS/class checks, and `git diff --check`: passed. (`make` is unavailable in this environment, so the two make-backed Python gate scripts were invoked directly.)

ui-web/src/shell/workdir.test.ts
Handsome-wzw: Added a file-level note before the imports describing what the module covers (the chip states, the menu, staging, the browse walk, and the two notes); the vitest directive stays on line 1 as the runner requires.

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed: the new note is file-level, precedes the imports, explains what the suite covers, and preserves the required Vitest directive on line 1. This resolves my finding.

gloryfromca (commented):

The new file-level note fixes the only prior blocker: it documents the test module's purpose, remains before the imports, and preserves the Vitest environment directive on line 1. The thread I opened is resolved, and I found no new findings.

Covered this round: the complete `github/main...HEAD` diff and the exact delta from the previously reviewed head; AGENTS.md, CLAUDE.md, and the context map; the previously reviewed callers, history, RPC/schema compatibility, security boundary, and architecture implications; and whether any test behavior was weakened. This revision changes documentation only, so none of those executable paths moved.

Verification:
- ui-web `gen:check`, `type-check`, and `vitest run`: 109 files, 1849 tests passed.
- Source-language and large-file gates plus `git diff --check`: passed.

ui-web/src/shell/workdir.ts
```diff
@@ -0,0 +1,364 @@
...
+    const empty = document.createElement('div')
+    empty.className = 'note'
+    empty.textContent = t('gui.wd.empty')
+    list.appendChild(empty)
+  }
+  for (const e of l.entries) {
+    const r = row(e.name, '', false, () => void browse(e.path))
+    r.setAttribute('role', 'button')
+    r.removeAttribute('aria-checked')
+    if (!e.ok) {
```
LivXue, line 264: **Nonblocking - let a row navigate even when its `ok` is false, and leave `ok` to the "use this folder" button.**

This line disables the row itself, so a directory that answers `ok: false` cannot be entered. But `ok` is `validate_override`'s verdict on whether a session may be PINNED there, and that validator also refuses every ANCESTOR of agent home (`if resolved in home.parents`, raven/agent/workdir.py). The two questions come apart for any directory that merely contains raven's own data: it may not be a working directory, while the folders inside it are perfectly good ones. The backend already says exactly this -- `test_fs_dirs_lists_subdirectories_only_and_marks_agent_home` carries the comment "the listing itself says it cannot be used even though its children can" -- so the handler is right and only the picker conflates the two.

Reproduced by calling `validate_override` directly on three layouts. The UI half is code-reading, at this line.

    agent home <tmp>/home/alice/.raven/workspace
      ok=False  <tmp>/home/alice            ok=True  <tmp>/home/alice/projects
    agent home <tmp>/vol/team/raven-home/workspace
      ok=False  <tmp>/vol/team/raven-home   ok=True  <tmp>/vol/team/raven-home/projects
    agent home <tmp>/workspace
      ok=False  <tmp>/workspace             ok=True  <tmp>/workspace/myproject

On the stock layout the greyed row is the reader's own home directory, met by pressing "up" out of it; the Home button still rescues that, so it costs a moment. Where agent home is not tucked under a dot directory -- `workspace_path` is `raven_home()/workspace`, or the configured `agents.defaults.workspace` used as written -- the valid folders beneath the greyed row have no other route, since the panel offers no typed path. They cannot be chosen at all.

Two smaller things sit on the same line. The tooltip is `gui.wd.blocked`, "it is inside its own data", which is the reverse of true for an ancestor: that directory contains raven's data rather than sitting in it. And the case has no test, because the fixture in `workdir.test.ts` greys out an entry named `.raven-data`, while `fs_dirs` skips `c.name.startswith(".")` -- a dotted entry never reaches `entries`, so the ancestor is the only kind of greyed row that can actually occur.

Selection is already refused at the destination by `use.disabled = !l.ok` a few lines below, so letting the row navigate does not make anything pinnable that is not allowed today.

raven/rpc/methods/console.py
```diff
@@ -1564,6 +1564,61 @@ async def fs_list(params: dict, *, agent_loop_factory=None) -> dict:
...
+        raise ConfigValidationError(str(e)) from None
+    entries = []
+    for child in children:
+        try:
+            if not child.is_dir():
+                continue
+        except OSError:
+            continue
+        entries.append({"name": child.name, "path": str(child), "ok": allowed(child)})
+        if len(entries) >= _FS_MAX_ENTRIES:
```
LivXue: **Nonblocking - bound the scan, not only the result, or move the walk off the event loop.**

The 500 cap counts directories appended, and the loop runs to the end of `children` to reach it, so the number of `is_dir()` calls is the number of entries in the directory rather than 500. The twin bounds the input instead: `fs_list` slices `children[:_FS_MAX_ENTRIES]` before it stats anything, and is additionally fenced by `_resolve_inside(root, rel)` to the session's own workspace, so its caller cannot choose the directory. `fs.dirs` drops both bounds by design -- the path is caller-supplied and deliberately unrooted -- and the walk is synchronous on the event loop.

Measured at this head, with the package extracted by `git archive cd7968084029 raven` and loaded from that tree:

    fs.dirs  30000 files + 1 dir     228.1 ms   entries=1
    fs.dirs  1200 dirs                70.6 ms   entries=500

The file-heavy directory costs three times the dir-heavy one to return a single row, on a local tmpfs; a network mount or a cold cache widens that. `fs_read` in this same module already records why it matters, having moved its read to `asyncio.to_thread` because a large blocking read was "on the event loop, stalling every other client on the shared socket".

This is not a security finding: the dispatcher sits behind the per-boot auth token, and `fs_list`'s own blocking `iterdir` predates this branch. What is new is that the directory is now the caller's choice and the scan is no longer bounded by the cap that appears to bound it.

LivXue:
No blockers; two nonblocking findings are marked inline.

Scope: I read the full diff of this PR -- 29 files, +1284/-10, merge base 7faa2060e, head cd7968084 -- diffed locally rather than through the files API. I normally review the `raven/agent/**` slice of a PR; this one touches no file under it, so that fence bound nothing and the verdict below covers the whole change.

Examined: the new `fs.dirs` handler and its containment story against the existing `fs.*` verbs; `validate_override` and whether the handler's docstring claim about `session.create` holds; the `SessionListItem.workdir` addition through the OpenRPC document and both generated clients; the draft-to-session promotion path in the live layer, including what happens to a staged pick when the create is refused; the rail regrouping and every render site of the new `.wdt` class; the five new `window.*` verbs against every harness under `ui-web/scripts/`; and the i18n pair for all fifteen new keys.

Confirmed findings:

- The folder picker disables a listing row when its `ok` is false, but `ok` answers whether a session may be pinned there, and `validate_override` also refuses every ancestor of agent home -- so a directory that merely contains raven's data cannot be entered, and the valid folders under it become unreachable where agent home is not under a dot directory.
- `fs.dirs` caps the result at 500 directories but not the scan, so a caller-chosen directory is walked to its end with one `is_dir()` per entry, synchronously on the event loop; `fs_list` caps the input instead and is rooted at the session workspace.

Verification:

- `pytest tests/test_rpc_console.py tests/test_rpc_session.py tests/test_rpc_contract_shapes.py`: 274 passed. `pytest tests/test_rpc_schema_match.py tests/test_rpc_registration.py`: 374 passed, 2 skipped (the two skips want ui-tui sources, which I did not extract). Both runs against the head, from a tree produced by `git archive cd7968084029`, with the cwd outside the checkout so the editable install could not shadow it.
- `validate_override` driven directly on three agent-home layouts for the first finding, and `fs_dirs` timed at the head for the second. Numbers are in the inline notes.

- The docstring claim that each entry's `ok` is "the answer `validate_override` would give `session.create`" holds: both call `validate_override(path, load_config().workspace_path)`.
- Symlink containment is sound -- `validate_override` resolves physically before comparing, so `ok` cannot be spoofed by a link into the agent's data.
- The five new `window.*` verbs: only `boot-order.test.mjs` and `page-switch-live.test.mjs` reach the paths that call them, and both were updated. `model-refresh-live.test.mjs` reads `080-overrides.js` as a string and does not execute it.
- `openLiveSession` is only ever handed a real rail row (`sess(id)` resolves through `sessionRows()`), so the new `setSessionWorkdir` call cannot read a field off a synthetic row.
- `.wdt` is declared as `.sess .t .wdt` and `Row` always renders with `className="sess"`, so the rule covers the search-result render site too.
- `page.html` and `i18n/messages.json` already carried CJK at the merge base, so the added lines pass the source-language gate on the carrier rule.
- A cosmetic one I am leaving out: once every conversation has a folder, the second group still renders under `always` as an empty "No folder" heading. It is the pre-existing behaviour of that flag, not something this branch decided.
- The file-level documentation blocker on `workdir.test.ts` was filed by gloryfromca, fixed, and resolved at this head. Not re-filed.

ui-web/src/shell/workdir.ts
Handsome-wzw: Agreed on all three points; fixed in 64b330cc. The row is no longer disabled: it stays a way in, marked (greyed text, the note on hover), and only the pick is withheld by the use button on the listing's own ok, which it already did. The note now reads 'Raven's own data is here or beneath it; its folders may still be', true for an ancestor as well. The test drives the ancestor case on a row the gateway can actually produce (/srv/raven-home above /srv/raven-home/workspace): the row is enabled and marked, entering it shows use disabled, the good child is clean, and the pick lands from there.

raven/rpc/methods/console.py
Handsome-wzw: Agreed; fixed in 64b330cc. The walk now slices children[:500] before it stats anything, so the file-heavy case costs 500 stats rather than one per name, and the whole resolve/list/stat/judge half runs in asyncio.to_thread, the way fs_read does. The contract's description now says what the cap bounds ('only the first 500 names are examined'). Test: a directory of 600 files plus one directory sorted past them sees exactly 500 is_dir calls and answers no entries, which is the price fs.list pays for the same bound.

gloryfromca (commented):

At 64b330ccaa4f, the follow-up changes correctly separate navigating into a directory from selecting it as the workspace, while the backend now limits expensive per-entry checks and performs the filesystem walk off the event loop. I tried to refute both fixes against validate_override, the picker caller, error propagation, and the generated RPC contract, and found no concrete failure.

Coverage this round: the full github/main...HEAD diff; AGENTS.md/CLAUDE.md and the routed runtime context; touched callers and history; backward compatibility; generated RPC/i18n consistency; and whether the tests were weakened. The prior file-purpose fix remains present, and my thread is resolved.

Verification:
- uv run pytest tests/test_rpc_console.py: 102 passed
- uv run pytest tests/test_rpc_contract_shapes.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py tests/test_rpc_session.py: 549 passed
- ui-web gen:check and type-check: passed
- ui-web test: 109 files, 1849 tests passed
- ui-tui RPC/i18n generation checks and type-check: passed
- source-language, large-file, and git diff checks: passed

raven/rpc/methods/console.py
```diff
@@ -1564,6 +1564,75 @@ async def fs_list(params: dict, *, agent_loop_factory=None) -> dict:
...
+        except ValueError:
+            return False
+        return True
+
+    try:
+        children = sorted((c for c in target.iterdir() if not c.name.startswith(".")), key=lambda c: c.name.lower())
+    except OSError as e:
+        raise ConfigValidationError(str(e)) from None
+    entries = []
+    for child in children[:_FS_MAX_ENTRIES]:
```
LivXue: **Blocker - bound the scan without dropping directories: the slice is applied to a name-sorted list, so a subfolder that sorts past the first 500 names disappears from the picker.**

The `asyncio.to_thread` half of this commit is right, and it is the half that closed my earlier note: the walk no longer runs on the event loop, so a caller-chosen slow mount cannot stall the other clients on the socket. The slice added beside it is the problem.

`children` here is sorted by name only (`key=lambda c: c.name.lower()`, three lines up). Taking `[:500]` of that list drops every directory whose name sorts after the first 500 entries. `fs.dirs` then answers `entries: []` for a directory that has subfolders, the panel renders `gui.wd.empty` ("No subfolders"), and since the panel offers no typed path, that subtree cannot be reached at all. Nothing tells the reader the listing was cut, and whether it happens depends on the folder's name, so the same shape of directory behaves differently from one machine to the next.

Reproduced at this head, with the package extracted by `git archive 64b330ccaa4f raven` and loaded from that tree:

    fs.dirs  600 files + 'zip-extracts'   ->  entries=[]
    fs.dirs  30000 files + 'onedir'       ->  100.9 ms  entries=[]

A `Downloads` with six hundred files and one `zip-extracts/` in it is enough.

On the justification, in the reply above and repeated in the new test's comment -- "the price fs.list pays for the same bound". `fs_list` does not pay it. Its sort key is `(not p.is_dir(), p.name.lower())` (console.py:1543), which puts directories first, so its own `[:500]` keeps them. Measured on the same directory:

    fs_list sort key, is 'zip-extracts' within the first 500 -> True

What `fs_list` does pay is the stat bill this commit set out to avoid: that key calls `is_dir()` on every child while sorting, so its cap bounds the answer and never bounded the work. The two functions cannot both be cited as the same bargain; `fs_list` buys a correct answer at O(N) stats, and `fs.dirs` now buys O(500) stats at the cost of the answer.

`test_fs_dirs_looks_at_the_first_500_names_and_no_more` pins the lost directory with `assert r["entries"] == []` on a root that contains `zzz-dir`, so correcting this later is a test change rather than an addition.

Remedy direction, not a patch: with the walk already off the event loop the slice is no longer load-bearing, so the cap can go back to counting directories FOUND, the way it did before, if the type comes from the dirent instead of a stat. `os.scandir` gives `is_dir()` off `d_type` on the filesystems this runs on. Measured, same 30000-file directory, cap on directories found:

    os.scandir walk  ->  19.1 ms  found=['onedir']

That is five times faster than this commit and twelve times faster than the version I first complained about, with the answer intact.

Disclosure: this head already carries a clean verdict from another reviewer. I am filing against it rather than around it -- the lens that finds this is not validate_override or the contract, it is asking what the new bound does to the answer.

raven/rpc/methods/console.py
LivXue: Half of this closes the note and half of it does not, so I am leaving the thread open.

The `asyncio.to_thread` move is the part I was after: the walk is off the event loop, so a caller-chosen slow mount can no longer stall the other clients on the socket.

The `children[:500]` added beside it introduces a worse problem than the one it bounds. The list is sorted by name only, so a subfolder sorting past the first 500 names is dropped and the picker answers "No subfolders" for a directory that has them, with no typed path to reach it by. And `fs_list` is not the precedent it is cited as -- its sort key puts directories first, so its own slice keeps them.

Filed as a blocker at console.py:1620 with the measurements and a remedy that keeps both the bound and the answer.

LivXue:
Blocking: one finding needs correction before this revision can merge; it is marked inline. This supersedes my "No blockers" note at cd7968084029, which was a verdict on that head.

Scope: the full diff of this PR, diffed locally, now 29 files and +1348/-10 at 64b330ccaa4f. I normally review the `raven/agent/**` slice; this PR touches no file under it, so that fence bound nothing and this verdict covers the whole change.

Round delta since cd7968084029: one commit, 64b330cca, answering both of my notes.

The first is closed, and I checked the fix rather than the commit subject. The row is no longer `disabled`, it carries `off`, and `page.css` gained `.pop .prow.off .txt` and `.pic` rules -- so "enterable but marked" is really rendered and is not a logic change with no paint behind it. `gui.wd.blocked` now reads true for an ancestor, in both locales. The new frontend case drives an ancestor row the gateway can actually produce, which the previous fixture could not: it greyed out a dotted name, and `fs_dirs` filters those out before they reach `entries`.

The second is half closed. Moving the walk into `asyncio.to_thread` is the part that answered my note, and it stands. The `children[:500]` added beside it is the blocker: the list is sorted by name only, so a subfolder sorting past the first 500 names is dropped, `fs.dirs` answers an empty listing for a directory that has subfolders, and the panel reports "No subfolders" with no typed path to reach it by. The `fs_list` precedent cited for it does not hold -- that function sorts directories first, so its slice keeps them, and its cap never bounded the work in the first place. `test_fs_dirs_looks_at_the_first_500_names_and_no_more` now pins the dropped directory, so this gets harder to correct the longer it stands.

Verification, all at 64b330ccaa4f from a tree produced by `git archive`, cwd outside the checkout so the editable install could not shadow it:

- `fs.dirs` on 600 files plus `zip-extracts/`: `entries=[]`. On 30000 files plus `onedir/`: 100.9 ms, `entries=[]`.
- `fs_list`'s sort key on the same directory: `zip-extracts` is within the first 500. So the two functions do not make the same bargain.
- An `os.scandir` walk taking the type from the dirent, capped on directories found: 19.1 ms, `found=['onedir']` -- faster than this commit and correct, which is why I do not think the bound and the answer are actually in tension here.
- Frontend fix read against `page.css`, the i18n pair, and the new test; not separately executed, since the earlier suite run covers the file and the change is a class swap plus two rules.

One note on process: this head already carries a clean verdict from another reviewer, given after an attempt to refute both fixes. I am filing against that rather than around it. The lens that reaches this one is not `validate_override`, the picker caller, error propagation or the generated contract -- all of which are fine -- it is asking what the new bound does to the answer the caller gets back.

gloryfromca (commented):
Blocking: the 500-name slice can hide existing subdirectories from the folder picker.

The later finding on this revision changes my stance. I independently reproduced it with 600 name-sorted files followed by zip-extracts/: _walk_dirs returned entries=[] even though that subdirectory existed. The picker then says there are no subfolders and provides no typed-path route into the omitted subtree.

The off-event-loop change remains sound; the blocker is specifically slicing before directory detection. The focused command uv run pytest tests/test_rpc_console.py -k fs_dirs reports 7 passed, but the new cap test currently codifies the failure by asserting that the later directory is absent.

I am not duplicating or resolving LivXue's open finding; that thread owns the objection.

raven/rpc/methods/console.py
Handsome-wzw: You are right, and thank you for checking the fix rather than the subject: the slice bought speed with the answer, and the fs_list precedent I cited does not hold since its sort key puts directories first. Fixed in 90e2bc4f along your remedy. The walk is os.scandir with is_dir taken from the dirent, the cap counts directories FOUND (sorted by name, first 500), and the to_thread half stays. The test that pinned the dropped folder is replaced by one that requires it: 600 files plus zip-extracts/ answers ['zip-extracts']; a second test holds the cap at 500 directories on a 600-directory root. Measured here on 30000 files plus one directory: 15.4 ms, entries=['zip-extracts'] (was 100.9 ms and [] at 64b330cc). The contract description now says 'at most the first 500 found'.

raven/rpc/methods/console.py
Handsome-wzw: Closing the other half here as well: 90e2bc4f drops the children[:500] slice. The walk is os.scandir with the type from the dirent, the cap counts directories found, and the to_thread move stays. Details and numbers are on the blocker thread below.

gloryfromca (commented):

At 90e2bc4ff57f, the prior blocker is fixed: fs.dirs now scans with os.scandir, keeps the filesystem work off the event loop, counts directories found toward the 500-entry cap, and no longer loses a real directory merely because files sort before it. I repeated the 600-files-plus-zip-extracts reproduction and received ['zip-extracts'].

Coverage this round: the full github/main...HEAD diff; AGENTS.md/CLAUDE.md and routed runtime context; the changed handler, UI caller, generated contracts, history, error behavior, and backward compatibility; and whether the corrected tests genuinely reversed the weakened expectation. I found no new concrete failure.

Verification:
- uv run pytest tests/test_rpc_console.py: 103 passed
- uv run pytest tests/test_rpc_contract_shapes.py tests/test_rpc_schema_match.py tests/test_rpc_registration.py tests/test_rpc_session.py: 549 passed
- ui-web gen:check and type-check: passed
- ui-web test: 109 files, 1849 tests passed
- ui-tui RPC/i18n generation checks and type-check: passed
- source-language, large-file, and git diff checks: passed

LivXue (approved):
Approving at 90e2bc4ff57f. This supersedes my "Blocking" note at 64b330ccaa4f, which was a verdict on that head: both findings are fixed here, and I checked the fixes rather than the commit subject.

Scope: the full diff of this PR, diffed locally -- 29 files, +1372/-10. I normally review the `raven/agent/**` slice; this PR touches no file under it, so that fence bound nothing and this covers the whole change.

Re-ran my own reproductions at this head, from a tree produced by `git archive 90e2bc4ff57f`, cwd outside the checkout:

    600 files + zip-extracts/      ->  ['zip-extracts']        (was [] at 64b330cc)
    30000 files + onedir/          ->  21.8 ms  ['onedir']     (228 ms correct, then 100.9 ms wrong)
    600 directories                ->  500 kept, d0000..d0499, name-ordered
    ancestor of agent home         ->  raven-home ok=False, its projects/ ok=True, its workspace/ ok=False

The third line is the one I had not asked for and am glad to see: sorting the dirents by name before the walk makes the cap take the alphabetically-first 500 rather than whatever order the kernel returned, so the same directory lists the same way on every machine. The fourth confirms the first finding's fix survived the rewrite.

Two things in the change I want to name rather than leave implicit. The test that pinned the dropped directory is gone and its successor asserts the inverse (`== ["zzz-dir"]`), so the guard now requires the folder instead of recording its absence. And the permission-error test moved its patch from `Path.iterdir` to `os.scandir` -- without that it would have kept passing while exercising nothing, which is the quieter half of changing which call does the work.

No further findings. The remaining merge gate is not mine to clear.

PR #543, opened 2026-09-20
fix(*): make the eight reload-only settings apply without a restart

gloryfromca:
## Summary

Eight keys on the settings page answered `Saved. Applies after the next gateway
reload or restart.` They do not any more, and the reload-only set is gone with
them -- there is nothing left for it to name.

They were all the same shape: the value was copied onto the objects the loop was
built with, and nothing read the file again. The fix is not a notification
mechanism. Each value is read from the file where it is used, so a write from
the page, the CLI or a hand edit is covered by construction rather than one
writer at a time.

**The two subsystem pins** -- `context.curatorModel` and
`skillForge.llmGateModel`, each with its provider -- are resolved per call
instead of once in the context-engine factory. Both consumers already funnelled
through one method (`CuratorSegmentBuilder._curator_binding`,
`LLMGateFilter._binding`) and already read the turn's binding when unpinned;
only the pin itself was frozen. `permissions.judgeModel` and
`sessionTitle.model` already read their pins this way, which is why those two
were never reload-only -- so the part that re-reads, binds through the pool and
reports an unusable pair once is shared as `providers.pool.live_pin_resolver`.

**`agents.defaults.maxToolIterations`** becomes a property over the live file,
falling back to what the loop was built with. A turn reads it once at its start,
so a change lands on the next turn rather than halfway through one.

**`agents.defaults.contextWindowTokens`** was the interesting one. It had two
copies with different readers: one frozen on the loop that only the usage report
read, and one on each `ModelBinding` that the budget read. A write reached
neither until a restart -- and after a model switch the two disagreed, because
the pool rebuilds its bindings when the number changes while the loop's copy
never moves. Measured on a real gateway before this change: trimming ran on the
new number while the page reported the previous model's catalogue size.

There is one source now, on the binding. The turn entry resolves it from the
file once and holds it for the whole turn tree, the same contract the
provider/model pair already has (`AgentLoop._with_live_window`): the budget is
read several times while a turn runs -- history trimming, the compaction check,
the usage report -- and a number that moved between those reads would leave one
turn disagreeing with itself. Outside a turn the default binding answers, which
is the rule `context_window_tokens` already followed.

Two consequences a reviewer should weigh:

- **The file is now the source for these values, not the object passed at
  construction.** In every production path they are the same thing --
  `EngineWiring` is filled from the same loader -- but a host that builds a loop
  with a synthesised `ContextConfig` or window and no config file behind it now
  gets the file's answer (or none). Clearing a key therefore takes effect too,
  which is what made `contextWindowTokens: null` on the settings page do
  anything at all.
- **`set_provider` carries the current default's window forward** rather than
  re-reading, because that entry point exists for a caller holding a provider of
  its own, which need not be a process this config was loaded for.

Not in this change: `assembler.set_context_window` / `ContextEngine.set_context_window`
have no caller in `raven/`, so the out-of-turn fallback windows held by the
curator and the history trimmer have never been updated after construction.
Pre-existing, untouched, and invisible to a turn (both read the active binding).

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Real host, one `raven serve` process, one session that never switched model, no
restart at any point:

```
turn 1                 context_max=0
settings.set 50000     {"applied": true, "previous": null}        <- no warning
turn 2                 context_max=50000
settings.set null      {"applied": true, "previous": 50000}
turn 3                 context_max=0
```

The same host earlier, on the code before this change, answered `context_max=0`
on turn 2 and carried `Saved. Applies after the next gateway reload or restart.`
in the reply; a restart on the same RAVEN_HOME was what produced 50000. The
disagreement described above was measured the same way: after `config.set model`
in that process the report read 131072 while a pool-built binding carried 50000.

`agents.defaults.maxToolIterations` was checked the same way on a real host
before this branch existed: the gateway log goes from `Iteration 1/40` to
`Iteration 1/3` across a `settings.set`, in one process.

The two pins are covered by tests rather than by a live host: the curator's slow
path and the skill gate did not run on a short turn, so there was nothing to
observe. The cases do a real factory build, rewrite the config file underneath
it and assert the next call follows.

- `uv run pytest tests/ -q`: 23325 passed, 4 failed, 111 skipped. All four
  failures reproduce on a clean checkout of the base commit in the same
  environment (`test_agent_loop_token_budget`, the two
  `test_agents_*_launcher` tool-face cases, `test_ppt_engine_image_search`);
  the launcher pair looks environment-dependent (a `SERPER_API_KEY` in the
  developer environment makes `web_search` available).
- After the rebase onto the current base: the same sweep over
  settings/context/skill-forge/loop/pool/usage-sink/onboarding, 2781 passed,
  1 failed (the pre-existing token-budget case).
- `npm run type-check` clean; `npx vitest run` 185 files, 2568 tests passed.
- Eighteen new or changed cases, each mutation-verified on its own -- among
  them: caching the resolver's first answer, warning on every call, reading one
  spelling of a block or a field, a turn rebuilding its window when the number
  has not changed, a running turn following a mid-turn edit, `set_provider`
  dropping a pinned window, and a settings write still claiming a restart is
  needed.

## Risk

User-visible: these eight writes take effect where they are read -- the next
call for a pin, the next turn for the cap and the window -- instead of after a
restart, and the settings surface no longer says otherwise. A turn already
running is never moved.

Contract: `settings.set` still returns `warning`; the only producer left is an
embedding model change, which the page renders the same way.

Cost: the added reads are a byte comparison on one small file (`LiveConfig`
re-parses only when the bytes change) plus a pool cache hit for a pin already
built. Nothing on this path builds a provider that was not going to be built.

Rollback: revert the commit. The two deleted holders (`AgentLoop._configured_window`,
`_RELOAD_ONLY_KEYS`) come back with it, and no other branch reads either.

## Related Issues

N/A


raven/context_engine/segments/curator.py
```diff
@@ -142,19 +143,13 @@ def _curator_binding(self) -> ModelBinding:
...
         """
-        if self.config.curator_model and self._pin is None and not self._pin_warned:
-            self._pin_warned = True
-            logger.warning(
-                "context.curator_model={!r} has no usable credentials of its own; "
-                "the curator follows the conversation's model instead",
-                self.config.curator_model,
-            )
-        return resolve(self._pin, self._fallback)
+        pin = self._pin_resolver() if self._pin_resolver is not None else None
```
gloryfromca, line 151: Blocking: resolve the curator pin once for the curation, not once per curator step.

`_slow_path` is one multi-request tool-calling conversation (up to 12 steps), and each request carries the preceding assistant tool calls and tool results. Since `_curator_binding()` now re-reads the file inside that loop, saving `context` between two steps sends the continuation to a different binding. That splits one in-flight turn across models/providers and makes the new model continue the old model's partial plan. Freeze the pair for the whole slow path (or the whole turn) so a live edit applies to the next curation instead.

raven/agent/loop/wiring.py
```diff
@@ -393,6 +393,42 @@ def context_window_tokens(self) -> int:
...
+        """The ReAct cap a turn runs under when its session pinned none.
+
+        Read live rather than frozen at build for the reason the permission
+        mode is: a cap is a sentence about the next turn, not work to redo.
+        The turn reads it once at its start, so a change lands on the next
+        turn rather than halfway through one.
+        """
+        from raven.config.live import max_tool_iterations
+
+        return max_tool_iterations(self._live_config) or self._default_max_iterations
```
gloryfromca: Blocking: snapshot `maxToolIterations` when the turn starts.

The live property is first read by `_run_agent_loop`, only after context assembly, whose curator slow path can make up to 12 network calls. If an operator changes the setting while that work is underway, this already-running request takes the new cap despite the new contract saying the edit lands on the next turn; lowering it can prematurely stop the active request, while raising it extends that request. Capture the cap at the `run_turn` boundary alongside the binding and carry that frozen value into the agent loop.

gloryfromca (commented):
Blocking: freeze the live curator pin and iteration cap at their promised turn boundaries.

I found two correctness issues, both marked inline. Each lets a settings edit made after a turn starts alter that in-flight turn; the curator case can switch model/provider between steps of one tool-calling conversation.

Verification:
- `uv run pytest tests/test_config_live.py tests/test_provider_pool.py tests/test_context_engine_factory.py tests/test_skill_forge_gate.py tests/test_agent_loop_model_switch.py tests/test_agent_loop_session_model.py tests/test_agent_loop_usage_sink.py -x`: 189 passed.
- The initial combined run including `tests/test_rpc_settings.py` stopped after 189 passes with 3 setup errors because `raven_everos` was not installed. Re-running that module with its declared dev extra, `uv run --extra dev pytest tests/test_rpc_settings.py -x`, passed all 87 tests.
- The web type-check was unavailable because `ui-web/node_modules` is absent.

Coverage included AGENTS.md and the Runtime terms routed by CONTEXT-MAP.md, the full diff, surrounding callers, relevant history/blame, backward-compatibility surfaces, whether tests were weakened, and the Model binding / Live preference / Generation architecture constraints. I did not raise lint/style matters or pre-existing issues on untouched lines.

raven/context_engine/segments/curator.py
gloryfromca: Confirmed and fixed in b4f67a8a0.

You are right about where the pin is asked for: `binding = self._curator_binding()` is inside `for step in range(1, self.max_steps + 1)`, and `messages` carries the previous steps' tool calls and results, so an edit between two steps hands the continuation to another provider with the old model's partial plan in it.

The fix holds the value for the whole turn rather than for the slow path alone, because the gate has the same shape one subsystem over: `run_turn` now enters `hold_for_this_turn()` beside `use_binding`, and `live_pin_resolver` answers from that hold. The first ask inside a turn resolves; every later one in the same turn gets the same pair; outside a hold every call is live, which is what a caller below `run_turn` and the background paths get.

A ContextVar for the reason the binding is one: work detached during a turn copies the context, so a sub-agent keeps the pair its turn started on.

Pinned by `test_a_pin_is_resolved_once_for_the_turn_that_holds_it` in tests/test_provider_pool.py, which changes the configured pair between two calls inside one hold and asserts the second still answers with the first. Mutation-verified: dropping the hold from the resolver reds it.

raven/agent/loop/wiring.py
gloryfromca: Confirmed and fixed in b4f67a8a0.

The ordering is as you describe: `_assemble_context_messages` runs at turn_path.py:2375 and `_run_agent_loop` at :2457, and the cap is read at :913 inside the latter -- so the first read lands after assembly, which is where the curator spends up to twelve requests. An edit in that window changed the request already running, which is the opposite of what this change claimed.

Resolved at the boundary now, not at first read: `_turn_scope()` calls `max_tool_iterations` itself and passes the value into `hold_for_this_turn(...)`, and `run_turn` enters it beside `use_binding`. A lazy hold would not have been enough here for exactly the reason you gave -- it would still be read after assembly.

`test_the_cap_is_resolved_when_the_turn_opens_not_at_its_first_read` pins it: it opens the scope, rewrites the config, then takes the first read and asserts the pre-edit value. Mutation-verified -- replacing the eager resolve with an empty `hold_for_this_turn()` reds it, which is the lazy version failing the case.

raven/context_engine/segments/curator.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified: `run_turn` opens the hold around context assembly, and `live_pin_resolver` caches each subsystem pin in that hold. Repeated curator steps therefore reuse one binding, while the next turn resolves the file again. The new provider-pool boundary test passes. This settles the finding.

raven/agent/loop/wiring.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified: `_turn_scope()` eagerly resolves `maxToolIterations` before `_run_turn` and context assembly, so the later `_run_agent_loop` read receives the held value even if the file changes during assembly. The new eager-boundary test passes. This settles the finding.

gloryfromca (commented):

The two prior blockers are fixed: subsystem pins remain stable across every call in one turn, and `maxToolIterations` is eagerly held before context assembly. Both threads have been replied to and resolved.

Verification: `uv run --extra dev pytest tests/test_config_live.py tests/test_provider_pool.py tests/test_context_engine_factory.py tests/test_skill_forge_gate.py tests/test_agent_loop_model_switch.py tests/test_agent_loop_session_model.py tests/test_agent_loop_usage_sink.py tests/test_rpc_settings.py -x` passed all 281 tests; `git diff --check github/refactor/ui_web_architecture...HEAD` also passed.

This round covered the rebased full diff and fix commit, AGENTS.md and Runtime terminology, affected callers and history, backward compatibility, test-strength changes, and the Model binding / Live preference / Generation architecture constraints. No new findings survived refutation.

0xKT (approved):

PR #544, opened 2026-09-20
fix(*): tell a handed-over memory write from a lost one at shutdown

LivXue:
## Summary

At teardown the store pipeline counted every unsettled record as lost, including one
a worker had already handed to the memory service. Cancelling that worker does not
cancel its request, so the service finished the write anyway, and the host told the
user the turn was gone while it was being indexed.

Measured on three one-shot CLI runs against a live EverOS: the client gave up at its
budget and `atomic_facts_extracted` landed for the same session 32 to 48 seconds
later, every time. The message was wrong twice over -- the turn was written, and the
service was not unavailable.

Review found that the first version of this fix replaced that error with its mirror.
The mark is set on entering `MemoryBackend.store()`, which is not a handoff: a
backend that persists only after an await, cancelled mid-await, leaves
`DrainOutcome(lost=0, in_flight=1)` with the backend's persistence list empty, and
the notice then said the turn had reached long-term memory. A slow connection
cancelled before its request body goes out has the same shape. `MemoryBackend`
declares no delivery guarantee, so no evidence from the transport is available
without extending the protocol and asking every backend to implement it. This branch
takes the other route and preserves the uncertainty instead.

What changed:

- `StorePipeline.drain` returns `DrainOutcome(lost, in_flight)` instead of one count.
  A record is marked in flight around the backend call, and the mark is read before
  the collect await that unwinds the cancellation, because that unwinding clears
  exactly what is being counted.
- `AgentLoop.drain_backend_stores` returns the outcome and logs the two apart: a
  warning for a turn this side watched fail, and for an unsettled one a line saying
  the drain stopped waiting and cannot see whether the service finished it.
- `report_dropped_memory_writes` becomes `report_memory_write_outcome`, since it no
  longer reports only what was dropped. An unsettled turn reads "were still being
  written when shutdown stopped waiting; whether the memory service finished them is
  not known here" -- the outcome, not a verdict on it.
- `DrainOutcome` is published on the memory engine's face beside `StorePipeline`, so
  callers reach it the way `test_memory_engine_face` requires. Its docstring now
  carries both cases, with the EverOS measurement scoped to the one it measured.
- A test drives the `raven agent` teardown that renders the outcome. It had no
  coverage at all: the shared harness stubs `maybe_build_memory_backend` to `None`,
  so the branch guarding those two lines never ran.

`in_flight` keeps its name deliberately. This repository already uses the word for
started and not settled -- `swap_in_flight`, `outbound.in_flight`, a DAG node in
flight -- with no delivery sense, so renaming it would detach the field from a
vocabulary that is already right. The delivery claim lived in the prose, and that is
what changed.

Deliberately unchanged: the 2.0s drain budget. It is far shorter than the write it
waits on -- one store measured 16.98s end to end, of which the flush leg was 12.1s --
and raising it to cover that would park every CLI exit for as long while buying
nothing, since the service completes a request it received either way and nothing
needs to recall the turn that just ended. That reasoning sits beside the constant,
because the gap between the two numbers reads like a defect and the next reader would
otherwise close it.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run on this branch's head, Python 3.12.13, all extras synced, rebased onto the
current default branch.

- `pytest tests/test_agent_loop_backend_dispatch.py tests/test_providers_factory.py
  tests/test_cli_agent_commands.py tests/test_memory_engine_face.py
  tests/test_cli_gateway_commands.py` -- 171 passed, 0 failed.
- Full suite: CI on this head runs the suite in four shards: 23742
  passed, 117 skipped, 0 failed. A local full run reports one failure,
  `tests/test_install_script.py::test_resolve_node_dir_answers_each_case_it_exists_for`,
  which was run again in a clean tree extracted from the default branch at the
  commit this branch is based on and fails there with the same assertion and the
  same value; the CI runner, which is not root, passes it. 0 introduced either way. The one failure is
  `tests/test_install_script.py::test_resolve_node_dir_answers_each_case_it_exists_for`.
  It was run again in a clean tree extracted from the default branch at the commit
  this branch is based on, and fails there with the same assertion and the same
  value. 0 introduced.
- Diff coverage against the default branch: `Diff coverage: 93.94% (31/33 executable changed lines)`,
  measured by the `coverage gates` job on this head. The ratchet moved with it:
  line +1.81pp, branch +2.74pp. The remaining uncovered pair is
  the same teardown inside `gateway_commands.run`, a 634-line function whose harness
  would cost far more than the one in `agent_commands`.
- `make lint-python`, `make lint-imports`, `make lint-types` -- all exit 0.
- `make check-commits`, `make check-large-files`, `make check-source-language` --
  all exit 0.
- The rebase was checked with `git range-diff`: every pre-existing commit replays
  `=`, and none of the commits the base gained touches a file this branch touches, so
  the changed-line set the coverage gate measures is unchanged.

Neither round's tests are theatre, and each was checked on its own:

- First round: the tests were watched failing on the missing attribute and the
  missing function. Two mutants were then run against them -- moving the in-flight
  read below the collect await reddens two of them, which is the constraint the
  comment beside that read now names; moving it below `cancel()` alone leaves them
  green, correctly, because `cancel()` only schedules.
- Second round: the notice test was watched failing on the delivery claim before the
  wording changed, and the end-to-end test was watched failing on the log line.
  Removing the `report_memory_write_outcome` call from the `agent` teardown turns the
  new CLI test red on its stdout assertion; restoring it turns it green with
  `git diff HEAD` empty.

One thing that would have gone unnoticed: `capsys` captures nothing from loguru here,
because the sink holds the real `sys.stderr` from when the handler was added, so the
first version of the end-to-end test passed while reading an empty string. It adds
its own sink now and asserts the sink saw something, so the log half cannot pass
unread.

No docs change: the only user-facing text is the shutdown notice itself, and no
document quotes it.

## Risk

User-visible change is one line of CLI output at exit, and it now states an unknown
outcome rather than a good one. A user who reads it has nothing to redo -- re-sending
the turn could duplicate a write the service did complete -- so the line is
informational by design, not an alarm.

The exit path is otherwise untouched: the budget constant is byte-identical to the
one on the default branch, and the diff of that file contains no non-comment line, so
exit timing does not move.

`drain_backend_stores` changes its return type, which is internal. Five callers:
three render the notice and are updated here, two await it and ignore the value as
they already did.

Rollback is a revert of the five commits on this branch. Nothing persists and no
format changes. Reverting only the last two would restore the delivery claim while
keeping the split, which is the state review rejected, so revert all or none.

- An existing assertion changed meaning and is called out rather than left in the
  diff. The test pinning the wording "still finishing in the background", whose
  docstring stated the write had reached the service, is renamed and rewritten; its
  two negative assertions were right and are kept, and a new sibling pins the other
  direction.
- The memory plugin still prints its own `final-flush sweep hit its 5.0s budget` line
  to stderr at shutdown. It belongs to a different layer and is factually correct;
  folding it into this notice would widen the diff into the plugin.
- The same teardown in `gateway_commands` stays uncovered. Its enclosing function is
  634 lines against 60 in `agent_commands`, and the gate passes without it.
- No domain term is coined. `StorePipeline` appears in `CONTEXT.md` only as a passing
  mention with no entry of its own, and `DrainOutcome` is a module-local type rather
  than a concept the map tracks.

## Related Issues

N/A


raven/memory_engine/store_pipeline.py
```diff
@@ -190,7 +209,11 @@ async def _worker(self, session_key: str) -> None:
                         # other one out of the global budget.
                         async with self._slots:
                             with trace.use_context(record.trace_ctx):
-                                landed = await self._store_once(session_key, record.messages, attempt=attempt)
+                                self._in_flight.add(session_key)
```
gloryfromca, line 222: Blocking: entering `MemoryBackend.store()` does not prove that a service received the write.

This marker is set before the generic backend call has established any handoff. For a compliant plugin that persists only after an await (including the new `_Wedged` test backend), `drain()` cancels the coroutine before persistence, but `abandoned - in_flight` removes that record from `lost` and the CLI tells the user it "reached long-term memory." I reproduced exactly that result: `DrainOutcome(lost=0, in_flight=1)` while the backend's persistence list remained empty. A slow HTTP connection cancelled before its request body is transmitted has the same gap. The live EverOS measurement establishes what happened for those requests, but the `MemoryBackend` contract supplies no corresponding delivery guarantee. The distinction therefore needs evidence from the backend/transport (or must preserve the uncertainty); merely entering `store()` cannot support the new success-facing notice.

gloryfromca (commented):
Blocking: a backend call can be cancelled before persistence but is reported as having reached long-term memory.

I found one concrete blocker, marked inline. I covered the full diff; all `drain_backend_stores` callers; the generic `MemoryBackend` contract and bundled EverOS path; cancellation, retry, repeated-drain, and shutdown accounting; branch history; backward compatibility; the changed assertions for weakening; and the repository's AGENTS/CLAUDE and runtime-context architecture rules. I found no separate issue in those areas.

Verification: the first focused run without optional extras stopped during collection because `raven_everos` was not installed. Re-running through the declared project setup, `uv run --frozen --all-extras pytest tests/test_agent_loop_backend_dispatch.py tests/test_providers_factory.py tests/test_generation_swap.py tests/test_agent_loop_everos_pipeline.py tests/test_rpc_bootstrap.py tests/test_core_runtime_swap.py tests/test_cli_tui_commands.py tests/test_everos_backend.py tests/test_agent_loop_run_emit.py tests/test_memory_engine_face.py -x`, passed all 402 tests. A minimal compliant cancellable backend reproduced the blocker as `DrainOutcome(lost=0, in_flight=1)` with no persisted record.

raven/memory_engine/store_pipeline.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed on the new head. The pipeline now describes this bucket as unsettled rather than delivered, and both user-visible reporting paths say its fate is unknown. The regression test reproduces the cancellable backend that persists nothing and asserts neither the log nor the CLI notice says the write reached memory. My original objection no longer holds.

gloryfromca (commented):

The new revision fixes my prior blocker by reporting a store call that outlives the drain as unsettled, without claiming either loss or delivery. The cancellable-backend reproduction now protects that boundary, and the added one-shot CLI test reaches the host rendering path that was previously unexercised. I found no new findings.

I reviewed the rebased full diff and its new commits; all drain callers; store cancellation and accounting; the generic backend contract and EverOS behavior; branch history and backward compatibility; the changed assertions for weakening; and the AGENTS/CLAUDE, runtime-context, and memory-engine-face constraints. The focused suite passed: `uv run --frozen --all-extras pytest tests/test_agent_loop_backend_dispatch.py tests/test_providers_factory.py tests/test_cli_agent_commands.py tests/test_generation_swap.py tests/test_agent_loop_everos_pipeline.py tests/test_rpc_bootstrap.py tests/test_core_runtime_swap.py tests/test_cli_tui_commands.py tests/test_everos_backend.py tests/test_agent_loop_run_emit.py tests/test_memory_engine_face.py -x` (438 passed). My prior thread is replied to and resolved.

raven/memory_engine/store_pipeline.py
LivXue: Reproduced before changing anything, and took the second of the two routes you named.

A backend that persists only after an await, cancelled mid-await by the drain:

    DrainOutcome(lost=0, in_flight=1)
    backend.persisted = []

So the mark is not evidence of a handoff, and the notice built on it was a
success-facing claim this side cannot make. Your `_Wedged` reading is right for the
same reason: it sleeps and returns, and never persists, so the branch's own
`(in_flight, lost) == (1, 0)` assertions were pinning the gap rather than the
behaviour.

I did not add transport evidence. `MemoryBackend` declares no delivery guarantee, so
that route means extending the protocol and asking every backend to implement it,
which is out of proportion to this fix. What changed instead is the claim, in all
three places that made it -- the shutdown notice, the teardown log, and four
docstrings that said the service had already been handed the record:

    reached long-term memory; indexing is still finishing in the background
    -> were still being written when shutdown stopped waiting; whether the
       memory service finished them is not known here

The set stays as it is. "Was inside the backend call" is computable and worth
reporting apart: an outcome nobody observed is not a loss, and it is not a win. The
EverOS measurement that motivated the split is kept and scoped to the case it
measured, next to the opposite case, in `DrainOutcome`'s own docstring.

`in_flight` keeps its name deliberately. This repo already uses the word for started
and not settled -- `swap_in_flight`, `outbound.in_flight`, a DAG node in flight --
with no delivery sense, so renaming it would detach the field from a vocabulary that
is already right. The delivery claim lived in the prose.

Two things worth flagging rather than leaving in the diff:

1. An existing assertion changed meaning. `test_a_handed_over_turn_is_not_announced_
   as_unwritten` pinned the wording "still finishing in the background" and its
   docstring stated the write had reached the service. It is renamed and rewritten;
   its two negative assertions were right and are kept, and a new sibling pins the
   other direction, so the notice is now held from both sides.
2. `capsys` captures nothing from loguru here -- the sink holds the real `sys.stderr`
   from when the handler was added, so a first version of the end-to-end test passed
   while reading an empty string. It now adds its own sink and asserts the sink saw
   something, so the log half cannot pass unread.

Also in this push: a test for the `raven agent` teardown that renders the outcome.
It had no coverage because the shared harness stubs `maybe_build_memory_backend` to
`None`, so the backend branch never ran; removing the render call turns the new test
red, restoring it turns it green. That closes the `coverage gates` red as a side
effect -- diff coverage is now 93.94% (31/33).

0xKT:
**Not a blocker.** The split is the right call and it is pinned: reverting
`max(0, abandoned - in_flight)` to the old `self._dropped += abandoned` turns 5
tests red, moving the `in_flight` snapshot after the collect await turns 4 red,
and deleting either rendered sentence in `cli/_helpers.py` turns 2 red. Three
observations, all latent or inherited. Read on head 9e11dd4b against base
0cb8b799.

## 1. The subtraction reads its two operands on opposite sides of an await

In `store_pipeline.py`, `in_flight` is measured at :300, `task.cancel()` runs at
:302, `await asyncio.wait(leftover, ...)` at :310, and `abandoned` at :319. The
snapshot and the thing it is subtracted from are therefore taken before and
after a real yield point. Any worker that reaches `self._active.pop(...)` at
:258 inside that window shrinks `abandoned` while `in_flight` still counts it,
and the surplus comes off some other session's genuine loss.

Reaching it needs a worker that survives the cancel, and the design deliberately
prevents that for the backend that ships: :258 sits in the `while` body rather
than a `finally` -- the comment above it says so, "A cancelled worker leaves its
entry behind on purpose, so the drain can count it" -- and `raven_everos`'s
`store` has no `except CancelledError`, no `except BaseException` and no
`asyncio.shield` anywhere (the only hit for those three in the whole package is
a type annotation at `backend.py:672`). So nothing in tree can produce it today.
A third-party `MemoryBackend` can: `raven/contracts/memory.py` says nothing
about cancellation, so a backend that swallows it is contract-conforming.

One correction to anyone acting on this: **`max(0, ...)` is not the bug and
removing it makes things worse.** In the reproduction `abandoned` is 1 and
`in_flight` is 4, so without the clamp the line subtracts 3 from a running
`_dropped` total and erases losses already counted earlier in the session. The
clamp is protective; the skew between :300 and :319 is the thing.

## 2. Of the three render sites this PR rewired, one is driven by nothing

`tests/test_cli_gateway_commands.py` contains zero mentions of
`drain_backend_stores`, `report_memory_write_outcome`, or the new wording --
deleting `report_memory_write_outcome(outcome, console)` at
`gateway_commands.py:1161` leaves the suite green. The control says the
selection really does reach CLI render sites: the same deletion at
`agent_commands.py:371` turns
`test_a_one_shot_shutdown_tells_the_user_what_the_drain_left` red, and
`test_cli_agent_commands.py:177` is where that path is stubbed.

Base has the same shape, so this is inherited rather than introduced, and it is
not why the PR should change. It is worth one line because the PR is rewiring
exactly that call and adding the test that covers its sibling.

## 3. One phenomenon, two different measurements

`raven/agent/loop/_shared.py:93` says "one extraction landed **32s** after the
drain gave up". `raven/memory_engine/store_pipeline.py:68` says an extraction
"was measured landing **34-48s** after the drain gave up". Both are written in
the grammar of measurement and both are the evidence for the same claim -- that
the service finishes whether or not this side is listening. Nothing in the tree
says whether that is one run described twice or two separate runs, so a later
reader cannot tell which number is the observation. These comments are precisely
the record that stops the next person re-litigating the drain budget, so it is
worth making the two agree.

gloryfromca:

The verifier note changes my stance from clean to nonblocking, not the merge decision. I reproduced observation 1 with a contract-conforming backend that suppresses cancellation: after four calls were snapshotted in flight, three workers settled during the collect await, and the drain returned `DrainOutcome(lost=0, in_flight=4)` while one genuinely queued record remained. The skew is real, but reaching it requires a third-party backend that suppresses cancellation, saturated store concurrency, and a queued record, so it belongs in follow-up work under the stated reachability bar.

Observation 2 is an inherited gateway-test coverage gap, and observation 3 is a prose measurement inconsistency; neither is a plain error that should hold this change. The focused shutdown and CLI suite remains green: `uv run --frozen --all-extras pytest tests/test_agent_loop_backend_dispatch.py tests/test_providers_factory.py tests/test_cli_agent_commands.py tests/test_cli_gateway_commands.py -x` (169 passed).

0xKT (approved):

PR #546, opened 2026-09-20
docs: refresh the readme overview workflow figures

LivXue:
## Summary

Refresh the overview figure at the top of the What is Raven section in both
READMEs with re-rendered versions. The diagram is unchanged in subject: one
prompt entering Raven, fanning out across the built-in agents and two external
agents, and converging on a single delivered result.

The image files are not committed. AGENTS.md section 7 rejects an added image by
extension at any size, so each figure is uploaded as a GitHub attachment and
referenced from the README by absolute URL. The new pair is hosted on the same
thread that already hosts the pair it replaces, so this figure keeps one
provenance trail. The superseded assets stay reachable, so any fork, mirror or
cache of the old text still resolves them.

This slot is language-specific: README.md and README.zh-CN.md carry different
asset ids here, unlike the benchmark charts, which share one id across both
files. Each file therefore takes its own image.

Checked and deliberately left unchanged: the alt text and the caption under each
figure. Both still describe the new render, and the English caption's "generates
DAGs and orchestrates multiple specialized agents" is if anything a closer fit to
the branching the diagram shows.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Byte equality of each uploaded asset against its local render, fetched back over
HTTP:

    curl -sL <asset-url> | sha256sum

Both matched their source exactly. A HEAD request is not sufficient here: the
attachment host answers HEAD with 403, which reads as a failed upload.

Repository gates, run on the committed range:

    make check-commits check-large-files check-source-language

All passed. The two file gates were re-run with a three-dot range and agreed. The
base had not moved under the branch: git rev-list --left-right --count reported
zero commits on the left.

Stale-reference sweep for every superseded asset id, over all tracked files:

    git grep -n -e <old-id>

No hits. The two READMEs are the only carriers of these assets; no page under
docs-site or docs references them, and no test pins them.

No test covers the README figures, so no suite was run for this change.
tests/test_living_docs.py globs raven/**/README.md and ui-tui/README.md, and does
not reach the root READMEs.

## Risk

The user-visible change is the rendered figure at the top of both READMEs.
Nothing executable changes. The assets are public attachments on a repository
that is already public and carry no credential or private detail, so no surface
is widened. Rollback is a revert of this commit; the previously referenced assets
remain hosted and resolve as before.

## Related Issues

N/A


gloryfromca (commented):
Blocking: correct the malformed Claude model id in the commit's Co-authored-by trailer before merge

The raw commit object records `Claude (claude-opus-5[1m])`; the literal `[1m]` is present in the bytes, so this is not terminal styling. AGENTS.md section 3.3 requires the actual current-session model id, and GitHub will carry the commit trailer into the squash commit, leaving the permanent co-author attribution inaccurate. Please rewrite the trailer with the actual model id.

Apart from that metadata defect, I found no issue in the README change. I reviewed the complete diff and surrounding README/history, checked AGENTS.md and CONTEXT-MAP.md terminology, fetched and visually inspected both old and new assets, verified all four URLs return valid PNGs, considered backward compatibility and the external-attachment architecture, and confirmed no tests were changed or weakened. The exact underlying `check-large-files`, `check-source-language`, and commit-message commands passed; `make` itself was unavailable, so I did not claim the wrapper command ran.

ZuyiZhou (approved):
Approving. Small diff, and every claim in it is checkable.

- Both superseded asset ids have zero hits across tracked files.
- The new ids are correctly language-specific: `f400d693` appears once in `README.md` and not at all in the Chinese file, `ed33ed0f` the reverse. That is the right handling for this slot, which unlike the benchmark charts does not share one id.
- All four assets return HTTP 200 `image/png`, the two superseded ones included, so the rollback claim holds without a re-upload.
- The re-renders are a genuine resolution increase: 4285x577 and 4303x577, against 2172x369 and 2172x368 before.

I also looked at both figures rather than only their URLs, because the caption is what a reader trusts. They agree with each other and with the text: one prompt into Raven, Raven-Research first, then a branch to Raven-Design and to Claude Code, Raven-Oncall after the latter, both arms converging on Codex and then a single delivered package. Two external agents, as described, and the two language renders are structurally identical with translated labels. The existing caption about generating DAGs does fit the branch-and-converge shape, so leaving it alone was right.

One observation, not a request. The aspect ratio moved from roughly 5.9:1 to 7.4:1, so at `width="100%"` the figure renders about 20% shorter than the one it replaces and the node labels sit smaller in the README column. The extra pixels mean it holds up on a high-dpi screen and the figure is wrapped in a link to the full asset, so this is a composition call rather than a defect -- worth knowing you made it.

PR #548, opened 2026-09-20
fix(*): apply the settings that were saved and then ignored

gloryfromca:
## Summary

#543 removed the reload-only set, which named eight keys. That set was not the
whole list. Several more settings are copied onto the objects the loop is built
with and never read again -- and those never said anything at all: the surface
answered `Saved.` and the process kept the value it started on. This is the rest
of the list, found by walking what `build_runtime` consumes and asking of each
writable key whether anything reads it again.

**`agents.defaults.reasoningEffort` is the one with a control on the settings
page** (`providers/Roles.tsx` writes it). It rides every model call of a turn,
and the provider's copy is frozen when the provider is constructed -- on
purpose: the comment on `ResolvingProvider` explains that a credentials refresh
must not import a live `agents` section, or a dormant routing edit takes effect
at a key-rotation boundary. So the loop sends the effort explicitly instead,
underneath the session's pin and a hook's override, which is the path a
session's own effort already travels. The provider keeps its default for
everyone who sends nothing.

**`tools.deepResearch.apiKey` was half done already.** `_maybe_promote_deep_research`
runs once per turn and swaps the offer stand-in for the working tool when a key
first appears. Replacing one was not covered: the working tool holds the section
it was built with, so a rotation left every call on the old credential until a
restart. It now rebuilds when the section changed and returns when it did not --
the skip matters, because a rebuild per turn would throw away the manager a
channel's async delivery depends on.

**The rest are read where they are used**: `agents.defaults.enablePersonalization`,
`memory.memoryTopK`, `tools.exec.timeout`, and the two `tools.web` vendors. The
vendor pair is worth naming: the keys beside them have been read live since they
landed, so the vendor was the half that still owed a restart.

All of them answer from the turn's hold (`hold_for_this_turn`, added in #543), so
one turn reads each once and every repeat costs a dict lookup instead of a file
read -- measured on a 2 KB config: 14.8 us against 0.05 us. Outside a turn every
read is live, which is what a caller below `run_turn` and the background paths
get.

**Not in this change, and why.** `agents.defaults.temperature` rides the same
frozen `GenerationSettings`, but making it live means sending an explicit
temperature on every call, and some reasoning models refuse one -- a real risk
for a key no surface renders. The three llm timeouts in that object are not
writable from any surface at all, so a restart is their honest semantics rather
than a lie. The same is true of the things that would need a stack rebuild
(memory backend, sandbox backend, workspace root): none is reachable from a
settings write, so there is no "saved and nothing happened" to fix.

**One property a reviewer should weigh.** The hold is a ContextVar, so work
detached during a turn copies it: a sub-agent spawned inside a turn keeps that
turn's values for its whole life, past the parent turn's end. Verified by
experiment. That is the semantics the model binding already has and for the
stated reason -- a sub-agent spawned before a switch must not finish on the
model chosen after it -- so a long-running sub-agent takes an edited exec
timeout or vendor at its next spawn, not mid-run.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/ -q`: 23375 passed, 4 failed, 111 skipped. All four
  failures reproduce on a clean checkout of the base commit in the same
  environment (`test_agent_loop_token_budget`, the two `test_agents_*_launcher`
  tool-face cases, `test_ppt_engine_image_search`); the launcher pair looks
  environment-dependent (a `SERPER_API_KEY` in the developer environment makes
  `web_search` available).
- After the cherry-pick onto the current base: the sweep over
  config-live/auth/deep-research/subagent/sandbox/shell/context-engine/memory/
  usage-sink/loop/settings/pool/console, 4324 passed, 1 failed (the pre-existing
  token-budget case).
- `npm run type-check` clean; `npx vitest run` 186 files, 2544 tests passed.
- `uv run ruff check` and `ruff format --check` over the full lint target set:
  clean.
- Ten new cases, each mutation-verified on its own: reading one spelling of the
  effort, resolving the effort lazily instead of at the boundary, ignoring the
  file for personalization, keeping the built recall depth, keeping the built
  vendors, sending a call without the configured effort, ignoring a rotated
  deep-research section, rebuilding on an unchanged one, keeping the built exec
  ceiling, and dropping the hold so a turn straddles an edit.
- The cost figures above are measured, not estimated: a micro-benchmark over a
  2 KB config, 20k reads held and 5k live.

## Risk

User-visible: a reasoning-effort change on the settings page reaches the next
turn; a rotated deep-research section reaches the next turn; the four
RPC-writable settings reach their next use. A turn already running is never
moved, and neither is a sub-agent already detached.

`ExecTool.timeout` becomes a property over the value the tool was built with.
Nothing assigns to it; the constructor argument is unchanged and still answers
when the file has no opinion.

The deep-research rebuild is the one place that does work rather than answer a
question: it unregisters and re-registers a tool. Guarded on the section having
actually changed, so an unchanged config costs one comparison per turn.

Rollback: revert the commit. Every reader falls back to the value its object was
constructed with, which is what it used exclusively before.

## Related Issues

N/A


raven/config/live.py
```diff
@@ -395,6 +417,72 @@ def max_tool_iterations(live: LiveConfig) -> int | None:
...
+
+    def one(*keys: str) -> str | None:
+        for key in keys:
+            value = live.get(key)
+            if isinstance(value, str) and value:
+                return value
+        return None
+
+    return (
+        one("tools.web.search.provider"),
```
gloryfromca, line 507: Blocking: the running main web tools and subagent key source must follow the saved vendor.

`web_providers()` is only consumed by `SubagentManager`. The main loop has already registered `WebSearchTool` and `WebFetchTool` with its startup `self.web_*_provider`, so changing `tools.web.search.provider` or `tools.web.fetch.provider` does not move those tools. I reproduced the normal search case with both keys present: after rewriting the live config from Serper to Exa, `loop.tools.get("web_search").provider` was still `serper` and its key was still `serper-key`. The new subagent path reads the vendor live, but still hands `RavenLoopBackend` the manager's startup `web_provider_keys`, so a key added or rotated in the same settings flow is missed there too. This leaves an ordinary settings action in the saved-and-ignored state this change is meant to remove.

raven/agent/loop/main.py
```diff
@@ -847,9 +852,23 @@ def _maybe_promote_deep_research(self) -> None:
             return
         if not DeepResearchTool.is_configured(cfg):
```
gloryfromca, line 852: Blocking: clearing the deep-research key must replace or disable the working tool.

The settings Clear action writes an empty key, but this return happens before the existing real tool or `deep_research_config` is changed. I reproduced the transition by promoting with `sk-old`, making `get_deep_research()` return an empty key, and calling this method again: the registry still held `DeepResearchTool` and `deep_research_config.api_key` was still `sk-old`. The supposedly removed credential therefore remains callable for paid requests until restart. The real-to-offer transition needs to be handled as well as offer-to-real and real-to-real rotation.

gloryfromca (commented):
Blocking: live web-provider changes and deep-research credential removal must take effect in the running loop.

I found two reachable settings writes that still leave the process on the old value; both are marked inline.

Covered: the full target-to-head diff, AGENTS.md plus the Runtime/Web UI context contracts, changed callers and consumers, the four-commit history, backward compatibility, and whether tests were weakened. The new tests add coverage but omit the two opposite-direction transitions reproduced inline.

Verification:
- Targeted runtime/RPC sweep: 513 passed.
- `uv run --all-extras pytest tests/test_rpc_settings.py -q`: 87 passed.
- Web UI: 186 files / 2,544 tests passed; TypeScript type-check passed.
- Source-language gate and `git diff --check`: passed.

raven/config/live.py
gloryfromca: Confirmed and fixed in 826291dfc.

You are right about the asymmetry, and about where it bites: the key readers resolve against the selection (`_live_web_search_key` reads `self.web_search_provider`), so a frozen vendor kept the pair on the old endpoint and the new vendor's key reached nothing.

What changed:

- `WebSearchTool`, `ImageSearchTool` and `WebFetchTool` take the vendor the way they already take the key -- a callable is the live form, a string stays a snapshot. An unknown name in the file falls back to the vendor the tool was registered with rather than raising: the file is read while a turn runs and a typo in it must not take the tool down mid-call.
- The loop's own `web_search_provider` / `web_fetch_provider` became properties over the file, held for the turn, so the key readers follow without further change.
- `WebFetchTool.effective_provider` (the Jina substitution) moved from registration to per call. Decided once it outlived the key that would have stopped it; it warns on the first substitution only, since it is now asked every call.
- The image vendor is derived from the selection, so both its halves follow rather than pinning to the boot choice.
- A spawn hands down `web_provider_keys` as the file has them, merged over the manager's boot set -- your second point. The vendor was read live and the keys were not, which is half a pair.

New cases, each mutation-verified: `test_the_main_web_tools_follow_the_saved_vendor` (vendor and key move together), `test_an_unknown_vendor_keeps_the_tool_working`, `test_a_spawn_hands_down_the_keys_the_file_has`.

One thing your comment sent me back to: the plural key reader I added first spelled `apiKey` itself and tripped `test_only_the_auth_module_decides_configuredness_from_a_key`. It goes through the singular schema-backed reader now, which is how that spelling stays in one place.

raven/agent/loop/main.py
gloryfromca: Confirmed and fixed in 826291dfc.

The unconfigured check returned before anything changed, so Clear left the working tool registered on the credential that had just been removed. Of the three transitions this method can see, only two were handled; real-to-offer is now the third: unregister the working tool, drop the manager, put the offer stand-in back, and record the cleared section.

Reproducing your case the same way -- promote on `sk-old`, then have `get_deep_research()` answer an empty key -- now leaves `DeepResearchOfferTool` registered with `deep_research_config.api_key == ""` and `deep_research_manager is None`. Pinned by `test_clearing_the_key_takes_the_working_tool_away`, mutation-verified.

Your comment also turned up a bug in my first cut of it, which is worth recording here because it is the more dangerous half: clearing unregistered whatever was registered under the name. That took away a test double and would have taken away a plugin shadowing `deep_research` just as happily -- three cases in `test_agent_loop_run_emit.py` caught it. Every branch keys off identity now (`isinstance` of the two tools this method itself registers) rather than the name, and a method called once per turn returns immediately when the entry is somebody else's. `test_an_unconfigured_process_stays_on_the_offer_tool` pins the no-op.

raven/config/live.py
gloryfromca: Blocking: clearing a saved web-provider key must also revoke the startup credential for new subagents.

Verified: the main tools now follow the saved vendor, and newly added or rotated keys reach subagents, so the original finding is settled. I am closing this thread; the remaining clear-key regression is anchored on the new code in the current review.

raven/agent/loop/main.py
gloryfromca: Blocking: the replacement offer must inherit the stored question broker after the stale credential is removed.

Verified: Clear now removes the working tool, clears its config and manager, and restores the offer, so the original stale-credential finding is settled. I am closing this thread; the broker regression in that replacement path is anchored on the new code in the current review.

raven/agent/subagent/manager.py
```diff
@@ -351,6 +356,37 @@ def refresh_agents(self) -> None:
...
+        """The per-vendor keys a spawn should hand down, as the file has them.
+
+        The vendor a sub-agent runs on is read live; handing it the keys this
+        manager was built with leaves a key added in the same settings flow
+        behind, which is the half of the pair this exists to close.
+        """
+        from raven.config.live import default_live, held, web_provider_keys
+
+        configured = held("tools.web.providers.keys", lambda: web_provider_keys(default_live()))
+        return {**(self.web_provider_keys or {}), **configured}
```
gloryfromca, line 381: Blocking: clearing a saved web-provider key must revoke the startup credential for newly spawned subagents.

`web_provider_keys()` drops empty values, then this merge restores any omitted vendor from `self.web_provider_keys`. On a manager built with `{'exa': 'exa-old'}`, saving `tools.web.providers.exa.apiKey` as an empty string makes the live reader return `{}`, and this method still returns `{'exa': 'exa-old'}`; I reproduced that exact transition. The next subagent can therefore keep spending a credential the settings UI removed until restart. The live representation needs to preserve an explicit empty slot, or this merge must otherwise distinguish a missing section from a cleared vendor and delete the boot entry.

raven/agent/loop/main.py
```diff
@@ -817,11 +826,42 @@ def _maybe_promote_deep_research(self) -> None:
...
+            if not working:
+                return
+            # Cleared on the settings page. Returning here would leave the
+            # working tool registered on the credential that was just removed,
+            # and it would keep spending against it until a restart.
+            self.tools.unregister("deep_research")
+            self.deep_research_config = cfg
+            self.deep_research_manager = None
+            self.tools.register(DeepResearchOfferTool())
```
gloryfromca: Blocking: the real-to-offer deep-research swap must preserve the configured question broker.

The gateway/TUI calls `set_deep_research_broker()` once during host wiring. `_register_real_deep_research()` explicitly reapplies that stored broker to later working tools, but this branch registers a fresh `DeepResearchOfferTool` directly and leaves its `_broker` as `None`. On the next research-shaped turn, `_ask_search_mode()` therefore returns `None` and the offer immediately degrades to regular search instead of showing the deep-vs-regular and setup prompts; it stays that way until restart. Apply `self._deep_research_broker` to the replacement offer (preferably through a shared registration helper) and pin the clear transition with a broker assertion.

gloryfromca (commented):
Blocking: saved key clearing must revoke subagent credentials, and the deep-research offer swap must retain its broker.

The two original failures are fixed on this revision, and those threads are resolved. Two transition regressions remain inline.

Coverage: repository rules (`AGENTS.md`, `CONTEXT-MAP.md`, the runtime `CONTEXT.md`, and `ui-web/CONTEXT.md`); the full target diff and the new-commit delta; affected callers and host wiring; history and backward compatibility; test changes (additive, with no weakened tests found); and the runtime architecture constraints.

Verification: `uv run --all-extras pytest tests/test_config_live.py tests/test_deep_research_tool.py tests/test_agent_loop_usage_sink.py tests/test_agent_loop_web_tools.py tests/test_agent_tools_web_providers.py tests/test_rpc_settings.py -q` passed 320 tests with 17 existing pytest-marker warnings; `uv run --all-extras pytest tests/test_agent_loop_run_emit.py tests/test_rpc_bootstrap.py -q` passed 66 tests; `git diff --check github/refactor/ui_web_architecture...HEAD` passed. A focused clear-key reproduction returned `{'exa': 'exa-old'}` from `_web_provider_keys_now()` after the saved Exa key was emptied.

raven/agent/subagent/manager.py
gloryfromca: Confirmed and fixed.

You are right, and the bug was mine one layer up: the singular reader already draws the distinction -- a present-but-empty slot is a revocation, not a miss -- and the plural one I built on top of it threw that away with `if value:`. The merge then did exactly what you describe, restoring the credential the settings surface had just removed.

`web_provider_keys()` keeps an explicit empty now: a cleared vendor is present with `""`, so the merge overwrites the boot entry, and only a vendor the file says nothing about is left to the boot value. That is the same contract the singular reader states, which is where it should have come from in the first place.

`test_a_cleared_vendor_key_revokes_the_one_a_spawn_booted_with` pins both halves of it -- the cleared vendor resolves to `""`, a vendor absent from the file keeps its boot credential. Mutation-verified: restoring the `if value:` filter reds it.

raven/agent/loop/main.py
gloryfromca: Confirmed and fixed.

`set_deep_research_broker` applies the broker to whatever is registered at the moment the host calls it, and a stand-in registered later is past that call -- so the replacement had `_broker is None` and `_ask_search_mode` degraded the offer to regular search silently, for the rest of the process.

Taken the way you suggested, as a shared helper: `_register_deep_research_offer()` builds the stand-in, hands it the stored broker when there is one, and registers it. Both registration sites go through it now -- the one at construction (where the broker is not wired yet, so it is a no-op there) and the clearing branch -- so the two cannot drift apart again.

`test_clearing_the_key_takes_the_working_tool_away` wires a broker before the promotion and asserts the replacement carries it, alongside the tool type, the cleared section and the dropped manager. Mutation-verified: registering a bare `DeepResearchOfferTool()` reds it.

raven/agent/subagent/manager.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified on `b152bec42f73`: a present empty Exa slot now survives `web_provider_keys()` and overwrites the boot value, while an absent Serper slot still falls back to its boot credential. The focused regression and surrounding settings/runtime suite pass. This finding is fixed and the thread is closed.

raven/agent/loop/main.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified on `b152bec42f73`: both construction-time and clear-time offer registration now use `_register_deep_research_offer()`, which reapplies the stored broker when present. The regression asserts the replacement keeps that broker, and the focused runtime suite passes. This finding is fixed and the thread is closed.

gloryfromca (commented):

The two prior blockers are fixed and their threads are resolved. The cleared-key path now preserves an explicit empty value so it revokes the boot credential, and the deep-research offer replacement now inherits the host's stored broker.

Coverage: the full target diff and the `826291dfccb4..b152bec42f73` delta; repository rules (`AGENTS.md`, `CONTEXT-MAP.md`, runtime `CONTEXT.md`, and `ui-web/CONTEXT.md`); affected callers and host wiring; history and backward compatibility; architecture constraints; and test changes, with no weakened tests found.

Verification: `uv run --all-extras pytest tests/test_config_live.py tests/test_deep_research_tool.py tests/test_agent_loop_usage_sink.py tests/test_agent_loop_web_tools.py tests/test_agent_tools_web_providers.py tests/test_rpc_settings.py tests/test_agent_loop_run_emit.py tests/test_rpc_bootstrap.py -q` passed 387 tests with 17 existing pytest-marker warnings. `git diff --check github/refactor/ui_web_architecture...HEAD` also passed.

0xKT (approved):

PR #552, opened 2026-09-20
fix(agent): exec refuses a typed ssh to a machine the registry knows

silverLXT:
Two field runs on 2026-09-14 put a registered machine's address in the task
statement, and the coding nodes typed "ssh -p <port> root@<ip> '... &'" from
the local shell 58 times to start GPU work. The look-at-it channel is capped
at 60 s and the on-call agent's job runner was not theirs to call, so the raw
address was the path of least resistance. The cap, the process-group sweep
that stops an orphan from holding a GPU, and the ledger all live on the other
two paths, so none of that work was metered.

The plain shell now refuses a command that names ssh and a host the registry
knows, and the refusal names both the machine and the two paths that are
meant for it. Everything else is untouched: an unknown host still works, scp
and rsync still work (they move files and start nothing on the far side), and
a registry that cannot be read recognises nothing, so ops being broken never
takes the plain shell down with it.

Type: fix

Verification:
- tests/test_shell_machine_channel.py: 32 passed, up from 28 on 0464aa1b1.
  The four new ones cover a typed ssh to a registered machine being refused
  with nothing run, an unknown host still running, a command that names the
  host without ssh still running, and a malformed registry refusing nothing.
- Wider run over the shell surface (machine channel, approval, background,
  policy reasons, rpc): 316 passed, 0 failed.

Risk: low, but it does close a path some prompt may rely on. The refusal is
text, not an exception, and it names the replacement, so a model that reaches
for the old path is told where to go. This does not add the missing channel
itself: work between 60 s and a few minutes still has no first-class home
outside the on-call agent, which is a separate design question.


gloryfromca (commented):
Blocking: the SSH detector must identify the executable and exact registered destination before refusing or naming a replacement.

I found three concrete failures in the new guard, detailed inline: ordinary path-qualified SSH invocations bypass it; same-host connections on different ports are resolved to the first row; and an unregistered address sharing a registered address's prefix is falsely refused.

Verification: `uv run pytest tests/test_shell_machine_channel.py -q` (32 passed), and `uv run pytest tests/test_shell_comments.py tests/test_shell_approval.py tests/test_shell_background.py tests/test_rpc_spine.py tests/test_sandbox_unit.py -q` (454 passed). I also reproduced all three cases directly against `raw_ssh_target`; the current tests do not cover them. GitHub's full check suite is green.

Coverage: I checked the repository rules and canonical Machine definition, the full diff, `ExecTool` callers and permission ordering, relevant registry/transport code and history, backward-compatibility claims, test changes for weakening, and the machine-identity architecture constraint.

raven/agent/tools/machine_exec.py
```diff
@@ -96,6 +96,39 @@ def machines_registered() -> bool:
         return False
 
 
+_SSH_WORD = re.compile(r"(?:^|[\s;&|(`])ssh(?=\s)")
```
gloryfromca: Blocking: recognize path-qualified SSH invocations before allowing the command to execute.

This pattern only accepts `ssh` at the start or after its small delimiter set. Both `/usr/bin/ssh root@203.0.113.7 ...` and `\ssh root@203.0.113.7 ...` execute the SSH client in an ordinary shell, but `raw_ssh_target` returns `None` for both. After any required permission approval, `ExecTool` therefore runs them on the plain path, bypassing the 60-second cap, process-group sweep, and ledger this change is meant to enforce.

raven/agent/tools/machine_exec.py
```diff
@@ -96,6 +96,39 @@ def machines_registered() -> bool:
...
+        return None
+    try:
+        from raven.ops.connections import load
+
+        rows = load()
+    except Exception:  # noqa: BLE001 -- a malformed registry must leave the plain shell working
+        return None
+    for row in rows:
+        host = str(row.get("host") or "").strip()
+        if host and host in command:
```
gloryfromca: Blocking: resolve the registered machine by SSH port as well as host before naming the replacement.

The registry explicitly supports distinct machines at one address on different ports. With `conn_a` at `203.0.113.7:58717` followed by `conn_b` at `203.0.113.7:64101`, `raw_ssh_target("ssh -p 64101 root@203.0.113.7 true")` returns `conn_a`. The refusal then directs the model to `exec(machine='conn_a')`, so following the proposed recovery can execute the command on the wrong machine.

gloryfromca: Blocking: match the SSH destination as a complete host instead of an arbitrary substring.

With registered host `203.0.113.7`, the command `ssh root@203.0.113.70 true` is classified as `conn_gpu` because `.7` is a prefix of `.70`. That contradicts the stated compatibility guarantee that SSH to an unregistered host still runs, and the error falsely claims the command reaches the registered machine.

gloryfromca: Blocking: compound-shell and alternate-port forms still need to be covered.

This specific shared-host port-selection defect is fixed at `8ab4d06c380a`; I reproduced `-p 64101` selecting the second row rather than the first. I am resolving this thread, and the remaining current-revision defects will be reported separately.

gloryfromca: Blocking: compound-shell and alternate-port forms still need to be covered.

This specific host-prefix defect is fixed at `8ab4d06c380a`; I reproduced `203.0.113.70` remaining unregistered when only `203.0.113.7` is present. I am resolving this thread, and the remaining current-revision defects will be reported separately.

raven/agent/tools/machine_exec.py
gloryfromca: Blocking: compound-shell and alternate-port forms still need to be covered.

This specific path-qualified invocation defect is fixed at `8ab4d06c380a`; I reproduced successful recognition for both `/usr/bin/ssh` and `\\ssh`. I am resolving this thread, and the remaining current-revision defects will be reported separately.

raven/agent/tools/machine_exec.py
```diff
@@ -96,6 +97,100 @@ def machines_registered() -> bool:
...
+    executable may be written ``ssh``, ``/usr/bin/ssh`` or ``\\ssh`` (a
+    backslash suppresses alias lookup and still runs the client), and all three
+    reach the far side.
+
+    The port is read from ``-p`` because the registry holds several machines at
+    one address on different ports; returning the address alone picks whichever
+    row is listed first and names the wrong machine.
+    """
+    try:
+        tokens = shlex.split(command, comments=False)
```
gloryfromca: Blocking: inspect every shell command in a compound line before allowing registered SSH to execute.

`shlex.split` does not split unspaced shell operators, and this helper returns after the first SSH-looking token. Consequently `true&&ssh -p 58717 root@203.0.113.7 true` produces no destination, while `ssh root@198.51.100.9 true; ssh -p 58717 root@203.0.113.7 true` returns only the unregistered first destination. In both cases the valid shell line proceeds through the plain executor and the registered SSH command bypasses the cap, sweep, and ledger.

raven/agent/tools/machine_exec.py
```diff
@@ -96,6 +97,100 @@ def machines_registered() -> bool:
...
+        while position < len(rest):
+            word = rest[position]
+            if word.startswith("-") and len(word) > 1:
+                if word.startswith("-p"):
+                    value = word[2:] or (rest[position + 1] if position + 1 < len(rest) else "")
+                    if value.isdigit():
+                        port = int(value)
+                    position += 1 if word[2:] else 2
+                    continue
+                if len(word) == 2 and word[1] in _SSH_VALUE_FLAGS:
```
gloryfromca: Blocking: honor an explicit SSH Port option when matching a registered destination.

OpenSSH accepts `ssh -o Port=58717 root@203.0.113.7 true` and connects to port 58717 (confirmed with `ssh -G`). Here `-o` is consumed as a value-taking flag, but its `Port=58717` value is discarded, so the helper retains port 22 and `raw_ssh_target` returns `None` for a machine registered at 58717. That direct raw SSH invocation therefore passes the new refusal.

gloryfromca (commented):
Blocking: compound-shell SSH and explicit -o Port destinations still need to be refused.

The three findings from the prior revision are fixed and their threads are resolved. This revision still has two concrete bypasses, detailed inline: valid compound shell lines can hide a registered SSH execution, and OpenSSH's explicit `-o Port=...` form is treated as port 22.

Verification: `uv run pytest tests/test_shell_machine_channel.py -q` passed 37 tests, and `uv run pytest tests/test_shell_comments.py tests/test_shell_approval.py tests/test_shell_background.py tests/test_rpc_spine.py tests/test_sandbox_unit.py -q` passed 454 tests. Direct reproductions confirmed both bypasses. GitHub's checks otherwise pass, but the coverage gate currently fails at 87.72% diff coverage (50/57 changed executable lines).

Coverage: I checked the repository rules and canonical Machine definition, the old-to-new delta and full diff, `ExecTool` callers and permission ordering, registry/transport behavior and relevant history, backward compatibility, test changes for weakening, and the machine-identity architecture constraints.

gloryfromca (commented):
Blocking: the open compound-command and explicit-port findings still need code changes.

This revision adds four coverage tests only; `raven/agent/tools/machine_exec.py` is unchanged from the prior revision. Re-running the exact reproductions still returns `None` for both reported bypasses, so those two threads remain open. I found no new defect in the test-only delta, and the added tests do not weaken existing assertions.

Verification: `uv run pytest tests/test_shell_machine_channel.py -q` passed 41 tests, and `uv run pytest tests/test_shell_comments.py tests/test_shell_approval.py tests/test_shell_background.py tests/test_rpc_spine.py tests/test_sandbox_unit.py -q` passed 454 tests. At review time all completed GitHub checks pass; unit shard 4/4 is still running and the coverage gate has not yet reported.

Coverage: I checked the repository rules and canonical Machine definition, the prior-to-current delta and full diff, callers and permission ordering, relevant history, backward compatibility, test-strength changes, and the machine-identity architecture constraints.

raven/agent/tools/machine_exec.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed at `46960856729e`: punctuation-aware tokenization now recognizes the unspaced operator, and every SSH destination in the line is checked. I reproduced both original compound-command cases resolving to `conn_gpu`, so I am closing this thread.

raven/agent/tools/machine_exec.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed at `46960856729e`: `-o Port=58717` now updates the parsed port before registry matching. I reproduced the original command resolving to `conn_gpu`, so I am closing this thread.

gloryfromca (commented):

Both open findings are fixed in this revision: compound commands now expose every SSH destination, including after unspaced operators, and `-o Port=...` participates in registry matching. I reproduced the original failing cases, replied to both threads, and resolved them. A sweep confirms every thread I opened is now resolved.

Verification: `uv run pytest tests/test_shell_machine_channel.py -q` passed 46 tests, and `uv run pytest tests/test_shell_comments.py tests/test_shell_approval.py tests/test_shell_background.py tests/test_rpc_spine.py tests/test_sandbox_unit.py -q` passed 454 tests. At review time every completed GitHub check passes; unit shard 4/4 is still running.

Coverage: I checked the repository rules and canonical Machine definition, the prior-to-current delta and full diff, callers and permission ordering, relevant history, backward compatibility, test-strength changes, and the machine-identity architecture constraints. I also applied the fourth-round convergence bar to parser edge cases and found none that justify another blocker.

raven/agent/tools/machine_exec.py
```diff
@@ -96,6 +97,128 @@ def machines_registered() -> bool:
...
+            if word.startswith("-") and len(word) > 1:
+                value = ""
+                if word.startswith("-p") or word.startswith("-o"):
+                    value = word[2:]
+                    if not value and index < len(tokens):
+                        value = tokens[index]
+                        index += 1
+                    if word.startswith("-o"):
+                        name, _, setting = value.partition("=")
+                        value = setting if name.strip().lower() == "port" else ""
```
LivXue: **Blocker - accept the space-separated spelling of an `-o` option value before matching the
registry.** The `-o` branch takes its value and then requires the keyword before `=` to be
exactly `port`:

    name, _, setting = value.partition("=")
    value = setting if name.strip().lower() == "port" else ""

OpenSSH takes that option either way, and `ssh -G -o "Port 58717" root@203.0.113.7` prints
`port 58717`. Written with a space, `name` is `"Port 58717"`, the value is discarded, the port
stays 22, and `raw_ssh_target` returns None for a machine that is registered at 58717 -- so the
command runs on the plain shell path, past the cap, the process-group sweep and the ledger,
which is the bypass this guard exists to close. Reproduced through `ExecTool` with the registry
holding `203.0.113.7:58717`: the `-o Port=58717` control answers "registered as 'conn_gpu'",
while `-o "Port 58717"` executes and its own stderr reports `connect to host 203.0.113.7 port
58717`. This is the option fixed at 8ab4d06c in its other documented spelling rather than a new
form, so the value wants splitting on whitespace as well as on `=`.

raven/agent/tools/machine_exec.py
```diff
@@ -96,6 +97,128 @@ def machines_registered() -> bool:
         return False
 
 
+# Values ssh takes as a separate word, so the destination is not confused with
+# one of them. Only the flags that can precede a destination are listed.
+# What punctuation_chars hands back as its own token. A command ends at one of
+# these, so the words after it belong to the next command, not to this ssh.
+_SHELL_OPERATORS = frozenset({";", "&", "&&", "|", "||", "(", ")", "<", ">", ">>", "<<"})
```
LivXue: **Nonblocking - complete the operator set, or a token before the destination can stand in for
it.** `_SHELL_OPERATORS` omits `>&`, `<&`, `&>` and `|&`, which `punctuation_chars` hands back
as single tokens. A token that is not in the set is read as a word, so one appearing before the
destination becomes the destination, and the `if destination is not None: continue` below then
skips the host that follows it. `_ssh_destinations("ssh 2>&1 -p 58717 root@203.0.113.7 true")`
returns `[('2', 22)]`, and the same command with `&>/dev/null` returns `[('&>', 22)]`; either
way the registered machine goes unrecognised and the line runs on the plain path. Graded
nonblocking rather than blocking because it needs the redirection written ahead of the host,
where the measured usage types the address directly -- but those four classes are exactly what
the `_DETACH` note at line 63 already names as redirections, so this set is the cheap place to
close it.

LivXue:

Scope: I reviewed the `raven/agent/**` portion of this PR -- 2 of 4 changed files, which is the
whole of the review zone. The other two are `tests/test_shell_machine_channel.py` and
`CHANGELOG.md`, read as context. Ring 3 (`raven/agent/subagent/`) and ring 4 (the dispatch
tools and the DAG skill) are untouched, so no change here carries a sign-off.

I reviewed the new detector and its single call site, re-ran every reproduction the earlier
rounds filed, and probed the two syntax families they did not cover.

Confirmed findings:

- Blocker, `raven/agent/tools/machine_exec.py:168`: the `-o` value parser accepts only the
  `Port=<n>` spelling, so `-o "Port <n>"` discards the value, leaves the port at 22, and a raw
  ssh to a machine registered at that port on that host runs on the plain shell path.
  Reproduced through `ExecTool`; the executed command's own stderr names port 58717.
- Nonblocking, `raven/agent/tools/machine_exec.py:104`: `_SHELL_OPERATORS` omits `>&`, `<&`,
  `&>` and `|&`; a redirection token written before the destination is taken as the
  destination, and the real host is then skipped. Reproduced against `_ssh_destinations`.

Verification:

- Every reproduction from the earlier rounds, re-run at this head: path-qualified and
  backslash-escaped ssh, a second machine at the same address on another port, the
  longer-address case, unspaced compound operators, two ssh commands in one line, and
  `-o Port=`. All seven behave as those threads say they were fixed to.
- `uv run pytest tests/test_shell_machine_channel.py -q`: 46 passed. With
  `tests/test_shell_comments.py tests/test_shell_approval.py tests/test_shell_background.py
  tests/test_rpc_spine.py tests/test_sandbox_unit.py`: 454 passed.
- `lint-imports`: 10 kept, 0 broken.
- Both findings here were reproduced by running the code, not by reading it.

Checked and deliberately not reported: the detector is spelling-based, so `bash -c 'ssh ...'`,
a backticked command, and a host built from a shell variable all still reach the machine -- I
reproduced all three. Not filed as a defect: they need deliberate indirection, the docstring
frames this guard as closing the path of least resistance rather than as a sandbox, and no
token-level fix reaches them. Also checked and clean -- a trailing `2>&1`, `|&` or here-string,
where the destination has already been parsed; the guard sitting ahead of `_guard_command`,
which by design leaves no approve-then-run path for this shape; and the complete absence of
deletions in the diff, so no earlier guard was dropped.

raven/agent/tools/machine_exec.py
gloryfromca: Blocking: the space-separated -o Port spelling must participate in registry matching.

Confirmed on unchanged head `46960856729e`: `_ssh_destinations("ssh -o 'Port 58717' root@203.0.113.7 true")` returns `[('203.0.113.7', 22)]`, and `raw_ssh_target` returns `None` for a registry row at port 58717. The targeted suite still passes 46 tests, so this spelling is uncovered rather than protected. This factual reproduction changes my standing stance from clean to blocker; I am leaving the thread open for its author to resolve.

silverLXT:
Revision c1f84ccb4 answers both threads from the last round.

Blocker (machine_exec.py, the -o branch): the option value is now split on an
equals sign or on whitespace, so `-o "Port 58717"`, `-o 'Port = 58717'` and
`-o Port=58717` all reach registry matching. Reproduction from the thread:
`_ssh_destinations("ssh -o 'Port 58717' root@203.0.113.7 true")` now returns
`[('203.0.113.7', 58717)]`, and `raw_ssh_target` names `conn_gpu` for the
58717 row.

Nonblocking (the operator set): adding `>&`, `&>` and friends to the
terminator set was not enough on its own. `2>&1` tokenises as three words --
`2`, `>&`, `1` -- so the bare descriptor was taken for the destination before
the operator was ever seen, and treating `&>` as a terminator ended the scan
before the real host. The set is split in two: command separators still end
the ssh; redirections (`< > >> << <<< <& >& &> &>> >| <>`) and the file they
name are stepped over, as is a bare digit standing in front of one. A
redirection left dangling before a separator does not swallow the separator.
The one shape the tokens cannot distinguish -- a destination that is itself a
bare integer written with a space before `>&` -- is documented at the check;
no registry row has that shape.

Tests: 12 new cases (two parametrised ExecTool tests for the spaced -o
spelling and for a redirection ahead of the host, plus a direct
`_ssh_destinations` table for after-host, dangling, dangling-then-separator,
heredoc and later-option-wins). All 12 fail on the previous head and pass
here. `tests/test_shell_machine_channel.py` 58 passed; the five wider suites
454 passed; ruff, ruff format and ty clean. Every new line in machine_exec.py
is covered; the three uncovered lines in the file (196->199, 288-289, 346)
predate this PR.

raven/agent/tools/machine_exec.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed at `c1f84ccb4fd0`: the space-separated and spaced-equals Port spellings now resolve to port 58717 and match `conn_gpu`. The targeted suite passes all 58 tests. This withdraws my confirming blocker on this thread; resolution remains with the reviewer who opened it.

gloryfromca (commented):

The prior blocker and operator-set note both reproduce as fixed: spaced `-o` Port values now match the registered port, and redirections before the destination no longer hide it. I withdrew my confirming blocker on LivXue's thread; that thread remains for its owner to resolve.

Named nonblocking follow-up: repeated conflicting Port settings use the wrong precedence. OpenSSH reports port 58717 for `ssh -G -o Port=58717 -o 'Port 22' example.com` because the first obtained value wins, while `_ssh_destinations` and the new test choose 22. This requires contradictory duplicate settings and removing the later duplicate is an immediate working path, so it does not meet the late-round blocker bar.

Verification: `uv run pytest tests/test_shell_machine_channel.py -q` passed 58 tests, and `uv run pytest tests/test_shell_comments.py tests/test_shell_approval.py tests/test_shell_background.py tests/test_rpc_spine.py tests/test_sandbox_unit.py -q` passed 454 tests. GitHub had not reported checks for this force-pushed revision at review time.

Coverage: I checked the repository rules and canonical Machine definition, the prior-to-current delta and full diff, callers and permission ordering, relevant history, backward compatibility, test-strength changes, and the machine-identity architecture constraints. All threads I opened remain resolved.

gloryfromca (commented):

The rebase onto current `main` preserved the reviewed detector and its tests byte-for-byte; the PR diff is still the same four files, and the prior fixes for spaced Port options and shell redirections still reproduce.

Named nonblocking follow-ups: the repeated conflicting-Port precedence mismatch remains as documented in the prior review. Separately, the detector treats any token named `ssh` as an executable position: `echo ssh -p 58717 root@203.0.113.7` is refused as reaching the registered machine even though the shell only prints those words. That can obstruct diagnostics or script generation. This change introduces the false positive and it is reachable in ordinary use, but an operator has working alternatives and it does not reopen the raw-SSH bypass, so it does not meet the round-seven blocker bar.

Verification: `uv run pytest tests/test_shell_machine_channel.py -q` passed 58 tests; `uv run pytest tests/test_shell_comments.py tests/test_shell_approval.py tests/test_shell_background.py tests/test_rpc_spine.py tests/test_sandbox_unit.py -q` passed 454 tests. At review time, 17 GitHub checks passed, the coverage gate was still pending, and docs deploy was skipped.

Coverage: I checked the repository rules and canonical Machine definition, the full diff and rebase delta, the detector's single caller and permission ordering, relevant history, backward compatibility, test-strength changes, and the machine-identity architecture constraints. All threads I opened remain resolved; the two open LivXue threads are not mine to resolve.

raven/agent/tools/machine_exec.py
silverLXT: Fixed in ecc9cc11f (rebased onto main cf3c95fd7).

You were right that this is the option in its other documented spelling, so the
value is now split on an equals sign or on whitespace, whichever comes first,
rather than on the equals sign alone:

    _OPTION_SPLIT = re.compile(r"\s*=\s*|\s+")
    ...
    pair = _OPTION_SPLIT.split(value.strip(), maxsplit=1)
    value = pair[1].strip() if len(pair) == 2 and pair[0].lower() == "port" else ""

Your reproduction on this head:

    ssh -o 'Port 58717' root@203.0.113.7 true  -> [("203.0.113.7", 58717)]
    ssh -o "Port 58717" root@203.0.113.7 true  -> [("203.0.113.7", 58717)]
    ssh -o 'Port = 58717' root@203.0.113.7 true -> [("203.0.113.7", 58717)]

raw_ssh_target now names conn_gpu for all three, and a non-port option is still
left alone in either spelling (-o StrictHostKeyChecking=no and
-o "StrictHostKeyChecking no" both keep port 22).

Covered by test_the_port_option_is_read_when_a_space_separates_it, parametrised
over the three spellings; all three fail on the previous head. The docstring and
the commit message now name both spellings instead of "-o Port=".

raven/agent/tools/machine_exec.py
silverLXT: Fixed in ecc9cc11f, though not by the route the note suggested -- adding the four
tokens to the one set does not close it, and for one of them it makes the miss
worse. Writing that out since you graded it cheap.

`2>&1` tokenises as three words, and the digit comes first:

    ["ssh", "2", ">&", "1", "-p", "58717", "root@203.0.113.7", "true"]

So `2` is taken for the destination before `>&` is ever looked at, and putting
`>&` in the terminator set does not change that. For `&>/dev/null` the set does
change it, into a different miss: the scan breaks at `&>`, this ssh ends with no
destination, and the registered host after it is not read at all -- an empty
answer rather than a wrong one, still a bypass.

The distinction the reader was missing is that a redirection does not end a
command. So the one set is now two:

    _COMMAND_SEPARATORS = {";", "&", "&&", "|", "||", "|&", "(", ")"}
    _REDIRECTIONS = {"<", ">", ">>", "<<", "<<<", "<&", ">&", "&>", "&>>", ">|", "<>"}

A separator still breaks. A redirection and the file it names are stepped over,
and so is a bare digit standing in front of one, after which this ssh keeps
being read. `|&` is a pipe, so it went to the separators. The file is skipped
only when there is one, so a dangling redirection before a separator does not
swallow the separator.

On this head:

    ssh 2>&1 -p 58717 root@203.0.113.7 true        -> [("203.0.113.7", 58717)]
    ssh &>/dev/null -p 58717 root@203.0.113.7 true -> [("203.0.113.7", 58717)]
    ssh root@h >; ssh -p 9 root@k                  -> [("h", 22), ("k", 9)]

One shape stays unresolvable and is documented at the check rather than papered
over: the tokens do not say whether a space separated the digit from the
operator, so `ssh 2 >&1` -- a host literally named `2` -- reads as a descriptor.
No registry row has that shape, since an address carries dots or letters, and
the failure direction is to keep scanning and find the real host.

Covered by test_a_redirection_before_the_destination_does_not_hide_it (three
redirection shapes through ExecTool) and test_redirections_are_stepped_over_wherever_they_fall
(six positions directly against _ssh_destinations). All fail on the previous head.

silverLXT:
Revision ba233ba54: the named nonblocking precedence finding is fixed, and the
branch is merged up to current main.

You were right and my earlier test had it backwards. Measured on OpenSSH 9.9p2,
ssh takes the FIRST obtained value for every option, with -p and -o Port
queueing together:

    ssh -G -o Port=58717 -o 'Port 22' example.com  ->  port 58717
    ssh -G -o 'Port 22' -o Port=58717 example.com  ->  port 22
    ssh -G -p 2222 -o Port=58717 example.com       ->  port 2222
    ssh -G -o Port=58717 -p 2222 example.com       ->  port 58717

The detector overwrote instead, so it was not only the conflicting-option case:
`ssh -p 58717 -p 22 root@<registered>` read as port 22, which means a command
that really reaches the registered machine read as unregistered and ran on the
plain shell path -- past the cap, the process-group sweep and the ledger. That
is the direction this guard exists to close, so I took the finding as a defect
rather than a suggestion.

The prior case in the redirection table asserted 22 under the id
"later-option-wins"; it now asserts 58717 as "first-port-wins", joined by
test_a_repeated_port_keeps_the_first_value_as_ssh_does covering both spellings
in both orders and -p twice in both orders. All six fail on the previous head.

The second named follow-up -- `echo ssh -p 58717 root@203.0.113.7` being
refused although the shell only prints it -- I am leaving as is and recording
why: deciding that a token named ssh is not in executable position needs a real
parser, and the failure direction here is a false refusal with the reason
printed, not a missed bypass. I would rather over-refuse `echo ssh ...` than
under-refuse something that runs.

Verification on this head: tests/test_shell_machine_channel.py 64 passed; all
tests/test_shell_*.py plus test_rpc_spine and test_sandbox_unit 652 passed;
ruff, ruff format and ty clean. Merge-tree against main is clean and the branch
is 0 behind.

gloryfromca (commented):

The repeated-Port precedence finding is fixed. I compared all six new `-p`/`-o Port` orderings with OpenSSH and the parser now keeps the same first obtained value; the corrected expectation is backed by broader coverage rather than a weakened test. The merge-up preserves the prior fixes and, against the refreshed current `main`, the PR remains the same four-file scope.

The previously named executable-position false-positive remains the only nonblocking follow-up. I found no new defect on this revision.

Verification: `uv run pytest tests/test_shell_machine_channel.py -q` passed 64 tests; `uv run pytest tests/test_shell_comments.py tests/test_shell_approval.py tests/test_shell_background.py tests/test_rpc_spine.py tests/test_sandbox_unit.py -q` passed 454 tests. GitHub CI was still in progress at review time, with 16 checks passed and three unit shards pending.

Coverage: I checked the repository rules and current canonical Machine definition, the full diff and prior-to-current delta, the detector's caller and permission ordering, relevant history, backward compatibility, test-strength changes, and the machine-identity architecture constraints. All threads I opened remain resolved; the two open LivXue threads are not mine to resolve.

raven/agent/tools/machine_exec.py
```diff
@@ -96,6 +97,174 @@ def machines_registered() -> bool:
...
+                # descriptor here -- a shape no registry row has, since an
+                # address carries dots or letters.
+                index += 1
+                continue
+            index += 1
+            if destination is not None:
+                continue
+            if word.startswith("-") and len(word) > 1:
+                value = ""
+                if word.startswith("-p") or word.startswith("-o"):
```
LivXue, line 195: **Blocker - read a bundled short-flag group before its port argument becomes the destination.**
`-vp 58717` matches neither test here: `word.startswith("-p")` is false for `-vp`, and the
`elif len(word) == 2 and word[1] in _SSH_VALUE_FLAGS` branch below needs a two-character token.
The group is skipped as if it took no argument, so `58717` falls through to
`destination = word.rsplit("@", 1)[-1]` and becomes the destination:
`_ssh_destinations("ssh -vp 58717 root@203.0.113.7 true")` returns `[('58717', 22)]`, host and
port both wrong, and `raw_ssh_target` returns None for a machine registered at
`203.0.113.7:58717`. OpenSSH takes the bundled spelling -- `ssh -G -vp 58717 example.com`
prints `port 58717`, and so do `-4p` and `-vvvp` (OpenSSH_8.9p1).

Reproduced through `ExecTool` with the registry holding `203.0.113.7:58717`: the line runs on
the plain shell path and the executed command's own stderr reports `connect to host
203.0.113.7 port 58717`, with a `; touch <marker>` tail confirming the whole line executed. The
control `ssh -p 58717 root@203.0.113.7` is refused as `conn_gpu` at this same head.

This is the class already fixed twice on this PR -- the `-o Port=` spelling at `8ab4d06c` and
the spaced spelling at `ecc9cc11f` -- a spelling ssh honours that the guard reads as port 22.
Bundling is ordinary getopt rather than an evasion, so the flag scan wants to read a `-xyz`
group character by character instead of only a two-character token.

raven/agent/tools/machine_exec.py
```diff
@@ -96,6 +97,174 @@ def machines_registered() -> bool:
...
+                # the digit was taken for the destination and the real host,
+                # further along the line, was never read. The tokens do not say
+                # whether a space separated the digit from the operator, so a
+                # destination that is itself a bare integer is read as a
+                # descriptor here -- a shape no registry row has, since an
+                # address carries dots or letters.
+                index += 1
+                continue
+            index += 1
+            if destination is not None:
```
LivXue, line 191: **Blocker - keep reading options after the destination, as ssh does, until the first
non-option word.** This line ends the option scan at the destination, but OpenSSH re-enters its
option loop after parsing the host and keeps reading until the first non-option word. So
`ssh root@203.0.113.7 -p 58717 true` connects on 58717 while `_ssh_destinations` returns
`[('203.0.113.7', 22)]`; `raw_ssh_target` then finds no row for a machine registered at 58717
and the line runs on the plain shell path, past the cap, the process-group sweep and the ledger.

Measured with `ssh -G` on OpenSSH_8.9p1:

    ssh -G root@example.com -p 58717 true    -> port 58717
    ssh -G root@example.com -o Port=58717    -> port 58717
    ssh -G root@example.com true -p 58717    -> port 22
    ssh -G root@example.com -- -p 58717      -> port 22

The last two give the boundary the fix should stop at: reading ends at the first non-option
word and at `--`, which is why `ssh root@h true -p 58717` is correct here today. First obtained
value still wins across the destination -- `ssh -G -p 58717 root@example.com -p 22` prints
58717 -- so the `port_set` latch added at `a61113df6` already carries that half.

Reproduced through `ExecTool` with the registry holding `203.0.113.7:58717`, for both the `-p`
and the `-o Port=` spelling written after the host: each ran on the plain shell path and the
executed command's own stderr reports `connect to host 203.0.113.7 port 58717`.

raven/agent/tools/machine_exec.py
```diff
@@ -96,6 +97,174 @@ def machines_registered() -> bool:
...
+# word and taking it for the destination -- lost the registered host that came
+# after it (reviewed 2026-09-20). The operator and the file it names are
+# stepped over instead, and this ssh's own words keep being read.
+_REDIRECTIONS = frozenset({"<", ">", ">>", "<<", "<<<", "<&", ">&", "&>", "&>>", ">|", "<>"})
+
+# An -o option's name and its value are separated by an equals sign or by
+# whitespace; OpenSSH honours both spellings.
+_OPTION_SPLIT = re.compile(r"\s*=\s*|\s+")
+
+_SSH_VALUE_FLAGS = frozenset("bcDEeFIiJLlmOoQRSWw")
```
LivXue, line 118: **Nonblocking - add `B` to this set so its argument is not read as the destination.** `ssh(1)`
takes a value for `-B bind_interface`. The option string in the shipped binary is
`1246ab:c:e:fgi:kl:m:no:p:qstvxAB:CD:E:F:GI:J:KL:MNO:PQ:R:S:TVw:W:XYy`, where `B:` takes a
value and `P` does not; `B` is the only value-taking flag this set omits, `p` and `o` being
handled by the branch above. So `-B` is skipped without consuming its argument, and
`_ssh_destinations("ssh -B lo -p 58717 root@203.0.113.7 true")` returns `[('lo', 22)]` -- the
interface name becomes the destination and the registered host is never read.

Reproduced against `_ssh_destinations` at this head. `ssh -G -B lo -p 58717 example.com` prints
`port 58717` on OpenSSH_8.9p1, confirming ssh accepts the flag and consumes its value.

Graded nonblocking rather than blocking because `-B` is rare in ordinary use, unlike the
bundled group and the port written after the host. It is filed separately because the remedy is
a different one: this set is a transcription of ssh's option string and wants checking against
it, not a change to the scan.

LivXue:
Blocking: 2 findings need correction before this revision can merge; see the inline notes.

Scope: I reviewed the `raven/agent/**` portion of this PR -- 2 of 4 changed files, which is the
whole of the review zone. The other two are `tests/test_shell_machine_channel.py` and
`CHANGELOG.md`, read as context. Ring 3 (`raven/agent/subagent/`) and ring 4 (the dispatch tools
and the DAG skill) are untouched by this revision, so no change here carries a sign-off.

Both findings I filed at `46960856` are fixed at this head and I re-ran every reproduction from
the earlier rounds. One correction to my own last round: the remedy I named for the operator-set
note was wrong. The author's reply showing that adding those four tokens to the one set leaves
`2>&1` worse, because the bare descriptor is taken for the destination before the operator is
ever seen, is correct; the split into `_COMMAND_SEPARATORS` and `_REDIRECTIONS` with a step over
a bare descriptor is a better fix than the one I suggested. I am also not re-filing the repeated
`Port` precedence point, which is the other reviewer's and is fixed at `a61113df6`.

Confirmed findings:

- Blocker, `raven/agent/tools/machine_exec.py:195`: a bundled short-flag group such as
  `-vp 58717` matches neither the `-p`/`-o` test nor the two-character value-flag test, so it is
  skipped as a no-argument flag and its port argument becomes the destination.
  `ssh -vp 58717 root@203.0.113.7 true` reads as `[('58717', 22)]` -- host and port both wrong --
  and runs on the plain shell path. `-4p` and `-vvvp` behave the same, and OpenSSH honours all
  three.
- Blocker, `raven/agent/tools/machine_exec.py:191`: the option scan ends at the destination, but
  ssh re-enters its option loop after the host until the first non-option word.
  `ssh root@203.0.113.7 -p 58717 true` connects on 58717 and reads here as port 22.
- Nonblocking, `raven/agent/tools/machine_exec.py:118`: `B` is the one value-taking flag missing
  from `_SSH_VALUE_FLAGS`, so `ssh -B lo -p 58717 root@203.0.113.7` takes the interface name for
  the destination.

All three share one mechanism: an option shape the scan does not model yields not an unknown
destination but a wrong one, because the destination is assigned from any word not recognised as
a flag. For what it is worth, the stdlib `getopt` accepts ssh's own option string and reproduces
all four measured cases, including both boundary cases, when a second pass over the words after
the destination mirrors ssh's own re-entry.

Verification:

- Reproduced through `ExecTool` with the registry holding `203.0.113.7:58717`. For each bypass
  the line ran on the plain shell path and the executed command's own stderr reports `connect to
  host 203.0.113.7 port 58717`; a `; touch <marker>` tail confirms the whole line executed. The
  control `ssh -p 58717 root@203.0.113.7` is refused as `conn_gpu` at the same head.
- `ssh -G` comparisons on OpenSSH_8.9p1 for every form above, including the boundary cases
  `... true -p 58717` and `... -- -p 58717`, which both stay on port 22 and which this scan
  already reads correctly.
- Every reproduction from the earlier rounds, re-run at this head: path-qualified and
  backslash-escaped ssh, a second machine at the same address on another port, the
  longer-address case, unspaced compound operators, two ssh commands in one line, `-o Port=`,
  the spaced `-o "Port n"` spelling, and a redirection ahead of the host. All still behave as
  those threads say they were fixed to.
- `uv run pytest tests/test_shell_machine_channel.py -q`: 64 passed. `lint-imports`: 10 kept,
  0 broken. GitHub checks at review time: 20 passed, 0 failed, 2 skipped.
- All three findings were reproduced by running the code, not by reading it.

Checked and deliberately not reported: the documented deviation for a destination that is a bare
integer written before `>&` holds -- the registry does not validate the host field, but no real
address is purely numeric and the failure direction is to keep scanning for the real host. The
`echo ssh ...` over-refusal named by the other reviewer is a false refusal with its reason
printed rather than a missed bypass, and the author has recorded why it stays. `bash -c 'ssh
...'`, a backticked command and a host built from a shell variable all still reach the machine;
unchanged from my last round, they need deliberate indirection and no token-level fix reaches
them. Also checked and clean: the separated `-v -p 58717` spelling, which parses correctly; the
guard sitting ahead of `_guard_command`, which leaves no approve-then-run path for this shape;
and the complete absence of deletions in the diff, so no earlier guard was dropped.

raven/agent/tools/machine_exec.py
silverLXT: Fixed in 2ee41be38. A `-xyz` group is now read letter by letter until one takes a value; the rest of the group, or the next word, is that value (`_read_option_group`). `-vp 58717`, `-4p 58717`, `-vvvp58717` and `-vo Port=58717` all read port 58717, and `-o Port=58717 -vp 22` keeps 58717 (first value wins). Covered by `test_options_are_read_the_way_getopt_and_ssh_read_them`, and `test_the_table_agrees_with_the_ssh_on_this_computer` re-measures every row with `ssh -G`. 12 of the new cases fail on ba233ba54.

raven/agent/tools/machine_exec.py
silverLXT: Fixed in 2ee41be38. The scan no longer stops at the host: options are read on until the first non-option word (the remote command) or `--`, and `--` before the host means no re-entry. Measured with `ssh -G` on OpenSSH_9.9p2, same as your 8.9 results: `root@h -p 58717 true` -> 58717, `root@h -o Port=58717 true` -> 58717, `root@h true -p 58717` -> 22, `root@h -- -p 58717` -> 22, `-- root@h -p 58717` -> 22, and `-p 58717 root@h -p 22` -> 58717. Through ExecTool, `ssh root@203.0.113.7 -p 58717 true; touch <marker>` is now refused as conn_gpu and the marker is not created.

raven/agent/tools/machine_exec.py
silverLXT: Fixed in 2ee41be38. The set is derived from the getopt string itself (`_SSH_OPTSTRING`, copied from the OpenSSH_9.9p2 binary) instead of a hand-kept list, so `B` is in. One difference from your 8.9 string: `P:` takes a value since OpenSSH 9.2 (`-P tag`); the current release is followed and the comment says so. The ssh -G test skips the `-P` row on a client that predates it.

gloryfromca (commented):

I reproduced the three author-described parser corrections on this revision: bundled option groups now consume their values as getopt does, options after the destination affect the connection only until the remote command or `--`, and the value-taking option table includes `-B` and current `-P`. The new compatibility table agrees with the installed OpenSSH client, so the tests strengthen rather than mask the behavior. The merge-up leaves the pull request at the same four-file scope.

The previously named executable-position false positive remains the only nonblocking follow-up. I found no new defect on this revision.

Verification: `uv run pytest tests/test_shell_machine_channel.py -q` passed 94 tests; `uv run pytest tests/test_shell_comments.py tests/test_shell_approval.py tests/test_shell_background.py tests/test_rpc_spine.py tests/test_sandbox_unit.py -q` passed 475 tests. GitHub checks are green except unit shard 4/4, which is still running.

Coverage: repository rules and the canonical Machine definition, the full pull-request diff and revision delta, the detector's caller and ordering, relevant history, backward compatibility, test strength, and the stated architecture constraints. All threads I opened remain resolved; the remaining review threads were opened by another reviewer.

0xKT (approved):

PR #553, opened 2026-09-20
fix(raven-oncall): a campaign bills only its own jobs under a shared rounds directory

silverLXT:
A campaign asked the shared rounds directory how much had been spent, and the
directory answered for every job in it, including jobs its sibling campaigns
had submitted. Rounds that keep one directory therefore bill each other, and
the bill is cumulative: round three opens already charged for rounds one and
two.

Measured on a run of 2026-09-11. Five campaigns declared the same rounds
directory. Round two had a budget of 130 and had actually used 34.6, but the
eighteen jobs the earlier rounds had left in the directory brought the reading
to 125.6, so the remaining 4.4 would not cover one more trial at 5.9 and the
round stopped. 142 GPU-minutes went unspent and that round's multi-seed
conclusion was lost. A sibling run that gave each round its own directory was
not affected, which is why this reads as a directory-hygiene problem until you
look at the meter.

Two changes. Spend is now counted from the campaign's own ledger, so a job a
sibling submitted is not billed here; the filter sits above the existing
timeline arithmetic, so same-device overlap is still charged once and two
cards in parallel are still charged twice. And declare refuses a rounds
directory a live sibling campaign already keeps, naming the sibling, so the
two rounds do not land in one directory in the first place.

Type: fix

Verification:
- All oncall tests: 775 passed, 1 failed. The one failure
  (test_the_products_tool_face_equals_the_forks_config_intent, a missing
  optional a2a_client dependency on this machine) also fails on 0464aa1b1, so
  this adds six passing tests and no new failures.
- The six new ones pin: a sibling's job is not billed; the campaign's own two
  trials overlapping on one device are still billed 20 minutes once; a job
  this executor just submitted is billed before the ledger has caught up (no
  refund window); billing_only leaves a foreign backend untouched; declare
  refuses a directory a live sibling keeps; a concluded sibling, and the
  campaign itself, do not count as a clash.
- The first test also pins the measured number: the same fixture directory
  reads 125.6 unfiltered, which is 34.6 of its own plus 91.0 of its siblings.

Risk: low for the meter, moderate for declare. The meter can only ever bill
less than before, and a run that already gives each round its own directory
reads the same. Declare now refuses something it used to accept, so a caller
that deliberately shares a directory between live campaigns will see an error
naming the sibling; concluded siblings and re-declaring the same campaign are
both still fine.


gloryfromca (commented):
Blocking: apply campaign ownership to OpenFOAM spend and prevent stale job adoption from concluded sibling directories.

Two concrete failures remain:

- `OpenFoamExecutor.spent_minutes()` overrides the filtered implementation and still charges every job under the shared directory.
- Reusing a concluded sibling's directory can collide on the deterministic idempotency key, causing the new campaign to inherit both an old bill and an old result.

Review coverage: I read `AGENTS.md`, `CLAUDE.md`, `CONTEXT-MAP.md`, and the applicable runtime context; inspected the full diff and all three commits; traced backend construction and spend callers, the watcher, ledger semantics, config-key generation, and submit/restart behavior; checked backward compatibility and the resource-admission architecture; and checked that existing tests were not weakened. `git diff --check` is clean. `uv run pytest tests/test_agents_oncall_flow_*.py -q` passed all 758 tests, but focused reproductions exposed the two gaps marked inline.

agents/raven-oncall/plugins/oncall-flow/oncall_flow/process_backend.py
```diff
@@ -282,6 +286,40 @@ async def _arun(self, cmd: str) -> tuple[int, str]:
 
     # ---- budget ----
 
+    def restrict_spend_to(self, own: Ledger | Iterable[str]) -> None:
```
gloryfromca, line 289: Blocking: apply this ownership filter in `OpenFoamExecutor.spent_minutes()`.

`OpenFoamExecutor` inherits `restrict_spend_to()`, so `billing_only()` appears to configure it, but its overridden `spent_minutes()` loops over every scanned row without calling `_counts_toward_spend()`. With an owned 20-core-minute row and a sibling 80-core-minute row, `restrict_spend_to({'mine'})` still returned 100 rather than 20 in a focused reproduction. OpenFOAM campaigns sharing a rounds directory therefore retain the cross-campaign billing failure this change is meant to fix. Please filter before populating `_cores` and the timeline, and cover this backend.

agents/raven-oncall/plugins/oncall-flow/oncall_flow/tools/ops_declare.py
```diff
@@ -58,6 +58,34 @@ def _round_dir_inside_case(remote_dir: str, staged_case: str) -> tuple[str, str]
...
+        return None
+    from oncall_flow.tools.ops import _campaign_meta
+
+    mine = Path(remote_dir).expanduser()
+    try:
+        siblings = sorted(d for d in ops_home.iterdir() if d.is_dir())
+    except OSError:
+        return None
+    for sib in siblings:
+        if sib == cdir or (sib / "concluded.json").exists():
```
gloryfromca: Blocking: prevent a concluded sibling's stale jobs from being adopted by the new campaign.

Skipping concluded siblings makes their `remote_dir` available even though its job directories remain. Job IDs are deterministic from the config, so once the new ledger records the same key, the new filter treats the predecessor's directory as its own. I reproduced a completed 12-minute job named `same`, then opened a new campaign ledger containing `same`: the new executor reported 12.0 spent before its first submit, and `submit()` sees the old `result.json` as an idempotent completion instead of launching the new run. Repeated or no-parameter configs therefore inherit both old spend and old results. Keep the directory exclusive while stale jobs remain, or isolate ownership strongly enough to disambiguate them.

gloryfromca (commented):
Blocking: the two existing review threads still reproduce and must be addressed.

This revision's only delta replaces the ledger attribute probe with runtime `isinstance` narrowing; I traced its callers and found no separate regression in that change. It does not alter either outstanding failure path, and both focused reproductions still fail with the same measured results.

Coverage for this revision: I rechecked the repository rules and applicable context, the full branch diff and new delta/history, backend and tool callers, ledger/config-key/restart compatibility, the resource-admission architecture, and test coverage; no existing tests were weakened. `git diff --check github/main...HEAD` is clean. `uv run pytest tests/test_agents_oncall_flow_*.py -q` passed all 758 tests. The original two threads remain open because neither is settled.

agents/raven-oncall/plugins/oncall-flow/oncall_flow/process_backend.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed on `5ed0d5d642a5`: the OpenFOAM override now applies `_counts_toward_spend()` before updating `_cores` or the timeline, and the added 20/80 core-minute regression verifies the restricted result is 20. The full on-call suite passes.

agents/raven-oncall/plugins/oncall-flow/oncall_flow/tools/ops_declare.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed on `5ed0d5d642a5`: a concluded sibling with ledger records now keeps the shared directory, and an unreadable ledger is conservatively treated as occupied. The new regression covers the stale-key case while the empty concluded-sibling and same-campaign redeclaration paths remain allowed. The full on-call suite passes.

gloryfromca (commented):

The new commit fixes both prior findings and adds direct regressions for them. I replied to and resolved both threads after verifying the fixes.

Coverage: I rechecked the repository rules and applicable context, the full branch diff and revision history, all affected backend/tool callers, ledger and deterministic-key behavior, restart/backward-compatibility paths, the resource-admission architecture, and the tests for weakening or missing assertions. I found no new defect. `git diff --check github/main...HEAD` is clean, and `uv run pytest tests/test_agents_oncall_flow_*.py -q` passed all 761 tests.

0xKT (approved):

PR #556, opened 2026-09-20
docs(*): record what the settings page shipped, where the design said otherwise

gloryfromca:
## Summary

Two statements in the committed settings design read as fact and are not. This
corrects them in place; no other section changes.

**`features/model/vendors.ts` was never created.** The design names it four
times as the home of the vendor mark table, with the table re-exported through
that domain's `types.ts` so the settings page could reach it. None of that was
needed: the table already lives in `components/ProviderMark.tsx`, which is the
same layer the shared picker is in, so the picker imports it directly and no
cross-domain edge arises.

This is the sentence that cost the picker its marks. The shared picker shipped
drawing provider names as plain text, on the reading that the edge was
forbidden -- a conclusion taken from this document rather than from the tree.
It was put back in #540. The Decisions entry now states where the table is and
keeps the planned path as the route not taken, so the next reader does not
re-derive it.

**The meta-tool entry was true but partial.** It said `tool_search` and
`tool_call` are greyed, not switchable, and left out of the counter, which
still holds. It did not carry the distinction #540 had to fix -- hidden from
the schema is not the same as withheld from the model, so the DAG controls were
drawn as fixed while their switch worked -- and it did not describe a default
install, where `tools.toolSearch.enabled` is off and a card named after tool
search holds the one meta-tool that is not tool search.

Two other decisions were deviated from during execution and both were reversed
back to what this document already said, so neither needed an edit here: the
picker is a popover again, and the About row offers the upgrade as its own
step.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
make check-source-language   -> pass
make check-large-files       -> pass
```

Every path this document now names was checked against the tree at the target
tip: `components/ProviderMark.tsx` exists and holds the table,
`features/model/vendors.ts` does not exist. The four occurrences of the missing
path are gone except one, which is now explicitly the plan that was not
followed.

## Risk

Documentation only; no code path changes. Rollback is the one commit.

## Related Issues

N/A


gloryfromca (commented):
Blocking: correct the stale outside-domain review count.

One factual inconsistency remains inline. The substantive shipped-state corrections otherwise match the implementation and history: `ProviderMark.tsx` predates the settings rewrite, `ModelPicker.tsx` imports `ProviderIcon` after #540, and `builtin` now distinguishes the fixed meta-tool rows while the DAG controls remain switchable.

Covered the repository rules and Web UI vocabulary, the complete target diff and surrounding document, the shipped implementation and callers, #494/#540 history, backward compatibility, test changes, and the relevant architecture boundaries. This patch changes no tests or runtime behavior. `git diff --check` passed. `make` is unavailable in this environment, so I ran the exact `check-source-language` and `check-large-files` recipes directly with `uv`; both passed. GitHub's docs, page, contract, repository, commit, and other completed checks are green; two unit-test shards were still pending when reviewed.

docs/specs/2026-09-17-web-settings-page-design.md
```diff
@@ -419,16 +418,16 @@ path and storage location with copy buttons.
...
 | `ui-web/src/rpc/fixtures/settings.ts` | 12 methods answered | every method the source calls |
 | `i18n/messages.json` | 302 `gui.set.*` keys | `gui.settings.*` for the new page; the `gui.set.*` keys other files reference stay |
 
 ## Cost
 
 Who pays: this change's author for the backend batch, the front-end rewrite,
 fixtures, tests and the manual pass; the architecture-branch owner for two
-reviews outside the settings domain (`components/ModelPicker.tsx`,
-`features/model/vendors.ts`) and for the rename pass if the conventions
+reviews outside the settings domain (`components/ModelPicker.tsx`) and for
```
gloryfromca, line 429: Blocking: change the stale outside-domain review count from two to one.

The removed `features/model/vendors.ts` entry was what made this plural: after this patch the parenthetical names only `components/ModelPicker.tsx`, and history confirms that `ProviderMark.tsx` predates #494 while #540 changed `ModelPicker.tsx`, not `ProviderMark.tsx`, to consume it. As written, the cost record still asks the architecture owner for two outside-domain reviews when only one outside-domain file shipped. Please make the count singular, or identify the second review if the number is intentional.

0xKT (approved):

PR #557, opened 2026-09-20
docs: open the showcase with a game run and give the zh readme the section

LivXue:
## Summary

Open the Showcase with a Raven-Code run that picked, built and judged an FPS
game over 42 rounds, and give README.zh-CN.md the Showcase section it never had.

The entry sits in its own table above the existing ones: the left cell carries
the run's task graph and a 30s clip of the result, and the right cell is left
empty for a second sample. It is a separate table rather than a row in the deck
grid, because a playable game is not a deck and an empty cell inside that grid
would read as a broken row.

The clip is a GitHub attachment referenced by a bare URL on its own line. That
is the only form that renders a player: a video element is removed by the
markdown sanitizer, which was confirmed by rendering both forms through the
markdown API before writing either one. The same rendering also shows that the
player header displays the name the file carried at upload, so the clip was
re-uploaded under its intended name rather than left under a working name.

README.zh-CN.md previously had eight sections against the English nine, and the
missing one was Showcase. Rather than hand-copying it, the section was derived
from the English one and its entry titles translated, so the two files stay
structurally aligned; alt text is left in English, matching every other alt
attribute in that file. Both files now render the same structure: one player,
seventeen images, fourteen cells, one empty cell.

The section's opening sentence is reworded in both files while the section is
being touched: it now says Raven generates the orchestration rather than
chooses it, and the Chinese copy names the panel as the multi-agent
orchestration graph.

- The task-graph alt text describes three nodes running "across 42 game rounds".
  The panel actually shows round 42 of a capped loop, with those three nodes
  being that one round's pick, do and judge steps, so the wording is loose. The
  author chose to keep it; it is recorded here rather than silently altered.
- The first upload of the clip is still attached to its hosting issue under the
  earlier filename. Nothing references it, and attachments are not removable
  once posted, so it stays as upload history.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Rendering, through GitHub's own markdown API rather than by eye, on the exact
section text of each file:

    gh api -X POST /markdown --input -   # mode gfm, repo context

Both files returned one video element, seventeen images, fourteen cells and one
empty cell, with the player header reading the intended filename and the first
entry title reading as intended in each language.

Byte equality of every uploaded asset against its local source, fetched back
over HTTP:

    curl -sL <asset-url> | sha256sum

All matched. A HEAD request is not usable here: the attachment host answers HEAD
with 403, which reads as a failed upload.

Tests that actually read these two files, found by grepping for them rather than
assumed absent:

    uv run --frozen --all-extras pytest tests/test_cli_onboard_commands.py tests/test_docker_runtime.py -q
    344 passed

Repository gates:

    make check-commits check-large-files check-source-language

All passed. The two file gates were also run against a bare revision, which
compares the working tree, because a commit-range form reports nothing while an
edit is still uncommitted.

No test asserts anything about Showcase content, asset URLs or section parity
between the two READMEs. The 344 above cover one heading lookup and one
whole-file read; nothing would fail if an entry were malformed. The rendering
check above is what stands in for that, and it is manual.

## Risk

Documentation only; nothing executable changes. The assets are public
attachments on a repository that is already public and carry no credential or
private detail. The clip is 53 MB, so the Showcase section is heavier to load
than before, and the player appears only in the rendered file view: the diff and
compare views show the bare URL as text. Rollback is a revert of these commits;
the assets stay hosted either way.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the full `github/main...HEAD` diff and surrounding README sections, all four branch commits, the relevant README-reading tests, backward compatibility, and the repository rules in AGENTS.md. I also checked for weakened tests (none changed) and architecture impact (none; this is documentation-only).

Verification: GitHub's Markdown API renders each Showcase section with one video, 17 images, 14 cells, and one intentionally empty cell; both language versions reference the same asset set. The new PNG and MP4 URLs return the expected content types, and I inspected the task-graph image against its caption. `git diff --check`, commitlint, and the large-file, source-language, and commit-message gate scripts pass. `uv run --frozen --all-extras pytest tests/test_cli_onboard_commands.py tests/test_docker_runtime.py -q` reports 344 passed.

userName20260323 (approved):

PR #559, opened 2026-09-20
feat(*): take the agents page to the agent hub prototype

0xKT:
## Summary

Takes the web UI's agents page (rail entry "Agent Hub", page "Agents") to the prototype's agent hub, and gives the wire what the prototype needed and did not have: a default model on an agent row.

Base: `refactor/ui_web_architecture`, the long-lived architecture branch (#475), not `main`. The diff shown here is the diff that merges into that branch.

What the page does now:

- Three sections, never folded: connected, available to connect, not installed on this machine. One control per row (connect / disconnect / retry / install), a status dot only where there is something to say, and a one-line summary from a small catalogue of what each agent is and who makes it.
- A 520x500 sheet in the shared detail host: what the agent is good at (editable, saved on blur), the API key where one is needed, the install command and site for an absent agent with a re-check, a Test button, and the model pill.
- A write in flight is held on the row as "connecting" for its length; a refusal -- a model pick included -- stays on the row as red text with Retry rather than as a toast. A running test keeps a Stop control.

The model pill, by the row's own rule (`model_source`, a new wire field):

- The built-in row picks from raven's connected providers. The pick is stored naming its provider (`stored_model_id`, the way `config.set model` stores the host's), validated against a provider raven knows that holds a usable credential, and paired with that credential on every dispatch through `ProviderPool.bind_pin` -- the backend resolves the pin per run, so a cached backend never runs one conversation's key against another's model. The write and the dispatch read the provider off the stored id through one function, `stored_provider_name` (the inverse of `stored_model_id`), so a section raven has no spec for resolves to that section on both sides.
- An acp row, raven's own shipped products included, picks from the choices its handshake advertised (`model_choices`) and the value is pushed over `session/set_config_option`, through one resolver (`session_model_for`) for a spawn, a direct chat and a DAG node (`model_for` threaded through the runner).
- An openai or cli row has no menu; the pill is disabled. Unset reads by ownership: one of raven's own "follows the main Raven", a third party runs on "its own default". Ownership is the wire's `own` (a new field: built-in, vendored, or an acp row whose handshake named Raven), with the catalogue as the fallback for a server that predates it. The picker offers listed ids only; there is no typed entry, since an acp row's write takes exact menu values and a built-in row's adds to no provider.

Deviations from the prototype, deliberately:

- No cancel while connecting; the row shows "connecting" until the server answers.
- Third parties do not pick from raven's providers: there is no path that would deliver raven's credential to them, so they pick from their own advertised menu.
- Raven-PPT is a hidden row and is not listed; "build an agent" is not in this PR.
- The old page's rename control, transport chips and test-cost notes are gone; the prototype has none of them.
- The built-in row's description stays read-only in the sheet (the server now takes the write; the field follows later).

Install data (commands and sites) comes from each vendor's own documentation: Claude Code, Codex, OpenCode, Hermes, OpenClaw, MiroThinker, GitHub Copilot, Qwen Code, CodeBuddy, Qoder, Grok, Kimi Code, Pi.

Wire changes: `SubagentRow.own / model / model_choices / model_source`; `subagents.update` takes `model`, `provider`, `clear_model` (clear wins) and `description: null` (back to the factory text; a built-in row goes back to following its seed); `subagents.test` takes `source: "vendored"`; `ThirdPartyAcpSubagentConfig.model` (and `schemas/subagent.schema.json` re-exported for it); `SubagentManager(provider_pool=...)` is injected by the loop and the playbook CLI.

Reviewed in two passes (three read-only reviewers plus a refuter per finding): ten findings confirmed, all fixed in the backend commit -- among them a dispatch that withheld the model from a cached built-in backend, a `model_source` that promised a write the server refused, and a built-in override edit that would have shadowed a legacy-spelled row.

Review round (gloryfromca, LivXue): all five of the first review's blockers and the second review's two blockers plus its two "fix or file" items where a fix was in reach are addressed in four commits -- the one-resolver read of a stored built-in model (`stored_provider_name`) with the spec-less section case tested on both sides and an unreadable config raised rather than blamed on the pick, the unusable-pin warning keyed per pin, a refused acp push said once per value instead of once per turn, the wire's `own` flag read by the page, the stored-prefix spelling matched to the host slug, no typed picks, a Stop for a running test, the model op held on the row, and the stale narrative in the manifest, the CONTRIBUTING ledger rows and two comments. The first reviewer's two later nonblocking follow-ups are in as well: the Claude Code preset no longer requires a `claude` on PATH (the adapter's SDK pin carries the CLI), and a stored id whose prefix names no provider section is read as part of the id, so a hand-written gateway-served id keeps its pin. The second review's finding 4 is fixed on the record side: a failed manual Test writes its verdict over the previous record's capabilities rather than erasing them, and a row recorded before the menu existed reads "attention, model menu not measured" until a re-verify lands instead of a stored "ready" (`CapabilitySnapshot.model_menu_measured`). Deferred by the owner as a feature the prototype does not draw: a row field saying whether the stored model is in effect (pin resolved / push accepted), which would be the durable answer to the second review's "silently" in findings 1 and 2. Not in this PR and not tracked as an issue for now.

Four pre-existing defects the live regression surfaced are fixed here as well, each in its own commit:

- `raven acp` answered `session/set_config_option` (and `session/new`) with the process default as the model option's `currentValue`, because it asked `model.options` without a session id. A client that renders `currentValue` as the selected item snapped back to the old model after every switch. Both call sites now ask about the session whose answer it is.
- The acp probe resolved `argv[0]` alone, and a shim-launched preset's command is an `npx` one that resolves on any machine with node, so Pi read as connectable wherever `pi` was not installed and the connect failed a minute later inside the adapter. `SHIM_REQUIRED_EXECUTABLES` (in `presets.py`) names the local agent a shim reaches for with its install command; the probe asks after it too and reports the row missing with the install beside it. Only Pi is listed: `codex-acp` bundles its agent, and `claude-agent-acp` runs the CLI its `@anthropic-ai/claude-agent-sdk` pin carries as a per-platform optional dependency (verified in the 0.66.0 package: the SDK's platform package holds the `claude` binary, and the live run's "Claude Code 2.1.220" is that bundled build, not the machine's), so neither wants an install.
- The Claude Code preset could not connect for an account whose Claude settings name a recent model: `claude-agent-acp@0.66.0` bundles CLI 2.1.220 through its SDK pin, and the CLI refuses such a model ("version 2.1.251 or newer is required"). The pin is now 0.79.0 (CLI 2.1.274), measured: handshake with resume, fork and load, six model choices, five modes; connect from the row in 9 s where 0.66.0 failed after 155 s.
- A row connected from the page recorded no capability snapshot, so it read "capabilities not recorded -- run a test", stateless and menuless, until a Test or a restart. The connect now records the handshake the Test would (`record_capabilities`, shared by both), only when the record is missing, stale or from before the menu.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Backend (worktree venv with the dev extra):

```
uv run pytest tests/test_rpc_subagents.py tests/test_rpc_subagents_model.py tests/test_rpc_schema_match.py \
  tests/test_subagent_manager.py tests/test_subagent_probe.py tests/test_subagent_dag_runner.py \
  tests/test_subagent_acp.py tests/test_subagent_builtin_agents.py tests/test_rpc_registration.py \
  tests/test_agent_loop_session_model.py tests/test_agent_loop_subagent_role.py \
  tests/test_subagent_direct_chat.py tests/test_subagent_workdir.py tests/test_agent_loop_web_tools.py \
  tests/test_cli_playbook_commands.py tests/test_config_live.py -q -p no:randomly
1471 passed
uv run pytest tests/test_acp_methods.py tests/test_acp_config_options.py tests/test_subagent_probe.py \
  tests/test_subagent_third_party.py tests/test_rpc_subagents.py -q      # the two fixes: 581 passed
uv run pytest tests/test_rpc_subagents_model.py tests/test_subagent_manager.py tests/test_provider_pool.py \
  tests/test_provider_wire_model.py tests/test_subagent_acp.py tests/test_rpc_subagents.py tests/test_acp_methods.py \
  tests/test_acp_config_options.py tests/test_subagent_probe.py tests/test_subagent_third_party.py -q
1345 passed                              # after the review round; 948 passed on the five files the rebase's auto-merge touched
make test-python                         # 24016 passed, 109 skipped (the full unit suite)
make lint-python                         # clean
```

Web and TUI:

```
npm run gen:check --prefix ui-web        # generated.ts matches the contract (189 methods)
npm run type-check --prefix ui-web       # clean
npm run lint --prefix ui-web             # 0 errors (5 pre-existing warnings in other domains)
npm test --prefix ui-web                 # 190 files, 2550 tests passed (after the rebase onto the current base)
node ui-web/scripts/check-class-namespace.mjs   # OK, extAgents pinned at 0 / 0
npm run lint:rpc --prefix ui-tui         # in sync
npm run lint:i18n --prefix ui-tui        # up to date
```

Served page (a test gateway from this branch, an isolated RAVEN_HOME copied from a real one): roster of 5 connected / 4 available / 9 not installed with display names; Raven-Research's pill lists the 55 choices its handshake advertised, a pick lands on the row and reads back from `subagents.list`; the built-in row's pick is stored as `openrouter/anthropic/claude-opus-5` and clears back to null; the three products whose snapshots predated the menu key were re-verified at startup and got their menus.

Live regression on the same gateway: a product row switched off and back on through the row control (the readiness ping relaunches it); MiroThinker connected with a key typed in the sheet; an absent row's install block, copy and re-check ("still not found"); a pick and a clear on Raven-Code's own menu. Then, with the session on DeepSeek and Raven-Research's row set to `deepseek/deepseek/deepseek-v4-flash` from its menu, one direct-chat turn to a Raven-Research instance: the host sent `session/set_config_option {configId: model}` with that value (acp frame journal), the child logged `Session model: deepseek/deepseek-v4-flash` and ran the turn on it (`Iteration 1/60 model=deepseek/deepseek-v4-flash`), and answered in 15 s. Connecting Claude Code, Codex and Pi from the row each ended in the server's refusal shown red on the row with Retry -- the machine's Claude Code CLI is too old for the adapter's model, Codex is not logged in, Pi's executable is not installed -- so their Test buttons were not reached.

## Risk

- Behaviour change: the agents page is redrawn; a built-in row's `model` in config (previously masked by the turn binding on every dispatch) now takes effect, paired with its own credential, and falls back to the conversation's model when no credential can serve it (one warning in the log). An acp row's `model` is new and off by default.
- The Pi row now reads "not installed" on a machine without the `pi` executable, where it read "available" before; a hand-written row is not held to this, only a row carrying the preset's provenance.
- An acp row whose capability snapshot predates the model menu reads "attention" with a "run a test" detail until the boot-time re-verify or a manual Test records the menu; before, it read the old "ready" with a disabled model pill. A failed manual Test no longer drops a row's recorded menu and statefulness.
- A built-in row's bare stored id (written by hand before this branch) is now paired by keyword on read and re-stored prefixed on its next write; a pick under a spec-less section is accepted only for an id that section lists. A hand-written id whose prefix names no provider section (`deepseek-ai/DeepSeek-V3`) keeps its provider derived by the pool -- the configured gateway, as such an id always ran.
- Rollback: revert the PR; the new config fields (`model` on an acp row, a built-in override row) are ignored by the previous code.

## Related Issues

N/A


ui-web/src/features/extAgents/catalogue.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (byPreset) return byPreset
+  return Object.hasOwn(CATALOGUE, row.name) ? CATALOGUE[row.name] : undefined
+}
+
+/* One of Raven's own, whichever way this install registered it. The wire flags
+   cover a built-in row and a folder discovered under `agents/`; an install that
+   registered a shipped product as a config row (its `install.py` does that)
+   carries neither flag, and the row is still Raven's. */
+export function isOwnRow(row: ExtAgentRow): boolean {
+  return !!row.builtin || !!row.vendored || entryOf(row)?.by === OWN_BY
```
gloryfromca: Blocking: honor the server-provided ownership flag when classifying rows.

`subagents.list` now sends `own` specifically for configured ACP rows whose handshake identifies Raven, including rows whose user-editable name is not in this catalogue. `extAgentRowOf` preserves it, but this predicate ignores it. A renamed or otherwise uncatalogued Raven row therefore gets generic branding, sorts with third parties, and labels its unset model as the agent default instead of following the main Raven. Please include `row.own` in this decision, retaining any old-server fallback that is still needed.

ui-web/src/features/extAgents/ExtAgentsPage.tsx
```diff
@@ -48,421 +49,528 @@ const kindText = (kind: string): string =>
...
+          </button>
+        ) : null}
+      </span>
+      {open ? (
+        <ModelPicker
+          anchor={pill.current}
+          current={shown ? { model: row.model_source === 'agent' ? row.model || '' : shown.id, provider: shown.provider } : null}
+          emptyNote={t(row.model_source === 'raven' ? 'gui.agent.model_no_provider' : 'gui.agent.model_managed')}
+          onClose={() => setOpen(false)}
+          onPick={(model, provider) => {
```
gloryfromca, line 427: Blocking: prevent the agent picker from offering model values that its save path cannot honor.

The shared `ModelPicker` always offers the typed search text when it is not an exact option, but this callback drops its `typed` argument and sends the value straight to `subagents.update`. For an ACP row, every unadvertised value is rejected because the server accepts only exact `model_choices`, so the visible custom row (and Enter on it) can never succeed. For a built-in row the same control says it will add the model to the provider, but this path never performs the provider `add_model` write. Constrain typed picks here or implement the persistence promised by the shared picker.

ui-web/src/features/extAgents/ExtAgentsPage.tsx
```diff
@@ -48,421 +49,528 @@ const kindText = (kind: string): string =>
...
+function shownModel(row: ExtAgentRow): { id: string; by: string; provider: string } | null {
+  if (!row.model) return null
+  if (row.model_source === 'agent') {
+    const hit = (row.model_choices || []).find((c) => c.value === row.model)
+    return { id: hit?.name || row.model, by: hit?.group || '', provider: hit?.group || row.name }
+  }
+  const cut = row.model.indexOf('/')
+  const head = cut > 0 ? row.model.slice(0, cut) : ''
+  const known = hostProviders().find((p) => p.id === head)
+  return { id: cut > 0 ? row.model.slice(cut + 1) : row.model, by: known ? known.name : head, provider: head }
```
gloryfromca: Blocking: normalize stored provider prefixes before deriving the picker's current selection.

`stored_model_id` persists public routing prefixes such as `openai-codex`, `github-copilot`, and `minimax-global`, while `model.options` keys those providers by config slugs such as `openai_codex`. Comparing the raw head here leaves `known` undefined and passes an unmatched provider to `ModelPicker`; reopening the menu selects its first provider and shows no current tick, while the pill prints the raw prefix rather than the provider label. Use the canonical provider-name conversion/shared wire helper before the lookup.

ui-web/src/features/extAgents/ExtAgentsPage.tsx
```diff
@@ -48,421 +49,528 @@ const kindText = (kind: string): string =>
...
+      </button>
+    )
+  } else if (shown === 'on') {
+    actions = row.builtin ? null : (
+      <>
+        <button className="mini danger" onClick={() => store.disconnectRow(row)}>
+          {t('gui.agent.disconnect')}
+        </button>
+        {canTest(row) ? (
+          <button className="mini" disabled={testing} onClick={() => void store.runTest(row)}>
```
gloryfromca: Blocking: keep a stop action available while an agent test is running.

This redesign leaves `subagents.test_cancel` and `store.stopTest` in place but removes their only UI call site: while `testing` is true, the sole Test button is disabled. A CLI test may run for 120 seconds and spend the agent's quota, so after starting one a web user cannot cancel it; closing the sheet does not invoke the endpoint. Restore a Stop control for the running state.

ui-web/src/features/extAgents/catalogue.test.ts
```diff
@@ -0,0 +1,68 @@
+import { beforeEach, describe, expect, it } from 'vitest'
```
gloryfromca, line 5: Blocking: add the required file-level purpose documentation to this new test module.

AGENTS.md section 1.1 requires every new code file to document its purpose in English with a module docstring or equivalent file-level documentation. This new module begins directly with imports, so it does not satisfy that hard repository rule.

gloryfromca (commented):
Blocking: the agent picker and test sheet still have user-visible paths that cannot behave as presented.

I found four functional blockers and one repository-rule blocker, marked inline. I covered the full merge-base diff, the extAgents callers and prior implementation, model dispatch through spawn/direct-chat/DAG/playbook, generated wire compatibility, commit history, AGENTS.md, CONTEXT-MAP.md, ui-web/CONTEXT.md and CONTRIBUTING.md, and the rewritten tests for removed coverage.

Verification: `uv run pytest` over the PR's 16 backend suites: 1471 passed; `npm test` in ui-web: 188 files / 2501 tests passed; focused extAgents suite: 5 files / 86 tests passed; type-check and generated-client check passed; ESLint completed with 0 errors and 5 warnings in untouched domains; class-namespace and large-file checks passed.

gloryfromca (commented):
Blocking: the five previously reported findings still need to be addressed.

The sole change since my previous review translates the stylesheet comment into English. That fixes the source-language gate failure, but it does not modify the implementation or test files behind any of the five open threads. Direct inspection confirms that all five findings still stand; I found no additional defect in this comment-only delta.

Verification:
- `uv run python scripts/check_source_language.py github/refactor/ui_web_architecture...HEAD` — passed.
- `npm test -- --run src/features/extAgents` — 5 files and 86 tests passed.
- `npm run type-check` — passed.

Coverage: I reviewed the new delta and commit history, rechecked the existing findings against the current implementation and callers, and carried forward the prior full-diff review of project rules, backward compatibility, architecture boundaries, and test integrity. No test was weakened in this revision. The five threads remain open because none is settled.

raven/rpc/methods/subagents.py
```diff
@@ -450,45 +463,136 @@ async def subagents_add(params: dict, *, agent_loop_factory: "AgentLoopFactory |
...
     if preset_name and preset_name in THIRD_PARTY_SUBAGENT_PRESETS:
         return third_party_subagent_preset(preset_name)["description"]
-    return ""
+    from raven.agent.subagent.vendored_agents import discover_product_rows
+
+    discovered = next((cfg for cfg in discover_product_rows() if getattr(cfg, "name", None) == name), None)
+    return (getattr(discovered, "description", "") or "") if discovered is not None else ""
+
+
+def _host_pair(model: str) -> str | None:
```
LivXue, line 497: Both halves of this PR read a stored id back by resolving the provider with `find_by_model`. This one also compares the whole incoming id against each section's bare `models` list -- and with `provider` sent the id was already prefixed at `:637` (`stored_model_id("mylocal", "my-model")` -> `"mylocal/my-model"`), so a section holding `models: ["my-model"]` can never match it.

The picker does offer exactly that pick: `_listable_provider_names` adds config extras on purpose and `_build_provider_entry` handles `spec is None`, so a user reaches this refusal -- and it blames their model and their key for a pick that is fine.

The docstring promises the check mirrors "the pairing the dispatch will make (`ProviderPool.bind_pin` with the provider the id names)". The dispatch half does not pass the name (see the comment on `_row_pin` in `raven/agent/subagent/manager.py`). Resolving through the section lookup here, and comparing the un-prefixed tail against `section.models`, would close this half.

raven/agent/subagent/manager.py
```diff
@@ -78,6 +79,25 @@
...
+    provider it was picked under, and the pool handed only the id would let a
+    configured gateway take the pin instead (``ProviderPool.bind_pin``) -- the
+    reader's credential choice, silently swapped for another bill.
+    """
+    from raven.providers.registry import find_by_model
+
+    model = getattr(config, "model", None)
+    if not model:
+        return None, None
+    spec = find_by_model(model)
```
LivXue: `_row_pin` returns `(model, None)` whenever `find_by_model` misses -- which is exactly the id `_host_pair` stores for a section no spec matches. `bind_pin` then takes its gateway branch (the pin binds under whatever credential the configured gateway names, the swap this docstring says the pair exists to prevent), or, with no gateway, `_resolve_provider_name` answers `"auto"` and `_has_credentials("auto", ...)` is False by construction, so the pin is dropped and the row runs the conversation's model.

Splitting the stored id at the first `/` and looking the head up as a provider section when `find_by_model` misses would close this half; either way, both halves should agree on one resolver.

raven/agent/subagent/manager.py
```diff
@@ -397,14 +422,29 @@ def build_builtin_backend(self, row: "AgentRow", build: Any = None) -> "RavenLoo
...
+        The row's own ``model`` is a pin the backend resolves per dispatch and
+        pairs with its own credential through the pool (``live_pin_resolver``),
+        never a value baked in here: this backend is cached across bindings, so
+        a model fixed at construction would be whichever one the manager
+        happened to be on when the row was first dispatched, and a bare id has
+        no key of its own to be sent with. Unusable, the row follows the
+        conversation's binding.
         """
         confine = getattr(row.config, "restrict_to_workspace", None)
+        pin = live_pin_resolver(
```
LivXue, line 446: The warning this resolver emits fires once per resolver, and the closure outlives the config that produced it: a second, different unusable pin logs nothing at all. Keying the once-guard on the pin tuple would keep one line per problem instead of one per backend.

raven/agent/subagent/backends/raven_loop.py
```diff
@@ -324,6 +332,14 @@ async def _run(
         # for callers that drive a backend directly.
         provider = provider or self.provider
         model = model or self.model
+        # The row's own model, when it has one, over the pair the dispatch
+        # brought: a per-agent model is a fact about this agent, and the turn's
+        # binding is the fallback it was always meant to be. Read per run
+        # because this backend is cached across bindings; a pin that resolves
+        # to nothing usable leaves the pair alone (see `live_pin_resolver`).
+        pinned = self._pin() if self._pin is not None else None
```
LivXue, line 340: The pin is authoritative here, and its failure is invisible: the row still reports the stored id with `model_source: "raven"` (`subagents.py:346`), so the pill shows a model the turns do not run. Whether the pin resolved is worth carrying on the row.

raven/rpc/methods/subagents.py
```diff
@@ -450,45 +463,136 @@ async def subagents_add(params: dict, *, agent_loop_factory: "AgentLoopFactory |
...
+    a write has no reason to wait on.
+    """
+    from raven.config.loader import load_config
+    from raven.config.schema import section_has_credentials
+    from raven.providers.registry import find_by_model, find_by_name
+    from raven.providers.wire import stored_model_id
+
+    try:
+        providers = load_config().providers
+    except Exception:  # noqa: BLE001 - a config that cannot be read serves nothing
```
LivXue: This renders "the config could not be read" as "your model or key is wrong": `PermissionError`, a malformed config and a loader failure all land here, none logged, while the caller's message asserts a fact about the user's model and credential. Log the exception, and treat an unreadable config as a server-side error rather than a field-validation one.

raven/agent/subagent/probe.py
```diff
@@ -502,7 +502,7 @@ async def _test_acp(cfg: Any, *, source: Source, elapsed: Any) -> TestResult:
     from raven.acp_client.capabilities import SnapshotStore, verify_agent
 
     snapshot = await verify_agent(cfg)
-    if source == "config":
+    if source != "preset":
```
LivXue: Recording for everything but `preset` is what a vendored row needs. Note that the record below this line is unconditional: a manual Test that fails on a flaky machine replaces a good snapshot and costs the row its menu and its statefulness until a later success. The verdict stays visible; the loss of the previous record does not.

raven/agent/subagent/probe.py
```diff
@@ -565,14 +565,23 @@ async def _verify_missing_snapshots(manager: Any, rows: list[Any]) -> None:
...
+    has_model_menu = getattr(store, "has_model_menu", None)
     recorded = False
     for row in rows:
         cfg = getattr(row, "config", None)
         if cfg is None:
             continue
         try:
             snapshot = acp_snapshot_for(cfg)
-            if snapshot is not None and not snapshot.stale:
+            outdated_menu = (
```
LivXue, line 664: When the re-verify does not return ready, nothing is recorded and the old snapshot -- the one with no `modelChoices` -- survives, so `_probe_acp` reports its stored `ready` and the sheet draws a disabled pill for an agent that does offer a menu. The INFO line below is the only trace.

`has_model_menu` already tells "never measured" from "measured, none"; carrying that distinction on the wire is what would make it visible.

raven/rpc/models.py
```diff
@@ -147,6 +147,16 @@ class SubagentRow(_Strict):
             "not offer a switch, a test or a delete for it."
         ),
     )
+    own: bool = Field(
```
LivXue, line 150: Added to the wire, read once (`source.ts:57`), consumed by nothing: ownership is decided client-side by `isOwnRow` (`ui-web/src/features/extAgents/catalogue.ts:108`). Either render the pill's unset label from this field, or drop it -- the server is the one that knows a shipped product was registered as a plain config row, which is the case `isOwnRow` approximates from the catalogue.

ui-web/src/features/extAgents/store.ts
```diff
@@ -111,13 +143,100 @@ export async function run(op: ExtAgentOp, row?: ExtAgentRow, args?: ExtAgentActA
...
+/* What this agent is good at, as the reader words it. A row that exists takes
+   it as a write; a preset holds it until the connect that writes the row. */
+export function describe(row: ExtAgentRow, text: string): void {
+  if (row.configured || row.vendored) void act(row, 'update', { description: text })
+  else set({ drafts: { ...get().drafts, [row.name]: text } })
+}
+
+/* The row's own model, or back to the agent's default. A refusal is toasted
+   rather than held on the row: the row is not broken, the pick was. */
+export function setModel(row: ExtAgentRow, model: string, provider?: string): Promise<string | null> {
```
LivXue: This bypasses `act`, so a refusal on the model op is a transient toast while every other verb holds its failure on the row (`act`, `:167`) -- and the description promises the held form. Same shape as the review's finding 2: an acp agent that refuses the push is re-attempted on every turn.

ui-web/src/features/extAgents/source.ts
```diff
@@ -54,6 +54,12 @@ export function extAgentRowOf(r: ExtAgentRowWire): ExtAgentRow {
     last_test_detail: r.last_test_detail || '',
     last_test_at_ms: r.last_test_at_ms || null,
     test_running: !!r.test_running,
+    own: !!r.own,
```
LivXue, line 57: The wire `own` reaches the client here and stops: nothing reads the client field either, since the page's ownership comes from `isOwnRow` (`catalogue.ts:108`). See the comment on `SubagentRow.own` in `raven/rpc/models.py`.

ui-web/src/features/extAgents/catalogue.ts
```diff
@@ -0,0 +1,130 @@
...
+ * the prompt text the dispatching model reads, and `probe_detail` is a verdict,
+ * not an introduction.
+ *
+ * Keyed by preset first and by name second. A preset is the package behind a
+ * row and survives a rename; Raven's own agents carry no preset, and their
+ * names are the manifest's and not the reader's to change, so the name is the
+ * stable key there. A hand-written row matching neither gets nothing, and the
+ * page falls back to what the wire does say.
+ *
+ * The install data is the vendor's own documentation (sources are listed in
```
LivXue: "sources are listed in the PR that added them" points at something a reader of the tree cannot open. The commands themselves are worth having here; the pointer is not a source. The other half of this sentence -- that they are the commands the presets hint at -- is exactly the kind of provenance worth keeping.

LivXue (changes requested):
Request changes

Thanks -- the page reads well, and the lane work behind the pill is the right shape: one `session_model_for` for a spawn, a chat and a DAG node, and the acp-only keyword guarded by signature rather than by a kind check. Two findings block, both on the built-in row's model path; the two after them should be fixed or filed. Everything below was read at the head commit; `raven/providers/pool.py` is unchanged by this PR, so its behaviour is taken as given.

## Blocking

**1. A built-in row's stored model and the model the dispatch runs can disagree, silently.** `subagents.update` stores a provider-qualified id, and both halves this PR adds read it back by resolving the provider with `find_by_model` alone -- which fails for an id whose prefix names a section raven carries no spec for. The write side then refuses the picker's own pick with a message blaming the model or the credential, both of which are fine; the bare-id route stores the id, and every dispatch afterwards runs the conversation's pair (or the gateway's key) while the row keeps reporting the stored id. Nothing on the row or the wire says the pin was dropped, and the one log line fires once per resolver, so a later, different unusable pin logs nothing at all. Fix direction: one resolver on both sides (see the inline comments on `_host_pair` and `_row_pin`), a refusal that names its actual cause, and ideally a row field for "the pin resolved".

**2. An acp model push the agent refuses is retried every turn, and stays on the row as if it landed.** `_set_model` runs on every route into a session, so a permanent refusal is re-attempted as if transient, one warning per attempt with no bound. The row (`subagents.list`) and the `instances` report keep showing the pick, and the refusal is a transient toast -- not the held red text with Retry the description promises. Fix direction: the same held-failure treatment the other verbs get, plus a row field (`model_push_error`, or `model_in_effect`); failing that, log once per problem rather than once per turn.

## Important

**3. A config that cannot be read is reported as a credentials problem** (`raven/rpc/methods/subagents.py:511`). The bare `except Exception` returns None, which the caller renders as "it names no provider raven knows, or that provider has no usable credentials". A `PermissionError`, a malformed config, a loader failure: none of it is logged, and the user is told their model or key is at fault. An unreadable config is a server-side error, not a field-validation one.

**4. The menu backfill can leave a "ready" row with an unmeasured menu, and a failed Test overwrites a good snapshot.** `_verify_missing_snapshots` (`raven/agent/subagent/probe.py:579`) records only on `ready`; when the re-verify does not return ready, the old snapshot -- the one with no `modelChoices` -- survives and `_probe_acp` reports its stored `ready`, so the sheet draws a disabled pill and "managed by itself" for an agent that does offer a menu, with an INFO line as the only trace. Same area: `_test_acp` records unconditionally (`probe.py:509`), so a manual Test that fails on a flaky machine replaces a good snapshot and costs the row its menu and its statefulness until a later success. `has_model_menu` already tells "never measured" from "measured, none"; a wire field carrying that distinction would be the durable fix.

## Nits

- `ui-web/src/features/extAgents/manifest.ts` (untouched here) still says the domain's detail-card button is `.xaedit` and that the class prefix "waits on the CSS step". This PR is that step: five `page.css` rules gone, prefix renamed, `LEGACY_LOCAL.extAgents` down to 0.
- `ui-web/CONTRIBUTING.md`: the `LEGACY_BORROWED` row still reads "one of the eleven, `warn`, is SetupSheet.tsx..." and "11 in the shared components", but this PR's ledger edit moved `warn` off that list (components 11 -> 10, `LEGACY_CHROME_EXPR` 10 -> 11). That row's totals look stale independently (the two lists hold 23 names; the row says 34) -- worth re-deriving in the same edit.
- `ui-web/src/styles/page.css:4143`: the specificity comment still cites `.sukey input`, whose rules this PR deleted (6 occurrences at base, 1 at head, and it is the comment).
- `raven/rpc/models.py:150` (`own`): added to the wire, read once (`source.ts:57`), consumed by nothing -- ownership is decided client-side by `isOwnRow` (`catalogue.ts:108`). Either render from the field or drop it; the server is the one that knows a shipped product was registered as a plain config row. The description's "Unset reads by ownership (`own`, a new wire field)" describes a read the code does not make.
- `ui-web/src/features/extAgents/catalogue.ts:13` and `styles.css:17`: both point at something a reader of the tree cannot open -- "sources are listed in the PR that added them", and a `my_docs/temp/.../proto_app.css` path. The prototype reference is useful provenance; the PR pointer is not a source.
- `ui-web/src/features/extAgents/store.ts:304`: `stopTest` has no caller.

## Process

- The base is `refactor/ui_web_architecture` (#475, still open). Please say so in the description, or land #475 first -- otherwise the diff a reviewer sees is not the diff that merges.
- Second description/code mismatch, alongside `own`: "a refusal stays on the row as red text with Retry" does not hold for the model op (finding 2).

## What is good here

- The key removal is clean: 17 i18n keys gone (34 leaves with the en/zh pairs), with no dangling reference left in either front end.
- The class-namespace ledger moves are consistent and self-enforcing -- the pin assertion (`check-class-namespace.mjs:859`) fails the next PR that leaves a pin behind.
- The lane unification is the right shape, and the acp-only keyword travels only to a `run` that declares it or takes `**kwargs`, so a backend written against the older paper is not handed it.
- The contract artifacts (`openrpc.json`, both `generated.ts`, the JSON schema) move in the same change as the models they describe.
- The verification section shows the acp path exercised end to end against a live gateway. The built-in path is where it stops short: the pick recorded there is `openrouter/...`, whose prefix `find_by_model` resolves, so the spec-less section looks like the case neither the tests nor the live run covered.

gloryfromca (commented):
Blocking: the five previously reported findings remain unaddressed.

Neither new commit modifies the implementation or test files behind my five open threads, and direct inspection confirms that each still stands. I found no additional blocker in the ACP session-model or shim-probe delta under the fourth-round bar.

Named nonblocking follow-up: `SHIM_REQUIRED_EXECUTABLES` now requires a PATH-level `claude` for the Claude Code preset, but the pinned `@agentclientprotocol/claude-agent-acp@0.66.0` resolves its native Claude CLI from the platform-specific optional dependency of `@anthropic-ai/claude-agent-sdk`. A Node-only installation can therefore run the adapter while this probe reports it missing. This change introduces the false negative and it is reachable in ordinary use, but the displayed global-install command is a working escape hatch, so it does not meet the fourth-round blocker threshold.

Verification:
- `uv run pytest tests/test_acp_config_options.py tests/test_acp_methods.py tests/test_subagent_probe.py tests/test_subagent_third_party.py -x`: 475 passed on Python 3.12.3.
- `npm test -- --run src/features/extAgents`: 5 files and 86 tests passed.
- Source-language, large-file, and `git diff --check` gates: passed. The environment has no `make`, so I ran the large-file target command from the Makefile directly.

Coverage: reviewed the full revision diff and the two-commit delta, surrounding callers and history, AGENTS.md / CLAUDE.md / CONTEXT*.md rules, backward compatibility, Runtime/Web UI architecture boundaries, and whether tests were weakened. The new tests add coverage rather than weakening existing assertions. Existing threads remain open because none is settled.

ui-web/src/features/extAgents/ExtAgentsPage.tsx
gloryfromca: Blocking: the catalogue test module still needs the required file-level purpose documentation.

Fixed here: while a test is running, the action bar now replaces Test with an enabled Stop button wired to `store.stopTest` / `subagents.test_cancel`. The interaction test exercises the cancellation call. Closing this thread.

ui-web/src/features/extAgents/ExtAgentsPage.tsx
gloryfromca: Blocking: the catalogue test module still needs the required file-level purpose documentation.

Fixed here: the stored public prefix is compared in normalized slug form and the matched provider id is returned to the picker, restoring both the provider label and current-selection tick. The hyphenated `openai-codex` case is covered. Closing this thread.

ui-web/src/features/extAgents/ExtAgentsPage.tsx
gloryfromca: Blocking: the catalogue test module still needs the required file-level purpose documentation.

Fixed here: this picker passes `allowTyped={false}`, and the shared picker now suppresses both the custom row and Enter fallback when typed values are disabled. The focused tests cover both paths. Closing this thread.

ui-web/src/features/extAgents/catalogue.ts
gloryfromca: Blocking: the catalogue test module still needs the required file-level purpose documentation.

Fixed here: `isOwnRow` now consumes the server-provided `row.own` flag while retaining the catalogue fallback for older servers, and the renamed uncatalogued Raven case is covered. Closing this thread.

gloryfromca (commented):
Blocking: add the required file-level purpose documentation to ui-web/src/features/extAgents/catalogue.test.ts.

Four of my five original findings are fixed; I replied to and resolved those threads. The remaining new test module still begins directly with imports, so it does not satisfy AGENTS.md section 1.1. At this fifth-round bar: the PR introduces the file, the violation is unconditional rather than dependent on a rare runtime combination, and the repository explicitly treats this as a hard rejection rule with no compliant workaround other than adding the file-level documentation.

Named nonblocking follow-ups:
- The previously reported Claude adapter probe false-negative remains unchanged.
- The new `stored_provider_name` resolver can reinterpret a pre-existing built-in override such as `deepseek-ai/DeepSeek-V3`, which previously ran through the configured gateway, as an explicit `deepseek_ai` provider and drop the pin. Reselecting the model under the gateway stores `openrouter/deepseek-ai/DeepSeek-V3`, so operators have a working recovery and this does not meet the late-round blocker threshold.

Verification:
- `uv run pytest tests/test_provider_pool.py tests/test_provider_wire_model.py tests/test_rpc_subagents_model.py tests/test_subagent_acp.py tests/test_subagent_manager.py -x`: 764 passed.
- `npm test`: 188 files and 2505 tests passed.
- `npm run type-check` and `npm run gen:check`: passed.
- `npm run lint`: passed with 0 errors and 5 pre-existing warnings on untouched files.
- Source-language, large-file, class-namespace, and `git diff --check` gates: passed.

Coverage: reviewed the full target diff and this four-commit delta, surrounding callers and history, AGENTS.md / CLAUDE.md / CONTEXT*.md rules, backward compatibility, Runtime/Web UI architecture boundaries, and test integrity. The added tests strengthen coverage; no test was weakened.

gloryfromca (commented):
Blocking: add the required file-level purpose documentation to ui-web/src/features/extAgents/catalogue.test.ts.

The new delta correctly stops treating a globally installed `claude` executable as a prerequisite for `claude-agent-acp`, and its regression test matches the adapter packaging contract. The remaining blocker is unchanged: this PR-added test module still begins directly with imports, contrary to the mandatory file-level purpose documentation in AGENTS.md section 1.1. At this sixth round it meets all three blocking criteria: this PR introduced the file, every normal checkout contains the violation, and there is no compliant workaround without changing the file.

The previously named legacy configured-gateway model-id compatibility case remains a nonblocking follow-up.

Verification: `uv run pytest tests/test_subagent_third_party.py tests/test_subagent_probe.py -x` (293 passed); source-language checker passed; large-file checker passed via its underlying `uv` command because `make` is unavailable; `git diff --check github/refactor/ui_web_architecture...HEAD` passed. I reviewed the revision delta and target diff context, affected callers and history, project rules and domain context, backward compatibility, architecture boundaries, and whether tests were weakened.

ui-web/src/features/extAgents/catalogue.test.ts
0xKT: Fixed in 79196f4a: the module now opens with its file-level purpose comment (what the catalogue says about a row beyond the wire, keyed by preset then name, and the ownership rule).

raven/rpc/methods/subagents.py
0xKT: Fixed in 35bf0f26 (plus 79196f4a). providers.wire.stored_provider_name is the inverse of stored_model_id and both sides read the provider through it: a spec'd provider by its route names, a spec-less section by the prefix the write put there. _host_pair stores every accepted id prefixed, and a spec-less section serves only the ids it lists, compared on the un-prefixed tail. The refusal for an unreadable config is gone (see the thread at :512). Tests: test_rpc_subagents_model.py test_update_accepts_a_pick_under_a_section_no_spec_matches_when_sent_with_its_provider and test_update_rejects_an_id_a_section_no_spec_matches_does_not_list -- note that "custom" carries a spec in the registry, so the tests use an extra section ("mylocal").

raven/agent/subagent/manager.py
0xKT: Fixed in 35bf0f26 and 79196f4a. _row_pin reads the provider with the same stored_provider_name the write used, so a spec-less section resolves to that section and the pool binds under its credential. 79196f4a adds the other direction: handed the pool's provider table, a prefix that names neither a spec nor a section (deepseek-ai/DeepSeek-V3 written by hand) is part of the id, the pin goes to the pool with no provider, and the gateway branch serves it as before. Tests: test_subagent_manager.py test_a_builtin_rows_pin_under_a_section_no_spec_matches_names_that_section and test_a_prefix_that_names_no_section_is_left_to_the_pool_to_derive; a round trip over every registered provider in test_provider_wire_model.py.

raven/agent/subagent/manager.py
0xKT: Fixed in 35bf0f26: the once-guard in live_pin_resolver is keyed on the (model, provider) pin, so a later, different unusable pin gets its own line and the same one still gets one. Test: test_provider_pool.py test_a_later_different_unusable_pin_is_reported_on_its_own.

raven/agent/subagent/backends/raven_loop.py
0xKT: Not changed here, and being weighed with the owner as a feature rather than a fix: the stored pick is validated at write time on both kinds (_host_pair against the credential, the acp menu against the handshake), so the pin and the dispatch only drift after a later config or agent change; the same silent fallback serves the other three pinned subsystems (context engine, skill forge, the settings role pins) with no wire signal either; and the acp half is live per-session state with no defined row-level meaning yet. A row field for "in effect" is the durable answer and is proposed as its own issue. The narrower halves are fixed here: the resolver mismatch that dropped a stored pick (35bf0f26, 79196f4a), the per-pin warning, and the once-per-value refusal log.

raven/rpc/methods/subagents.py
0xKT: Fixed in 35bf0f26: _host_pair no longer swallows the exception. An unreadable config raises out of the handler and is logged by the dispatcher as the server's failure, so the reader is not told their model or key is wrong. Test: test_rpc_subagents_model.py test_host_pair_lets_an_unreadable_config_raise_rather_than_blame_the_pick.

raven/agent/subagent/probe.py
0xKT: Fixed in 79196f4a and 5bc8c6aa. _test_acp records the verdict always and the capabilities only when the verify reached them: an unusable result is written as the previous record's capabilities under the new status, detail, timestamp and fingerprint (probe.py _test_record), the same trust SnapshotStore.load already extends to a stale row -- and the previous record is read with allow_stale, so a failed Test right after a config edit keeps what the roster was trusting instead of erasing it. A row whose login really died still shows the failed verdict; a session it cannot load fails at session/load and starts fresh. Tests: test_subagent_acp.py test_a_failed_manual_test_keeps_the_capabilities_the_last_record_measured, test_a_failed_test_after_a_config_edit_keeps_the_capabilities_the_roster_was_trusting, test_a_first_test_that_fails_is_recorded_as_it_is.

raven/agent/subagent/probe.py
0xKT: Fixed in 79196f4a, on the probe side rather than by recording the failed re-verify (which would turn a flaky boot connect into a red row, and -- before the to_row change below -- would have written a modelChoices key that ends every later backfill retry). CapabilitySnapshot now knows whether its menu was measured (model_menu_measured, set from the row's keys at load, written as the key's absence so a verdict re-recorded over an old row does not mint a measured-empty menu), and _probe_acp reports such a row as attention with "its model menu has not been measured yet -- run a test" instead of the stored ready; attention keeps the row in the installed group. The boot backfill still re-measures it. Residual, accepted: the pill beside that status still reads "managed by itself", since the page cannot tell an unmeasured menu from a measured-empty one without the wire field discussed at raven_loop.py:340. Test: test_subagent_acp.py test_a_row_recorded_before_the_menu_reads_attention_until_it_is_measured.

raven/rpc/models.py
0xKT: Fixed in 47493932: isOwnRow reads row.own first, with the catalogue as the fallback for a server that predates the field, so a renamed acp row whose handshake named Raven sorts and words itself as Raven's. Test: catalogue.test.ts "trusts the wire's own flag for a Raven row this catalogue does not know". The description's sentence on ownership now describes the read the code makes.

ui-web/src/features/extAgents/store.ts
0xKT: Fixed in 47493932 and 18994e39. setModel / clearModel go through act, so a refused pick is held on the row in red with Retry and is not toasted (store.test.ts "holds a refused pick on the row with the pick to retry"). On the agent side, a refused push is said once per value rather than once per turn (acp_agent._say_model_refused; test_subagent_acp.py test_a_refused_model_is_said_once_per_value_not_once_per_turn). Whether the push landed is not yet on the row; see the thread at raven_loop.py:340 for the follow-up.

ui-web/src/features/extAgents/source.ts
0xKT: Fixed in 47493932; the field is read by isOwnRow now (see the thread on raven/rpc/models.py).

ui-web/src/features/extAgents/catalogue.ts
0xKT: Fixed in 5c70ce80: the pointer to the pull request is gone, the provenance sentence (the vendor's own documentation, the same commands the presets hint at) stays. The styles.css pointer to a local scratch path went with it.

0xKT:
Review round, status by finding.

LivXue:
- Blocking 1 (built-in stored model vs dispatch): fixed in 35bf0f26 and 79196f4a. One resolver (providers.wire.stored_provider_name) on the write and the dispatch side, the spec-less section tested on both, the unreadable-config refusal removed, the unusable-pin warning keyed per pin, and a prefix that names no section left to the pool to derive (the gateway case).
- Blocking 2 (refused acp push): the write refusal is held on the row with Retry (47493932); the per-turn re-attempt is said once per value (18994e39). A row field for "in effect" is not in this PR -- see Important 4.
- Important 3 (unreadable config reported as a credentials problem): fixed in 35bf0f26.
- Important 4 (menu backfill and failed Test overwriting a snapshot): fixed in 79196f4a and 5bc8c6aa. A failed manual Test records its verdict over the previous record's capabilities instead of erasing them (probe.py _test_record), and a row recorded before the menu existed reads attention with "model menu has not been measured yet" until a re-verify lands, instead of the stored ready (CapabilitySnapshot.model_menu_measured). The in-effect row field (findings 1 and 2) stays a proposed follow-up.
- Nits: manifest text, CONTRIBUTING ledger rows (re-derived from the tool: 125 / 15 / 23, warn on the expression list), page.css comment, the two pointers, stopTest's caller -- 5c70ce80 and 47493932.
- Process: the description states the base branch in its first paragraph; the model-op sentence now matches the code.

gloryfromca:
- The five first-round findings: four resolved by you; the catalogue test module's file-level comment is in 79196f4a.
- Nonblocking, Claude adapter probe: fixed in ac97e890 (claude_code is out of SHIM_REQUIRED_EXECUTABLES; the SDK pin carries the CLI).
- Nonblocking, legacy gateway-served id: fixed in 79196f4a (stored_provider_name takes the config's provider table; a prefix naming no section is part of the id and the pin goes to the pool's gateway branch, with tests on both sides).

Verification for the round: uv run pytest over the touched backend files: 1345 passed before the follow-ups, then 667 / 607 / 365 passed on the files each follow-up touches (test_subagent_acp.py, test_subagent_probe.py, test_rpc_subagents.py, test_subagent_manager.py, test_provider_wire_model.py, test_rpc_subagents_model.py, test_provider_pool.py); npm test in ui-web, 188 files / 2505 tests; type-check, lint (0 errors), gen:check, check-class-namespace and the TUI catalogue check all pass.

ui-web/src/features/extAgents/catalogue.test.ts
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed: the file now opens with an English purpose comment that documents the catalogue mapping and ownership precedence, satisfying AGENTS.md section 1.1. Closing this thread.

gloryfromca (commented):

The file-level purpose documentation now satisfies the remaining repository-rule blocker. This revision also closes both follow-ups I had named: the bundled Claude adapter is no longer rejected for lacking a global executable, and legacy gateway-served model ids remain on the pool derivation path while configured spec-less provider sections resolve explicitly. The snapshot updates preserve prior capabilities across a failed test without collapsing an unmeasured model menu into a measured-empty one. I found no new plain error in the delta, and every thread I opened is now resolved.

Verification: `uv run pytest tests/test_provider_wire_model.py tests/test_provider_pool.py tests/test_rpc_subagents.py tests/test_rpc_subagents_model.py tests/test_subagent_acp.py tests/test_subagent_manager.py tests/test_subagent_probe.py -x` (926 passed); `npm test -- --run src/features/extAgents/catalogue.test.ts` (6 passed); source-language, large-file, and `git diff --check` gates passed. This pass covered the revision delta and full target-diff context from the prior rounds, affected callers and history, project rules and domain context, backward compatibility, architecture boundaries, and test integrity.

raven/agent/subagent/backends/raven_loop.py
0xKT: Decision from the owner: this is a feature, not a fix -- the prototype the page was built to draws no "in effect" state, the same silent fallback serves the context engine, the skill forge and the settings role pins with no wire signal either, and a fresh pick is validated at write time on both kinds (credential for the built-in row, handshake menu for an acp row), so the stored model and the dispatch only drift after a later config or agent change. Deferred: not in this PR, and not tracked as an issue for now.

0xKT:
Owner's decision on the one open item: the row field for whether a stored model is in effect (LivXue findings 1 and 2, the raven_loop.py:340 thread) is a feature the prototype does not draw, so it is deferred -- not in this PR and not tracked as an issue for now. Every other finding in the round is fixed or answered on its thread.

raven/rpc/methods/subagents.py
LivXue: The prefix case is closed. Verified at `5bc8c6aa`: `stored_provider_name` is the inverse of `stored_model_id` -- a spec'd provider by its route names, otherwise the prefix the write put there -- `_host_pair` reads the provider through it, and a spec-less section is compared on the un-prefixed tail. I separately checked that `stored_model_id` does not double-prefix an id that already carries the provider, which is what the section branch would otherwise have done to `mylocal/my-local-model`. Ran `tests/test_rpc_subagents_model.py` on this head: the pick under a spec-less section and its negative both pass. Resolving.

raven/agent/subagent/manager.py
LivXue: Both halves now resolve through one function. Verified at `5bc8c6aa`: `_row_pin` reads `stored_provider_name(model, providers=pool.config.providers)`, and I traced that the two configs are the same one -- `load_runtime_config(None, None)` calls `load_config(None)`, which is what `_host_pair` calls -- so the write and the dispatch cannot disagree about the provider. I also confirmed the attribute chain holds on the real pool and not only on the stub: `ProviderPool.config` is a property returning `Config`, and the runtime injects `ProviderPool(lambda: load_runtime_config(None, None))`. The updated `test_the_manager_pairs_a_builtin_rows_model_through_the_pool` pins the change at `("vendor/pinned", "vendor")` rather than `None`. Resolving.

raven/agent/subagent/manager.py
LivXue: Fixed and verified at `5bc8c6aa`: `warned` is a set keyed on the `(model, provider_name)` tuple, so the same unusable pin still gets one line and a later, different one gets its own. `test_a_later_different_unusable_pin_is_reported_on_its_own` asserts exactly two lines for two distinct pins. Resolving.

raven/agent/subagent/backends/raven_loop.py
LivXue: I accept the deferral, and I want the residual risk on the record rather than the thread closed as fixed. Nothing in the code is wrong: a pick is validated at write time on both kinds, so the stored model and the pair the dispatch runs can only drift after a later config or agent change -- a credential removed, a gateway repointed. When that happens the row still reports the stored id with `model_source: "raven"` and the one warning line is the whole signal, which is the part I flagged. Taking the owner's decision that the prototype draws no such state, so this is a feature and not a fix. Not tracked as an issue: noting that as the state of the record, not as agreement that it needs no tracking. Resolving.

raven/rpc/methods/subagents.py
LivXue: The `except Exception` is gone. Verified at `5bc8c6aa`: `load_config()` is called unguarded, so a `PermissionError` or a malformed config raises out of `_host_pair` and the caller's field-validation refusal is no longer reached -- the reader is not told their model or key is wrong for a file the server could not read. `test_host_pair_lets_an_unreadable_config_raise_rather_than_blame_the_pick` pins it. I also checked what that raise costs on the dispatch side, since `_row_pin` now reads config where it previously only matched keywords: `bind_pin` on the next line already read the same config, so no new failure mode. Resolving.

raven/agent/subagent/probe.py
LivXue: Verified at `5bc8c6aa`. `_test_record` writes this test's verdict always and keeps the previous record's capabilities when the verify never reached them, and the previous record is read with `allow_stale=True`, so a failed Test right after a config edit keeps what the roster was trusting. I also confirmed the mechanism that keeps it safe in the other direction: `to_row()` deletes `modelChoices` when the menu was not measured, so a verdict re-recorded over an old row cannot mint a measured-empty menu, which would have ended every later backfill retry. Resolving.

raven/agent/subagent/probe.py
LivXue: Verified at `5bc8c6aa`. `model_menu_measured` is set at load from `"modelChoices" in row` and never persisted as a field of its own, so the distinction is carried without being minted, and `_probe_acp` reports such a row as attention with the run-a-test detail instead of its stored `ready`. That closes the disabled-pill case. Your choice not to record the failed re-verify is the better half of what I suggested and I am not asking for the other: a flaky boot connect turning a row red would be worse than the trace it replaces. Resolving.

raven/rpc/models.py
LivXue: The field is read now. Verified at `5bc8c6aa`: `isOwnRow` tests `!!row.own` ahead of `builtin`, `vendored` and the catalogue, so a renamed acp row whose handshake named Raven sorts and words itself as Raven's, with the catalogue as the fallback for a server that predates the field rather than as the rule. Resolving.

ui-web/src/features/extAgents/store.ts
LivXue: Verified at `5bc8c6aa`. `setModel` and `clearModel` go through `act`, so a refused pick is held on the row like every other verb instead of toasted, and the description's red text with Retry now holds for this op. The agent half is covered too: `_model_refused` says a refusal once per value and is discarded on success, so the re-assertion on every route in no longer produces a line per turn. The whether-it-landed half is the thread at `raven_loop.py:340`, deferred there. Resolving.

ui-web/src/features/extAgents/source.ts
LivXue: Same fix as the thread on `raven/rpc/models.py`: the field is read by `isOwnRow` now. Verified at `5bc8c6aa`. Resolving.

ui-web/src/features/extAgents/catalogue.ts
LivXue: Verified at `5bc8c6aa`: the pointer to the pull request is gone and the provenance sentence kept, and the `styles.css` pointer to a local scratch path went with it. Resolving.

0xKT:
Two more pre-existing defects found by the real-machine acceptance on this branch, fixed in faefa84e and 3d34c9d4: the Claude Code preset's adapter pin moves from 0.66.0 to 0.79.0 (0.66.0 bundles CLI 2.1.220, which refuses any recent Claude model with "version 2.1.251 or newer is required", so the row could not connect for such an account; 0.79.0 measured: handshake with resume/fork/load, six model choices, five modes, connect from the row in 9 s, a tool-using turn rendered with the same _meta.claudeCode.toolName / rawOutput / fence shapes the dialect reads), and a row connected from the page now records its capability snapshot at connect (record_capabilities, shared with the manual Test; only when the record is missing, stale or from before the menu), where before it stayed stateless and menuless until a Test or a restart. Tests: test_rpc_subagents.py test_add_records_the_capabilities_of_an_acp_row_that_answered and siblings, test_subagent_acp.py test_the_connect_records_what_a_test_would.

gloryfromca (commented):

The acceptance-test delta is sound. The Claude ACP preset now pins a published adapter whose SDK dependency matches the documented compatibility rationale, while retaining the measured dialect shapes. The add/enable path records capabilities only when the snapshot is missing, stale, or predates model menus; it shares the manual-Test writer, treats recording as best effort after a successful ping, and leaves concurrent config changes detectable through the existing fingerprint. I found no new plain error.

Verification: `uv run pytest tests/test_rpc_subagents.py tests/test_subagent_acp.py tests/test_subagent_probe.py tests/test_subagent_third_party.py tests/test_acp_dialects.py -x` (665 passed); Ruff passed on all touched Python files; source-language, large-file, and `git diff --check` gates passed. I also verified the npm package metadata for `@agentclientprotocol/claude-agent-acp@0.79.0`. This pass covered the new delta plus the previously reviewed target-diff context, affected callers and history, project rules and domain context, backward compatibility, architecture boundaries, and test integrity.

gloryfromca (commented):

This is a rebase onto target commit #570. Range comparison shows 16 of the 17 PR commits patch-identical; the only adjusted commit correctly retains the target-added `importSync: 0` class-namespace ledger entries alongside the reviewed agent-hub reductions. The current target has since gained #577 in overlapping live-activity code; `git merge-tree` reports a clean merge, and the merged callers preserve the target's record-address keying together with this PR's settled DAG accounts. I found no new plain error.

Verification: full UI suite 190 files / 2550 tests passed; UI type-check, generated-client check, class-namespace gate, source-language gate, large-file gate, and `git diff --check` passed. A synthetic merge with the current target passed 275 focused activity/tasks/manager tests. A broader synthetic-merge run was not fully green: two memory-poller tests failed because the disposable environment discovered no optional EverOS plugin; both exact tests passed (2/2) in the configured review environment. This pass covered the semantic rebase delta, current-target overlap, previously reviewed full diff, affected callers/history, project rules and domain context, backward compatibility, architecture constraints, and test integrity.

gloryfromca (approved):

gloryfromca (commented):

This revision merges the current target into the branch. Its tree (`5eabd6ec712f`) exactly matches the conflict-free synthetic merge reviewed on the preceding head, so it contains no additional hand-resolved or semantic delta. The target's record-address live-activity fix and this PR's settled DAG accounting remain composed as previously assessed.

Verification on the actual head: `uv run pytest tests/test_rpc_tasks.py tests/test_rpc_subagent_calls.py tests/test_subagent_activity.py tests/test_subagent_manager.py -x` (275 passed); class-namespace, source-language, large-file, and `git diff --check` gates passed. This pass covered the merge commit/tree identity, overlapping callers, target compatibility, the previously reviewed full diff and history, project rules and domain context, architecture constraints, and test integrity.

LivXue (approved):

PR #562, opened 2026-09-20
feat(playbook): wire generated participant functions

ypflll:
## Summary
- add a checked-in harness-generation capability configuration and reject enabled-but-unwired entries
- generate, validate, carry, bind, and execute `intake`, `advise`, `judge`, and `salvage`
- enforce disabled fields at schema, host validation, payload, and worker boundaries
- support both in-process Raven and forked ACP worker paths with a pure-data `StepView` and AST gate

Depends on #558.

## Verification
- GitHub CI: 20/20 checks passed
- all four Python unit/coverage shards passed
- diff coverage: 90.48% (304/336 changed executable lines; threshold 90%)
- focused post-sync suites: 211 architecture/runtime tests and 240 boundary tests passed
- 6 real OpenRouter/ACP E2E tests passed
- kernel wheel smoke confirms the capability loader and JSON are packaged


raven/agent/harness_capabilities.py
```diff
@@ -0,0 +1,95 @@
...
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any, Literal
+
+ModuleName = Literal["memory", "planning", "capability", "action"]
+FunctionKind = Literal["self", "participant"]
+
+_PATH = Path(__file__).parents[1] / "playbook" / "harness_generation.json"
```
gloryfromca, line 12: Blocking: the new contracts module breaks the standalone kernel contract and leaves the required gates red.

This module is automatically included in the standalone `raven-core` wheel, but `_PATH` points at `raven/playbook/harness_generation.json`, and the core-wheel builder ships no `raven/playbook` tree. In a clean wheel environment, importing this module and calling `parameter_enabled("memory", "systemPrompt")` raises `RuntimeError` caused by `FileNotFoundError`. The placement also fails the repository's paper rules: both CI and a local run report `no __tier__ declared` and `raven/contracts is 3637 lines against a ceiling of 3570`. Thus the current artifact exposes unusable functions and the required test suite cannot pass.

gloryfromca (commented):
Blocking: the capability loader must stop breaking the standalone kernel contract and the required gates.

I found one blocker, marked inline. I reproduced it against a clean `raven-core` wheel, and the current GitHub CI run reports the same two architecture-test failures seen locally.

Coverage: I read the complete three-dot diff and relevant surrounding callers, the participant/harness history, AGENTS.md and the Runtime context rules; checked both in-process and ACP/forked paths, backward compatibility, test changes, packaging, and architecture boundaries. I found no weakened tests.

Verification: `uv run pytest tests/test_harness_generation_capabilities.py tests/test_subagent_charter.py tests/integration/test_subagent_charter_e2e.py tests/test_participant_hook.py tests/test_harness_modules.py -x` passed (150 tests). `uv run pytest tests/test_contracts_two_tier_ledger.py tests/test_kernel_budget.py` failed (2 failed, 12 passed) on the new contracts module. The six credentialed real-LLM tests were collected but not executed locally.

raven/agent/harness_capabilities.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed by moving the loader to `raven.agent`, updating all callers, and keeping it out of the standalone kernel artifact. The previously failing contract ledger and paper-budget gates now pass locally and in the completed CI shard; a clean full-wheel install still loads the JSON-backed switches successfully.

gloryfromca (commented):

The new revision resolves my earlier blocker by moving the JSON-backed capability loader out of the standalone contracts kernel, with every caller updated. I found no new findings, replied to my prior thread, and resolved it.

Coverage: I re-read the full target diff and the revision delta; checked AGENTS.md and Runtime architecture rules, callers and history, both worker paths, backward compatibility, packaging, and the test changes. No tests were weakened.

Verification: the focused plus architecture suite passed (`164 passed`). A clean full-wheel install loaded the capability switch successfully, and a clean `raven-core` wheel contained no agent capability loader. All completed GitHub checks are green; one unrelated unit shard remains in progress.

gloryfromca (commented):

This revision adds boundary coverage only. The new tests exercise malformed capability catalogs, disabled and malformed generated functions, runtime failure-as-silence behavior, participant-note message shapes, and dispatch payload propagation without changing production behavior or weakening existing assertions. I found no new findings, and my prior thread remains resolved.

Coverage: I reviewed the revision delta and full target diff, project and Runtime architecture rules, affected callers and history, backward compatibility, test strength, and the in-process and ACP/forked paths.

Verification: `uv run pytest tests/test_harness_generation_capabilities.py tests/test_subagent_charter.py tests/test_subagent_manager.py tests/integration/test_subagent_charter_e2e.py tests/test_participant_hook.py tests/test_harness_modules.py tests/test_contracts_two_tier_ledger.py tests/test_kernel_budget.py -x` passed (`327 passed`). `git diff --check` also passed; all completed GitHub checks are green while the remaining shards run.

LivXue:
Reviewed at head `a36ed5b6`. This PR is stacked on `refactor/participant_full_roster` (#558), so I reviewed only its own contribution: `git diff refactor/participant_full_roster...feat/harness_generation_capabilities`, 16 files, +1346/-59.

What I ran: the PR's four test files (257 passed), the architecture pair plus the harness and participant suites (70 passed), and scratch probes against a copy of the tree for (a) the catalogue guard with three malformed documents, (b) the four enforcement boundaries with `capability.tools` disabled, (c) the worker boundary with a payload carrying a disabled field, and (d) the empty-brief payload below. A separate full unit run came back 23775 passed / 126 skipped; the only failure anywhere was `tests/test_install_script.py::test_resolve_node_dir_answers_each_case_it_exists_for`, which is environment-dependent (a system node without npm) and unrelated - neither the test nor `scripts/install.sh` is touched here.

No blocking defect. One regression worth fixing before merge, four behaviours that silently differ from what the design says and need a decision, then a set of cheap fixes and some coverage gaps.

## 1. The generated instructions are injected twice when `brief` is empty (regression)

`build_payload` omits `brief` for an empty brief but still writes the legacy `prompt` (`playbook/agent_generator.py:301-303`), and `parse` uses `prompt` as the brief when `brief` is absent while reading `instructionAddendum` separately (`agent/subagent/charter.py:136-143`), so `Charter.task_brief` (`:107-110`) concatenates the same text twice:

    SubPlaybook(memory.systemPrompt="ONLY THE GENERATED INSTRUCTIONS"), brief=""
      -> payload keys ['instructionAddendum', 'prompt']
      -> task_brief 'ONLY THE GENERATED INSTRUCTIONS\n\nONLY THE GENERATED INSTRUCTIONS'

Before this change the payload carried `prompt = system_prompt or brief`, so the text appeared once. `brief` is optional in the emitted schema (`agent_generator.py:210`), so the case is reachable from the real flow, and `task_brief` is rendered into the worker's own identity (`agent/harness/memory.py:155-157` -> `context_engine/segments/identity.py:24-32`), which is why a doubled line is not merely cosmetic. No test covers an empty brief. Fix: take the `prompt` fallback only when `instructionAddendum` is absent, or do not emit `prompt` for a brief-less worker.

## 2. Four behaviours that silently differ from the stated design (a decision, please)

**(a) A malformed capability catalogue turns the whole worker table off instead of refusing.** `emit_tool` is called outside `generate`'s repair loop (`agent_generator.py:468`), so a `RuntimeError` from the loader reaches the broad catch at `loop/wiring.py:735-737` and becomes `None` plus one warning: no workers, no charters, no generated functions, and the user sees an ordinary answer that is indistinguishable from "the model chose not to delegate". That is the opposite of the headline intent, "reject enabled-but-unwired entries". Validating the catalogue once at startup (or when `agentHarness: generate` is configured) would make the refusal visible.

**(b) An `intake` in the hook/ACP lane sees an empty step.** `turn_path.py:1827-1833` builds the `before_user_inbound` context with five fields, while `hook/participant.py:223-242` reads `turn_question`, `messages`, `tools`, `max_iterations` and `context_window_tokens` through `getattr` defaults. Same source, same phase, two lanes: the in-process lane reports a question, a transcript and a tool array; the forked lane reports `''`, empty and None - measured side by side. A function that branches on `step['question']` or walks `step['transcript']` therefore does nothing in the ACP lane without raising, and the description the generating model was shown (`agent_generator.py:224-231`) promises those keys with no phase qualifier.

**(c) A disabled `capability.tools` at the worker boundary drops the field instead of refusing it, and dropping it widens.** `charter.py:150` yields `Charter.tools is None`, and `None` means "no narrowing", so a dispatch that asked for `["read_file"]` keeps every tool. `CONTEXT.md:145` states the invariant the other way: a charter may take a tool away and may not hand one back. Host and worker agree today, so this needs host/worker catalogues to disagree - the skew the surrounding docstrings explicitly anticipate. Either refuse the dispatch or log the drop.

**(d) The charter hook precedes the entrances' own adapters, and a short-circuit ends the phase.** `core/runtime.py:203-211` passes `extra_hooks=[charter_hook, *host.hooks]`, `core/hooks_stack.py:56-59` appends them in order, and `hook/composite.py:160-161` returns immediately on a short-circuit. An `intake` that ends a turn therefore also skips the entrance's own inbound observers: the ACP worker path runs `OnUserInboundAdapter` (`engine_stack.py:74-78` via `rpc/bootstrap.py:85`), which records engagement and resets cron activity. The new test asserts exactly that short-circuit shape, so this is a consequence to confirm rather than to discover later. The same change makes `len(self.hooks) > 0` true on every install, which is what turns the inbound/observer machinery on by default; I checked that the other consequences are no-ops (`rolls_back=False` keeps the draft gate off).

## 3. Smaller fixes

- A `salvage` that raises is reported as `sub-agent used all 15 rounds and produced no answer` (`subagent/backends/raven_loop.py:655-662`), so the dispatcher is handed the wrong cause and will retry the wrong remedy.
- An `intake` short-circuit in the in-process lane returns before the run's tail (`raven_loop.py:454-455`, ahead of `:666-677`), so the node completes with no transcript and an empty panel.
- `advise` is appended to the last message and never taken back out (`raven_loop.py:63-73, 474-476, 615-617`), while the hook lane replaces its addendum each round (`hook/participant.py:395-427`). After a tool round the note lands inside a `role: "tool"` message, and up to 15 rounds x 2 notes stack in the saved transcript - for a verb whose contract is "guidance for the next model call".
- Over-long `functions` and `code` are truncated in silence (`charter.py:181, 192`), although the docstring right above promises "truncated and logged" (`:129-131`); the first call then blames the author's syntax. The gate's refusal message also says "not allowed in a charter judge" for every verb (`charter_code.py:123`).
- The seat is built into every runtime and `_step_payload(step)` is evaluated as an argument before the "is this verb declared" check (`charter.py:405-413`), so an install that never enables the feature still pays a payload walk per verb call (measured at roughly 0.6 ms with 50 messages and 10 tools, 3.2 ms at 400/60, up to twice per iteration). An early return before building the payload removes it.
- A JSON-valid but misshapen catalogue raises `AttributeError` instead of the loader's own message (`harness_capabilities.py:43, 48`, called outside the `try` at `:58-69`): `parameters: []` gives `'list' object has no attribute 'get'`.
- `action.parameters.checksImpl` is read (`agent_spec.py:167`) but is not in the loader's wired set, so enabling it refuses to load and that read can never be true. Intended placeholder, or should the reader go?
- The verb-to-module map exists four times (`agent_generator._PARTICIPANT_FUNCTIONS`, `charter.parse`'s `function_modules`, `agent_spec._enabled_fields_only`, and the JSON). They agree today; nothing keeps them agreeing.
- `CONTEXT.md:138-146` still defines a Charter as "a prompt ..., the tools ..., declarative checks ..., an optional judge as code, and a deadline", which the code no longer matches (brief, instruction addendum, functions), and the new capability vocabulary is undocumented anywhere - AGENTS.md section 6.

## 4. Coverage

- `agent/harness/memory.py:157`: reverting `task_brief=charter.task_brief` to `charter.prompt` fails no test (514 unit plus 31 e2e stay green), so nothing verifies that an addendum ever reaches the worker's turn.
- `core/runtime.py:209`: dropping the charter hook fails no test, so neither its presence, nor its position, nor `rolls_back=False` is pinned - flipping that flag to the default would gate the incremental reply for every runtime.
- The payload boundary has no CI-collected assertion: `brief`/`instructionAddendum` are asserted only in `tests/integration/test_subagent_charter_e2e.py`, and the unit shards do not collect `tests/integration` at all.
- `test_generated_functions_execute_in_the_in_process_worker_loop`, added by this PR, lives in `tests/integration/` but uses a stub provider and a temp directory, so by AGENTS.md section 5.2 it is not an integration test; CI never collects it, and the 14 changed lines it covers do not reach the coverage gate. Moving it into the unit suite would put them under the gate.

## 5. Process

- Being stacked on `refactor/participant_full_roster`, the PR is not under `main`'s ruleset while it stays that way; after #558 merges and this retargets, the gate is two approvals (one plus the extra approval for unattributed commits), and `dismiss_stale_reviews_on_push` means they have to come after the last push.
- All four commits are authored by `yaopengfei@shanda.com`, unlinked to a GitHub account and unsigned. The commit types match the PR title, and the merge commit does not survive the squash.

Checked and sound, so the lists above are the whole of it: the loader's guard (three malformed catalogues refused, each naming the offending path, and `enabled` required to be a real boolean); the four boundaries, including each of the eight enabled entries disabled one at a time and then fed to every layer, with per-field attribution; the AST gate and the signature enforcement for all four verbs; `compile_judge` kept as a wrapper so the existing judge path is untouched; the salvage call placed after the built-in no-tools wrap-up, so it cannot preempt a real answer; `parse`'s legacy `prompt` alias keeping an older host or worker talking; the JSON present in both the wheel and the sdist with `raven-core` unaffected; and `lint-imports` clean, with no new kernel edge.

gloryfromca (commented):

A new reviewer note changes my standing assessment from clean to nonblocking. I independently reproduced the empty-brief compatibility case: `build_payload("", SubPlaybook(memory=...))` emits the same generated text in both `prompt` and `instructionAddendum`, and `parse(...).task_brief` contains it twice. The focused suites still pass (`73 passed`), which confirms this case is not covered.

This change introduces the duplication and an empty brief is supported, but the worker retains a working path: its guidance is repeated rather than lost or refused. On this fifth round, that does not meet the required blocker condition that an operator has no working way out. I am carrying @LivXue's already-posted observation as a named follow-up rather than opening a duplicate inline thread.

Covered this turn: the new record evidence, the payload/parser call path, a direct reproduction, and focused tests. My earlier full review of this same revision is otherwise unchanged.

gloryfromca (commented):

This rewritten revision is patch-equivalent to the three commits already reviewed: each stable patch ID matches its prior counterpart. After refreshing the force-updated target ref, the requested diff is the intended 16-file change (`+1346/-59`) and `git diff --check` is clean. I found no new defect. The already-recorded empty-brief instruction duplication remains the one named nonblocking follow-up, so I have not opened a duplicate inline thread.

Verification on this exact head: `uv run pytest tests/test_harness_generation_capabilities.py tests/test_subagent_charter.py tests/test_subagent_manager.py tests/integration/test_subagent_charter_e2e.py tests/test_participant_hook.py tests/test_harness_modules.py tests/test_contracts_two_tier_ledger.py tests/test_kernel_budget.py -x` -> `327 passed in 7.72s`. At the final CI snapshot, every completed check was green; unit shard 2/4 was still pending.

Covered: AGENTS.md, CLAUDE.md, and CONTEXT terminology; the full target diff; rewritten history; callers and compatibility behavior; test-strength changes; and the standalone-wheel and participant architecture constraints. My earlier thread remains resolved.

raven/agent/subagent/charter.py
```diff
@@ -101,18 +133,29 @@ def parse(payload: Any) -> Charter | None:
     """
     if not isinstance(payload, dict):
         return None
-    prompt = str(payload.get("prompt") or "")[:MAX_PROMPT_CHARS]
-    stop_when = str(payload.get("stopWhen") or payload.get("stop_when") or "")[:MAX_PROMPT_CHARS]
+    legacy_prompt = payload.get("prompt")
+    raw_brief = payload.get("brief") if "brief" in payload else legacy_prompt
+    prompt = str(raw_brief or "")[:MAX_PROMPT_CHARS]
+    instruction_addendum = (
```
LivXue, line 139: **Nonblocking - do not let an empty brief inject the generated instructions twice.** With a non-empty `memory.systemPrompt` and an empty brief, `build_payload` omits `brief` but still writes the legacy `prompt` (`playbook/agent_generator.py:301-303`), so `parse` takes the addendum as the brief here and reads `instructionAddendum` again just below, and `Charter.task_brief` (`:107-110`) concatenates the same text twice into the worker's own identity (`agent/harness/memory.py:155-157` -> `context_engine/segments/identity.py:24-32`). Before this change the payload carried `prompt = system_prompt or brief`, so the text appeared once. `brief` is optional in the emitted schema (`agent_generator.py:210`), so the case is reachable from the real flow. Reproduced at this head: payload keys `['instructionAddendum', 'prompt']`, and `task_brief` == `'ONLY THE GENERATED INSTRUCTIONS\n\nONLY THE GENERATED INSTRUCTIONS'`. No test covers an empty brief. Take the `prompt` fallback only when `instructionAddendum` is absent, or stop emitting `prompt` for a brief-less worker.

raven/agent/subagent/charter.py
```diff
@@ -101,18 +133,29 @@ def parse(payload: Any) -> Charter | None:
...
+        else ""
+    )
+    stop_when = (
+        str(payload.get("stopWhen") or payload.get("stop_when") or "")[:MAX_PROMPT_CHARS]
+        if _parameter_enabled("memory", "stopWhen")
+        else ""
+    )
     tools: tuple[str, ...] | None = None
-    raw_tools = payload.get("tools")
+    raw_tools = payload.get("tools") if _parameter_enabled("capability", "tools") else None
```
LivXue, line 150: **Question - should a disabled `capability.tools` drop the field, or refuse the dispatch?** Dropping it yields `Charter.tools is None`, and `None` means "no narrowing" at `narrowed_tools` (`:254-265`), so a dispatch that asked for `["read_file"]` keeps every tool. That inverts the invariant `CONTEXT.md:147` states: "a charter may take a tool away and may not hand one back". The switch-off path is therefore the one path that widens. Host and worker have to disagree on the catalogue for this to bite, which is exactly the skew the surrounding docstrings anticipate, so it is not reachable today on a single consistent install. Code-reading only: the gate on this line, the early return in `narrowed_tools`, and the CONTEXT invariant; no run. Refusing the dispatch, or logging the drop, would both keep the invariant true.

raven/agent/subagent/charter.py
```diff
@@ -239,20 +296,68 @@ def _prior_satisfied(
...
-class CharterParticipant:
+def _pure_data(value: Any) -> Any:
+    """Detach the JSON-like part of a StepView from host runtime objects."""
+    if value is None or isinstance(value, (str, int, float, bool)):
+        return value
+    if isinstance(value, Mapping):
+        return {str(key): _pure_data(item) for key, item in value.items()}
+    if isinstance(value, Sequence) and not isinstance(value, (str, bytes, bytearray)):
+        return [_pure_data(item) for item in value]
+    return str(value)
```
LivXue, line 307: **Nonblocking - give a generated function a readable tool call rather than a Python repr.** `_pure_data` falls back to `str(value)` for anything that is not a scalar, mapping or sequence, and `ToolCallRequest` is none of those, so `step['response']['tool_calls']` reaches model-written code as a list of strings. Reproduced at this head: `tool_calls[0]` has type `str` with value `"ToolCallRequest(id='call_1', name='write_file', arguments={'path': 'out/a.txt', 'content': 'x'}, ...)"`. The `advise` description the generating model is shown promises "a read-only JSON-like dict" (`agent_generator.py:224-231`), and the allow-listed Python subset the gate enforces cannot parse a repr back, so the exact call an `advise` would want to comment on is the one thing it cannot read. Detaching the known fields of a tool call into a dict would keep the promise the description makes.

raven/agent/subagent/charter.py
```diff
@@ -239,20 +296,68 @@ def _prior_satisfied(
...
+    Generated pure-data functions implement only their declared verbs; every
+    other verb keeps the base participant's no-op answer.
     """
 
     __slots__ = ()
 
+    async def intake(self, text: str, step: StepView) -> Any:
+        return _generated_answer("intake", text, _step_payload(step))
+
+    async def advise(self, step: StepView) -> Any:
```
LivXue, line 355: **Nonblocking - do not build the step payload before checking that the verb is declared.** `_step_payload(step)` is evaluated as an argument, so it runs before `_generated_answer` can find that this charter declares no such function. The seat is unconditional: `build_runtime` passes the `CharterParticipant` class itself as the factory (`raven/core/runtime.py:203`), and `ParticipantHook` calls `self._factory()` directly (`agent/hook/participant.py:170,177`), which bypasses the `charter_participants()` guard that would have returned no participant. So an install with no charter and no generated functions still walks the whole transcript and tool list on every phase, twice per iteration for `advise`. Reproduced at this head with `current_charter()` None and `charter_participants()` empty: `_step_payload` built 1.0 times per `advise()` call, 0.39 ms per call at 400 messages and 60 tools. An early return before the payload is built removes it.

raven/agent/harness_capabilities.py
LivXue, line 12: **Question - should a missing catalogue fail the dispatch, or degrade it?** The loader lives in `raven/agent` while its document lives in `raven/playbook`, and it sits on the dispatch path (`parse` is reached from `agent/loop/wiring.py:658` and `subagent/backends/raven_loop.py:248`). With the file absent, `parse` raises instead of returning a charter. Reproduced at this head: `parse({"brief": "hi"})` raised `RuntimeError: invalid harness generation capability file: [Errno 2] No such file or directory`. The packaging pair is fine, and I checked it: `pyproject.toml` now ships `raven/playbook/*.json`, and the standalone kernel roster carries neither the loader nor the data, so the earlier wheel blocker stays fixed. The open question is the cross-shelf data edge from agent cargo to a playbook-owned file, and whether a data-file problem should take the dispatch down or leave every switch at its default.

raven/agent/harness_capabilities.py
```diff
@@ -0,0 +1,95 @@
...
+_CONNECTED_PARAMETERS = frozenset(
+    {
+        ("memory", "systemPrompt"),
+        ("memory", "stopWhen"),
+        ("capability", "tools"),
```
LivXue, line 17: **Question - should the `checksImpl` reader exist, or should the wiring arrive?** `raven/playbook/agent_spec.py:167` reads `parameter_enabled("action", "checksImpl")`, but `checksImpl` is absent from `_CONNECTED_PARAMETERS` here, so the switch is false forever and that branch can never be taken. Turning it on does not help: the catalogue's own guard then refuses to load at all. Reproduced at this head: enabling `action.parameters.checksImpl` raises `RuntimeError: harness generation capability is enabled but not wired: action.parameters.checksImpl`. Either the reader is a placeholder that should go, or the parameter should join the wired set. Asking rather than filing, because which one is right is a design call.

raven/agent/harness_capabilities.py
```diff
@@ -0,0 +1,95 @@
...
+    if not isinstance(entry, dict) or type(entry.get("enabled")) is not bool:
+        raise RuntimeError(f"invalid harness generation capability file: {path}.enabled must be a boolean")
+    return entry["enabled"]
+
+
+def _validate_connected(root: dict[str, Any]) -> None:
+    for module, body in root.items():
+        if module == "note" or not isinstance(body, dict):
+            continue
+        parameters = body.get("parameters", {}).get("items", {})
```
LivXue, line 43: **Nonblocking - let a misshapen catalogue report the loader's own error, not an attribute error.** This line assumes every level is a mapping, so a JSON-valid document with `"parameters": []` raises before the loader's own message can be formed. Reproduced at this head: `AttributeError: 'list' object has no attribute 'get'`, where a well-formed document returns cleanly and an unwired entry correctly reports `enabled but not wired`. `_validate_connected` is called at `:69`, outside the `try` at `:58-69` that maps `OSError`, `JSONDecodeError`, `KeyError` and `TypeError` onto `invalid harness generation capability file`, and `AttributeError` is not in that tuple in any case. The same shape applies to `functions.get(kind, {}).get("items", {})` at `:50`. This is the diagnostic path of the guard itself, so it is the one place a raw attribute error costs the most.

raven/agent/harness_capabilities.py
```diff
@@ -0,0 +1,95 @@
...
+        cache_key = (_PATH, stat.st_mtime_ns, stat.st_size)
+        if _CACHE_KEY == cache_key and _CACHE_ROOT is not None:
+            return _CACHE_ROOT
+        document = json.loads(_PATH.read_text(encoding="utf-8"))
+        root = document["harnessGeneration"]
+    except (OSError, json.JSONDecodeError, KeyError, TypeError) as exc:
+        raise RuntimeError(f"invalid harness generation capability file: {exc}") from exc
+    if not isinstance(root, dict):
+        raise RuntimeError("invalid harness generation capability file: harnessGeneration must be an object")
+    _validate_connected(root)
```
LivXue, line 69: **Nonblocking - make a catalogue failure visible instead of silently emptying the worker table.** `emit_tool` is called outside `generate`'s repair loop (`playbook/agent_generator.py:468`), so the `RuntimeError` this line can raise leaves `generate` and meets the broad `except Exception` at `agent/loop/wiring.py:735-737`, which logs one warning and returns `None`. The turn then runs with no workers, no charters and no generated functions, and the user sees an ordinary answer that is indistinguishable from the model choosing not to delegate. That is the opposite of this module's stated intent, which is to reject enabled-but-unwired entries: the rejection happens and nothing above a log line can tell. Reproduced for the raise at this head; the swallow at `wiring.py:735-737` is code-reading. Validating the catalogue once at startup, or when `playbooks.agentHarness: generate` is configured, would make the refusal visible where an operator can act on it.

raven/agent/subagent/backends/raven_loop.py
```diff
@@ -262,7 +276,7 @@ async def run(
         finally:
             IN_SUBAGENT_RUN.reset(token)
 
-    async def _run(
+    async def _run(  # noqa: C901 -- this is the bounded in-process worker loop
```
LivXue, line 305: **Nonblocking - include the generated functions in the two lines that prove a charter crossed the boundary.** Anchored at the nearest changed line; the site is `raven_loop.py:253-258`, which this PR does not touch, and its twin is `agent/loop/wiring.py:666-671`. Both still enumerate `tools`, `checks` and `judge` only, so a dispatch whose payload is generated functions and nothing else logs as `charter applied to this run (all tool(s), 0 check(s))`, which is indistinguishable from an empty charter. The comments directly above both lines name exactly this failure: "a charter that was dropped on the way looks exactly like one that was never written". This PR adds a fourth thing a charter can carry and does not add it to either line. Code-reading only: I read both sites and every `charter.*` attribute they touch; I did not run a dispatch and read a real log.

raven/agent/subagent/backends/raven_loop.py
```diff
@@ -404,14 +418,50 @@ def allowed(name: str) -> bool:
                 }
             ]
         )
+        iteration = 0
+        participants = charter_mod.charter_participants()
+        original_task = task
```
LivXue, line 456: **Nonblocking - bind the question a generated function reads to the authored task, not to the charter-prefixed one.** This line captures the task after `_WorkerBackend` has already prefixed the charter onto it (`subagent/dag_tool.py:234-241`), and publishes it as `StepView.question` at `:442`, which is what a generated `intake`, `advise` or `salvage` reads. The clean value arrives in the same call and is discarded: `_WorkerBackend.run` now forwards `authored_task` through `optional_keyword`, and `RavenLoopBackend.run` declares that parameter and never reads it. Reproduced at this head with a recording backend: the inner backend received `task` == `'You are the summariser worker. Keep to 3 bullets.\n\nSummarise the Q3 revenue table.'` and `authored_task` == `'Summarise the Q3 revenue table.'`; `inspect.getsource` counts exactly one occurrence of `authored_task` in the class, the declaration. Disclosure: the unused parameter is pre-existing, present before this PR; what is new is that this PR adds the first consumer of "the question" in this function and picks the other string, and that the DAG lane only began delivering a value there in the target branch. The hook lane fills the same field from the turn's own question (`agent/hook/participant.py:234`), so the two lanes now disagree on what `question` means. No test asserts what the worker lane puts there. Either take the authored value when it arrives, or say on the paper that `question` in this lane includes the charter, so generated code can be written against it.

raven/agent/subagent/backends/raven_loop.py
```diff
@@ -420,6 +470,10 @@ def allowed(name: str) -> bool:
         cut_at_ceiling = False
         while iteration < self._MAX_ITERATIONS:
             iteration += 1
+            if participants:
+                advice = await participant_answer("advise", participant_step("iteration"))
+                if isinstance(advice, str) and advice:
+                    _append_participant_note(messages, advice)
```
LivXue, line 509: **Question - should an `advise` note stay in the transcript after the call it was written for?** The note is appended to `messages[-1]` and never taken back out, while the forked lane strips and re-splices its addendum each round (`agent/hook/participant.py:395-427`). Two consequences follow. Notes accumulate: `advise` is asked before each model call and again after a round of tool calls (`:470-477`, `:611-618`), so up to two per round across fifteen rounds stay in the saved transcript. And after a tool round the note lands inside a `role: "tool"` message, which is not where guidance addressed to the model belongs. The verb's own contract is guidance for the next model call, which reads as single-use. Code-reading only: I read the four call sites, `_append_participant_note`, and the loop's message construction; I did not run a multi-round turn against a real provider.

LivXue:

Scope: I reviewed only the `raven/agent/**` portion of this PR - 6 of 16 changed files. The remaining 10 files were read as context but are not covered by this verdict.

This is a second round. The head moved from `a36ed5b6` to `f49cdef2`, and the PR's own contribution did not change: both three-dot diffs are 1896 lines with the same sha256, and all 16 file blobs hash identically, so this revision is a rebase rather than a fix. What moved is the target branch, which gained 13 commits over 68 files with no overlap against this PR's files.

Already on the record, and not new work: I posted a PR-level note at the previous head listing most of what follows, and gloryfromca independently reproduced the empty-brief case and carried it as a named follow-up. The inline notes here anchor those observations at specific lines so they can be resolved one by one. Exactly one finding is new this round, and it only became checkable at this base.

Corrections to my earlier note, caused by the rebase rather than by the author: `CONTEXT.md:145` is now `:147`, `dag_tool.py:225` is now `:235`, and the worker-resolution block I cited as `dag_tool.py:1516-1525` is now `:1542-1551`.

What I examined: the full 16-file diff against the merge base; every caller of the reshaped payload, including both dispatch tools; the capability loader and all four of its enforcement boundaries; both worker lanes; the packaging pair and the standalone kernel roster; the seat contract; and the `CONTEXT.md` charter definition.

Confirmed findings, anchored inline:

- An empty brief injects the generated instructions twice into the worker's identity, where before this change they appeared once.
- The step payload is built before the check that the verb is declared, and the participant seat is unconditional, so an install with no charter pays the walk on every phase.
- The two log lines that prove a charter crossed the boundary still enumerate tools, checks and judge only, so a functions-only charter logs as an empty one.
- Proposed tool calls reach a generated function as Python repr strings, which the allow-listed subset cannot parse, against what the description promises.
- A misshapen catalogue raises an attribute error instead of the loader's own message, in the guard's own diagnostic path.
- A catalogue failure is swallowed into an empty worker table with one log line, which is the opposite of the module's stated intent.
- The question a generated function reads carries the charter prefix, while the authored task arrives in the same call and is discarded.

Questions rather than findings: whether the `checksImpl` reader should exist or be wired; whether a missing catalogue should fail the dispatch or degrade it; whether an `advise` note should persist after the call it was written for; and whether a disabled `capability.tools` should drop the field, given that dropping it widens where `CONTEXT.md:147` says a charter may only narrow.

Three further items have no anchorable line and are stated here instead:

1. `tests/integration/test_subagent_charter_e2e.py` gained the only test that exercises the in-process generated-function wiring, but it uses a stub provider and a temp directory, so by the repository's own test rules it is not an integration test. `norecursedirs` keeps `tests/integration` out of the unit run, so CI never collects it and the lines it covers do not reach the coverage gate. It runs offline in under seven seconds; moving it into the unit suite would put those lines under the gate. It also trips the idle-ceiling warning.
2. The reshaped payload is forwarded to workers by `spawn_tool.py:297-309` and `:526` and by `dag_tool.py:235` and `:1592`, so model-written Python now runs inside both dispatch tools even though neither file changed here. Raised for visibility; the maintainer has ruled that this is acceptable.
3. The charter hook is prepended to `extra_hooks` at `raven/core/runtime.py:209`, ahead of the entrance's own inbound observers, and `CompositeHook` returns immediately on a short-circuit (`agent/hook/composite.py:160`), so an `intake` that ends a turn also skips the engagement and cron-activity bookkeeping that `OnUserInboundAdapter` does (`core/engine_stack.py:78`). The PR's own test asserts that short-circuit shape, so this is a consequence to confirm rather than to discover later.

Verification, all at this head:

- `tests/test_harness_generation_capabilities.py test_subagent_charter.py test_subagent_manager.py test_participant_hook.py test_harness_modules.py` -> 282 passed.
- `tests/test_contracts_two_tier_ledger.py test_kernel_budget.py tests/integration/test_subagent_charter_e2e.py` -> 45 passed.
- The empty-brief case, the payload walk, the repr tool calls, the misshapen catalogue, the unwired `checksImpl`, the absent catalogue, and the question binding were each re-run here rather than carried over from the previous round. The log lines, the note accumulation, the tool widening and the hook order are code-reading only, and say so inline.
- GitHub checks at this head: 18 passing, 2 skipping, none failing.

Checked and deliberately not reported: the loader's guard refuses three malformed catalogues, each naming the offending path, and requires `enabled` to be a real boolean; each of the eight enabled entries was disabled one at a time and fed to every layer, with per-field attribution; the AST gate and signature enforcement hold for all four verbs; `compile_judge` stays a wrapper so the existing judge path is untouched; `salvage` is placed after the built-in no-tools wrap-up, so it cannot preempt a real answer; the legacy `prompt` alias keeps an older host or worker talking; the JSON ships in both the wheel and the sdist with the standalone kernel unaffected; `lint-imports` is clean with no new kernel edge; and the target branch's new `worker_table_for` returning None keeps stored Playbook nodes away from turn-scoped generated workers, which narrows this feature's reach rather than widening it.

ypflll:
PR #558 is merged and this branch is now synchronized with the latest `main` at `60d44bb7`. The latest scoped review reports no blockers; the remaining inline notes are explicitly nonblocking questions/follow-ups and are preserved for subsequent hardening rather than expanding this reviewed change. All current checks have no failures. @LivXue @0xKT please submit fresh `APPROVE` reviews on this head when available; branch policy requires two approvals.

gloryfromca (commented):

This revision merges the now-landed target stack and current `main`; the scoped 16-file PR diff remains `+1346/-59`, passes `git diff --check`, and has the same stable patch ID as the already-reviewed change. The only overlapping `main` change adds attachment scoping beside the existing turn-tier scope, and the merged call path remains coherent. I found no new defect.

The existing inline questions and hardening notes remain valid nonblocking follow-ups, including the reproducible empty-brief instruction duplication. The author has explicitly preserved those for subsequent hardening, so they do not hold this revision.

Verification on this exact head: the focused harness, charter, participant, integration, contract-ledger, and kernel-budget suite passed (`339 passed in 8.36s`); all current GitHub checks pass, including all four unit shards, coverage gates, kernel-wheel smoke, lint, and repository checks.

Covered: AGENTS.md, CLAUDE.md, and CONTEXT terminology; `git diff github/main...HEAD`; merge history and overlapping callers; backward compatibility; test-strength changes; and standalone-wheel and participant architecture constraints. My only prior thread remains resolved; the open threads belong to another reviewer.

ypflll:
Resolved the latest main conflict at 4dc79767. The sole conflict was an insertion collision in raven_loop.py; both the generated Participant advice helper and main’s tool-availability helper are preserved. Verification: 429 focused Harness/Charter/Subagent/Browser tests passed; ruff, ty, diff check, and all 10 import contracts passed. @gloryfromca @LivXue @0xKT please re-review and submit the required approvals on this head.

gloryfromca (commented):

I reviewed the `main` synchronization and its sole conflict in `raven/agent/subagent/backends/raven_loop.py`. The resolution preserves both generated-participant advice handling and `main`'s dynamic tool-availability filtering, with the required imports and call sites intact. The automatic composition at exhaustion is also consistent with the main-loop contract: the no-tools wrap-up runs first, and generated `salvage` remains the fallback when that produces no answer. I found no new defect.

The scoped PR diff remains the same 16 files and `+1346/-59`, and `git diff --check` passes. Existing inline hardening items remain explicitly deferred, nonblocking follow-ups.

Verification on this exact head: `467 passed in 21.81s` across harness generation, charter, participant, subagent manager/backend, browser/tool-availability, integration, contract-ledger, and kernel-budget tests. At the final snapshot, every completed GitHub check was green; unit shards 1/4, 3/4, and 4/4 were still pending.

Covered: AGENTS.md, CLAUDE.md, and CONTEXT terminology; the full `github/main...HEAD` diff; both merge parents and conflict resolution; callers and backward compatibility; test-strength changes; and standalone-wheel, participant, and tool-registry architecture constraints. My prior thread remains resolved; the open threads belong to another reviewer.

0xKT (approved):

raven/agent/subagent/charter.py
ypflll: Accepted as a concrete nonblocking regression and tracked in #603. Keeping it out of this already-approved wiring change avoids invalidating the current review with a broader compatibility revision.

raven/agent/subagent/charter.py
ypflll: Decision: catalogue skew must fail closed rather than widen the tool set. The compatibility behavior and regression coverage are tracked in #603; this remains nonblocking for the same-install path reviewed here.

raven/agent/subagent/charter.py
ypflll: Accepted. Generated participants should receive a JSON-like tool-call object, not a Python repr. Tracked with a focused serialization test in #603.

raven/agent/subagent/charter.py
ypflll: Accepted as a performance hardening item. The undeclared-verb fast path and a payload-construction regression test are tracked in #603.

raven/agent/harness_capabilities.py
ypflll: Decision: a missing checked-in catalogue is a corrupt installation and should fail fast, not silently enable defaults. Documentation and a clearer operator-facing diagnostic are tracked in #603.

raven/agent/harness_capabilities.py
ypflll: The reader is intentional as an enforcement gate: checksImpl is reserved and declared disabled, so a stored artifact that tries to use it is rejected until wiring exists. The naming/documentation clarification is tracked in #603.

raven/agent/harness_capabilities.py
ypflll: Accepted. Misshapen but JSON-valid catalogue shapes should become the loader RuntimeError rather than leak AttributeError. Tracked with malformed-shape coverage in #603.

raven/agent/harness_capabilities.py
ypflll: Decision for this PR: generation remains fail-soft for turn availability and logs the setup failure. Startup/config-time validation or stronger operator surfacing is tracked in #603.

raven/agent/subagent/backends/raven_loop.py
ypflll: Accepted. Both Charter-boundary log lines should include the generated-function count. Tracked in #603.

raven/agent/subagent/backends/raven_loop.py
ypflll: Accepted. StepView.question should use authored_task when supplied and fall back to task only for older callers. Tracked with worker-lane coverage in #603.

raven/agent/subagent/backends/raven_loop.py
ypflll: Accepted. Advise guidance should be single-use and should not be persisted inside a tool message. The lifecycle change and multi-round regression coverage are tracked in #603.

PR #565, opened 2026-09-20
feat(ui-web): split model providers from model settings, one picker

gloryfromca:
## Summary

The settings dialog's Model page answered two questions at once -- which accounts this install holds, and which model each job runs on -- and it listed the connected providers only, so the fifty-odd vendors Raven supports were reachable through a dropdown inside an "add" form and nowhere else. This splits the page in two, gives every model a kind, and merges the two model pickers into one.

**Model providers** is a two-column page. The left column is every provider `model.options` returns (fifty-five on a stock registry, `hosted_vllm` and `custom` among them) with a search box and six filters -- connected, direct vendors, aggregators, browser sign-in, local -- connected rows first and each carrying its vendor mark. The right column is that provider's connection, beside the list rather than paged into.

**Model settings** keeps the roles card alone, and each of its eleven slots offers its own kind: the embedding slot lists embedding models, image lists image models, and a chat model appears in neither.

**Adding a model** is a popover under its button. It fetches the vendor's list, groups a gateway's ids by vendor, counts the kinds present over the current search result and filters to the one pressed, and writes one model per click rather than collecting ticks for a confirming press. An id the vendor does not list is typed into the same box with a chip saying what kind it is.

**One picker.** `components/ModelPicker` (164 lines, the roles card's) is deleted; the roles card opens the composer's. What an opening lists is now an `Offer`: a kind, the providers the caller allows, a title, the pair it marks, and what a pick means.

### Two decisions worth reading

**The kind narrows each provider's column, never the provider list.** A tester reported that a provider with a working key but no model list built vanished from the composer's picker, and the model the chip named -- set by onboarding, never "added" -- was in no column to be marked. So a provider with nothing of the offered kind is listed with a count of zero and a column that says so plus a row to type an id into; and the model a slot or the conversation currently holds is listed in its provider's column even when that column does not carry it. Verified on a real host: with `agents.defaults.model` set to an id absent from `providers.deepseek.models`, the picker opens on DeepSeek and lists it first, marked, and writes nothing.

**One classifier, in Python.** `registry_data.kind_of` files a model by what it writes, and `ModelLabel.kind` carries that answer to the page; nothing in TypeScript derives a kind from capabilities or modalities. The one exception is the chip on a typed id, which guesses from the name -- a translation of `inferred_tags`, with a nine-row table asserted on both sides so the copy cannot drift.

### Backend

Two read-only, additive fields on `model.options`: `ModelLabel.kind` and `ModelOptionProvider.gateway` (the registry's `is_gateway`, which no client can derive from a slug). No new method, no new config key. Contract first: `openrpc.json`, then pydantic, then the handlers, then `npm run gen`.

### Not in this change

The three navigation entries the prototype also adds (channels, cron, memory); the prototype's OAuth pane, which tells the reader to run a CLI command -- the browser device flow stays; backend enforcement of kinds, which remains a filter the page applies; the onboarding wizard, which keeps rendering `providers/Providers.tsx` untouched; the TUI.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Rebased onto `refactor/ui_web_architecture` at 6062b3678 and re-run after the rebase.

```
uv run pytest tests/test_rpc_model.py tests/test_provider_registry_data.py tests/test_rpc_schema_match.py -q
630 passed in 12.44s

(cd ui-web && npx vitest run)
Test Files  186 passed (186)
     Tests  2552 passed (2552)

(cd ui-web && npx tsc --noEmit && npm run gen:check && npm run build && python3 build.py)
generated.ts matches the contract (189 methods)
boot-snapshot: OK (250 nodes match golden)
boot-snapshot: OK (251 nodes match golden)

make check-source-language; make check-large-files; npm run lint:i18n --prefix ui-tui
i18n: generated catalogue is up to date

uv run ruff check raven tests
All checks passed!
```

Mutation checks on the two new backend fields: commenting out the `"gateway"` line fails `test_options_rows_carry_the_gateway_flag` on `KeyError: 'gateway'`; commenting out the `"kind"` key fails `test_model_labels_carry_a_kind` on `KeyError: 'kind'`. Restored, `git diff` clean.

Driven on a real host (`raven serve` on its own home and port 18899, real keys for DeepSeek / OpenRouter / Gemini, a real browser):

- Model providers draws 55 rows in two columns of 286px and 462px, the three connected ones first with a status dot, DeepSeek carrying the default badge, 54 vendor marks and one lettered tile (`custom`).
- The embedding slot's picker lists all three connected providers with counts 1 / 0 / 0 and exactly one model. It is at the body, not inside the dialog, at z-index 46 over the veil's 40.
- The add-model popover against OpenRouter's live catalogue: 534 models, kind tabs reading all 534 / text 445 / image 52 / embedding 33 / audio 4, five vendor groups, the four seeded models checked. Typing `my-team/bge-custom` offers it with an "embedding" chip, which is what `inferred_tags` answers for that name.
- The composer's chip lists the same three providers with OpenRouter counted 1: its embedding, reranker and image models are not offered for chat.

## Risk

Composer users see two changes: embedding and reranker models leave the model chip's picker, and typing an id there now adds it to the provider before switching. Settings users learn two navigation entries where there was one; `settingsTab.id` values written elsewhere still land on a model page, since the roles page keeps the id `model`.

A model the registry classifies wrongly -- an embedding model with no `embedding` tag whose name the two patterns miss -- is invisible to the embedding slot until an overlay states its kind. The add-model popover's chip is the way to state it.

Rollback: the two wire fields are additive and optional on the reader's side; the section split is a data change in `SECTIONS`; the deleted picker is one `git revert` away.

One gate pin moved: `fixture-shape`'s `UNSENT` list loses its six `model_labels` entries, because the offline fixtures now send them (a page that filters by kind needs a kind offline too). The gate's own message asks for exactly this.

## Related Issues

N/A


ui-web/src/features/settings/SettingsApp.tsx
```diff
@@ -63,7 +69,7 @@ function Nav({ tab }: { tab: SectionId }): JSX.Element {
 }
 
 const PAGE: Record<SectionId, () => JSX.Element> = {
-  general: General, usage: Usage, model: Model, skills: Skills,
+  general: General, usage: Usage, provider: Provider, model: Model, skills: Skills,
```
gloryfromca, line 72: Blocking: route the provider-opening helpers to the new provider section.

Now that `PAGE.model` draws only `Roles` and `PAGE.provider` owns the connection UI, the existing `openModels()` and `openProviderModels()` helpers in `settings/store.ts` still assign `settingsTab.id = 'model'`. The ordinary recovery callers therefore land on the wrong page: `openModelsForMissingProvider()` is used by the composer before-send path, and the shared picker calls `openProviderModels()` for a restricted unavailable provider. In both cases the user sees Model settings instead of the catalogue/detail needed to connect an account, so retrying cannot resolve the missing-provider state. Point those entry paths at `provider` (retaining the selected slug where applicable) and cover the post-split destination.

ui-web/src/features/settings/providers/Roles.tsx
```diff
@@ -171,62 +176,61 @@ async function clearRole(r: Role): Promise<SettingsSnapshot | void> {
...
+  /* No provider this role may use is connected: there is nothing to open onto,
+     and the way out is the providers page rather than an empty popover. A
+     provider that is connected but has nothing of this kind added is NOT this
+     case -- the picker lists it, says the column is empty and offers a row to
+     type an id into. */
   if (!provs.length && !val) {
     return role.media
       ? <button type="button" className="mini ghost" onClick={() => store.set({ provider: MEDIA_PROVIDER, provAdd: MEDIA_PROVIDER })}>{t('gui.settings.roles.connect_openrouter')}</button>
-      : <Rov>{t('gui.settings.roles.no_provider')}</Rov>
+      : <Rov>{t(role.everos ? 'gui.settings.roles.no_key_provider' : 'gui.settings.roles.no_provider')}</Rov>
```
gloryfromca: Blocking: make each no-provider role state navigate to Model providers.

This branch says the way out is the providers page, but the non-media case renders a passive `Rov`, while the media button only writes `provider`/`provAdd` and leaves the active tab on `model`. On a fresh install every role is therefore a dead end; with OpenRouter disconnected, the image/speech/video shortcut also stays on the roles card and never renders the selected provider. Render an actionable control for both branches and switch to the `provider` section when it is used.

ui-web/src/features/model/store.ts
```diff
@@ -131,12 +164,37 @@ export async function setProtocol(model: string, provider: string, protocol: Api
...
   const scope = at.scope
+  const { offer } = at
   close()
+  /* A slot writes through its own path -- a role key, an EverOS section, a
+     media pair -- and never through the conversation switch below, whose scope
+     logic answers a different question. Returned before it: the two are
+     alternatives, not steps. */
+  if (offer.pick) {
+    try {
+      await offer.pick(m, provider, typed)
```
gloryfromca: Blocking: carry the typed model kind through an offer-specific pick.

`ModelPicker` passes the displayed `typedKind` into `choose`, but this early branch forwards only `(model, provider, typed)`. The role callback then sends `model.add_model` with only the id. For an unknown id whose name does not trigger backend inference (for example `acme-vector-v1` typed into the embedding slot), no capability/output overlay is stored; the next `model.options` has no label entry, `modelKind` falls back to text, and the just-selected model disappears from the embedding picker on reopen. Extend the offer callback to receive the chosen kind (or otherwise apply the role kind with `statedTags`) before adding the model.

gloryfromca (commented):
Blocking: route missing-provider recovery and preserve typed non-text model kinds.

Reviewed the full target diff and the later head delta, including AGENTS.md / CLAUDE.md / CONTEXT*.md rules, the changed tests, callers, relevant history, backward compatibility, and the ui-web layering constraints. Three concrete blockers are marked inline.

- `uv run pytest tests/test_rpc_model.py tests/test_provider_registry_data.py -q`: 244 passed.
- `ui-web: npm test -- --run`: 186 files, 2552 tests passed.
- `ui-web: npm run gen:check && npm run lint && npm run type-check`: passed (five existing warnings, no errors).
- `ui-tui: npm run lint:rpc && npm run lint && npm run type-check`: passed (existing warnings, no errors).
- source-language and large-file gates: passed.

ui-web/src/features/settings/SettingsApp.tsx
gloryfromca: Right, and it was the worse half of the two: `openProviderModels` is what the composer's own dead end calls, so "OpenRouter has no model added" was landing the reader on a page with no way to add one.

Both doors now open `provider` (36207a26a). `openModels` too -- it answers "nothing is connected", which is the same page.

Pinned at the layer that owns it, in `settings/store.test.ts`: the picker's own test stubs `openProviderModels`, so nothing was asserting which tab either door sets. Reverting the two lines fails the new case on `expected 'model' to be 'provider'`.

ui-web/src/features/settings/providers/Roles.tsx
gloryfromca: Agreed, and on a fresh install this is the whole first-run path: every slot is in this branch, and a sentence naming a page the reader has to go and find themselves is not a way out.

Both states are buttons now, and both switch the tab; a media slot also selects OpenRouter, since that is the only vendor it may use (36207a26a).

One thing your reading surfaced that I had not: the tab switch has to come first. `setTab` clears the section's drawers, `provider` among them, so `set({ provider })` before it names the row into state the switch then discards. The second case below fails on exactly that ordering.

Two cases, both red without the fix:
- a slot with no provider renders a BUTTON and clicking it leaves `store.get().tab === 'provider'` (reverting to `Rov` fails on `expected 'SPAN' to be 'BUTTON'`)
- a media slot with OpenRouter off lands on `['provider', 'openrouter']`

ui-web/src/features/model/store.ts
gloryfromca: Correct, and the failure is exactly the one the chip exists to prevent: an id whose name neither pattern matches, typed into a slot, stored as text, gone from the slot it was typed into on the next read.

`Offer.pick` now takes the kind and `choose` forwards it (`kind ?? offer.kind`); `setRole` spreads `statedTags(kind)` into the `add_model` params (36207a26a).

The case types `our-finetune` into the embedding slot and asserts the call carries `capabilities: ['embedding'], output_modalities: ['vector']`. Mutating the forward to a literal `'text'` fails it.

ui-web/src/features/settings/SettingsApp.tsx
gloryfromca: No blockers; this can merge as far as I am concerned.\n\nVerified at 36207a26a4fd: both recovery helpers now select the provider section, and the provider-specific path preserves the requested slug through redraw and refresh. The new store regression test passes in the full web suite.\n\n

ui-web/src/features/settings/providers/Roles.tsx
gloryfromca: No blockers; this can merge as far as I am concerned.\n\nVerified at 36207a26a4fd: both empty-role states are actionable, and the media path switches sections before selecting OpenRouter, so setTab cannot clear the selection. Both navigation cases pass in the full web suite.\n\n

ui-web/src/features/model/store.ts
gloryfromca: No blockers; this can merge as far as I am concerned.\n\nVerified at 36207a26a4fd: the displayed kind crosses Offer.pick and reaches statedTags before the typed model is added. The unknown embedding-id regression case passes and asserts the vector metadata on the wire call.\n\n

gloryfromca (commented):

The current revision fixes all three prior blockers: both recovery doors reach Model providers, every empty role state has a working route there, and typed role picks preserve their stated kind. The corresponding regression tests exercise the concrete failures, and all three review threads are resolved.

Coverage included the repository rules in AGENTS.md and ui-web/CONTEXT.md, the full target diff and the current revision delta, affected callers and history, backward compatibility at the Offer/source and RPC seams, test changes for weakening, and the documented UI architecture boundaries. No new findings survived verification.

Verification:
- ui-web: 186 test files, 2,556 tests passed
- ui-web: RPC generation check, ESLint, and TypeScript passed (5 existing warnings, 0 errors)
- backend: 244 focused model/provider tests passed
- ui-tui: RPC generation check, ESLint, and TypeScript passed (23 existing warnings, 0 errors)
- source-language and large-file gates passed

gloryfromca (commented):

The current revision restores the text-only registry fallback needed by the first-run model step, preserves empty non-text offers, keeps hand-added model kinds and gateway grouping intact, and removes the hidden store-to-component dependency. The subsequent head delta fixes the import ordering and documents the fallback; no new review finding survived verification.

Coverage included AGENTS.md/CLAUDE.md, CONTEXT-MAP.md and the relevant Runtime, Web UI, and TUI context rules; the full target diff plus both new deltas; affected callers and commit history; backward compatibility for older model sources and additive RPC fields; test changes for weakening; and the UI/RPC architecture boundaries.

Verification:
- ui-web: 186 test files, 2,559 tests passed; RPC generation and TypeScript checks passed
- ESLint: the full run found one import-order error plus 5 existing warnings; the latest head fixes that error, and its changed source passes ESLint directly
- backend: 244 focused model/provider tests passed
- ui-tui: RPC generation, ESLint, and TypeScript passed (23 existing warnings, 0 errors)
- source-language and large-file gates passed against the latest head

gloryfromca (commented):

The new qualified/bare-id comparison fixes the reported duplicate rows and unreachable batch count for the direct-provider case. One nonblocking follow-up remains: `isAdded()` now trusts `ModelCandidate.added`, but that flag is the result of the one-time `model.fetch_models` call. After a catalogue row is removed, `run()` refreshes only the provider snapshot/configured list; the open sheet retains `m.added: true`. The row therefore remains checked and excluded from Add all until the popover is closed and reopened. Reopening refetches the catalogue and is a working escape hatch, so in this fifth review round this does not meet the blocking bar; it is named here rather than left as an open thread.

Verification at `737b718b5072`: focused provider suite 18/18 passed; full web suite 186 files and 2,560 tests passed (with the suite's existing aborted localhost-fetch diagnostics); TypeScript check, changed-file ESLint, and full/delta `git diff --check` passed.

Coverage: repository rules (`AGENTS.md`, `CLAUDE.md`, `CONTEXT-MAP.md`, and Web UI context/conventions), the full branch diff as previously reviewed plus the complete new delta, affected callers and backend identity/removal paths, relevant history, backward compatibility with stored spellings, test-strength review, and Web UI source/store architecture. No prior settled finding was reopened, and all reviewer-started threads remain resolved.

gloryfromca (commented):

The new delta correctly shares model-identity comparison between the provider popover and the picker, suppresses the qualified/bare duplicate in the covered path, and restores the current model to a provider column when it is absent from that provider's list. No new finding survived review. The sole nonblocking follow-up is the stale remove-state item already named in the previous revision's summary; this delta does not touch that path, so it is not repeated here.

Verification at `3ea41bff2d4d`: affected picker/provider suites 63/63 passed; full web suite 186 files and 2,562 tests passed (with the suite's existing aborted localhost-fetch diagnostics); TypeScript check, changed-file ESLint, and full/delta `git diff --check` passed.

Coverage: repository rules and Web UI context/conventions, the previously reviewed full branch plus the complete new delta, picker/popover callers, backend model-identity history and compatibility spellings, test-strength review, and source/store architecture. All reviewer-started threads remain resolved.

gloryfromca (commented):

This test-only delta adds meaningful coverage for vendor-group ordering, counts, unheaded bare rows, and group-scoped batch addition. It changes no production behavior and introduces no new finding. The sole nonblocking follow-up remains the item already named on the previous revisions; this delta does not touch that path, so it is not repeated here.

Verification at `51629cdb16cf`: provider-detail suite 19/19 passed; full web suite 186 files and 2,563 tests passed (with the suite's existing aborted localhost-fetch diagnostics); TypeScript check, changed-file ESLint, and full/delta `git diff --check` passed.

Coverage: repository rules and Web UI context/conventions, the previously reviewed full branch plus the complete test delta, the grouped-row renderer and batch caller, relevant history and compatibility behavior, test-strength review, and source/store architecture. All reviewer-started threads remain resolved.

gloryfromca (commented):

The rebase is sound. Range-diffing the old and new series shows the reviewed commits are preserved, while the target branch's agent-specific `components/ModelPicker` remains for its new caller. The rebase repair restores that component's `anchorRow` implementation and CSS from the target, and the surviving picker tests cover its placement and no-typed-id contract. No new finding survived review. The sole nonblocking follow-up remains the item already named on prior revisions and is not repeated here.

Verification at `a42981481462`: full web suite 191 files and 2,580 tests passed; focused backend suites 244 tests passed; TypeScript and generated-client checks passed; full lint completed with zero errors and five pre-existing warnings on untouched files; full/delta `git diff --check` passed.

Coverage: repository rules and Web/Runtime context, the full current PR diff plus old-to-new range-diff, rebase conflict resolution, affected picker callers and history, backward compatibility, test-strength review, and component/domain/source architecture. All reviewer-started threads remain resolved.

gloryfromca (commented):

This documentation-only delta accurately records why the rebased target retains its agent-specific picker and the concrete contract needed for later consolidation. It changes no runtime behavior, tests, compatibility, or architecture boundaries, and no new finding survived review. The sole nonblocking follow-up remains the item already named on prior revisions and is not repeated here.

Verification at `34281ba1380f`: full web suite 191 files and 2,580 tests passed; TypeScript and generated-client checks, changed-file ESLint, and full/delta `git diff --check` passed.

Coverage: repository comment rules and Web UI context/conventions, the previously reviewed full branch plus the complete documentation delta, both picker callers and their history, compatibility assumptions, test-strength review, and component/domain architecture. All reviewer-started threads remain resolved.

gloryfromca (commented):

At `2860baa64bde`, the 14 pull-request commits are patch-for-patch identical to the previously reviewed series after rebasing onto target commit `649ea4728` (every `git range-diff` entry is `=`). I checked the target's new settings refresh changes against the provider/model split and found no semantic collision or new finding. The sole nonblocking follow-up already named on this change remains unchanged.

Verification: `npm test` (192 files, 2,583 tests), `npm run type-check`, `npm run gen:check` (189 methods), and `uv run pytest tests/test_provider_registry_data.py tests/test_rpc_model.py -q` (244 tests) all pass. `npm run lint` passes with 0 errors and the same 5 warnings on untouched files; `git diff --check` passes.

Coverage included `AGENTS.md`/`CLAUDE.md` and the UI context rules, the target diff and old/new range-diff, touched callers and history, RPC/generated-client and TUI backward compatibility, test-strength changes, and the documented picker/provider architecture constraints. All previously opened review threads are settled and resolved.

0xKT (approved):

PR #566, opened 2026-09-20
fix(agent): bound every tree walk and take it off the event loop

ZuyiZhou:
## Summary

On 2026-09-20 one `find` call froze the whole gateway for 2h21m. The model asked for `find(pattern="**/gps_trilateration/index.html", path="/Users/admin")`, and `FindTool` ran `Path.glob` synchronously on the event loop, filtered the noise directories only after walking them, and had no deadline. The registry's 300s ceiling is an `asyncio.wait_for`, which cannot preempt synchronous code, so every session and every subagent on the gateway stopped until the walk came back (`gateway.log`: `Tool result: find duration=8476519ms`). `list_dir` with `recursive=true` had the same shape through `rglob`.

The fix gives the three tree-walking tools one walk, `raven.agent.tools.tree_walk`, with these bounds:

- noise directories (`node_modules`, `.git`, ...) are pruned before they are entered, not filtered afterwards -- the pruning tests spy on `os.scandir` and assert the noise directory is never opened;
- the walk yields one entry at a time and checks a 20s wall-clock budget (the value grep's fallback already used) before every entry, so whatever the caller does with one entry -- the `stat` per `find` match, the read per grep file -- is covered by the check on the next. The budget cannot be overrun by one large or slow directory: the probe with eight matches, a 30ms stat each and a 50ms budget returns 2 matches and the notice. When the budget is hit, `find` and `list_dir` append a `PARTIAL result: ... traversal budget ...` clause to their trailer, and a deadline that fires before any match is not reported as a clean `No files found` / `Directory is empty`;
- every caller runs the walk in `asyncio.to_thread`, so the loop keeps serving other sessions while the walk runs. Cancelling the thread does not stop the worker, which is why the deadline lives inside the walk. The pure-Python `grep` fallback already had the prune and the deadline; it moves onto the shared walk and off the loop with the others.

`find` matches each path component with `fnmatch.translate`, with a component that is exactly `**` spanning any number of components; the matcher is a position-set simulation (`_advance`), linear in components times segments, so a pattern carrying many `**` cannot make one match call slow. A pattern's leading literal components decide where the walk starts: `src/**/*.py` begins inside `src` and never lists a sibling of it, a pattern that is all literal (`src/lib/c.py`, `src/lib`) names exactly one path and answers without a walk, and the prefix itself is a match when the rest of the pattern can match nothing (`src/**` still lists `src/`). It answers exactly what `Path.glob` answered for the same pattern: pinned by a parametrized oracle test over twenty-seven patterns on a fixture and nineteen on a fixture with links (a directory link out of the tree, a link cycle, a file link, a noise directory beneath a link), and checked by hand against `Path.glob` over `raven/` for 27 patterns including bracket edge cases and trailing slashes, 0 mismatches. Two deliberate differences: a trailing `**` also lists files beneath its prefix (3.12's `Path.glob` listed only directories there), and a pattern with `..` or a leading `/` is refused instead of followed, since a pattern must not reach outside the directory the fence resolved. A trailing slash keeps `Path.glob`'s directory-only meaning (`*/`, `src/**/`). A symbolic link to a directory is entered where a single component names or matches it and never under `**`, which is how `Path.glob` read a pattern: `*/util/helper.py` reaches through a linked `vendor`, `**/helper.py` does not, and a link cycle is entered once per component the pattern spends on it, so `*/*/*/b.py` enters `src/loop -> src` twice and stops. The walker asks its caller per link (`follow`), `find` answers from the pattern and hands the link positions of a path to the matcher so `**` cannot claim them, and `list_dir` and the grep fallback never ask, staying as `rglob` and `os.walk` left them. A third deliberate difference from the merge base: a noise directory inside a pattern's literal prefix (`node_modules/*.js`, `src/node_modules/*.js`) is walked, since the pattern asked for it, where `Path.glob`'s filter dropped it; one met below the prefix is pruned as before (`*/node_modules/*.js` answers nothing), `list_dir` on such a path lists it, and the `find`, `list_dir` and code-flow `glob` descriptions state that rule instead of an unconditional "skipped". Review found an earlier wording of the two `find` descriptions one component short ("except one the pattern starts with"), while the oracle helper already pruned below the prefix; the sentences now say what the helper checks, and the oracle has the `src/node_modules` shape on both sides.

The raven-code `glob` tool (`agents/raven-code/.../tools/search.py`) parses trunk `find` output line by line and strips the trailer before merging brace alternatives. It learns the new clause, so a cut-short child is reported as cut short rather than as a tree that lacks the files; its `is_notice` grammar covers the new line so it is never pasted into a listing as a path. The code-flow tool guide (`prompts/TOOLS_CODE.md`), the wrapper's description and its `limit` parameter told the model that a PARTIAL `glob` result means "raise `limit` or narrow the pattern"; a traversal-budget PARTIAL is now a result `glob` can return and a higher limit does nothing for it, so all three now tell the two notices apart. The seeded copy of the guide refreshes through the receipt `seed_guide` keeps beside it, so no legacy digest is added.

Fixed in passing, and now pinned: recursive `list_dir` filtered `rglob` results on the components of the absolute path, so a workspace beneath a directory named `build`, `dist`, `venv` or another noise name listed as "Directory ... is empty" (reproduced at the merge base under `<tmp>/build/proj`; this head lists it). The shared walk prunes below the listed path only; `find` and the grep fallback already did.

Not changed: the 20s value. It is the budget grep's fallback already ran under; this PR does not re-measure it, it makes it hold for every entry of every walk.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

At head bb836988, on origin/main at 608cedf1 (nine commits: the walk, the grep pin, the changelog, the per-entry budget with the prefix start, the literal-directory case, the symbolic-link reading with the noise-prefix wording, the noise-name-above-the-path pin, the guide wording, and the prefix-rule wording; `git range-diff` against 9e17d346: the eight earlier commits replayed `=` onto the moved main, whose new commits touch no file this PR touches):

```
uv run pytest tests/test_file_search_traversal_guard.py tests/test_agents_code_tools_plugin.py \
  tests/test_search_tools.py tests/test_tools_workdir.py tests/test_tool_registry_execute.py tests/test_filesystem_tools.py -q
# 203 passed
COLORTERM=truecolor uv run --frozen --python 3.12 --all-extras pytest -q --ignore=tests/integration   # in a throwaway worktree
# 24146 passed, 110 skipped in 274.82s (0:04:34) at head bb836988, in a throwaway worktree with COLORTERM=truecolor
uv run pytest tests/test_agents_code_launcher.py tests/test_agents_code_tools_plugin.py -q   # the guide's seeding and wording tests, the glob wrapper
# 119 passed
git diff --check origin/main...HEAD                         # clean
scripts/check_source_language.py origin/main...HEAD         # exit 0
scripts/check_large_files.py origin/main...HEAD             # exit 0
git merge-tree --write-tree HEAD origin/main                # clean
```

Revert-to-red, with the tests kept (101 tests in `test_file_search_traversal_guard.py` + `test_agents_code_tools_plugin.py`):

```
the three production files reverted to the previously reviewed head (fd5663ee)   19 failed
file_search.py + filesystem.py reverted to origin/main, tree_walk.py removed      collection error
```

```
deadline checked once per directory (the reviewed placement)   10 failed
walk starts at the root instead of the pattern's prefix          2 failed
trailing slash matches files too                                 4 failed
a pattern naming one file still walks                            1 failed
the prefix itself is never a match                               7 failed
a pattern naming one directory walks beneath it                  1 failed
control (nothing mutated)                                      101 passed
```

From the first review round (still holding at this head): find without `to_thread` 1 failed; list_dir without `to_thread` 1 failed; grep fallback without `to_thread` 1 failed; pruning removed 6 failed; fork drops the PARTIAL clause 1 failed.

Fourth round, the symbolic-link reading (`test_file_search_traversal_guard.py` + `test_agents_code_tools_plugin.py`, 128 tests, `-n 0`; a mutant that lets `**` sweep a link, or forgets the link positions, re-enters the cycle until the OS refuses the path and answers paths `Path.glob` never did):

```
baseline                                                 128 passed
MUTANT  ** may sweep a link                                8 failed   (7 oracle cases, and the link the pattern ends at is opened)
MUTANT  find never asks to enter a link                    6 failed   (the six oracle patterns that pass through a link)
MUTANT  a link is entered even when the pattern ends at it 1 failed   (scandir spy on **/vendor)
MUTANT  ancestor links forgotten (links_of empty)          1 failed   (a cycle case answers a path Path.glob never did)
MUTANT  the walker consults follow but never enters        7 failed
MUTANT  each followed root gets a fresh deadline           9 failed   (every deadline test, and the walker's one-budget test)
REVERT  production files to reviewed head d67184c8         7 failed
CONTROL restored                                         128 passed
```

Revert-to-red for the noise-name pin: with `raven/agent/tools/filesystem.py` at the merge base and the test kept, `test_a_noise_name_above_the_searched_path_hides_nothing` fails on the `Directory ... is empty` answer (1 failed); at this head it passes. The `agents/**` change is two prose strings and a guide line, outside the coverage gate's `raven/**` and `plugins-dist/**` pathspecs; the guide's name-list test and its seeding tests (`test_agents_code_launcher.py`) cover the file.

The prefix rule is pinned by mutation, the other policy (prune a noise directory inside the prefix) against the two guard files (131 tests, `-n 0`):

```
baseline                                                 131 passed
MUTANT  a noise directory inside the prefix is pruned    5 failed
      test_find_matches_what_path_glob_matched[node_modules/**/*.py], test_find_matches_what_path_glob_matched[node_modules/*/], test_find_matches_what_path_glob_matched[node_modules/*/d.py], test_find_matches_what_path_glob_matched[node_modules/x/d.py], test_find_matches_what_path_glob_matched[src/node_modules/*.py]
CONTROL restored                                         131 passed
```

The symlink oracle is the reviewer's own test shape over a fixture with links: `vendor -> outside/real`, `src/loop -> src`, `alias.py -> a.py`, a `node_modules` beneath the link; 19 patterns, 0 mismatches against `Path.glob` on 3.12, plus a scandir spy pinning that `**/vendor` never opens the link, a walker test pinning that the default walk yields a link and does not enter it while a `follow`ed link is walked after the tree under one deadline, and a `list_dir` and grep test pinning `rglob`/`os.walk` parity through a link.

- [x] User-facing docs or screenshots are updated when needed (CHANGELOG `Unreleased / Fixed` entry)

## Risk

User-visible behaviour changes, all in tool output the model reads:

- `find` and `list_dir` (recursive) can now return a partial listing with a `PARTIAL result` clause in the trailer where they previously either returned everything or hung. The clause tells the model the absence of a match is not conclusive and to narrow the search.
- A path-prefixed `find` pattern spends its budget only inside its prefix, so on a large tree it finds what a root-anchored walk used to miss.
- `find` refuses patterns containing `..` or starting with `/` with an `Error running find:` line; `Path.glob` used to raise on the second and could climb on the first.
- `find` with a trailing `**` lists files as well as directories beneath the prefix; a trailing `/` lists directories only, as before.
- `find` reaches through a symbolic link to a directory where a literal or single-wildcard component names or matches it, as it did at the merge base; the earlier revisions of this PR had lost that and returned a clean miss with no notice. `**` never enters a link, as before.
- A noise directory inside a `find` pattern's literal prefix, or named as the `list_dir` path, is walked: `find('node_modules/*.js')` and `find('src/node_modules/*.js')` answer its files and `list_dir('node_modules', recursive=true)` lists it, where the merge base answered nothing and "empty". One met below the prefix is pruned as before. Both descriptions state the rule.
- Recursive `list_dir` on a workspace beneath a directory named like a noise directory (`/tmp/build/proj`) lists it instead of answering "is empty".
- The code-flow `glob` description, its `limit` parameter and the tool guide tell a limit cut from a traversal-budget PARTIAL; the notices themselves are unchanged.
- `list_dir` flat listings are unchanged; the empty-directory message is unchanged.

Rollback: revert the squash commit. The new module has no other consumers, and no state or schema is involved.

- [x] Security impact considered (a pattern can no longer climb out of the fenced directory via `..`; a directory link is entered exactly where `Path.glob` entered it at the merge base, one non-`**` component, and never under `**`, so the fence's reach is what it was and not wider; `list_dir` and grep never enter one)
- [x] Backward compatibility considered (the fork's trailer grammar is extended, not changed; existing fork tests pass unchanged)

## Related Issues

N/A


raven/agent/tools/file_search.py
```diff
@@ -432,25 +469,49 @@ async def execute(
...
+
+        Runs in a worker thread. An entry is matched on its path components
+        relative to ``base``, and a matched directory is shown with a trailing
+        slash. The second value says whether the walk hit its deadline, in
+        which case the list is what was found before it did.
+        """
+        matches: list[tuple[float, str]] = []
+        base_parts = len(base.parts)
+        try:
+            for root, dirs, names in tree_walk.walk(base):
```
gloryfromca: Blocking: anchored find patterns can time out in unrelated sibling subtrees.

`_search` always starts at `base`, even when the compiled pattern begins with literal components. In a fixture where `unrelated/` is visited before `src/` and the budget expires on the second directory, the previous `base.glob("src/**/*.py")` returns `src/a.py`, while this implementation returns a partial miss because it enters `unrelated/` first. `src/**/*.py` is one of the documented ordinary query shapes, and the partial notice's advice to narrow the pattern does not help because it is already path-prefixed. A path-prefixed query must not let unrelated sibling trees consume the budget before its named subtree is reached.

raven/agent/tools/tree_walk.py
```diff
@@ -0,0 +1,56 @@
...
+    ``None``); whatever the caller collected before that is its partial result.
+    """
+    if deadline_s is None:
+        deadline_s = WALK_DEADLINE_S
+    deadline = time.monotonic() + deadline_s
+    for root, dirs, names in os.walk(base):
+        if time.monotonic() > deadline:
+            raise TimeoutError(f"traversal deadline of {deadline_s:g}s exceeded under {base}")
+        dirs[:] = [d for d in dirs if d not in IGNORE_DIRS]
+        yield root, dirs, names
```
gloryfromca: Blocking: one large or slow directory can still overrun the advertised traversal deadline.

The deadline is checked before yielding an entire directory, and the generator does not regain control until the caller finishes that directory's full `names` batch. During that gap `FindTool` stats every match, fallback grep reads every file, and recursive `list_dir` accumulates the batch. In a bounded probe with eight top-level matches, a 30 ms `stat` per entry, and a 50 ms deadline, `find` took 244 ms and returned all eight matches without a `PARTIAL` notice. On a large flat directory or slow mount this can run through the 20-second budget to the registry ceiling; cancelling `asyncio.to_thread` then does not stop the worker. The deadline needs to cover per-entry processing, not only transitions between directories.

gloryfromca (commented):
Blocking: anchored searches can spend their budget outside the requested subtree, and per-directory work can run past the deadline.

I reviewed the full `github/main...HEAD` diff, relevant history, and the built-in `find`/`grep`/`list_dir` callers, the raven-code `glob` wrapper, and the registry's 300-second timeout boundary. I also checked the repository rules in `AGENTS.md`/`CLAUDE.md` and `CONTEXT-MAP.md`/`CONTEXT.md`, backward compatibility with the prior `Path.glob`/`rglob` behavior, and whether tests were weakened. The branch and commit naming, source language, test-file placement, dependency/assets constraints, and domain vocabulary are consistent; I found no weakened tests.

Verification: `uv run pytest tests/test_file_search_traversal_guard.py tests/test_agents_code_tools_plugin.py tests/test_search_tools.py tests/test_tool_registry_execute.py tests/test_tools_workdir.py -x` passed all 149 tests. `git diff --check github/main...HEAD` passed, and the GitHub CI checks are green. The two inline failures are not covered by those tests.

Nonblocking follow-up: the matcher also drops `Path.glob`'s directory-only meaning for trailing-slash patterns, so patterns such as `src/**/` can now return files (and `*/` can match top-level files). This is a narrow compatibility corner and is not part of my block.

gloryfromca (commented):
Blocking: the two previously reported traversal-budget defects remain unchanged.

This revision adds only the Unreleased changelog entry; it does not change either code path or add coverage for the two open findings. I re-read the full `github/main...HEAD` diff and the new commit, and rechecked the project rules in `AGENTS.md`/`CLAUDE.md` and `CONTEXT-MAP.md`/`CONTEXT.md`, the affected callers and history, backward compatibility, test strength, and the registry/threading architecture. I found no additional defect beyond the existing threads and the previously named nonblocking trailing-slash compatibility follow-up.

Both blockers still reproduce on this head: an anchored `src/**/*.py` search can spend its budget in an unrelated sibling and return a partial miss, while the flat-directory probe with a 50 ms budget took 242 ms and returned eight matches without a `PARTIAL` notice. Accordingly, the new changelog statement that the shared walk stops at a 20-second wall-clock budget currently overstates the guarantee; it is the documentation consequence of the same open deadline finding, not a new finding. The two threads remain open because neither is settled.

Verification: `uv run pytest tests/test_file_search_traversal_guard.py tests/test_agents_code_tools_plugin.py tests/test_search_tools.py tests/test_tool_registry_execute.py tests/test_tools_workdir.py -x` passed all 149 tests. `git diff --check github/main...HEAD` also passed. No tests were weakened.

raven/agent/tools/file_search.py
ZuyiZhou: Reproduced on the reviewed head (fd5663ee) with your recipe: a fixture where `unrelated/` is listed before `src/`, budget expiring on the second directory, `find("src/**/*.py")` returned the PARTIAL notice and no `src/a.py`.

Fixed in 5f14180a. `_compile_pattern` now returns a `_Pattern(prefix, segments, dirs_only)`: the run of leading components with no wildcard is the prefix, and `_search` starts the walk at `base/<prefix>` instead of at `base`. `src/**/*.py` begins inside `src` and never lists a sibling; a pattern that is all literal answers from the prefix alone with no walk; and the prefix itself is a match when the rest of the pattern can match nothing (`src/**`), which is what keeps `src/` in that listing. While there I also restored the trailing-slash directory-only meaning you noted as nonblocking (`*/`, `src/**/`), since it falls out of the same compile step.

Tests (all in `tests/test_file_search_traversal_guard.py`): `test_find_starts_at_the_patterns_literal_prefix` builds 50 sibling directories under `unrelated/`, spies on `os.scandir`, and asserts neither `unrelated` nor the search root itself is ever listed while `src/b.py` and `src/lib/c.py` are found; `test_find_with_a_literal_pattern_answers_without_a_walk` monkeypatches the walk to fail the test if called; `test_find_keeps_a_trailing_slashs_directory_only_meaning`; and the `Path.glob` oracle table gained `src/lib/c.py`, `src/lib`, `src/lib/`, `*/`, `src/**/`, `src/*/`, `missing/*.py`, `src/nope.py`. The same oracle run over `raven/` (27 patterns, 886 entries) still shows 0 mismatches against `Path.glob`.

Mutation, each alone, control restored between runs (101 tests in the two files):

```
walk starts at the root, not the prefix     2 failed
trailing slash matches files too            4 failed
literal file pattern still walks            1 failed
prefix itself never a match                 7 failed
control                                     101 passed
```

raven/agent/tools/tree_walk.py
ZuyiZhou: Reproduced on the reviewed head (fd5663ee) with your probe shape: eight matches in one flat directory, a 30ms `stat` per entry, a 50ms budget; `find` returned all eight with no PARTIAL notice. The deadline was checked once per directory and the caller's per-entry work sat between checks.

Fixed in 5f14180a. `tree_walk.walk` now yields one entry at a time, `(root, name, is_dir)`, and reads the clock before each yield, so whatever the caller does with one entry (the stat in `find`, the read in the grep fallback, the append in `list_dir`) is covered by the check on the next. All three callers were moved onto the per-entry shape; the fork's glob tool needed no change beyond its test's tick count. The same probe now returns 2 matches and the PARTIAL trailer.

Tests: `test_find_deadline_covers_the_work_done_per_entry` and `test_grep_fallback_deadline_covers_the_work_done_per_file` drive that probe deterministically (a clock the stat/read advances by 30ms, budget 50ms) and assert 2 results plus the notice; `test_the_walk_checks_the_deadline_before_every_entry` pins the walker itself. The existing tick-based tests were re-derived for the per-entry clock reads and the fixture for the limit-plus-deadline notice gained a file under `late/` so the walk is actually cut short.

Mutant "deadline checked once per directory" (the old placement): 10 failed. Reverting the three production files to the reviewed head: 19 failed. Control: 101 passed.

The changelog line was rewritten to say what now holds ("checks a 20s wall-clock budget before every entry it yields"); the PR description follows in the same push.

raven/agent/tools/file_search.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

I confirm the original anchored-search failure is fixed: the current implementation starts `src/**/*.py` at `base/src`, the focused test proves neither the root nor `unrelated/` is opened, and the ordinary query returns the expected matches.

One narrower follow-up: the statement that every all-literal pattern answers without a walk currently holds for files, but not directories. For `src/lib`, `_search` records `src/lib/` and then still walks its descendants; with an already-expired test budget it returns that correct match plus a `PARTIAL` trailer, whereas `Path.glob` answers the exact directory directly. The match remains usable and `list_dir` on the parent is an escape hatch, so this does not meet the fourth-round blocker bar. The original thread is settled from my side.

raven/agent/tools/tree_walk.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

I confirm the reported batch overrun is fixed: the walker now returns control for a clock check between entries, and both the focused tests and my rerun of the eight-entry probe stop after two matches with a `PARTIAL` notice.

A narrow remainder is that work on the final yielded entry has no subsequent check. With one file, a 50 ms budget, and its `stat` advancing the clock by 100 ms, `find` still returns `only.py` without `PARTIAL`. This is not introduced by this revision, requires the slow work to land on the final entry, and still yields a usable result, so it does not meet the fourth-round blocker bar; it is a follow-up rather than a reason to keep the original thread open. The original thread is settled from my side.

gloryfromca (commented):

This revision fixes both prior blockers: path-prefixed searches begin at their literal prefix instead of traversing unrelated siblings, and the shared walker checks its budget between yielded entries so caller-side work in a large flat directory is accounted for. I replied to both threads with the verification details; both settled threads are resolved. The earlier trailing-slash compatibility follow-up is also fixed.

I reviewed the full `github/main...HEAD` diff and this revision's delta, the `find`/fallback-`grep`/recursive-`list_dir` callers and raven-code `glob` wrapper, relevant history, the registry/thread cancellation boundary, backward compatibility against `Path.glob`, and whether tests were weakened. I also rechecked `AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md`/`CONTEXT.md`, commit/source language, test placement, dependencies/assets, and domain vocabulary. The new tests strengthen rather than weaken the coverage, and I found no new merge blocker.

Named follow-ups: an all-literal directory pattern such as `src/lib` records the correct match but still walks below that directory and can add a needless `PARTIAL` trailer; and work performed on the final yielded entry has no next deadline check, so one final slow `stat`/read can finish over budget without the trailer. At this fourth review round neither qualifies as blocking: the former has a direct `list_dir` escape hatch and the latter is a narrow residual not introduced by this revision, requiring the slow operation to be the final entry while still returning a usable result.

Verification: `uv run pytest tests/test_file_search_traversal_guard.py tests/test_agents_code_tools_plugin.py tests/test_search_tools.py tests/test_tool_registry_execute.py tests/test_tools_workdir.py -x` passed all 163 tests. `git diff --check github/main...HEAD` passed. All completed GitHub checks are green; `coverage gates` was still running when reviewed.

raven/agent/tools/file_search.py
ZuyiZhou: Reproduced: `find("src/lib")` with an already-expired budget returned `src/lib/` plus the PARTIAL trailer, where `Path.glob("src/lib")` answers the directory and descends into nothing.

Fixed in dc56ff5c: a pattern with no wildcard names exactly one path, so `_search` returns after recording the prefix match whether it is a file or a directory; there is nothing to walk and no budget to run out of. `test_find_with_a_literal_pattern_answers_without_a_walk` now covers `src/lib` and `src/lib/` (the walk is monkeypatched to fail the test if called) alongside the file cases. Mutant "walk beneath a named directory again": 1 failed; control: 52 passed. The `Path.glob` oracle over `raven/` still shows 0 mismatches for the literal-directory shapes (`agent/tools`, `agent/tools/`, `agent`).

raven/agent/tools/tree_walk.py
ZuyiZhou: Reproduced: one file, a 50ms budget, a stat that advances the clock by 100ms -- `find` returns `only.py` with no trailer.

I left this one as it is, deliberately, and want to say why rather than let it pass as an oversight. The trailer means "entries were skipped, absence is not conclusive". After the final entry nothing was skipped: the result is complete, and a clock check after the last yield would attach a PARTIAL notice to a complete answer, which is the mislabel the trailer exists to avoid. What the residual actually is: the work on one entry (one `stat`, one file read) cannot be interrupted from inside the walk, so the bound is "the budget plus one entry's work" for every entry, not only the last -- a slow entry in the middle also finishes over budget before the next check raises. The check before each entry keeps that overrun to one entry's work; nothing inside a synchronous walk can make it zero. If a bound on a single blocking syscall is wanted, that is the worker-thread cancellation question, which section 16 of the review skill already records as not achievable by cancelling `to_thread`.

raven/agent/tools/file_search.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed. The `not compiled.segments` return now covers exact directories as well as files, and the strengthened test makes any call to the walker fail for `src/lib` and `src/lib/`. The previous needless traversal and trailer are gone, so this follow-up is closed from my side.

raven/agent/tools/tree_walk.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Agreed. `PARTIAL` describes omitted entries and an inconclusive absence; after the final yielded entry the enumeration is complete, so adding the trailer would be incorrect. The cooperative bound is necessarily the budget plus at most one synchronous entry operation, while the per-entry precheck fixes the original unbounded batch of caller work. I withdraw the residual suggestion and consider this thread fully settled.

gloryfromca (commented):

The exact-directory follow-up is fixed: an all-literal file or directory pattern now returns its one answer without entering the walker, and the test fails if a walk is attempted. I also accept the author's final-entry explanation and withdrew that follow-up: `PARTIAL` means entries were omitted and absence is inconclusive; once the final entry is processed the enumeration is complete, while the cooperative runtime bound is necessarily the budget plus at most one synchronous entry operation. Both threads I opened remain resolved.

I reviewed the full `github/main...HEAD` diff and the `5f14180a..dc56ff5c` delta, the `find`/fallback-`grep`/recursive-`list_dir` callers and raven-code `glob` wrapper, relevant history, backward compatibility with `Path.glob`, test strength, and the worker-thread/registry timeout boundary. I also rechecked `AGENTS.md`/`CLAUDE.md`, `CONTEXT-MAP.md`/`CONTEXT.md`, commit/source language, test placement, dependencies/assets, and domain vocabulary. Tests were strengthened, not weakened, and I found no remaining plain error or merge blocker.

Verification: `uv run pytest tests/test_file_search_traversal_guard.py tests/test_agents_code_tools_plugin.py tests/test_search_tools.py tests/test_tool_registry_execute.py tests/test_tools_workdir.py -x` passed all 163 tests. `git diff --check github/main...HEAD` passed. No completed GitHub check was failing; three unit shards were still running when reviewed.

ZuyiZhou:
Rebased onto main at cf3c95fd to clear the changelog conflict: this branch and main had both inserted at the top of Unreleased / Fixed, so the only manual resolution was the order of the entries; nothing else conflicted. `git range-diff` against dc56ff5c: three commits replayed `=`, and the two that carry changelog text differ only in that hunk.

Rerun on the rebased head 62ab017d: the six focused files 173 passed; the full suite 23996 passed, 110 skipped in its own worktree; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. The description's head and numbers are updated to this head.

gloryfromca (commented):

I independently checked the rebase onto `cf3c95fd`. The complete non-changelog patch is byte-for-byte identical to the previously reviewed patch, the substantive corresponding commits have identical stable patch IDs, and the only touched-file change on the new base was `CHANGELOG.md`; the branch's existing entry is unchanged and now precedes the newly landed entries. Both threads I opened remain resolved.

For this rebased revision I rechecked the full `github/main...HEAD` diff, range-diff and history, the affected callers and worker-thread/registry boundary, backward compatibility, and test strength. I also rechecked the repository constraints in `AGENTS.md`/`CLAUDE.md` and `CONTEXT-MAP.md`/`CONTEXT.md`, including source/commit language, test placement, dependencies/assets, and domain vocabulary. No tests were weakened and I found no new plain error.

Verification: `uv run pytest tests/test_file_search_traversal_guard.py tests/test_agents_code_tools_plugin.py tests/test_search_tools.py tests/test_tool_registry_execute.py tests/test_tools_workdir.py tests/test_filesystem_tools.py -x` passed all 173 tests. `git diff --check github/main...HEAD` passed. Current GitHub checks were passing, with two unit shards still pending when reviewed.

raven/agent/tools/file_search.py
```diff
@@ -432,25 +492,70 @@ async def execute(
...
+            # A pattern that is all literal names exactly one path, file or
+            # directory; what lies beneath a named directory is not asked for,
+            # so there is nothing to walk and no budget to run out of.
+            if not start.is_dir() or not compiled.segments:
+                return matches, False
+        start_parts = len(start.parts)
+        last_root: str | None = None
+        parents: list[str] = []
+        try:
+            for root, name, is_dir in tree_walk.walk(start):
```
LivXue: **Blocker - keep a symlinked directory reachable, or say in the parity claims that it is not.**

`_search` walks through `tree_walk.walk`, which is `os.walk` at its default `followlinks=False`, so a directory symlink is yielded as an entry but never descended. `Path.glob` did descend one for a literal or single-`*` component and refused only under `**`. A pattern that has to pass *through* a symlinked directory therefore loses matches the merge base returned, and because the walk ran to completion there is no `PARTIAL` trailer: the model reads the absence as conclusive, which is the one reading this PR exists to prevent.

The exclusion is also asymmetric inside this revision, because `os.walk` always scans its own top. The same symlink is traversed when it lands in the pattern's literal prefix and skipped when it does not:

```
find('vendor/util/helper.py')   base hit    head hit    all-literal, answered from the prefix
find('vendor/*/helper.py')      base hit    head hit    the symlink IS the walk root
find('*/util/helper.py')        base hit    head MISS   same symlink, same file, no notice
find('**/helper.py')            base miss   head miss   Path.glob never followed under **
```

Reproduced with this branch's own oracle, `test_find_matches_what_path_glob_matched`, run unchanged except for a fixture that holds a directory symlink whose target sits outside the workspace. Four cases passed against the merge base and two failed against this head, one of them:

```
pattern='*/util/helper.py'  glob=['vendor/util/helper.py']  find=[]
```

`_glob_fixture` builds no symlink, and the 27-pattern run over `raven/` has none either, so the oracle that backs the parity claim cannot reach this shape.

Two directions close it. Descend a symlinked directory for a non-`**` component, which restores what `Path.glob` answered. Or keep the walk as it is and make the exclusion explicit: the changelog's "answers what `Path.glob` answered for the same pattern" and `_compile_pattern`'s "Follows what `Path.glob` accepted" would then need the exception named, and the prefix case should stop following the link so one policy holds everywhere. The second is the smaller change and a defensible policy. What does not hold is the current state, where the answer depends on where the walk happened to start.

raven/agent/tools/file_search.py
```diff
@@ -432,25 +492,70 @@ async def execute(
...
+        prefix -- ``base/src`` for ``src/**/*.py`` -- and an entry is matched
+        on its path components from there; a matched directory is shown with
+        a trailing slash. The prefix itself is a match when the rest of the
+        pattern can match nothing at all (``src/**``, or a pattern that is all
+        literal), which is also how a pattern naming one file finds it without
+        a walk. The second value says whether the walk hit its deadline, in
+        which case the list is what was found before it did.
+        """
+        matches: list[tuple[float, str]] = []
+        start = base.joinpath(*compiled.prefix)
```
LivXue, line 582: **Nonblocking - make the noise-directory rule one rule, or change what the tools tell the model about it.**

`_search` starts the walk at `base/<literal prefix>`, and `tree_walk.walk` prunes `IGNORE_DIRS` among a directory's children but never at its own root. A pattern whose literal prefix names a noise directory therefore walks inside it, while the same directory reached any other way is still pruned:

```
find('node_modules/*.js')       base: no files    head: node_modules/y.js
find('node_modules/**/*.js')    base: no files    head: node_modules/y.js, node_modules/pkg/x.js
find('**/node_modules/*.js')    base: no files    head: no files
```

The same divergence shows up in this branch's own oracle, whose expected set filters `IGNORE_DIRS` out:

```
pattern='node_modules/*.js'  glob=[]  find=['node_modules/y.js']
```

`ListDirTool` gained the same opening: `list_dir('node_modules', recursive=true)` answered "Directory node_modules is empty" at the merge base and lists the directory here.

I am not asking for the old answer back. Naming a noise directory explicitly and getting its contents is arguably the better behaviour, and it is what stops `list_dir` being useless on a path the caller chose on purpose. What does not survive the change is the sentence the tools hand the model: `FindTool.description` at line 445 still reads "Noise directories (.git, node_modules, etc.) are skipped", and `ListDirTool.description` says they "are auto-ignored", both without condition, and neither was touched by this PR. Either prune the walk root as well, or let those two sentences say the pruning applies to what the pattern did not name. A case in the oracle parametrisation would be worth having either way, since that is the assertion that would have caught this.

LivXue:

Scope: I reviewed only the `raven/agent/**` portion of this PR -- 3 of 7 changed files (`tools/file_search.py`, `tools/filesystem.py`, and the new `tools/tree_walk.py`). The remaining 4 -- the two test files, `CHANGELOG.md`, and the raven-code `glob` wrapper -- were read as context but are not covered by this verdict.

The design is the right one, and the earlier rounds landed it well. Reading the deadline before every yield rather than once per directory is what makes the bound real: it turns an unbounded batch of caller-side work into "the budget plus at most one entry's operation", which is the tightest bound a cooperative synchronous walk can offer. Compiling a pattern into a literal prefix plus per-component matchers is the other half, and it is what stops an unrelated sibling spending a path-anchored search's budget. Both are better than what they replace, and the new tests pin them with `os.scandir` spies and event-loop turnaround assertions rather than with timing alone.

What I examined: the in-scope diff at `cf3c95fd..62ab017d`; the `find`, recursive `list_dir` and fallback-`grep` call sites; every reader of the moved `_IGNORE_DIRS` and `_WALK_DEADLINE_S` symbols; the four notice strings the engine can now emit against the raven-code wrapper's three regexes, including the zero-matches-plus-timeout case; the import-linter seat contracts; `CONTEXT.md` and `CONTEXT-MAP.md` for claims about these tools; and the threads already settled on this PR, so as not to re-file them.

Confirmed findings:

- `file_search.py:546` (blocking) -- a directory symlink reached by any component other than the pattern's literal prefix is yielded but never descended, so `find` loses matches `Path.glob` returned, silently and with no `PARTIAL` trailer. The same symlink is still traversed when it lands in the prefix, because `os.walk` always scans its own top, so the exclusion is an artifact of where the walk starts rather than a policy.
- `file_search.py:530` (nonblocking) -- a pattern whose literal prefix names a noise directory now walks inside it, while the same directory reached any other way is still pruned. `FindTool.description` and `ListDirTool.description` still tell the model the pruning is unconditional.

Verification:

- Reproduced, both findings: this branch's own `test_find_matches_what_path_glob_matched` oracle, run unchanged against each side, with a fixture holding a directory symlink whose target sits outside the workspace. 4 passed at the merge base, 2 failed at this head (`pattern='*/util/helper.py' glob=['vendor/util/helper.py'] find=[]`, and `pattern='node_modules/*.js' glob=[] find=['node_modules/y.js']`).
- `python -m pytest tests/test_file_search_traversal_guard.py tests/test_search_tools.py tests/test_tools_workdir.py tests/test_filesystem_tools.py` against a `git archive` extraction of this head: 97 passed. No test weakened.
- `lint-imports` at this head: 10 contracts kept, 0 broken.

- The `_walk_entries` docstring claims it reproduces `sorted(rglob(...))` order while sorting on `rel.parts`. I expected a mismatch; it is not one. `PurePath.__lt__` compares `_parts_normcase`, so the two orders agree, and base and head print byte-identical listings on a fixture built to separate them. The docstring is correct.
- A symlinked prefix component is joined onto `base` with no re-resolution, so a pattern can walk outside the resolved fence. Pre-existing: `Path.glob` followed the same link at the merge base. Raised as context, not as a finding against this branch.
- `_matches` backtracks exponentially in the number of `**` components, and the deadline is read only between entries, so one call cannot be interrupted. Measured 0.20s for a pattern carrying seven `**` against a 22-deep path of colliding names; reaching seconds needs roughly twelve of them, and it is worker-thread time, not loop time. Not credible as a hang.
- The sweep behind "bound every tree walk" is complete. The only recursive walks left in `raven/agent/` are the three this PR converted; the two surviving `glob()` calls (`vendored_agents.py:540`, `background_exec.py:220`) are single-level over small known directories.
- Neither finding duplicates a thread already settled here. Those covered the sibling-budget and per-directory-deadline defects and their two follow-ups; no round reached symlinks or a noise-directory prefix.

Three things outside the reviewed scope, disclosed rather than filed:

- `agents/raven-code/plugins/code-flow/prompts/TOOLS_CODE.md:71` tells the model that a PARTIAL `glob` result means "raise `limit` or narrow the pattern". A traversal-budget PARTIAL cannot be fixed by raising `limit`, and that is now a result `glob` can return. The wrapper's own merged notice already says "narrow the path or pattern", so this prompt line is the one place left describing the old set of causes.
- Recursive `list_dir` filtered on the absolute path's components at the merge base, so any workspace whose absolute path contained a component named `build`, `dist`, `venv` or the like reported its whole tree as "Directory . is empty". This revision fixes that, and the changelog does not mention it. Reproduced on both sides in a workspace at `/tmp/build/proj`; worth a line, since anyone who hit it would have concluded their workspace was empty.
- The branch is conflicting with `main` again. Three commits landed since the rebase onto `cf3c95fd`, and intersecting the two file lists leaves `CHANGELOG.md` as the only overlap, so it is the same resolution as last time.

raven/agent/tools/file_search.py
gloryfromca: Blocking: directory symlinks reached after wildcard components are silently skipped without a partial notice.

Independent reproduction confirms the reported compatibility break. With `vendor` symlinked to an external directory containing `util/helper.py`, merge-base `Path.glob("*/util/helper.py")` returns `vendor/util/helper.py`, while this head returns `No files found matching pattern.` The literal-prefix forms still hit and `**/helper.py` misses on both sides, matching the reported asymmetry. Because the walk completes, the false negative carries no `PARTIAL` warning and invalidates my earlier clean stance.

I also reproduced the separate noise-prefix behavior (`node_modules/*.js` is pruned at the merge base but returned here when it becomes the walk root); I am leaving its nonblocking grade and both thread states to the reviewer who opened them. The six affected test files still pass all 173 tests, so the symlink shape is not covered. GitHub currently reports this head as `CONFLICTING` with `main` as well.

ZuyiZhou:
Rebased once more, onto main at 350bd5a6: #528 merged with its changelog entry at the same anchor, so this branch conflicted there again and only there. `git range-diff` against 62ab017d: three commits replayed `=`, the two that carry changelog text differ only in that hunk; main's commits since the last rebase touch no file this PR touches.

Rerun on the rebased head d67184c8: the six focused files 173 passed; the full suite 24049 passed, 110 skipped in its own worktree; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. The description's head and numbers are updated to this head.

gloryfromca (commented):
Blocking: the open symlink-traversal compatibility finding remains unchanged on this rebased revision.

I independently checked the rebase onto `350bd5a6`: the complete non-changelog patch is byte-for-byte identical to the previously reviewed patch, corresponding substantive commits are unchanged in the range-diff, and the only replay difference is changelog context/order. Changes now on `main` after that base touch none of this PR's files, and GitHub reports the head mergeable.

The existing blocker still reproduces on `d67184c8`: with `vendor` as a directory symlink containing `util/helper.py`, `Path.glob("*/util/helper.py")` returns `vendor/util/helper.py`, while `FindTool` returns a conclusive clean miss. I am carrying the open reviewer finding as the standing blocker, not re-grading or duplicating it; their symlink and noise-policy threads remain open, while both threads I opened remain resolved.

For this revision I checked the full `github/main...HEAD` diff, range-diff/history, affected callers, backward compatibility, test strength, and worker-thread/registry boundary, plus the repository rules in `AGENTS.md`/`CLAUDE.md` and `CONTEXT-MAP.md`/`CONTEXT.md`. No tests were weakened and I found no additional issue.

Verification: `uv run pytest tests/test_file_search_traversal_guard.py tests/test_agents_code_tools_plugin.py tests/test_search_tools.py tests/test_tool_registry_execute.py tests/test_tools_workdir.py tests/test_filesystem_tools.py -x` passed all 173 tests. `git diff --check github/main...HEAD` passed. GitHub's unit shards were still pending when reviewed.

raven/agent/tools/file_search.py
ZuyiZhou: Reproduced with your recipe on d67184c8: the oracle test run over a fixture whose `vendor` is a directory link to a tree outside the workspace gave `pattern='*/util/helper.py'  glob=['vendor/util/helper.py']  find=[]`, and the four-row table came out as you have it, the literal-prefix forms hitting because `os.walk` scans its own top. I also measured `Path.glob` on 3.12 over a wider fixture (the link out of the tree, a cycle `src/loop -> src`, a file link `alias.py -> a.py`): a literal or single-wildcard component enters a link, `**` never does, so `src/*/b.py` answers `src/loop/b.py`, `*/*/*/b.py` answers `src/loop/loop/b.py`, `src/**/b.py` answers `src/b.py` alone, and `**/vendor` answers the link itself without looking beneath it.

Fixed in cb780b7a, taking your first direction, since the PR's claim is parity with what `find` answered before and the second would have made `vendor/util/helper.py` a miss as well:

- `tree_walk.walk` gains a `follow(root, name)` hook, asked per directory link after the entry is yielded; a link it says yes to is walked once the tree that holds it is done, pruned the same way and under the same deadline. `list_dir` and the grep fallback pass nothing and stay as `rglob` and `os.walk` left them (pinned).
- `find` says yes where a single component consumes the link and the pattern still has a component to spend beneath it (`**/vendor` is a match, not a place to walk; pinned with a scandir spy). The matcher is now a position-set simulation, `_advance`, and takes the link positions of a path so that `**` cannot claim a link component. That is the `Path.glob` reading, and it is what makes a cycle finite: each entry into `src/loop` costs one component of the pattern, so `*/*/*/b.py` enters it twice and stops.

Tests: `test_find_reaches_through_a_symbolic_link_where_path_glob_did` is your oracle, unchanged in shape, over 19 patterns on that fixture (your four rows, the cycle, the file link, a noise directory beneath the link), with the walk budget set to 2s so a walk that did not end would show as a PARTIAL mismatch rather than a hang (a mutant that lets `**` sweep a link re-enters the cycle until the OS refuses the path, and fails on the paths `Path.glob` never answered); `test_find_does_not_look_inside_a_link_the_pattern_ends_at`; `test_the_walk_enters_a_link_only_when_told_and_under_the_same_budget` (the default walk yields a link and does not enter it; with `follow` the link's entries come after the tree's and the deadline is one for both); `test_list_dir_and_grep_do_not_enter_a_symbolic_link`.

Mutation, each alone, the files restored between runs (`test_file_search_traversal_guard.py` + `test_agents_code_tools_plugin.py`, 128 tests, `-n 0`):

```
baseline                                                 128 passed
MUTANT  ** may sweep a link                                8 failed   (7 oracle cases, and the link the pattern ends at is opened)
MUTANT  find never asks to enter a link                    6 failed   (the six oracle patterns that pass through a link)
MUTANT  a link is entered even when the pattern ends at it 1 failed   (scandir spy on **/vendor)
MUTANT  ancestor links forgotten (links_of empty)          1 failed   (a cycle case answers a path Path.glob never did)
MUTANT  the walker consults follow but never enters        7 failed
MUTANT  each followed root gets a fresh deadline           9 failed   (every deadline test, and the walker's one-budget test)
REVERT  production files to reviewed head d67184c8         7 failed
CONTROL restored                                         128 passed
```

Full suite at cb780b7a in its own worktree: 24076 passed, 110 skipped. The changelog entry and `_search`'s docstring now say how a link is read; the description's verification block is updated. One thing worth your eye: a link out of the workspace is entered exactly where `Path.glob` entered it before this PR, so the fence's reach is what it was at the merge base and not wider; I have not narrowed it here.

raven/agent/tools/file_search.py
ZuyiZhou: Agreed on both halves, and taken the second: the behaviour stays and the tools say it. Fixed in cb780b7a.

- `FindTool.description` now reads "Noise directories (.git, node_modules, etc.) are skipped, except one the pattern starts with: 'node_modules/*.js' looks inside it." The code-flow plugin's `glob` carries the same sentence, since it answers through the host tool and would otherwise have kept the unconditional one.
- `ListDirTool.description` now reads "Common noise directories (.git, node_modules, __pycache__, etc.) are skipped below the listed path; name one as the path to see inside it."
- The oracle prunes beneath the pattern's literal prefix only (`_glob_oracle`), which is the one rule the walk applies, and the parametrisation gains `node_modules/*/d.py`, `node_modules/**/*.py`, `node_modules/x/d.py`, `node_modules/*/` and `**/node_modules/*/d.py`; the last is the case that stays pruned. The changelog entry says the same in one sentence.

Not changed: the walk root is still not pruned, which is the behaviour you said is arguably better and the one that keeps `list_dir` useful on a path the caller chose; the tests above pin it rather than the merge base's answer.

ZuyiZhou:
Follow-up at 9e17d346, taking the three things LivXue disclosed outside the reviewed scope rather than filed:

- The code-flow tool guide (`prompts/TOOLS_CODE.md`) said a PARTIAL `glob` result means "raise `limit` or narrow the pattern". The wrapper's own description and its `limit` parameter said the same, so all three now tell a limit cut from a traversal-budget PARTIAL, which a higher limit does nothing for. The seeded copy refreshes through the receipt `seed_guide` keeps beside it, so no legacy digest was added; the guide's name-list and seeding tests pass unchanged.
- Recursive `list_dir` at the merge base filtered on the absolute path's components: reproduced under `<tmp>/build/proj`, the merge base answers `Directory ... is empty` and this branch lists `a.py`, `src/`, `src/b.py`. The changelog entry now says so, and `test_a_noise_name_above_the_searched_path_hides_nothing` pins it for `list_dir`, `find` and the grep fallback together (red with the merge base's `filesystem.py`, green here).
- Rebased onto main at 87f39286; the six earlier commits replayed `=` and main's new commits touch no file of this PR.

Also worth knowing for the `_matches` measurement: since cb780b7a the matcher is a position-set simulation (`_advance`), linear in components times segments, so the exponential backtracking you measured at 0.20s for seven `**` is gone with the symlink change.

At this head: the six focused files 201 passed; `tests/test_agents_code_launcher.py` 119 passed together with test_agents_code_tools_plugin.py; full suite 24086 passed, 110 skipped in its own worktree; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. The description is updated to this head.

raven/agent/tools/file_search.py
```diff
@@ -383,7 +490,7 @@ def description(self) -> str:
             "Find files by glob pattern (e.g. '*.py', 'src/**/*.ts'). Prefer this over "
             "running find/ls through exec. Returns paths relative to the search root, "
             "most-recently-modified first. Noise directories (.git, node_modules, etc.) "
-            "are skipped."
+            "are skipped, except one the pattern starts with: 'node_modules/*.js' looks inside it."
```
LivXue: **Nonblocking - the sentence is narrower than the rule it describes.**

You took the second of the two directions -- keep the behaviour and let the tools say so -- and the behaviour is now stated, one component short. `_compile_pattern` makes the prefix the whole run of leading components with no wildcard in them, and `_search` starts the walk at `base/<prefix>`, so a noise directory is entered wherever it sits *inside* that prefix, not only where the pattern opens with it:

```
node_modules/*.py       -> (none)                     pattern opens with a noise name
src/node_modules/*.py   -> src/node_modules/n.py      it does not, and the walk enters anyway
*/node_modules/*.py     -> (none)
**/node_modules/*.py    -> (none)
```

Reproduced on this head, against a fixture holding `ws/src/node_modules/n.py` and no top-level `ws/node_modules`. The discriminator is whether the noise name falls inside the literal prefix, so `ui-web/dist/*.js` and `packages/foo/node_modules/*.js` are entered on the same rule, while the description tells the model they are skipped.

Three places carry the same sentence and all three are short by the same step: `FindTool.description` here, the code-flow `glob` wrapper's description, and the changelog's "A noise directory the pattern names first". `ListDirTool.description` is accurate as written -- `list_dir` really does turn on the listed path, having no prefix to speak of.

Worth saying that the new test helper already has the rule right: `_glob_oracle` prunes `rel.parts[skip:]` with `skip = len(prefix)`, which is exactly "beneath the literal prefix". Code and oracle agree; only the prose is behind them. The parametrisation beside it is all either noise-first or `**`-prefixed, so it cannot reach the `src/node_modules/...` shape -- one case of that would pin the sentence as well as the rule.

ZuyiZhou: Reproduced on 9e17d346 with your fixture (`ws/src/node_modules/n.py`, no top-level `node_modules`): `src/node_modules/*.py` answers `src/node_modules/n.py`, `*/node_modules/*.py` and `**/node_modules/*.py` answer nothing, and the two descriptions said "except one the pattern starts with". You are right that the rule is the prefix, not the first component: `_compile_pattern` takes the whole run of leading literal components, `_search` starts the walk at `base/<prefix>`, and `os.walk` never prunes its own top, so a noise directory is entered wherever it sits inside that prefix and pruned only below it.

Fixed in bb836988, prose only, since the code and the oracle helper already had the rule:

- `FindTool.description` and the code-flow `glob` description now read "Noise directories (.git, node_modules, etc.) are skipped below the pattern's literal path prefix; one named inside that prefix, as in 'src/node_modules/*.js', is entered."
- The changelog says "A noise directory inside a pattern's literal prefix (`node_modules/*.js`, `src/node_modules/*.js`) is walked ... one met below the prefix is pruned as before." `ListDirTool.description` is unchanged, as you said it is accurate.
- `_glob_fixture` gains `src/node_modules/n.py`, and the oracle parametrisation gains `src/node_modules/*.py` (the prefix case, answered) and `*/node_modules/*.py` (a wildcard before the noise name, pruned), so the sentence is pinned on both sides; the existing cases still agree with `Path.glob` less the pruning below the prefix, including `src/**/*.py`, which now has a noise directory to prune beneath `src`.

Mutation, the other policy (prune a noise directory inside the prefix, i.e. refuse the walk), against the two test files (131 tests, `-n 0`):

```
baseline                                                 131 passed
MUTANT  a noise directory inside the prefix is pruned    5 failed
      test_find_matches_what_path_glob_matched[node_modules/**/*.py], test_find_matches_what_path_glob_matched[node_modules/*/], test_find_matches_what_path_glob_matched[node_modules/*/d.py], test_find_matches_what_path_glob_matched[node_modules/x/d.py], test_find_matches_what_path_glob_matched[src/node_modules/*.py]
CONTROL restored                                         131 passed
```

Full suite at bb836988 in its own worktree: 24146 passed, 110 skipped in 274.82s (0:04:34). Also rebased onto main at 608cedf1 (changelog anchor only; main's new commits touch no file of this PR).

gloryfromca (commented):

The prior symlink compatibility blocker is fixed and its opener has resolved the thread. I independently checked the new follow hook and matcher state: `find` now follows a directory link only when a single pattern component consumes it, does not let `**` sweep links, terminates on link cycles by consuming finite pattern components, and leaves recursive `list_dir` and fallback `grep` at their previous no-follow behavior. The nested noise-directory wording now matches the actual literal-prefix rule in both tool faces and the changelog. The raven-code brace wrapper also preserves traversal-budget partial notices and no longer advises raising a result limit when only narrowing the search can help.

I reviewed the full `github/main...HEAD` diff and the revision history, the `find`/fallback-`grep`/recursive-`list_dir` callers and raven-code wrapper, backward compatibility against `Path.glob`/`rglob`, the worker-thread and shared-deadline architecture, and whether tests were weakened. I also rechecked `AGENTS.md`/`CLAUDE.md` and `CONTEXT-MAP.md`/`CONTEXT.md`, including source and commit language, test placement, dependencies/assets, and domain vocabulary. The new oracle and boundary cases strengthen coverage; I found no remaining plain error or merge blocker.

Verification: `uv run pytest tests/test_file_search_traversal_guard.py tests/test_agents_code_tools_plugin.py tests/test_search_tools.py tests/test_tool_registry_execute.py tests/test_tools_workdir.py tests/test_filesystem_tools.py -x -n 0` passed all 203 tests. `git diff --check github/main...HEAD` passed. GitHub reports the head mergeable; the completed Python lint and four unit shards are green, while `coverage gates` and `repository files` were still running when reviewed.

ZuyiZhou:
Rebased onto main at d50a13c1: the conflict was the changelog anchor again (main gained entries at the top of Unreleased / Fixed), resolved by entry order only; main's commits since the last base touch no file this PR touches, and `git range-diff` against bb836988 shows all nine commits replayed `=`.

Rerun on the rebased head 8dade176: the six focused files 203 passed; the full suite 24652 passed, 110 skipped in its own worktree; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. The description's head and numbers are updated.

gloryfromca (commented):

I independently checked the rebase onto `d50a13c1`: the non-changelog patch is byte-for-byte identical to the clean revision I previously reviewed, all nine commits replay as patch-equivalent apart from changelog context, and the branch's changelog entry is unchanged while newly landed `main` entries surround it. GitHub reports the head mergeable, and every review thread remains resolved.

For this revision I rechecked the full `github/main...HEAD` diff, range-diff/history, affected callers, backward compatibility, test strength, and worker-thread/deadline architecture, plus the repository rules in `AGENTS.md`/`CLAUDE.md` and `CONTEXT-MAP.md`/`CONTEXT.md`. No tests were weakened and I found no new plain error.

Verification: `uv run pytest tests/test_file_search_traversal_guard.py tests/test_agents_code_tools_plugin.py tests/test_search_tools.py tests/test_tool_registry_execute.py tests/test_tools_workdir.py tests/test_filesystem_tools.py -x -n 0` passed all 203 tests. `git diff --check github/main...HEAD` passed. Completed GitHub checks were green; three unit shards and the newly queued commit-lint run were still pending when reviewed.

LivXue (approved):

Issue #461, opened 2026-09-17
A running sub-agent shows a clock and nothing else: tool calls are stored but never streamed


Handsome-wzw:
## Summary

While a sub-agent answers, the page can show only two things: how long the turn has been running, and whatever the agent says in prose. Every tool call it makes is collected and stored, and reaches the page only after the fact. For an agent whose work is mostly tool calls -- a deck build, a research sweep -- the live view is a clock and nothing else for minutes at a time.

## Steps to reproduce

1. Send a substantial task to a sub-agent instance (a deck rebuild will do).
2. Watch the instance surface while it works.

## Expected behavior

Some sign of what it is doing. The data exists: `subagents.instance.history` returns the tool rows afterwards, and the ACP protocol reports each tool call as it happens.

## Actual behavior

The turn clock counts (`ui-web/src/features/subagents/TurnClock.tsx`, new in #405) and prose appears when the agent speaks. Nothing else. Reading the history over `/rpc` during a run shows the work that was invisible on screen at the time -- "13 pages of structure settled", "wrote 4606 bytes to build/build.py" -- while the page showed a clock.

## Where it is decided

`AcpTurn`'s update dispatch splits three kinds of update three ways (`raven/acp_client/acp_agent.py`):

| update | appended to `self.events` (stored) | forwarded to `on_delta` (live) |
| --- | --- | --- |
| answer text | yes | yes |
| tool call / tool result (`_BREAKING_UPDATES`) | yes | **no** |
| thought (`_THOUGHT_UPDATES`) | yes | **no** |

Only the text branch calls `await self._on_delta(text)`. The other two branches build the record and return.

There is a second, smaller effect on top: the instance surface refreshes on a 2s `setInterval` (`ui-web/src/live/230-tabs.js:229`) rather than on an event, so even stored rows arrive on the next tick rather than as they land.

## Why this is a seam and not an oversight

`on_delta` is typed `Callable[[str], Awaitable[None]]`: a string, nothing else. A tool call cannot travel through it without widening the signature, and that signature is shared by every backend implementation (`acp_agent`, `openai_api`, `raven_loop`, `cli_agent`, `routing`, plus anything third-party). So this is a protocol-level change, not a UI one, which is presumably why it landed where it did.

The codebase has already named the same gap once, from the storage side -- `raven/agent/subagent/activity.py`:

[...]

PR #567, opened 2026-09-20
fix(*): live spawn record, live usage and a leaner node panel subtitle

0xKT:
## Summary

Two defects in the desk tasks tab's node panel, both reported from the page.

1. A spawn task's node panel read its record once, when opened, and never again. `useNodeRecord` re-read only on `store.nodeVersion`, which only a dag's `node_updated` frame bumps, and its key held no status; so a panel left open on a running spawn kept the first read's single step while the transcript's own spawn card moved on, and the final answer never landed either. The panel now re-reads a running spawn's record on the one-second beat the transcript card already reads `subagent.context` on (skipping a beat while a read is still out, and only while a tasks source is wired), and re-reads once on any status transition, which also covers a dag node settled by a `run_completed` frame that no `node_updated` announced.
2. The subtitle read `Raven-Code @raven-code-db8bbb / done / 2m55s / usage not reported / 14 tools / 1 failed`. It now reads agent, status word, duration and, only when the lane reported usage, the token total: `Raven / running / 1m50s / 4,910 tokens`. The handle stays on the board card and the work-order tab; the tool count stays on the board card, and the calls themselves are in the process fold (whose folded multi-call rows still say how many failed). The two catalogue keys nothing reads any more are removed, the TUI's generated copy is regenerated, and the design note says what the header shows and how often a record is re-read.

3. A running node's usage was invisible until it settled: `tasks.list` read tokens, tool counts and files from the record on disk, which the run writes when it finishes. The reader now overlays those facts from the activity being collected for the node in this process (the same in-memory account `subagent.context` and `dag.node` serve a transcript from), keyed by the record id for a spawn and by `node_live_key` for a dag node, only for a node that is running and only for what the lane has reported so far. A dag node that finishes while its run is still going keeps the account the runner set aside for it (`activity.record_settled`) until the manifest is written, so its total does not vanish between the two. On the page, each read of a running node's record also re-reads its row through `tasks.list(kind, id)`, so the subtitle's token total moves during the run (one row read out at a time, so a dag node's per-tool-call frames do not stack requests), and a dag node's own terminal frame re-reads the row too, so the node's final total lands even while a sibling keeps the run going; `reconcile` keeps a frame that landed while such a read was out, so a read started on the beat cannot put a settled row back to running.

No wire change: `TaskNode`'s fields and their null semantics are unchanged; a lane that has not reported usage yet still reads null. The in-process lane reports usage after every model call, so its total moves during the run; the acp lane reports once at the end of its turn, so its total appears when the run settles.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run in the fix worktree, base `refactor/ui_web_architecture` at 6062b367:

- `npm test --prefix ui-web -- src/features/tasks`: 8 files, 189 tests passed. Seven tests are new (the beat, the skipped beat while a read is out, no beat for a dag node, the settle re-read and the beat stopping, the beat stopping on close, the row re-read on the beat reaching the subtitle, `reconcile` keeping a frame that landed during its read) and three changed for the subtitle; the tests that reproduce each defect were run against the pre-fix source and failed there.
- `npm run lint --prefix ui-web`: 0 errors, 5 pre-existing warnings in other domains.
- `npx tsc --noEmit -p ui-web`: clean.
- `npm test --prefix ui-web -- scripts/gates`: 40 files, 158 tests passed.
- `npm run lint:i18n --prefix ui-tui`: generated catalogue up to date.
- `make lint-python` and `make test-python`: 23980 passed, 109 skipped.
- In `ui-tui`: `npm run lint`, `npm run lint:rpc`, `npm run type-check`, `npm test` (2074 passed), `npm run build`.
- `uv run pytest tests/test_rpc_tasks.py`: 43 passed (3 new: a running spawn's usage, counts and files off the live activity and nothing once the block closes; a live lane that has not spoken keeps its nulls; a running dag node's usage off `node_live_key`).
- `python3 ui-web/build.py`: both boot snapshots match their golden (250 and 251 nodes).
- Real gateway (`raven serve` from the worktree with the rebuilt page): a Raven-Code spawn dispatched over RPC, its node panel opened while running and left open. Sampled every 3 s, the process fold's tool rows went 1, 3, 4 without the panel being closed; the answer appeared when the run settled; the subtitle read `Raven-Code / running / 6s` and then `Raven-Code / done / 51s` (the acp lane reports usage once, at the end of its turn).
- Real gateway on an isolated `RAVEN_HOME` whose default model is deepseek, so the in-process `Raven` lane runs: a three-step spawn (each step a `sleep 15` shell command that needs the previous output). Sampled every 3 s with the panel left open, the subtitle read `Raven / running / 6s / 1,812 tokens`, then `3,769 tokens`, then `5,852 tokens`, and `Raven / done / 49s / 8,021 tokens` once settled; the board card's tool chip went 1, 2, 3; `tasks.list` reported the same growing usage over RPC.

- [x] User-facing docs or screenshots are updated when needed (docs/specs/2026-09-18-desk-tasks-list-design.md, sections 4 and 7)

## Risk

- [x] Security impact considered: none; the change is confined to rendering and to calling an existing read-only method (`subagent.context`) once a second per open panel while a spawn runs, the cadence the transcript card already uses.
- [x] Backward compatibility considered: the two removed catalogue keys had one consumer, removed in the same change; `gui.tasks.tools_n` stays in use by the board card.
- [x] Rollback path is clear for risky changes: revert the squash commit.

## Related Issues

N/A for closing. #461 is the instance surface's version of the same class of gap (a running sub-agent's tool calls not streamed live) and is not closed by this.


gloryfromca (commented):

I reviewed the full target diff and the surrounding task source, store, live-event reducers, node-panel lifecycle, transcript polling precedent, and relevant history. The running-spawn beat is scoped to an open panel, serializes reads, stops on close or settlement, and the status dependency provides the final refresh. The subtitle/catalogue changes have no remaining consumers and preserve the board/process presentation of tool data.

I also checked AGENTS.md, CONTEXT-MAP.md, ui-web/CONTEXT.md, and ui-web/CONTRIBUTING.md; backward compatibility; test changes for weakening; and the architecture gates. I found no plain error or blocker.

Verification:
- `npm test --prefix ui-web`: 186 files, 2551 tests passed.
- `npm test --prefix ui-web -- src/features/tasks`: 8 files, 187 tests passed.
- `npm test --prefix ui-web -- scripts/gates`: 40 files, 158 tests passed. An initial concurrent run timed out in `boot-order.test.mjs`; the isolated rerun and the complete rerun both passed.
- `npm run lint --prefix ui-web`: 0 errors, 5 warnings on untouched files.
- `npm run lint:i18n --prefix ui-tui`: generated catalogue is current.
- Large-file and source-language checks against `github/refactor/ui_web_architecture...HEAD`: passed.
- `git diff --check`: passed.

0xKT:
Pushed b97a49c2 on top of the reviewed head: the user asked for the running node's token total to move during the run, which needed one backend change, so the PR title now reads `fix(*)` and the description has a third item.

- `tasks.list` overlays a running node's `tokens_in` / `tokens_out`, tool counts and `files` from the activity being collected for it in this process (`run_activity.live(<record id>)` for a spawn, `run_activity.live(node_live_key(run_id, node_id))` for a dag node), only for what the lane has reported so far; a lane that has not spoken keeps its null. The wire shape and its null semantics are unchanged.
- The page re-reads a running node's row (`store.reconcile`) on each read of its record, so the subtitle's token total follows the beat and every `node_updated` frame. `reconcile` now keeps a frame that landed while its read was out (the rule `refresh` already follows), so a read started on the beat cannot put a settled row back to running.
- Tests: three backend (`tests/test_rpc_tasks.py`, 43 passed) and two frontend (tasks domain 189 passed). Real-gateway pass with the in-process lane: the subtitle read `Raven / running / 6s / 1,812 tokens`, then `3,769`, `5,852`, and `Raven / done / 49s / 8,021 tokens` once settled. The acp lane reports usage once at the end of its turn, so its total appears when the run settles.

Please take another look at the new commit.

gloryfromca (commented):

I reviewed the new live-usage commit and rechecked the complete target diff. The activity overlay uses the same spawn and DAG keys and lifetimes as the existing record readers, preserves null for facts a lane has not reported, and does not change the wire shape. On the page, the row reread follows the existing record cadence and the reconcile guard prevents a live frame from being overwritten by an older response.

I covered the changed callers and data flow through activity collection, `tasks.list`, the task source/store, the node panel, and the concurrent RPC transport; relevant history; AGENTS.md and the UI context/contributing rules; backward compatibility; test changes for weakening; and architecture constraints. I found no concrete failure or outstanding suggestion.

Verification:
- `uv run pytest tests/test_rpc_tasks.py`: 43 passed.
- `npm test --prefix ui-web -- src/features/tasks`: 8 files, 189 tests passed.
- `npm test --prefix ui-web`: 186 files, 2553 tests passed.
- `npm test --prefix ui-web -- scripts/gates`: 40 files, 158 tests passed.
- `npm run lint --prefix ui-web`: 0 errors, 5 warnings in untouched files.
- Relevant Python Ruff check and format check: passed.
- `npm run lint:i18n --prefix ui-tui`: generated catalogue is current.
- Large-file, source-language, and `git diff --check` checks: passed.

0xKT:
Pushed 6578c175, from a second read of the live usage commit; the description's third item says the same in prose.

- `_overlay_live` now applies to a running node only. The live index is keyed by a record id that is unique per conversation, so a finished spawn could have taken the numbers of another conversation's run collected under the same id. (The same key serves `subagent.context`'s live transcript read, which predates this PR; scoping that key by session would be its own change.)
- A dag node that finished while its run was still going read as reporting nothing again -- its account reaches disk with the manifest, written once the whole run is over, and `collecting()` drops the live entry as the node's block exits -- so the total the panel showed while the node ran vanished until the run ended. The runner now sets the account aside (`activity.record_settled`) at the node's end, `tasks.list` reads it there when the manifest has no entry yet, and the run forgets the copies once the manifest carries them, or on a hard stop.
- One row read is out at a time per open panel, so a dag node's per-tool-call frames do not stack `tasks.list` requests.
- Tests: `tests/test_rpc_tasks.py` 45 passed (settled account read and forgotten; a settled spawn ignores a live activity under its key), `tests/test_subagent_activity.py` and a two-node runner test (node b sees a's set-aside account mid-run; the index is empty after the manifest), tasks domain 191 passed (dag row re-read on `node_updated`; one row read at a time).

gloryfromca (commented):
Blocking: the terminal DAG node update must fetch the retained final usage instead of leaving the subtitle stale until the whole run ends.

I reviewed the new corrective commit and rechecked the complete target diff. I covered the activity index lifecycle, DAG runner event ordering and cleanup paths, `tasks.list`, the task live reducer/store, the node-panel effects, relevant history, backward compatibility, test changes for weakening, AGENTS.md and the UI architecture/domain rules. The settled-account backend path is sound, but the page does not consume it on the node's terminal event; the inline finding is the one blocker.

Verification:
- `uv run pytest tests/test_rpc_tasks.py tests/test_subagent_activity.py tests/test_subagent_dag_runner.py -k 'tasks or activity or finished_nodes_account_is_set_aside_until_the_manifest_is_written'`: 61 passed.
- `npm test --prefix ui-web -- src/features/tasks`: 8 files, 191 tests passed.
- `npm test --prefix ui-web -- scripts/gates`: 40 files, 158 tests passed.
- UI type check: passed.
- UI lint: 0 errors, 5 warnings in untouched files.
- Relevant Python Ruff check and format check: passed.
- TUI generated-catalogue, large-file, source-language, and `git diff --check` checks: passed.

ui-web/src/features/tasks/TasksPage.tsx
```diff
@@ -340,78 +340,102 @@ interface RecordLoad {
...
     if (!src) { setState((prev) => ({ loading: false, record: prev.record, failed: true })); return }
+    reading.current = true
+    /* The row too, while the node runs: its usage and tool counts grow on the
+       server as the lane reports them (`tasks.list` reads the live activity),
+       and no frame carries them -- so the subtitle's token total moves with
+       the record. One row read out at a time: a dag node's frames arrive once
+       per tool call, and stacking a read per frame would multiply requests
+       the way the record's own `reading` guard exists to prevent. Once the
+       node settles, the terminal frame's own reconcile brings the final copy. */
+    if (node.status === 'running' && !reconciling.current) {
```
gloryfromca, line 389: Blocking: refresh the row when a DAG node settles, even if another row read is already in flight.

`_run_node` stores `did.as_meta()` in the new settled index before it emits the terminal `dag.node_updated`. That event changes this node to `completed` and bumps `version`, so this effect runs but fails the `node.status === 'running'` condition; `onNodeUpdated` itself does not reconcile. If the previous running read is still out, `store.reconcile` correctly drops it because the terminal frame is newer, but nothing queues the replacement. Consequently a node's final model-call usage (and an ACP lane's end-of-turn usage) stays at the pre-terminal value until `dag.run_completed`; with another DAG node still running, that can be the rest of the run. Please cover a completed node while a sibling remains running, including the in-flight-read case, so the retained account actually reaches the subtitle.

0xKT:
| Finding | Verdict | Change |
| --- | --- | --- |
| Inline at `TasksPage.tsx:389`: a dag node's terminal `node_updated` frame did not re-read the row, so the settled account stayed off the page until `dag.run_completed` | accept | `applyNodeUpdated` now answers a `refetch` for a node's own terminal frame (`completed`, `failed`, `skipped`, `cancelled`, `interrupted`, `exception`), the way `applySubagentStatus` and `applyRunCompleted` already do; `onNodeUpdated` goes through `apply`, so the store reconciles the row on that frame. A row read that was out when the frame landed is still dropped by the `liveTick` guard, and the frame's own reconcile is the replacement. |

Verified: the frame is emitted after `_run_node` records the set-aside account (`activity.record_settled` in the block's `finally`, before the emit at the end of the node), so the read the frame triggers finds it.

Tests (pushed as 2612503d): `live.test.ts` pins which statuses ask for a reconcile and that a running frame does not; `store.test.ts` pins that a node's terminal frame reads the row and a running frame does not; `TasksPage.test.tsx` covers the case you asked for -- a node completes while its sibling runs, with the panel's own row read still out: the frame's reconcile lands the final `4,910 tokens` and the stale read, released afterwards, does not put the node back to running. Tasks domain 194 passed.

ui-web/src/features/tasks/TasksPage.tsx
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at 2612503d0c4a: terminal node frames now schedule the row reconcile through `apply()`, and the regression test covers a completed node with a sibling still running plus an older in-flight panel read. The finding is fixed.

gloryfromca (commented):

The terminal-frame change fixes the outstanding stale-usage failure: `applyNodeUpdated` requests reconciliation for terminal node statuses, `onNodeUpdated` routes it through the store's `apply` path, and the regression test covers a completed node while a sibling continues plus rejection of the older in-flight pane read.

Review coverage: the new commit and the full target diff; AGENTS.md/CLAUDE.md, CONTEXT-MAP.md, and ui-web/CONTEXT.md; callers and event/settled-record ordering; history and backward compatibility; the source/seam/store architecture constraints; and whether tests were weakened (they were not). No new candidate finding survived refutation.

Verification:
- `npm test --prefix ui-web`: 186 files, 2,558 tests passed
- `npm test --prefix ui-web -- src/features/tasks`: 8 files, 194 tests passed
- `npm test --prefix ui-web -- scripts/gates`: 40 files, 158 tests passed
- `npm run lint --prefix ui-web`: 0 errors, 5 pre-existing warnings in untouched files
- diff, source-language, and large-file checks: passed

gloryfromca (approved):

PR #569, opened 2026-09-20
fix(subagent): announce a cancelled run to its parent, with the reason

ZuyiZhou:
## Summary

A sub-agent run that was cancelled told nobody. `_run_subagent_inner`'s `CancelledError` branch wrote the record and emitted a `subagent.status` event, then re-raised; the status event is live-only and never replayed, and no announcement was injected into the parent conversation. On 2026-09-20 a poster run its parent had restarted was cancelled by `raven web --stop`, and the parent session kept its "started" receipt with nothing after it: the files that run had produced were never mentioned again.

What changes:

- **Every cancel names its reason.** The four cancel methods take a keyword `reason` and record it per task before cancelling; the run pops it on its way out. The gateway shutdown, the generation reload, the TUI exit, the served shutdown and the `/stop` command each pass a specific reason; the client-facing `subagent.interrupt` / `cancel_instance` / `cancel_session` methods fall back to the manager's defaults (`a user stopped this run`, ...). A cancellation that reached the run without a recorded reason (a bare `task.cancel()`, a parent task torn down) is announced as such rather than with an invented reason.
- **A run cancelled before its first step is reported too.** `spawn()` returns once the task exists, and asyncio closes an unstarted coroutine without entering it, so a cancel that lands before the first step runs neither handler in the body. `spawn` now puts the task text, summary and origin aside under the task id; `_run_subagent` pops that entry as its first statement; the cancel methods report a still-present entry themselves (`_finish_unstarted`), through the same helper the not-dispatched frame uses, so the registry row goes to `cancelled`, the status event is emitted and the parent is announced to. The not-dispatched frame also writes the `cancelled` registry row it previously left at `pending`.
- **The cancellation is announced** through the same `_announce_result` a returned or failed run uses: header `[Subagent '<label>' was cancelled]`, the reason, the instance handle to resume from, the record when the run wrote one. A run cancelled while still queued behind the dispatch gate (never dispatched, no record) is announced from the outer frame. The announce is guarded: a manager with no submit wired logs instead of asserting, and an announce that fails is logged, so the `CancelledError` always propagates.
- **The record carries the reason**: `error.md` reads `Cancelled: <reason>`, beside the `status: cancelled` the meta already had.
- **Every announcement names its working directory** (`Working directory: <path the run was dispatched to>`), so the parent can find what a run left on disk without searching for it. The redirect a design-engine run makes into its own `designs/<slug>-<hash>/` sub-directory is a separate change on the child side and is not in this PR.
- **`cancelled` joins the delivered-marker status** on the wire: `TurnStartedDelegated`, `SubagentDeliveredPayload` and `TranscriptDelegated` in `raven/rpc/models.py`, the three enums in `rpc-schema/openrpc.json` (edited by hand), the regenerated `ui-tui/src/rpc/generated.ts` and `ui-web/src/rpc/generated.ts`, the TUI's `deliveredMessageKey` and the web page's `DELIVERED` record, plus the new `gui.deleg.delivered_cancelled` string in `i18n/messages.json` (en + zh). Without it the web page normalised an unknown status to `ok` and drew a stop as "finished; its result just joined this conversation".

Not changed: what happens to the announcement when the process itself is exiting. At `raven web --stop` the announce is submitted before the scheduler drains (sub-agents are cancelled before the spines tear down, which is the existing order), so it lands in the parent's lane and gets the shutdown grace like any other queued turn; whether it runs to completion depends on that grace. Persisting an undelivered announcement across a restart is not attempted here.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

At head 6d313490, rebased onto origin/main at 350bd5a6, which carries #528's shutdown gate for the same manager. `git range-diff` against the previous head b037bee8: the second commit replayed unchanged; the first differs where it meets #528. `cancel_all` closes dispatch admission first, as #528 has it, then records each run's reason before cancelling; the gateway keeps #528's order (page mount torn down after the sweep) with this PR's reason on the sweep; `raven serve` had its sweep moved ahead of the seal by #528, so the reason travelled with it. Three of #528's tests named the bare `cancel_all()` and were adjusted the way this PR's first round adjusted the others: the gateway's second source-order scan looks for `cancel_all(reason=`, the TUI's AsyncMock side effect and the bootstrap's failing-cancel stub accept the keyword. Runs at this head:

```
uv run pytest tests/test_subagent_manager.py tests/test_rpc_schema_match.py -q -n 0   # 538 passed
grep -rlE "cancel_all|cancel_by_(id|session|instance)|_announce_|subagent\.delivered|delivered_cancelled|TurnStartedDelegated|SubagentDeliveredPayload" tests/*.py | xargs uv run pytest -q
# 1874 passed across the 26 files that touch the cancel paths, the announce, or the delivered marker
uv run pytest -q --ignore=tests/integration
# 24013 passed, 110 skipped at head 6d313490, in a throwaway worktree with COLORTERM=truecolor
make check-source-language check-large-files                # exit 0 (COMMIT_RANGE=origin/main..HEAD)
cd ui-tui && npm run type-check && npm run lint && npm run lint:rpc && npm run lint:i18n   # pass; the 23 lint warnings are pre-existing and in files this PR does not touch
cd ui-web && npm run type-check && npm run gen:check         # pass
NODE_OPTIONS=--no-experimental-webstorage npm test  # ui-tui: 2074 passed; ui-web: 1852 passed
```

Revert-to-red, each production file reverted to `origin/main` on its own with the tests kept:

```
raven/agent/subagent/manager.py reverted    5 failed
raven/rpc/models.py reverted                5 failed
rpc-schema/openrpc.json reverted            5 failed
```

```
inner CancelledError branch does not announce     3 failed
outer (never dispatched) branch does not announce 1 failed
reason never recorded (falls back to unexplained) 3 failed
cancelled announced with the failed wording       3 failed
no Working directory line                         2 failed
error.md written without the reason               1 failed
control (nothing mutated)                         521 passed
```

Second round, the never-started path (160 tests in `test_subagent_manager.py`):

```
manager.py reverted to the first-round head        3 failed
_cancel skips never-started runs                   1 failed
cancel_all skips never-started runs                1 failed
body does not claim its entry (double report)      4 failed
registry row left at pending                       1 failed
no cancelled status event                          2 failed
control                                            160 passed
```

Five existing tests changed with the fix, all for the new keyword: four pin the shutdown order by scanning the source for the literal `cancel_all()` call and now look for `cancel_all(reason=`; three stub `cancel_all(self)` and now accept `reason`; the `cancel_all` stub in `test_subagent_manager.py` gained the `_cancel_reasons` attribute the method now reads.

- [x] User-facing docs or screenshots are updated when needed (CHANGELOG `Unreleased / Fixed` entry)

## Risk

- The parent conversation now receives one more inbound turn per cancelled sub-agent run, which triggers a main-agent turn to summarise it. A `/stop` on a session with N running sub-agents produces N such turns.
- Every sub-agent announcement carries a new `Working directory:` line between the task and the result.
- Clients validating the wire contract see a fifth delivered status, `cancelled`. Both bundled UIs render it; a third-party client built against the previous enum would reject the event.
- The four cancel methods gain a keyword-only `reason`; positional callers are unaffected.
- A spawn cancelled before its task ever ran now produces the `cancelled` registry row, status event and announcement it previously produced nothing for. A bare `task.cancel()` from outside the manager on such a task still reports nothing; no caller in the repo does that.

Rollback: revert the squash commit. No stored data changes shape; a record written with `error.md` reads back under the old code as a cancelled run with an error file.

- [x] Security impact considered (the reason strings are chosen by the gateway, not by the sub-agent; the fenced result text is system-authored)
- [x] Backward compatibility considered (keyword-only parameter with defaults; enum extended, not changed)

## Related Issues

N/A


raven/agent/subagent/manager.py
```diff
@@ -2144,17 +2202,29 @@ def _inject(self, content: str, origin: dict[str, str], delegated: dict[str, str
...
 
-    async def cancel_by_session(self, session_key: str) -> int:
-        """Cancel all subagents for the given session. Returns count cancelled."""
-        tasks = [
-            self._running_tasks[tid]
+    async def _cancel(self, live: dict[str, asyncio.Task], reason: str) -> list[asyncio.Task]:
+        """Cancel ``live`` with ``reason`` on record for each run, and wait them out."""
+        for tid, task in live.items():
+            self._cancel_reasons[tid] = reason
+            task.cancel()
```
gloryfromca, line 2328: Blocking: handle cancellation before the task's first event-loop step.

`spawn()` returns after creating/tracking the task, so a caller can immediately call `cancel_by_id()` before the new coroutine runs. In that sequence, `task.cancel()` prevents `_run_subagent()` from entering at all; neither `CancelledError` handler executes. I reproduced this with `await spawn(...)` followed directly by `await cancel_by_id(...)`: cancellation returned `True`, the task disappeared, and the submit sink received zero announcements. The parent therefore retains exactly the stale started receipt this change is intended to close. Please make the cancel path cover a never-started task too and add a regression test that does not yield before cancelling.

gloryfromca (commented):
Blocking: an immediately cancelled run still leaves its parent with only the started receipt.

I reviewed the repository rules and context maps, the full diff, cancellation callers and lifecycle, relevant history, RPC/client backward compatibility, test changes for weakening, and the Runtime/TUI architecture boundaries. The focused Python suite passed (694 tests); the web transcript suite passed (156 tests), and its type/generated checks passed; the TUI message/chat-stream suites passed (91 tests), and its type/i18n/RPC-generation checks passed. The inline race remains reproducible and needs coverage before merge.

raven/agent/subagent/manager.py
ZuyiZhou: Reproduced on the reviewed head (1b4b8ed2) with your sequence: `await spawn(...)` then `await cancel_by_id(...)` with no await between them returned `True`, the task left `_running_tasks`, the submit sink saw nothing, and a spy on `_run_subagent` recorded that the body was never entered (`entered == []`).

Fixed in d453b478. `spawn` now puts the task text, the summary and the origin aside under the task id (`_unstarted`); `_run_subagent` pops that entry as its first statement, before any await, so an entry still present when a cancelled task is done means the body never ran. `_cancel` and `cancel_all` then report such a run themselves through `_finish_unstarted`, which shares one helper (`_report_undispatched_cancel`) with the not-dispatched frame: the registry row goes to `cancelled` (the not-dispatched frame left it at `pending` before; fixed in the same helper), the `subagent.status` event is emitted, and the parent gets the same `[Subagent '...' was cancelled]` announcement with the reason. A run whose body did start is untouched, because it claimed its entry first.

I took the report-at-the-call-site option rather than yielding once inside `spawn` to force the first step: the latter would have run `build_executor` and the record's file writes inside the caller's turn, which is a timing change the receipt does not promise.

Tests, all in `tests/test_subagent_manager.py`: `test_a_run_cancelled_before_its_first_step_is_still_announced` cancels with no await after `spawn`, asserts `entered == []` so it provably reached the unstarted case, and checks the announcement, the `pending -> cancelled` status events through a delivery sink, and the registry row; `test_cancel_all_reports_the_runs_it_stopped_before_they_started` covers the shutdown sweep; `test_a_run_that_did_start_is_reported_once` pins that the two paths do not both fire. The `cancel_all` stub in `test_cancel_all_gives_up_on_a_run_that_ignores_its_cancellation` gained the new attribute.

Mutation, each alone, control restored between runs (160 tests):

```
manager.py reverted to the reviewed head            3 failed
_cancel skips never-started runs                    1 failed
cancel_all skips never-started runs                 1 failed
body does not claim its entry (double report)       4 failed
registry row left at pending                        1 failed
no cancelled status event                           2 failed
control                                             160 passed
```

Residual I want to state rather than hide: a bare `task.cancel()` from outside the manager on a never-started task still reports nothing, since no cancel method runs. No caller in the repo cancels a spawn's task directly; every path goes through the four cancel methods.

gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in d453b4781cb0. I reran the exact no-yield sequence: the cancelling path now emits the cancelled registry/status state and submits one parent announcement, while a run that entered its body is still announced only once. The three dedicated regression tests pass.

gloryfromca (commented):

The new pre-start metadata handoff fixes the reported no-yield cancellation race, updates the pending registry row, emits the terminal status and parent announcement, and avoids duplicate reporting after the task body starts. I reviewed the repository rules and context maps, the full diff and revision delta, cancellation callers and lifecycle/history, RPC/client compatibility, tests for weakening, and the Runtime/TUI architecture boundaries.

Verification: the changed manager suite passed 160/160; the three new regression tests passed 3/3; and the focused manager/RPC/CLI/runtime suite passed 697/697. A broader 1,265-test run stopped after 505 passes with two failures in unchanged optional-memory-backend tests; both reproduced individually because this environment has no `everos-memory` provider (one also attempted its pre-existing `/subagents` path). I did not weaken or skip them to claim a green broader run. The previously blocking thread is replied to and resolved.

ZuyiZhou:
Rebased onto main at cf3c95fd to clear the changelog conflict: this branch and main had both inserted at the top of Unreleased / Fixed, so the only manual resolution was the order of the entries; nothing else conflicted. `git range-diff` against d453b478: the second commit replayed `=`, the first differs only in that changelog hunk.

Rerun on the rebased head b037bee8: `tests/test_subagent_manager.py tests/test_rpc_schema_match.py` 526 passed; the 26 files that touch the cancel paths, the announce or the delivered marker 1855 passed; ui-tui type-check, lint (23 pre-existing warnings, 0 errors), lint:rpc, lint:i18n and 2074 tests pass; ui-web type-check, gen:check (179 methods) and 1852 tests pass; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. Full suite in its own worktree: 23960 passed, 110 skipped. A first full run, taken while three other full suites shared the machine, had one failure in `tests/test_shell_machine_channel.py::test_the_cap_fires_on_the_machine_and_takes_the_whole_group`, a one-second process-group cap timing test in a file this PR does not touch; it passed 8/8 rerun alone and in the clean full run above. The description's head and numbers are updated to this head.

gloryfromca (commented):

I reviewed the rebased revision itself: the range-diff from the previously clean head, the changelog conflict resolution, the intervening changes on main in touched RPC/CLI/UI files, the full resulting PR diff, caller and architecture compatibility, repository rules/context, and whether tests or coverage were weakened. The implementation and regression-test commits are patch-equivalent; only the changelog insertion context changed, and both sets of entries are preserved. The previously resolved cancellation finding remains fixed.

Verification on a detached worktree at exact head b037bee874a4: `uv run pytest tests/test_subagent_manager.py tests/test_rpc_schema_match.py -x` passed 526/526. `git diff --check github/main...github/pr/569` also passed.

ZuyiZhou:
Rebased onto main at 350bd5a6, which now carries #528's shutdown gate for the same manager, so this time the conflict was real and not only the changelog. Four files conflicted and each was resolved by keeping both intents:

- `manager.py`: `cancel_all` closes dispatch admission first (#528), then records each run's reason before cancelling it (this PR). The second commit, the never-started report, replayed unchanged on top.
- `gateway_commands.py`: #528's order stands (the page mount is torn down after the sweep); the sweep carries this PR's reason.
- `bootstrap.py`: #528 moved the sweep ahead of the spine seal; this PR's reason travelled with it, and the old block this branch had edited in place is gone with #528's move.
- `CHANGELOG.md`: entry order only.

Three of #528's tests named the bare `cancel_all()` and would have broken without a textual conflict: the gateway's second source-order scan now looks for `cancel_all(reason=`, the TUI's AsyncMock side effect and the bootstrap's failing-cancel stub accept the keyword. That is the same adjustment this PR's first round made to the tests that existed before #528.

Rerun on the rebased head 6d313490: `tests/test_subagent_manager.py tests/test_rpc_schema_match.py` 538 passed; the 26 files that touch the cancel paths, the announce or the delivered marker 1874 passed; #528's own suites beside them (`test_cli_gateway_commands`, `test_cli_tui_commands`, `test_rpc_bootstrap`, `test_subagent_dag_runner`, `test_core_runtime_swap`, `test_generation_swap`) 1223 passed in one run with the manager tests; ui-tui type-check, lint, lint:rpc, lint:i18n and 2074 tests pass; ui-web type-check, gen:check and 1852 tests pass; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. Full suite in its own worktree: 24013 passed, 110 skipped. The description's head, the rebase note and the numbers are updated to this head.

gloryfromca (commented):

I reviewed the real conflict resolution against #528: `cancel_all` closes dispatch admission before taking its snapshot, records each task's reason before cancellation, and every host keeps its spine open until that sweep completes. The never-started-task fix replayed unchanged. I also covered the repository rules and context maps, full resulting diff, range-diff/history, all cancellation and DAG-admission callers, RPC/client backward compatibility, architecture boundaries, and test changes for weakening. The previously resolved finding remains fixed and its thread is still resolved.

Verification: manager/schema suites passed 538/538; gateway/TUI/RPC/runtime/delegation suites passed 184/184; targeted DAG shutdown/admission cases passed 5/5; and `git diff --check github/main...HEAD` passed. A broader run stopped after 810 passes with three failures in unchanged DAG memory-record tests because this environment has no optional `everos-memory` provider; I did not weaken or omit those failures when reporting the run.

0xKT:
**Not a blocker**, and nothing I found stands as a defect. Two notes about why it stands, because in both cases the code is correct for a reason that is not written down anywhere, and a later change could take that reason away without anyone noticing.

Both started as findings in my pass and both died under their own reproduction. I am passing on the measurements rather than the claims, because the measurements are the useful part.

(1) The window between `dispatched = True` and the inner `try` is silent -- and it is safe only because nothing in it yields.

`_run_subagent` hands responsibility for reporting a cancellation to the inner frame at manager.py:1525 (`dispatched = True`), but the inner `try` whose CancelledError handler announces does not open until 1664. A cancellation landing in between is caught by the outer handler at 1530, skipped by `if not dispatched:`, and re-raised with nothing said -- and `_finish_unstarted` cannot cover it either, since the body ran and popped its `_unstarted` entry at 1505. That is exactly the "started receipt with nothing after it" this PR exists to abolish.

It cannot happen today, and the reason is worth stating: the two `await` expressions in that gap are not suspension points.

* `await ensure_node_claimed(...)` (1622) is handed a `LocalFileBackend()` whose three IO methods (prompt_backend.py:24-38) are `async def` with no `await` in the body. Awaiting a coroutine that never awaits a future does not return control to the loop, so a cancel cannot be delivered there.
* `async with index_guard(history_root)` (1621) is an `asyncio.Lock` whose `acquire` returns without yielding when uncontended, and it cannot be contended: the holder's critical section 1621-1624 runs to completion without yielding, on a single-threaded loop. Measured, with the longest real holder (dag_runner.py:399-417) run concurrently against three manager blocks on one session root: 0 waits. The same counter reports 1 the moment a yielding holder is substituted.

So the correctness of the silent window rests on "`LocalFileBackend`'s async methods happen not to await" and "no holder of this lock happens to yield". Make the history backend genuinely async -- the obvious next step for a file backend -- and the window opens, silently, in the code path this PR was written to make loud. Moving `dispatched = True` below the point where the inner frame can actually report, or having the outer handler announce whenever the inner one did not, would make the guarantee structural instead of incidental.

(2) The same shape on the double-announce path.

The inner handler announces `cancelled` and re-raises (manager.py:1736-1757). The CancelledError then unwinds through `async with executor:` at 1524, whose `__aexit__` awaits `stop()` (interfaces.py:114-120). A non-CancelledError raised there would REPLACE the CancelledError, `_run_subagent`'s `except Exception` at 1535 would fire, and `_announce_result(..., "error")` would tell the parent a second, contradictory thing about the same run.

It cannot happen today either: `DirectExecutor.stop()` is a no-op, and `boxlite_executor`'s (209-226) swallows every failure mode at 269-289, and `build_executor` (sandbox/__init__.py:80-103) returns only those two. I could only produce the double announcement by supplying an executor class that does not exist in this tree -- which is not evidence of a defect, and I am saying so rather than filing it. But the outer `except Exception` has no way to know the parent was already told, so if a third executor ever arrives whose teardown can fail, the contradiction appears with no test to catch it. A `cancelled` flag already exists in the inner frame; the outer handler could read the same fact.

WHAT I CHECKED AND FOUND SOUND. The discriminator the second commit rests on -- "an entry still present when a cancelled task is done means the body never ran" -- holds: `_run_subagent` pops at 1505 as its first statement with no await above it, `_finish_unstarted` pops-then-checks (2021), and `dict.pop` is atomic against the loop, so the two reporters are mutually exclusive by construction. The six `.cancel()` sites divide correctly: two cancel runs and both call `_finish_unstarted` (2333, 2480), and the other four cancel memory pollers, held reports, or the DAG path that settles with its caller (801). No leak: a created task either runs (popped at 1505) or is cancelled (popped at 2021).

One small thing on that: `spawn` writes `self._unstarted[task_id]` at 981, AFTER `asyncio.create_task` at 970. It is correct because the three statements between are synchronous, so the task cannot have started. But the other side of this invariant carries a comment explaining itself ("First, before any await", 1503-1504) and this side does not -- and an `await` inserted between 970 and 981 would leave a run that really ran carrying an entry that later makes a cancel report it as "stopped before its first step". Writing the entry before `create_task` would remove the invariant instead of relying on it.

GATES on head 6d31349072081d43b48bb4364a497af18fd4117e / tip 608cedf1: `GATES OK base=350bd5a6 ruff/lint-imports/commit/large/lang rc=0`. `pyt.sh tests/test_subagent_manager.py tests/test_rpc_schema_match.py tests/test_rpc_bootstrap.py tests/test_cli_gateway_commands.py tests/test_core_runtime_swap.py tests/test_generation_swap.py tests/test_cli_tui_commands.py` -> 705 passed.

0xKT (approved):

PR #572, opened 2026-09-20
feat(agent): pause a web tool whose vendor refused the key

ZuyiZhou:
## Summary

A 401 or 402 from a reader, or a 401, 402 or 403 from a search vendor, is about the key or the account, not the page or the query, so every call after it meets the same answer. The web tools kept asking: on 2026-09-20 Jina answered HTTP 402 on every `web_fetch` of a session, the tool returned one identical error envelope per call, and the model went on calling it.

`web_fetch`, `web_search` and `image_search` now remember a refusal per tool instance (`_VendorRefusal` in `raven/agent/tools/web.py`) and answer later calls without a request:

- the `error` string is the same for the refusing call and every paused call (`<Vendor> refused the key (HTTP 402)`), so `failure_class` reads them as one cause and the loop's stop-repeating nudge fires; `is_hard_tool_failure` reads the envelope as deterministic (none of the transient markers appear in the text);
- the `detail` says what the status means (401 rejected key, 402 out of credit, 403 not allowed), that this call was not sent, and what the user has to do: the vendor's `tools.web.providers.<vendor>.apiKey` slot, its env var, its sign-up page, or another vendor under `tools.web.<search|fetch>.provider`. The `Tell the user:` framing is deliberate: a model cannot fix a key or a bill, and the incident's model retried instead of saying so;
- the pause is per tool rather than per session because the key is shared by every session of the process, so one session's refusal is every session's; per-session marking would send each session into the same 402 once;
- all three tools read their key live, in both loops: `WebFetchTool` takes a callable key source the way `WebSearchTool` and `ImageSearchTool` do, the main loop hands each of them one through `_live_vendor_key`, and the sub-agent lane through one `live_key` helper over the same resolver on its module `LiveConfig` (it used to hand `web_search` and `image_search` the key it was constructed with, while their envelope told the user the slot is read on the next call). That resolver, `live_vendor_key` in `raven/config/live.py`, is one function the main loop's two live readers now delegate to (slot, then Serper's pre-vendor leaf, then the boot value; an empty-but-present slot is a revocation). A key set at `tools.web.providers.<vendor>.apiKey` therefore reaches the next call and lifts the pause without a restart, which is what the envelope tells the user to do; otherwise one real request goes through after `VENDOR_REFUSAL_PAUSE_S` (10 minutes) and re-arms the pause if refused again. The envelope's advice names what each route needs: the config slot is read on the next call, the env var and the vendor selection need a restart;
- each call reads the key once, at its start, and that one value is what the request carries and what a refusal is recorded against; the request builders take it as a parameter. Read through the live property again at `note` time, a key the user replaced while a request was in flight was recorded as refused without ever having been sent, and the next call was answered from the pause: in the main loop the tool serves every session of the process, so all of them lost it for ten minutes;
- which statuses count differs by tool, because what a status can mean differs by endpoint. A search vendor's API is the endpoint, so 401, 402 and 403 all speak of the account: Serper answers a rejected key with 403 (measured against the live endpoint). A reader's 403 speaks of the URL: Jina answers an anonymous request for a domain it has blocked with 403 for every URL under it (measured: `r.jina.ai/https://httpbin.org/status/200` answers 403 with Jina's own `AbuseAlleviationError` envelope, "Anonymous access to domain httpbin.org blocked ... due to previous abuse"), and Firecrawl answers 403 for a site its policy does not scrape ("This website is no longer supported"). So `SEARCH_REFUSAL_STATUSES` is `{401, 402, 403}` and `FETCH_REFUSAL_STATUSES` is `{401, 402}`; a reader's 403 is reported per URL exactly as before this PR;
- a request that carried no key is never paused, whatever the status, because there was no key to refuse. That is the default configuration, Jina without a key (`needs_key=False`), where review found that one blocked domain paused `web_fetch` for every URL for ten minutes and told the user to replace a key that does not exist;
- statuses outside a tool's set (500, 429, a reader's 403, ...) are reported exactly as before and pause nothing.

Not changed: which vendor is configured. The incident's Jina account is out of credit; whether to top it up or select Tavily/Exa under `tools.web.fetch.provider` is a configuration decision this PR only makes visible. Tavily reports an exhausted plan with its own 432 and 433 rather than 402; those stay outside the set and are reported per call. The research sub-agent's web tools are the fork's own classes (`agents/raven-research/.../research_flow/tools/web.py`, which imports only `raven.contracts.tool` and `raven.security.network`): they already classify these statuses as `quota_err`, carry no `_VendorRefusal`, and are untouched, so the fork has no pause to lift and gets none of this PR's behaviour. An earlier revision of this description said its pause lifts on the plugin's rebuild; there is no such pause.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

At head 38222344, four commits on origin/main at 608cedf1; the fourth is this round's. `git range-diff` against the previous head eba9f9d7: the three earlier commits replayed onto main with only context changed, where main's browser-tool registration now sits beside this PR's web-tool registrations in `raven_loop.py`:

```
uv run --frozen --python 3.12 --all-extras pytest -q -n 0 -p no:randomly tests/test_agent_tools_web_providers.py tests/test_agent_loop_web_tools.py tests/test_security_web_ssrf.py
  158 passed (17 pre-existing PytestWarnings: asyncio marks on sync tests in that file)
COLORTERM=truecolor uv run --frozen --python 3.12 --all-extras pytest -q      (full suite, in its own worktree at this head)
  24123 passed, 110 skipped in 348.21s (0:05:48)
make lint-python lint-imports lint-deps lint-types                            # all pass
PYTHONPATH=. uv run --frozen --python 3.12 --extra dev python scripts/check_source_language.py origin/main...HEAD   # exit 0
make check-commits; PR_TITLE=... make check-pr-title; COMMIT_RANGE=origin/main...HEAD make check-large-files       # exit 0
git diff --check origin/main...HEAD; git merge-tree --write-tree HEAD origin/main                                   # clean
```

The reviewer's probe, run through the real `WebFetchTool` (Jina, no key; a transport answering 403 for one host and a page for the others), at the previous head and at this one:

```
previous head 067f6232 (web.py swapped in):
  PAUSED https://blocked.example/a   Jina Reader refused the key (HTTP 403)
  PAUSED https://example.org/b       Jina Reader refused the key (HTTP 403)
  PAUSED https://example.net/c       Jina Reader refused the key (HTTP 403)
  requests sent: 1
this head:
  error  https://blocked.example/a   Jina Reader answered HTTP 403
  ok     https://example.org/b       len=4
  ok     https://example.net/c       len=4
  requests sent: 3
```

Revert-to-red: `raven/agent/tools/web.py` reverted to `origin/main`, or to the previous head, with the tests kept fails at collection (`AttributeError: module 'raven.agent.tools.web' has no attribute 'FETCH_REFUSAL_STATUSES'`, exit 2).

Mutation, third round, each mutant applied alone with the file restored between runs (the two web test files, 152 tests at this head, `-n 0`; a mutant that changes a status set also changes how many cases the parametrised tests collect, so the rows give the failures):

```
note() records a refusal from a request that carried no key    2 failed  (keyless reader, 401 and 402; its 403 case is held by the fetch set alone)
FETCH_REFUSAL_STATUSES keeps 403                                5 failed  (a reader's 403, all five readers)
FETCH_REFUSAL_STATUSES drops 402                                3 failed  (the 402 pause tests, including the loop-level key rotation)
SEARCH_REFUSAL_STATUSES drops 403                               1 failed  (image_search on Serper's 403)
web_fetch built on the search set                               5 failed
control (nothing mutated)                                     152 passed
```

Fourth round, one key read per call and the sub-agent lane's two tools (the two web test files, 155 tests, `-n 0`):

```
baseline                                                   155 passed
MUTANT  web_fetch's note() re-reads the key                1 failed   (the in-flight case for web_fetch)
MUTANT  web_search's note() re-reads the key               1 failed   (for web_search)
MUTANT  image_search's note() re-reads the key             1 failed   (for image_search)
MUTANT  web_fetch's request re-reads the key               1 failed   (the request carries a key the pause was not recorded against)
MUTANT  sub-agent lane: web_search on a boot key           1 failed   (the lane test)
MUTANT  sub-agent lane: image_search on a boot key         1 failed   (the lane test)
MUTANT  vendor switch loses its restart qualifier          34 failed  (every pause test, both tools)
REVERT  web.py + raven_loop.py to reviewed head 428cbe6b   38 failed  (the three in-flight cases, the lane test, the 34 wording cases)
CONTROL restored                                           155 passed
```

The reviewer's in-flight probe (the real `WebFetchTool` on Jina, a transport answering 402 and rotating the key source while the request is in flight), at the previous head and at this one:

```
### reviewed head 428cbe6b
  key-source reads during one execute(): 4  ['KEY-OLD', 'KEY-OLD', 'KEY-OLD', 'KEY-NEW']
  Authorization actually sent          : ['KEY-OLD']
  next call                            : answered from the pause, 0 request sent, carrying []  (api_key resolves to KEY-NEW)
### this head 38222344
  key-source reads during one execute(): 1  ['KEY-OLD']
  Authorization actually sent          : ['KEY-OLD']
  next call                            : served, 1 request sent, carrying ['KEY-NEW']  (api_key resolves to KEY-NEW)
```

Earlier rounds, kept for the record; each was measured at the head of its round, and the tests they name are unchanged since except for the parametrisations noted below. First round (131 tests in the two files, at f87ec320):

```
refusals never remembered                     25 failed
a new key does not lift the pause              1 failed
the pause never ends                           1 failed
a 500 pauses too                              19 failed
paused calls change the error string          23 failed
web_search never short-circuits                8 failed
image_search never short-circuits              1 failed
web_fetch never short-circuits                16 failed
control (nothing mutated)                     131 passed
```

Second round, the live key (133 tests, at 145344f0):

```
the four production files reverted to the first-round head   3 failed
WebFetchTool ignores its key source                          4 failed
main loop hands the reader a one-time key                    1 failed
sub-agent lane hands the reader a one-time key               1 failed
live_vendor_key never reads the file                         6 failed
control                                                     133 passed
```

Tests changed across the rounds: `test_a_new_key_lifts_the_pause_at_once` rotates the key through the callable source the loops hand the tool instead of a private field; the "vendor and status only, never the key" rule for search (was 401) and fetch (was 402) now uses a 500, since a refusal status renders differently on purpose; `test_a_reader_refusing_the_key_pauses_the_tool` is parametrised over the reader set (401, 402) and `test_a_search_vendor_refusing_the_key_pauses_the_tool` over the search set (401, 402, 403). New in the third round: `test_a_keyless_reader_is_never_paused_by_a_status[401|402|403]` (the probe above) and `test_a_readers_403_is_about_the_page_not_the_key` for every reader. Fourth round: `test_the_subagent_lanes_reader_reads_its_key_live_too` became `test_the_subagent_lanes_web_tools_read_their_keys_live_too` and covers all three tools; new: `test_a_key_rotated_while_a_request_is_in_flight_is_tried_before_it_is_paused[web_search|image_search|web_fetch]`; the two pause tests assert the remedy wording.

- [x] User-facing docs or screenshots are updated when needed (CHANGELOG `Unreleased / Fixed`)

## Risk

- After a 401 or 402 from a reader, or a 401, 402 or 403 from a search vendor, the tool answers for up to ten minutes (or until the key changes) without contacting the vendor, with an envelope/text that names the cause and the fix. A key set at the vendor's config slot is read on the next call and lifts the pause; a user who instead tops up the same account gets one retry at the window's end.
- A request that carried no key never pauses, and a reader's 403 never pauses: on the default configuration (Jina without a key) `web_fetch` behaves exactly as before this PR.
- All three tools in both loops read their key from the live config on every call (the main loop's `web_search` and `image_search` already did; `web_fetch` and the sub-agent lane's two now do), and once per call, so the key a request carried is the key its refusal is recorded against; a key removed from the file (an empty slot) is a revocation for them too.
- The refusal text says the vendor selection under `tools.web.<kind>.provider` takes effect on restart, as the env var does; only the config slot is read on the next call.
- The refusing call's own error text changes from `<Vendor> answered HTTP 402` to `<Vendor> refused the key (HTTP 402)` plus a detail, for the statuses in the tool's set. Anything matching the old text for those statuses would need updating; nothing in the repo does.

Rollback: revert the squash commit. The pause is process-local state; no config or stored data changes shape.

- [x] Security impact considered (the key is never echoed; the envelope names the config path and env var, not the value)
- [x] Backward compatibility considered (tool schemas unchanged; non-refusal statuses render exactly as before)

## Related Issues

N/A


raven/agent/tools/web.py
```diff
@@ -799,6 +892,9 @@ async def execute(self, url: str, extractMode: str = "markdown", maxChars: int |
             # The same rule as the handlers below: most of these reasons name the
             # hostname, and a reader whose every target is refused is one cause.
             return json.dumps({"error": "URL validation failed", "detail": error_msg, "url": url}, ensure_ascii=False)
+        if (refused := self._refusal.active(self.api_key)) is not None:
```
gloryfromca: Blocking: make `web_fetch` observe a replacement key before deciding that this pause is still active.

The main loop constructs this tool with the one-time value from `_web_key(...)`, and `api_key` always prefers that `_init_api_key` over the environment. Consequently, changing `tools.web.providers.<vendor>.apiKey` as this new error instructs does not change the value passed to `active`; exporting a replacement environment value cannot override an old configured value either. The tool stays paused, then sends the old key when the cooldown expires and re-arms the pause. Only rebuilding the generation or restarting recovers, contrary to the changelog and response claim that a new key lifts the pause at once. `test_a_new_key_lifts_the_pause_at_once` misses the production path because it directly assigns the private `_init_api_key` field. Please ensure the real config/key rotation path reaches this check (or make the user-facing recovery contract accurately require a reload).

gloryfromca (commented):
Blocking: `web_fetch` must observe the replacement key that the refusal message tells the user to set.

I found one blocking issue inline. I reviewed the full diff; the main-loop and sub-agent construction paths; tool instance lifetime and failure-streak handling; relevant history; AGENTS.md/CLAUDE.md and CONTEXT-MAP.md/CONTEXT.md; backward compatibility; and the changed tests for weakening. I also checked the batched image-search path: requests already in flight before the first refusal is known are not evidence of a broken pause, while queued and later requests do observe it.

Verification: `uv run pytest tests/test_agent_tools_web_providers.py tests/test_agent_loop_web_tools.py tests/test_agent_loop_failure_streak.py tests/test_agent_loop_tool_loop_break.py -q` passed (184 passed, 17 existing warnings). A focused reproduction confirmed that rotating `TAVILY_API_KEY` leaves a configured `WebFetchTool` on the old key and the refusal remains active. The GitHub test, lint, and repository checks shown at review time were green; the coverage aggregation job was still running.

raven/agent/tools/web.py
ZuyiZhou: Reproduced on the reviewed head (f87ec320) through the loop: a `web_fetch` built by `wire(...)` with `web_provider_keys={"tavily": "sk-boot"}`, a 402 from the fake vendor, then a new key written to `tools.web.providers.tavily.apiKey` in the live config file -- the next call was still answered from the pause and `tool.api_key` still read `sk-boot`. As you said, the test reached the private field and so passed.

Fixed in 145344f0. `WebFetchTool` now takes a callable key source, the way `WebSearchTool` already did, and reads it on every call (`api_key` prefers the source, then the env var). Both loops hand it one: the main loop through `_live_vendor_key(fetch_provider)`, the sub-agent lane (`raven_loop.py`, which had the same one-time value) through the same resolver on its module `LiveConfig`. That resolver is one new function, `live_vendor_key` in `raven/config/live.py`; the main loop's two live readers now delegate to it rather than each spelling the slot-then-leaf-then-boot order, so the rotation path has one definition. The envelope's advice was corrected to say what each route needs: the config slot is read on the next call, the env var needs a restart, or another vendor under `tools.web.<kind>.provider`. The CHANGELOG line says the same.

Not changed: the research sub-agent's own `WebFetchTool` construction in `agents/raven-research/.../plugin.py`, which passes its plugin's static `fetch_api_key`; that key comes from the plugin's own config lifecycle rather than the host file, and I did not want to decide its rotation contract inside this PR. Its pause therefore lifts on the plugin's rebuild, as before.

Tests: `test_a_new_key_lifts_the_pause_at_once` now rotates the key through the callable source and asserts the rotated key is what the next request carries; `test_a_reader_key_set_after_a_refusal_reaches_the_next_call` (`tests/test_agent_loop_web_tools.py`) drives the production path -- a loop from `wire(...)`, a 402 from the boot key, a paused call that is not sent, the config file rewritten, and the next call served with the rotated key; `test_the_subagent_lanes_reader_reads_its_key_live_too` does the same for the `RavenLoopBackend`-registered tool.

Mutation, each alone, control restored between runs (133 tests in the two web files):

```
the four production files reverted to f87ec320       3 failed
WebFetchTool ignores its key source                  4 failed
main loop hands the reader a one-time key            1 failed
sub-agent lane hands the reader a one-time key       1 failed
live_vendor_key never reads the file                 6 failed
control                                              133 passed
```

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed on `145344f0aee7`: `WebFetchTool` now reads the callable source on each call, both production construction paths supply the live vendor-key resolver, and the tests exercise an on-disk config rotation through the main loop plus the sub-agent registration path. The original blocker is fixed; the relevant review suite passed (223 tests, 17 existing warnings).

gloryfromca (commented):

The previous `web_fetch` key-rotation blocker is fixed. I reviewed the full `github/main...HEAD` diff and the new revision delta; main-loop and in-process sub-agent construction; the shared live-config resolver and its boot, empty-slot, and environment fallbacks; tool callers and relevant history; failure-streak behavior; backward compatibility; the changed tests for weakening; and the repository rules in AGENTS.md/CLAUDE.md and CONTEXT-MAP.md/CONTEXT.md. I found no new findings.

Verification: `uv run pytest tests/test_agent_tools_web_providers.py tests/test_agent_loop_web_tools.py tests/test_config_live.py tests/test_agent_loop_failure_streak.py tests/test_agent_loop_tool_loop_break.py tests/test_security_web_ssrf.py -q` passed (223 passed, 17 existing warnings). `git diff --check github/main...HEAD` and the merge-tree check against `github/main` were clean. GitHub lint, repository, docs, trajectory, RPC, page, bridge, and smoke checks visible during review were green; several matrix jobs were still running.

raven/agent/tools/web.py
0xKT, line 1: **Blocking:** on the default `web_fetch` configuration, one ordinary 403 page
disables the tool for every URL for ten minutes, and tells the user to fix it by
setting a key the vendor does not take.

The idea is right for the search tools, where the vendor's API *is* the endpoint
and a 401/402/403 really is about the account. The fetch path is different, and
the default backend is where it bites.

## The chain, each link measured

**1. Jina is the default fetch backend and needs no key.**
`DEFAULT_FETCH_PROVIDER = "jina"` (web.py:66) and its spec carries
`needs_key=False` (web.py:84-86). `effective_provider` also falls back to it
whenever a keyed vendor is selected without a key, so it is what most installs
run.

**2. Jina relays the target page's status.** Measured against the live service
just now, with a control:

    https://httpbin.org/status/403                    -> 403
    https://r.jina.ai/https://httpbin.org/status/403  -> 403
    https://r.jina.ai/https://example.com             -> 200

**3. `_provider_fetch` raises on it.** The Jina branch is
`r = await client.get(f"https://r.jina.ai/{url}"); r.raise_for_status()`, so a
403 page becomes `HTTPStatusError(403)`.

**4. That arms the pause.** `403 in REFUSAL_STATUSES`, so
`self._refusal.note(403, self.api_key)` records it -- with `self.api_key == ""`,
because Jina has no key.

**5. Every later call is refused.** Driven through the real `WebFetchTool` with
a stub that answers 403 for one host and 200 for the others:

    PAUSED  https://example.com/a   Jina Reader refused the key (HTTP 403)
    PAUSED  https://example.org/b   ... so this call was not sent
    PAUSED  https://example.net/c   ... so this call was not sent

CONTROL, same tree, `_VendorRefusal.note` stubbed to return False:

    ok  https://example.com/a   Jina Reader answered HTTP 403
    ok  https://example.org/b   len=9
    ok  https://example.net/c   len=9

So the pause is the cause, not the 403. And this is new: `grep -c
"_VendorRefusal\|REFUSAL_STATUSES"` on base's `web.py` is **0** -- before this PR
a 403 degraded that one URL and the next call worked.

**6. The remedy the user is given cannot be followed.** The detail reads
"Jina Reader needs attention -- set a working key at `<config_path>`", for a
vendor with `needs_key=False`. And because `active()` lifts the pause only when
`key != self.key`, and the key is `""` on both sides forever, "until the key
changes" never arrives -- the full `VENDOR_REFUSAL_PAUSE_S` elapses every time.
The model is told "Do not retry this tool until they have", so it stops asking.

## Why this is blocking rather than a note

It needs no opt-in (Jina is the default), the trigger is a page status that is
routine on the open web, the blast radius is every URL rather than the one that
failed, and the stated fix is impossible. The PR's own comment names the rule it
breaks: REFUSAL_STATUSES are "The statuses that are about the key, **not the
page** or the query" -- through a relaying reader, 403 is exactly about the page.
The same function says so twice, in opposite directions: on success it returns
`r.status_code` as the page's status, and on failure it reads the same number as
a verdict about the account.

## What would clear it

Any one of these, and I have not picked for you:

- Do not arm the pause for a vendor whose spec says `needs_key=False` -- there is
  no key to be refused, so no status from it can mean what the pause means.
- Treat the fetch path separately from search: for a backend that relays an
  upstream status (Jina does; the POST-based backends answer for their own API
  and are fine), a 4xx describes the page.
- Keep 401 and 402, which no relayed page can produce for a keyless reader, and
  drop 403 from the fetch tool's set.

The search half of this PR I have no objection to, and the live-key reading in
the second commit does answer the earlier review point -- I checked that
separately.

gloryfromca: Blocking: the default Jina fetch path must not treat a relayed page 403 as a key refusal.

I independently reproduced the factual chain, so this changes my prior clean stance. Live controls returned 403 for both `https://httpbin.org/status/403` and its `r.jina.ai` relay, while Jina returned 200 for `https://example.com`. With the real `WebFetchTool` and a deterministic relay stub, the first page 403 produced `Jina Reader refused the key`, the next unrelated URL returned `paused: true` / `not sent`, and the transport recorded only the first call. The focused suite still passes (136 passed, 17 existing warnings), so it does not cover this distinction. I agree that the keyless default reader needs to distinguish relayed page status from vendor credential refusal before merge.

ZuyiZhou: Reproduced on the reviewed head with your recipe: the real `WebFetchTool` on Jina with no key, a transport answering 403 for one host and a page for the others. The first URL returned `Jina Reader refused the key (HTTP 403)` and the next two were answered from the pause and not sent (one request in total). The same probe at ca0a3c4e: `Jina Reader answered HTTP 403`, then both pages served, three requests sent. The description's verification block carries both transcripts.

Two measurements from the live service while there, which sharpen the mechanism without changing your conclusion:

- Today `r.jina.ai/https://httpbin.org/status/200` answers 403 as well, with Jina's own envelope: `{"code":403,"name":"AbuseAlleviationError","status":40305,"message":"Anonymous access to domain httpbin.org blocked until ... due to previous abuse ..."}`. So the 403 there is Jina's per-domain verdict on an anonymous request, given for every URL under that domain, rather than the page's status relayed; a page's own 401, 402 or 403 (`httpbingo.org/status/401|402|403`, `api.github.com/user`) comes back as 200 text with a `Warning: Target URL returned error 403: Forbidden` line. Either way the status is about the page or its domain and never about a key, which is your point, and the "anonymous" in Jina's message says why a key the user does not have cannot be what was refused.
- With a bogus key Jina answers 401 `AuthenticationFailedError` on every URL, so 401 and 402 from a keyed reader are the reader's own verdict; that is the incident this PR started from (Jina's 402 on a keyed account). Serper answers a rejected key with 403, which is why the search set keeps it. Firecrawl answers 403 for a site its policy does not scrape ("This website is no longer supported"), so the reader-side rule is not a Jina rule.

Fixed in ca0a3c4e, taking your first and third options together rather than a relay flag:

- `_VendorRefusal.note` records nothing when the request carried no key: with no key sent there was none to refuse, whatever the status. This alone closes the default configuration.
- The status set is per tool: `SEARCH_REFUSAL_STATUSES = {401, 402, 403}`, `FETCH_REFUSAL_STATUSES = {401, 402}`. A reader's 403 is reported per URL exactly as before this PR, for keyed readers too.

Why not the relay flag (your second option): a flag that switched the pause off for Jina would have lost the case the PR exists for, since Jina's 402 on a keyed account is Jina's own verdict and not a relayed one, and Firecrawl's 403 needs the rule without relaying anything. The two rules above say what a status can mean, not who answered it.

Tests, in the shape you asked for:

- `test_a_keyless_reader_is_never_paused_by_a_status[401|402|403]` is your probe through the real tool: Jina with no key, one host refused, the next two URLs fetched, three requests sent, no `paused` key, and the error text the tool gave before this PR.
- `test_a_readers_403_is_about_the_page_not_the_key` for each of the five readers, keyed: the next URL is fetched and neither envelope carries a pause.
- `test_a_reader_refusing_the_key_pauses_the_tool` now runs over the reader set (401, 402) only; `test_a_search_vendor_refusing_the_key_pauses_the_tool` over the search set (401, 402, 403), with Serper's 403 as the reason in its docstring.

Mutation, each alone, the file restored between runs (the two web test files, 152 tests, `-n 0`; a mutant that edits a set also changes how many cases the parametrised tests collect, so the rows give the failures):

```
note() records a keyless refusal          2 failed   (keyless reader, 401 and 402; its 403 case is held by the fetch set alone)
fetch set keeps 403                       5 failed   (a reader's 403, all five readers)
fetch set drops 402                       3 failed   (the 402 pause tests, including the loop-level rotation)
search set drops 403                      1 failed   (image_search on Serper's 403)
web_fetch built on the search set         5 failed
web.py at the reviewed head               collection error (no FETCH_REFUSAL_STATUSES)
control                                 152 passed
```

Full suite at ca0a3c4e in its own worktree: 24000 passed, 110 skipped. The changelog entry and the description now say which statuses count for which tool and that a keyless request never pauses. Not changed: Tavily's own 432/433 for an exhausted plan stay outside the set and are reported per call; the research plugin's own reader is untouched, as before.

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed independently on `ca0a3c4e2d26`: the default keyless Jina path now reports a 403 per URL and sends the next unrelated request; a keyed reader's 403 also remains per URL, while keyed reader 401/402 and search 401/402/403 still arm the pause. The focused loop, provider, live-config, failure-streak, and SSRF suite passed: 242 tests, with 17 pre-existing pytest-mark warnings. This clears my blocker stance.

gloryfromca (commented):

The Jina 403 blocker is fixed: keyless fetches and keyed-reader 403s remain per-URL failures, while keyed fetch 401/402 and search 401/402/403 retain the refusal pause. I reviewed the full `github/main...HEAD` diff and the revision delta, repository instructions and domain context, both production construction paths and their callers, commit/range history, response and key-source compatibility, test changes for weakening, and the live-config architecture. `git diff --check` and the merge-tree check were clean.

Verification: `uv run pytest tests/test_agent_tools_web_providers.py tests/test_agent_loop_web_tools.py tests/test_config_live.py tests/test_agent_loop_failure_streak.py tests/test_agent_loop_tool_loop_break.py tests/test_security_web_ssrf.py -q` passed (242 tests; 17 pre-existing warnings).

ZuyiZhou:
Rebased once more, onto main at 350bd5a6: #528 merged with its changelog entry at the same anchor, so this branch conflicted there again and only there. `git range-diff` against ca0a3c4e: the three commits differ only in that changelog hunk; main's commits since the last rebase touch no file this PR touches.

Rerun on the rebased head eba9f9d7: the three web files 155 passed; the full suite 24053 passed, 110 skipped in its own worktree; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. The description's head and numbers are updated to this head.

raven/agent/tools/web.py
```diff
@@ -827,6 +938,9 @@ async def execute(self, url: str, extractMode: str = "markdown", maxChars: int |
             # message that repeats the request URL.
             status = e.response.status_code
             logger.error("WebFetch error for {}: {} answered HTTP {}", url, self.spec.label, status)
+            if self._refusal.note(status, self.api_key):
```
LivXue: **Nonblocking - read the key once per call, and pause the value that was actually sent.**

`self.api_key` is a property over a live source, so every access re-reads
`tools.web.providers.<vendor>.apiKey`. One `execute()` reads it four times: for
`_refusal.active()` at line 910, twice inside `_provider_fetch` while building the
Jina `Authorization` header, and again here at `note(status, self.api_key)`. If the
user writes a replacement key while the request is in flight, `note` records the
pause against a key the vendor never saw, and the next call is answered from the
pause without being sent -- so the replacement is blocked for the full
`VENDOR_REFUSAL_PAUSE_S` without ever being tried.

Reproduced through the real `WebFetchTool` on Jina -- I ran it at `ca0a3c4e`, and
`web.py` is byte-identical at `eba9f9d7` -- with a transport that answers 402 and
rotates the key source while the request is in flight:

    key-source reads during one execute(): 4  ['KEY-OLD', 'KEY-OLD', 'KEY-OLD', 'KEY-NEW']
    Authorization actually sent          : ['Bearer KEY-OLD']
    pause armed against key              : KEY-NEW
    next call                            : paused=true, 0 requests sent, api_key resolves to KEY-NEW

The window is one request's flight time (30s on the Jina client) and it closes once
the pause is armed, so this is much narrower than the rotation blocker fixed in
`145344f0`. But the user-visible outcome is the same one that blocker was about,
and in the main loop this tool is process-wide -- one `AgentLoop` per runtime serves
every session -- so a mis-armed pause takes `web_fetch` away from all of them for
ten minutes. `WebSearchTool.execute` (line 318) and `ImageSearchTool.one` (line 654)
have the same shape.

Resolving the key once at the top of the call and handing that one value to
`active`, to the request and to `note` would close it; I have not picked a shape
for you.

raven/agent/subagent/backends/raven_loop.py
```diff
@@ -387,8 +387,15 @@ def allowed(name: str) -> bool:
             fetch_provider = WebFetchTool.effective_provider(
                 self.web_fetch_provider, self._web_key(self.web_fetch_provider)
             )
+            # Live for the same reason the main loop's is: a refused key pauses
```
LivXue: **Nonblocking - the reason in this comment applies to the two tools above it.**

This lane registers `web_search` at line 374 and `image_search` at line 381 with
`api_key=self._web_key(...)`, a plain string resolved from `self.web_provider_keys`
/ `self.search_api_key` / `self.jina_api_key` -- all frozen when the backend was
constructed, and the backend is cached across runs. Only `web_fetch` gained the
live source. But this PR gives all three tools the same refusal pause and the same
envelope, and that envelope tells the user to set a key at
`tools.web.providers.<vendor>.apiKey` "(read on the next call)". On this lane that
sentence is false for two of the three.

Reproduced at `ca0a3c4e`, where both files are byte-identical to this head: one
tool class, the two construction shapes side by side, a 401 from the vendor, then
the key rotated.

    main loop   api_key=<callable>   api_key after rotation: KEY-NEW   requests sent (of 2): 2
    sub-agent   api_key='KEY-OLD'    api_key after rotation: KEY-OLD   requests sent (of 2): 1

The sub-agent lane's second call is answered from the pause and never sent, while
carrying advice it cannot honour.

Bounded, which is why I am not calling it blocking: the tools are built inside
`_run`, so the pause dies with the run rather than with the process. The cost is
one sub-agent run whose search is dead and whose remedy text misleads. Note also
that `test_the_subagent_lanes_reader_reads_its_key_live_too` covers `web_fetch`
here; the two siblings have no equivalent.

raven/agent/tools/web.py
```diff
@@ -116,6 +117,93 @@ def resolve_vendor_key(
...
+    outcome = (
+        f"{meaning}; no further request will be sent to {spec.label} until the key changes or "
+        f"{VENDOR_REFUSAL_PAUSE_S / 60:g} minutes pass"
+        if sent
+        else f"{meaning} on the last request, so this call was not sent"
+    )
+    detail = (
+        f"{outcome}. Tell the user: {spec.label} needs attention -- set a working key at "
+        f"{spec.config_path} in the config file (read on the next call), restart with {spec.env_var} set, "
+        f"or select another vendor under tools.web.{kind}.provider; sign-up at {spec.signup}. "
```
LivXue: **Nit - the third remedy needs a restart, and is the only one that does not say so.**

The two remedies before it are each qualified: the config slot is "(read on the
next call)", the environment variable is "restart with ... set". This one is bare,
and against those two the contrast reads as a promise that it also takes effect at
once. It does not -- `provider` is fixed at registration and feeds both `spec` and
the endpoint for the life of the tool, so selecting a different vendor under
`tools.web.<kind>.provider` needs the process restarted, exactly like the env var.

Code-reading only; I did not run this one. A qualifier matching the env-var branch
would settle it.

LivXue:

Scope: I reviewed only the `raven/agent/**` portion of this PR -- 3 of 7 changed
files. The remaining 4 (`raven/config/live.py`, `CHANGELOG.md`, and the two test
files) were read as context but are not covered by this verdict.

What I examined: the full `350bd5a6..eba9f9d7` diff; the refusal machinery in
`web.py` and all six of its call sites; both production construction lanes, plus
every other site in `raven/`, `agents/`, `plugins-dist/` and `bridge/` that builds
these three tools; the `wiring.py` extraction against the two bodies it replaces;
and the loop-side contracts the changelog leans on.

Confirmed findings, all nonblocking:

- `web.py:941` -- `note()` re-reads the live key, so a key rotated while a request
  is in flight gets paused without ever having been sent. Reproduced.
- `raven_loop.py:374,381` -- the sub-agent lane's `web_search` and `image_search`
  keep a boot-snapshot key while the new envelope tells the user the config slot is
  read on the next call. Reproduced.
- `web.py:201` -- of the three remedies in the refusal text, only the vendor switch
  carries no qualifier, and it is the one that needs a restart. Code-reading.

Verification:

- 34 refusal envelopes (5 readers x 2 statuses, 8 search vendors x 3) put through
  `is_hard_tool_failure` and `failure_class`: all read as hard failures, and every
  `web_fetch` envelope gets its own stable streak class. My concern that "Do not
  retry this tool" in the detail would trip the transient-marker scan does not
  hold -- those markers are `429 / rate limit / timed out / timeout /
  no healthy upstream / 502 / 503`.
- `uv run lint-imports`: 10 contracts kept, 0 broken.
- `uv run pytest tests/test_agent_tools_web_providers.py
  tests/test_agent_loop_web_tools.py tests/test_config_live.py`: 184 passed,
  2 failed. Both failures are `test_the_settings_page_*` reading
  `ui-web/src/features/settings/SettingsPage.tsx`, which the `git archive` tree I
  ran against did not contain -- my harness, not this branch.

Checked and deliberately not reported: `probe()` bypassing the pause (correct --
it is how a new key gets tested); `_VendorRefusal` concurrency (no await inside
`active` or `note`, so atomic under asyncio); the `"paused": true` field on the
call that was in fact sent (inaccurate, but nothing in the repo reads it);
`effective_provider` still being computed from the boot key (unchanged by this
PR); and the search tools' string returns all collapsing to the `other` streak
class (also unchanged -- the changelog's claim survives because the refusing call
and the paused ones share that class).

One correction to the record, out of scope so noted here rather than inline: the
previous round's response says the research plugin's `WebFetchTool` keeps a static
key and "its pause therefore lifts on the plugin's rebuild, as before".
`agents/raven-research/.../plugin.py` imports `WebFetchTool` from
`research_flow.tools.web`, a fork that pulls in only `raven.contracts.tool` and
`raven.security.network` and carries no `_VendorRefusal` at all, so that tool has
no pause to lift. The conclusion is unaffected -- nothing to change there -- but
the fork simply does not get this PR's protection.

I found both earlier blockers already fixed at this head and have re-filed
neither. The rebase onto `350bd5a6` landed while I was drafting, so I re-checked
against the new head before posting: all four production files are byte-identical
to `ca0a3c4e`, `git range-diff` shows the three commits differing only in the
CHANGELOG hunk, and every anchor above still lands on the line it names.

raven/agent/tools/web.py
ZuyiZhou: Reproduced with your probe shape on the reviewed head (`web.py` unchanged since ca0a3c4e): the real `WebFetchTool` on Jina, a transport that answers 402 and rotates the key source while the request is in flight, then a second call.

```
  key-source reads during one execute(): 4  ['KEY-OLD', 'KEY-OLD', 'KEY-OLD', 'KEY-NEW']
  Authorization actually sent          : ['KEY-OLD']
  next call                            : answered from the pause, 0 request sent, carrying []  (api_key resolves to KEY-NEW)
```

Fixed in 38222344, in the shape you described: each `execute` (and `WebSearchTool.probe`) reads the key once, at its start, and hands that one value to `active`, to the request builder (`_provider_request`, `_search_images` and `_provider_fetch` take it as a parameter and read the property no more) and to `note`. The same probe at this head:

```
  key-source reads during one execute(): 1  ['KEY-OLD']
  Authorization actually sent          : ['KEY-OLD']
  next call                            : served, 1 request sent, carrying ['KEY-NEW']  (api_key resolves to KEY-NEW)
```

All three tools, since `WebSearchTool.execute` and `ImageSearchTool.one` had the shape you named. For `image_search` the one read serves the whole batch: a key rotated between two queries of one call is what the next call reads, and the pause is recorded against the key those queries carried.

Tests: `test_a_key_rotated_while_a_request_is_in_flight_is_tried_before_it_is_paused[web_search|image_search|web_fetch]` drives each tool through the rotation and asserts one read per call, the old key sent and refused, and the next call sent with the new key and served rather than answered from the pause. The `_VendorRefusal` docstring now says which key `note` and `active` are handed and why.

Mutation, each alone, the two files restored between runs (the two web test files, 155 tests, `-n 0`):

```
baseline                                                   155 passed
MUTANT  web_fetch's note() re-reads the key                1 failed   (the in-flight case for web_fetch)
MUTANT  web_search's note() re-reads the key               1 failed   (for web_search)
MUTANT  image_search's note() re-reads the key             1 failed   (for image_search)
MUTANT  web_fetch's request re-reads the key               1 failed   (the request carries a key the pause was not recorded against)
MUTANT  sub-agent lane: web_search on a boot key           1 failed   (the lane test)
MUTANT  sub-agent lane: image_search on a boot key         1 failed   (the lane test)
MUTANT  vendor switch loses its restart qualifier          34 failed  (every pause test, both tools)
REVERT  web.py + raven_loop.py to reviewed head 428cbe6b   38 failed  (the three in-flight cases, the lane test, the 34 wording cases)
CONTROL restored                                           155 passed
```

Full suite at 38222344 in its own worktree: 24123 passed, 110 skipped in 348.21s (0:05:48).

raven/agent/subagent/backends/raven_loop.py
ZuyiZhou: Reproduced on the reviewed head through `RavenLoopBackend.run` with `web_search_provider="tavily"`, `image_search=True` and a key under `tools.web.providers.tavily.apiKey`: after the file was rewritten, `web_fetch` resolved the new key and `web_search` and `image_search` still answered the boot value.

Fixed in 38222344: the lane has one `live_key(vendor)` helper over `live_vendor_key` on its module `LiveConfig`, and all three tools are built on it; `web_fetch`'s inline lambda folded into the same helper. The registration-time withholding (`if web_search.api_key`) reads the same live value once, so a lane that had no key at boot is still withheld the way it was.

`test_the_subagent_lanes_reader_reads_its_key_live_too` is now `test_the_subagent_lanes_web_tools_read_their_keys_live_too`: it registers all three on one vendor, asserts each resolves the boot key, rewrites the slot, and asserts each resolves the rotated key. Mutation: putting `web_search` back on `self._web_key(...)` fails it, and so does putting `image_search` back (rows in the matrix on the first thread).

On the correction in your summary, thank you: the research plugin's `WebFetchTool` is the fork's class (`research_flow.tools.web`, importing only `raven.contracts.tool` and `raven.security.network`) and carries no `_VendorRefusal`, so there is no pause there to lift; the description said otherwise and now says the fork gets none of this PR's behaviour.

raven/agent/tools/web.py
ZuyiZhou: Agreed, and fixed in 38222344: the third remedy now reads "or restart with another vendor selected under tools.web.<kind>.provider", the qualifier the env-var branch has, since `provider` is fixed at registration and feeds both `spec` and the endpoint for the life of the tool. The two pause tests (`test_a_reader_refusing_the_key_pauses_the_tool`, `test_a_search_vendor_refusing_the_key_pauses_the_tool`) assert that sentence, and the changelog entry says which route takes effect when: the slot on the next call, the env var and the vendor on restart. Mutation: the old wording fails both tests (row in the matrix on the first thread).

ZuyiZhou:
Round four at 38222344, answering the three nonblocking notes (each thread has its own reply) and rebased onto main at 608cedf1: main's new commits touch `wiring.py` and `raven_loop.py` beside the lines this PR edits (the browser-tool registrations), so the three earlier commits replayed with context changes only; no conflict.

- One key read per call, in all three tools: the request carries the value that a refusal is recorded against, so a key replaced while a request is in flight is tried by the next call instead of paused unsent.
- The sub-agent lane hands `web_search` and `image_search` the live reader `web_fetch` already had, through one `live_key` helper.
- The refusal text says the vendor selection takes effect on restart, as the env var does.
- The description no longer claims the research plugin's reader has a pause to lift: its tools are the fork's own classes and carry none. Thank you for the correction.

At this head: the three web files 158 passed; full suite 24123 passed, 110 skipped in 348.21s (0:05:48) in its own worktree; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. The mutation matrix and the in-flight probe at both heads are on the first thread and in the description.

raven/agent/tools/web.py
0xKT: **Re-verified fixed** at 38222344, by me, with the recipe I filed.

The thread was already resolved when I got back to it, and someone else's resolve is not my verification -- so here is mine, run on 38222344 merged onto tip 608cedf1 in an isolated worktree, through the real `WebFetchTool` on the keyless Jina default with a transport that answers 403 for one host and a page for the others:

    https://example.org/blocked   -> {"error": "Jina Reader answered HTTP 403", ...}
    https://example.com/a         -> the page
    https://example.net/b         -> the page
    requests actually sent: 3

At the head I filed against (145344f0) the same three URLs gave one request and two answers from the pause, and the first error read `Jina Reader refused the key`. Both halves of the finding are gone: the message no longer blames a key on a path where there is none, and a page 403 no longer disables the tool for every other URL.

The fix is at the right level, which is why I am not asking for anything further. Splitting the constant in two -- `SEARCH_REFUSAL_STATUSES = {401, 402, 403}`, `FETCH_REFUSAL_STATUSES = {401, 402}` (web.py:130-131) -- keeps 403 meaning "the account cannot do this" where the vendor's API *is* the endpoint, and stops it meaning that where the vendor is relaying somebody else's page. The comment above it states that distinction in the terms the defect actually had, so the next reader gets the reason rather than the rule.

One note for the record, not a request: my first probe today reported all three URLs "answered" with zero requests sent. That was the SSRF host check refusing three unresolvable example hostnames before the fetch path was reached -- the giveaway being that even the URLs that should have been fetched sent nothing. The numbers above are from the re-run with hostnames that resolve.

I am closing this thread on my side. The PR is not graded yet: the standing verdict on the board is for ca0a3c4e and the head is now 38222344, so it waits for a verdict on the current tree before it can enter the lane.

gloryfromca (commented):

The new revision resolves the three later notes without regressing the previously reviewed refusal behavior. Each web-tool call now captures one key for the pause check, request, and refusal record; all three sub-agent tools use the live vendor-key resolver; and changing vendors is correctly described as requiring a restart. I checked the repository rules and runtime terminology, the full `github/main...HEAD` diff and revision range history, every production constructor and request-helper caller, compatibility of the changed private helpers, the live-config precedence and registration architecture, and the tests for weakened or implementation-only coverage. The earlier commits differ from the reviewed revision only in rebased context. All review threads I opened are already resolved.

Verification: `uv run pytest tests/test_agent_tools_web_providers.py tests/test_agent_loop_web_tools.py tests/test_config_live.py tests/test_agent_loop_failure_streak.py tests/test_agent_loop_tool_loop_break.py tests/test_security_web_ssrf.py -q` passed (245 tests; 17 existing warnings). `git diff --check` and the merge-tree check were clean, and all current GitHub CI checks pass.

raven/config/live.py
0xKT, line 1: **Blocking:** two ways this PR's live key reader loses a key the process had, both reached from the same decision -- the key became live and the provider stayed boot-decided.

I filed the 403 blocker on this PR and re-verified its fix myself on this head; that one is settled and I am not reopening it. These are different, and both are regressions against 608cedf1.

## 1. A Jina key at the pre-vendor leaf stops being sent the moment any `providers` section exists

`live_vendor_key` (raven/config/live.py:227-247) resolves the pre-vendor leaf for Serper alone:

    slot = web_provider_key(live, vendor)
    if slot: return slot
    if vendor == "serper": ... return leaf
    if slot == "": return ""          # <- cuts off before boot
    return boot or ""

`web_provider_key` answers `None` only when `tools.web.providers` is absent from the file. Once the subtree is present, a vendor with no entry in it gets `""` (WebProvidersConfig.key_for, raven/config/schema.py:1041-1045), so `slot == ""` returns early and `boot` -- which does carry the leaf -- is never consulted.

MEASURED, with two controls, through the real readers and no stubs:

    tools.web.jinaApiKey only                  boot='sk-jina-paid'      live='sk-jina-paid'
    same key + providers.serper                boot='sk-jina-paid'      live=''
    serper's legacy leaf + providers.firecrawl boot='sk-serper-legacy'  live='sk-serper-legacy'

The third row is the one that isolates it: identical config shape, identical code path, different vendor, different outcome -- so this is the missing branch and not my setup. `boot` answers the paid key in every row.

THE RULE IS THE REPO'S OWN, in four places. CONTEXT.md:475-477: "The pre-vendor leaves (`tools.web.search.apiKey`, `tools.web.jinaApiKey`) are still read, after the vendor slot, and only for Serper and Jina." `WebToolsConfig.vendor_key` (schema.py:1086-1094) has both branches. raven/config/update_tools.py:265-267 says a config written before the vendor layout keeps working. And `live_vendor_key`'s own docstring claims it resolves "in the order `WebToolsConfig.vendor_key` uses", which is the function it then diverges from.

REACHABLE BY THE ORDINARY PATH, and the config shape is one the product writes for itself: `update_tools.set_web_provider_key` adds the `providers` subtree and leaves the legacy leaf in place, and that is what the settings surface and the onboarding wizard call. So a user who has had a Jina key since before the vendor layout loses it the first time they set any other vendor's key -- silently. With `key == ""` the Jina branch omits the header entirely (web.py:984-987), so the paid account simply stops being used: no error, no log line, no pause.

INTRODUCED HERE: `git show 608cedf1:raven/config/live.py | grep -c live_vendor_key` -> 0. At base, wiring.py:846 and raven_loop.py:400/407/416 all passed `self._web_key(...)`, i.e. `resolve_vendor_key`, which has the Jina branch (base web.py:113-114). This PR moved web_fetch (main loop) and all three web tools (sub-agent lane) onto the new reader.

The fix is one branch, mirroring the Serper one.

## 2. A keyed backend whose key is cleared at runtime fails every call and can never pause

`effective_provider` (web.py:893-905) decides the backend once, from the boot key, and says so: "A keyed backend selected without a key would be offered and fail on every call, so it is replaced here, out loud, rather than registered." That invariant held while the key was also boot-resolved. This PR made the key live and left the provider boot-decided, so clearing `tools.web.providers.<vendor>.apiKey` at runtime -- exactly the file the new error message tells the user to edit -- leaves `web_fetch` registered on a keyed vendor with no key.

Three things then line up, each read at this head:

* `web_fetch` has no "key not configured" branch (contrast `WebSearchTool.execute`, web.py:272-285) and is not in `_config_gated_tools` (wiring.py:804 registers `web_search`, :815 `image_search`, and nothing registers `web_fetch`), so it is still offered.
* The keyed backends build the header unconditionally -- `headers={**json_headers, "Authorization": f"Bearer {key}"}` at web.py:996-1000 for tavily, and the same shape for its siblings -- so an empty key is sent as a literal `Authorization: Bearer `.
* `_VendorRefusal.note` (web.py:167-169) opens with `if status not in self.statuses or not key: return False`, so the 401/402 that comes back cannot arm the pause: there is no key to key it on.

The result is one identical vendor error per call, for the life of the process, with the pause this PR exists to provide unable to engage -- which is verbatim the failure the changelog entry says this PR ends.

I did not drive a live vendor for this one: the chain above is read from code, and the one behavioural half I ran is the `note()` refusal. Treat the severity as resting on the code chain, and say so if you think a link does not hold.

Either half of the split would close it: re-decide the provider when the key resolves empty (the same "out loud" fallback `effective_provider` already performs at boot), or give `web_fetch` the "key not configured" branch its search siblings have so the tool refuses instead of sending an empty credential.

## What these two have in common

Both are the same seam: the key is now read per call, the provider is still read once. Everything downstream of `effective_provider` -- which backend, whether a key is required, what the error text should say -- was written when those two facts were resolved together. The two findings above are the two places where they can now disagree; a third reader of the same pair may be worth a look while you are in there.

GATES on head 38222344 / tip 608cedf1: `GATES OK base=608cedf1 ruff/lint-imports/commit/large/lang rc=0`.

raven/agent/tools/web.py
ZuyiZhou: One more change to this sentence, at dbc8ffe0, because main moved: #543 and #548 read `tools.web.<kind>.provider` live (per call in the gateway loop, at spawn for a sub-agent), so "needs a restart" became false for the vendor while this thread was closed. The remedy now reads "set a working key at <slot> or select another vendor under tools.web.<kind>.provider (both are read from the config file without a restart), or restart with <env var> set" -- each route still says when it takes effect, which was your point; the two pause tests assert the new sentence.

raven/agent/tools/web.py
ZuyiZhou: A follow-up at dbc8ffe0, since the rebase onto main changed the ground here. Main's `WebFetchTool.provider` (#548) decides the Jina substitution per access and reads the key source to do it, so after the rebase `web_fetch` read the source four times per call again -- the pause identity was still the key the request carried, but the probe you wrote showed the count. Each tool now has a `_resolve()` that reads the vendor and the key once and returns the pair; `execute` and `probe` use it and hand `vendor` and `key` to the builders and normalisers, which read the properties no more. The one-read pin holds for all three tools again, and the pause is keyed on the pair, since the vendor is live too (`test_a_refusal_by_one_vendor_pauses_no_other`).

ZuyiZhou:
Rebased onto main at d50a13c1, and this time the conflict was real: main merged #543 and #548 while this PR was open, which made the web *vendor* live too -- the three tools take a callable `provider`, `WebFetchTool` decides the Jina substitution per call and gained a callable key source of its own, and the gateway loop hands all three live vendors and keys (`provider=lambda: self.web_fetch_provider`, `api_key=lambda: self._live_vendor_key(...)`). The sub-agent lane reads the vendor at spawn (`_web_search_provider_now`) and keeps it for the run.

How the four commits were replayed (`git range-diff` against 38222344):

- where main and this PR had written the same thing -- `WebFetchTool`'s callable key source, the gateway's `web_fetch` registration -- main's version stands; this PR's `live_vendor_key` resolver in `raven/config/live.py` stays, and the gateway's `_live_vendor_key` delegates to it as before;
- the pause, the per-tool status sets, the one-read rule and the lane's live keys replayed on top of main's constructors; no other hunk changed.

And one new commit, dbc8ffe0, because main's change moves the ground under two things this PR claimed:

- **The pause is keyed on the (vendor, key) pair the request carried**, not the key alone. With the vendor live, a refusal recorded against a key is a verdict on a vendor that may no longer be the one in use; another vendor given the same key string is now a new request (`test_a_refusal_by_one_vendor_pauses_no_other`).
- **Each call resolves the pair once**, through a `_resolve()` on each tool, and hands it to the request builder, the envelope and the refusal record. Main's `WebFetchTool.provider` reads the key source on every access to decide the substitution, so after the rebase the one-read pin failed for `web_fetch` (four reads again, the pause identity still right); `_resolve()` makes the selection, the key and the substitution one decision, and the pin holds for all three tools again. The builders and normalisers take `vendor` and `key` as parameters and read the properties no more.
- **The remedy text** (the thread on `web.py:201`) changes once more, because its previous wording -- the vendor switch needs a restart -- is false on main: it now reads "set a working key at <slot> or select another vendor under tools.web.<kind>.provider (both are read from the config file without a restart), or restart with <env var> set". The changelog says the same.

Mutation for the new commit, each alone, `web.py` restored between runs (the two web test files plus `test_config_live.py`, 219 tests, `-n 0`):

```
baseline                                                                   219 passed
MUTANT  the pause ignores the vendor                                       1 failed
MUTANT  web_fetch resolves the pair twice (key read again for the request) 1 failed
MUTANT  web_fetch's envelope names the vendor read again                   1 failed
MUTANT  the vendor switch loses its live qualifier                         34 failed
REVERT  web.py to the rebased head 9c788e0b (before the adaptation)        36 failed
CONTROL restored                                                           219 passed
```

At dbc8ffe0: the four focused files 222 passed; full suite 24630 passed, 110 skipped in 335.77s (0:05:35) in its own worktree; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. The description is rewritten for this head.

gloryfromca (commented):
Blocking: preserve the legacy Jina key when the providers subtree contains only other vendors.

The rebase adaptation correctly resolves each call's vendor/key pair once, scopes pauses to that pair, and now re-evaluates a cleared keyed fetch backend so it falls back to Jina. The first half of the existing open blocker at `raven/config/live.py` still holds, though: `live_vendor_key` consults the pre-vendor leaf only for Serper. With `tools.web.jinaApiKey=sk-jina-paid` plus a `providers.serper` entry, I reproduced `WebToolsConfig.vendor_key("jina") == "sk-jina-paid"`, `live_vendor_key(..., "jina", boot=...) == ""`, and `WebFetchTool._resolve() == ("jina", "")`. The ordinary settings migration shape therefore silently stops sending the paid Jina credential, contrary to the schema and `CONTEXT.md` compatibility contract. This is already tracked in the open 0xKT thread, so I am not duplicating it inline or resolving someone else's thread.

I covered the repository rules and runtime terminology, the full `github/main...HEAD` diff, the rebase/range history, all production constructors and request-helper callers, backward compatibility, test strength, and the live vendor/key architecture. Verification: `uv run pytest tests/test_agent_tools_web_providers.py tests/test_agent_loop_web_tools.py tests/test_config_live.py tests/test_agent_loop_failure_streak.py tests/test_agent_loop_tool_loop_break.py tests/test_security_web_ssrf.py -q` passed (275 tests; 17 existing warnings). `git diff --check` and the merge-tree check were clean; current CI is green except one unit shard still pending at review time.

raven/config/live.py
ZuyiZhou: Reproduced both halves on dbc8ffe0, and thank you for the third row: it is what told me the missing branch was the whole story for the first half. Your table through the real readers, `boot` from `WebToolsConfig.vendor_key` and `live` from `live_vendor_key`, with two rows added while I was there:

```
  tools.web.jinaApiKey only                    boot='sk-jina-paid'       live='sk-jina-paid'
  same key + providers.serper                  boot='sk-jina-paid'       live=''
  serper's legacy leaf + providers.firecrawl   boot='sk-serper-legacy'   live='sk-serper-legacy'
  a cleared slot beside the serper leaf        boot='sk-serper-legacy'   live='sk-serper-legacy'
  harness key, providers names other vendors   boot='tv-harness'         live=''
```

The fifth row is the same mechanism one step further out: `WebProvidersConfig.key_for` answers an empty key for a vendor the subtree does not name, the same answer it gives for one named with an empty key, and `live_vendor_key` read every empty as a revocation before it consulted the boot value. So beside the Jina leaf, a key a harness passed through with no file entry behind it was lost as soon as the file's `providers` named any other vendor. Same table at 40893f7f:

```
  tools.web.jinaApiKey only                    boot='sk-jina-paid'       live='sk-jina-paid'
  same key + providers.serper                  boot='sk-jina-paid'       live='sk-jina-paid'
  serper's legacy leaf + providers.firecrawl   boot='sk-serper-legacy'   live='sk-serper-legacy'
  a cleared slot beside the serper leaf        boot='sk-serper-legacy'   live='sk-serper-legacy'
  harness key, providers names other vendors   boot='tv-harness'         live='tv-harness'
```

**First half**, fixed in 40893f7f: `live_vendor_key` follows `WebToolsConfig.vendor_key` in full. The slot; then the pre-vendor leaf that is Serper's or Jina's, the latter through a new `web_jina_key` reader (`raven/config/live.py`) over `live_web_jina_key` (`raven/config/schema.py`), which validates the leaf alone the way `set_web_provider_key` validates one slot and is admitted with the same last-good memory as its twin; then a revocation only where the file names the vendor with an empty key (`_names_vendor`); then the boot value where the file says nothing about the vendor. The docstring now says so instead of claiming an order it did not follow.

**Second half**: you were right about the chain on 38222344, and it is closed at this head by the previous commit (dbc8ffe0), which main's #548 made necessary and possible. `WebFetchTool` now resolves the selected vendor, the key and the Jina substitution together, once per call (`_resolve()`), so clearing `tools.web.providers.<vendor>.apiKey` -- the edit the refusal text points at -- moves the next call onto Jina, keyless, instead of sending `Authorization: Bearer` with nothing after it; the envelope says `jina-reader`, and there is no 401 to arm anything. `effective_provider`'s docstring on main already says it is asked per call. Two tests pin it through the paths you named: `test_a_keyed_reader_whose_key_is_cleared_falls_back_to_jina_per_call` (the tool, a transport recording what Tavily and Jina were each sent) and `test_a_reader_whose_key_is_cleared_in_the_file_reads_through_jina` (the main loop's real wiring and the config file, `tool.provider == "jina"` afterwards).

On the "third reader of the same pair": `_resolve()` is now the only place the pair is read, and the request builder, the envelope and the refusal record all take it as parameters, so I believe the seam is closed rather than patched twice; the one-read pin (`test_a_key_rotated_while_a_request_is_in_flight_is_tried_before_it_is_paused`) holds for all three tools.

Tests for the first half, in `tests/test_config_live.py`: `TestLiveVendorKeyFollowsTheCanonicalOrder` runs your three rows and three more (a cleared slot beside the Serper leaf, a cleared slot with no leaf, a slot winning over its leaf) against `WebToolsConfig.vendor_key` as the oracle; `test_a_vendor_the_file_says_nothing_about_keeps_its_boot_key` pins the fifth row and its revocation twin; `test_a_revoked_leaf_is_an_answer_not_a_miss`; `test_a_leaf_the_schema_rejects_dispenses_no_new_answer`; and `test_a_rotated_leaf_reaches_the_next_read[jina|serper]`, which I added after the first matrix: with the Jina branch mutated away, your second row still passed, because the boot value carries the same leaf and the new boot fallback answered it. A leaf rotated in the file while a `providers` subtree names another vendor is what the branch is for (boot still says the old key), and that is what now pins it.

Mutation, each alone, the files restored between runs (the two web test files plus `test_config_live.py`, 232 tests, `-n 0`):

```
baseline                                                                     232 passed
MUTANT  the jina leaf is not consulted (serper's alone, as reviewed)         2 failed
MUTANT  an empty slot is a revocation whether or not the file names the vendor 1 failed
MUTANT  the leaf reader forgets its last admitted key on a rejected edit     1 failed
MUTANT  web_fetch runs on the selected vendor even when its key is gone      3 failed
REVERT  live.py + schema.py to the reviewed head dbc8ffe0                    5 failed
CONTROL restored                                                             232 passed
```

Full suite at 40893f7f in its own worktree: 24717 passed, 110 skipped in 260.48s (0:04:20). I did not change `CONTEXT.md`: its sentence (both leaves, after the slot, Serper and Jina only) was the rule the code now follows.

ZuyiZhou:
Round six at 40893f7f, on the blocker 0xKT filed on `raven/config/live.py` (reproduced by gloryfromca on dbc8ffe0): `live_vendor_key` consulted the pre-vendor leaf for Serper alone and read the slot reader's empty answer for an unnamed vendor as a revocation, so a Jina key at `tools.web.jinaApiKey` stopped being sent as soon as any other vendor's key was set, and a harness-supplied key was lost the same way. The reader now follows `WebToolsConfig.vendor_key` in full (slot, then either leaf, then a revocation only where the file names the vendor with an empty key, then boot), with a new live reader for the Jina leaf; the second half of that finding, a keyed reader sending an empty credential after its key is cleared, is closed by the per-call resolution the previous commit introduced and is now pinned by two tests. Details, the measured table at both heads and the mutation matrix are on the thread.

At 40893f7f: the four focused files 235 passed; full suite 24717 passed, 110 skipped in 260.48s (0:04:20) in its own worktree; lint-python, lint-imports, lint-types, lint-deps, `git diff --check`, source-language, commit and title checks, large-files all pass. The description is updated for this head.

gloryfromca (commented):

The blocking compatibility regression is fixed on 40893f7fba64. I independently exercised the canonical/live key matrix: the legacy Jina leaf now survives an unrelated providers entry, the Serper leaf still does, a harness-supplied key survives when the file names only another vendor, an explicit empty slot still revokes, and a nonempty slot still wins. I also confirmed the keyed-fetch half: clearing the selected keyed vendor now resolves the next call through keyless Jina rather than sending an empty Authorization header.

One low-reachability follow-up is worth preserving here rather than as an open thread: after an empty vendor slot has been admitted as a revocation, a later schema-invalid providers edit that omits that vendor makes _names_vendor inspect the rejected raw dict and restore the boot key instead of keeping the last admitted revocation. I reproduced empty Tavily slot -> misspelled-only providers subtree -> sk-boot. It requires an explicit clear followed by a malformed manual edit, and correcting the config is an immediate way out, so at this eighth round it does not meet the blocking bar; the live-config last-good discriminator can be hardened separately.

Coverage this turn: AGENTS.md/CLAUDE.md and the Runtime terms in CONTEXT-MAP.md/CONTEXT.md; the full main...HEAD diff; the 832f0b478..HEAD fix delta; both production wiring lanes and the tool resolver; range-diff/history and backward compatibility; and the added tests for weakening or false-green coverage. The focused loop/provider/live-config/failure-streak/SSRF suite passed: 288 tests, with 17 existing pytest-mark warnings. git diff --check and every reported GitHub check pass.

0xKT:
**Not a blocker.** Both halves of my blocking finding are fixed, and I verified each myself on `40893f7f` rather than taking the thread for it. Two things left, both about promises the new refusal text makes.

### The blocker is closed -- what I ran

**First half.** Nine rows through the real readers, `WebToolsConfig.vendor_key` as the oracle and `live_vendor_key` as the subject, all nine agreeing. The rows carrying the proof are two I added on top of the table I filed:

```
row                                        vendor   file says          boot                live reader
jina leaf ROTATED, providers.serper        jina     'sk-jina-NEW'      'sk-jina-paid'      'sk-jina-NEW'
serper leaf ROTATED, providers.firecrawl   serper   'sk-serper-NEW'    'sk-serper-legacy'  'sk-serper-NEW'
```

Boot holds a *stale* value there, so only a reader that actually consults the leaf can answer. Mutating the Jina branch away leaves my original three rows green and fails exactly these two -- the same thing your `test_a_rotated_leaf_reaches_the_next_read` says, arrived at separately. My filed table could not have caught a missing branch; yours can.

**Second half.** The real `WebFetchTool`, real `_resolve`, real `_provider_fetch`, only the socket replaced:

```
CONTROL  key present  -> _resolve=(tavily,'tv-live')  request api.tavily.com  Authorization: 'Bearer tv-live'
SUBJECT  key cleared  -> _resolve=(jina,'')           request r.jina.ai       <no Authorization header>
```

With `_resolve` mutated back to a boot-decided provider the subject reproduces what I filed verbatim: `Authorization: 'Bearer '` to `api.tavily.com`. Closed.

---

### 1. The new Jina leaf reader knows one spelling; the loader accepts two

`web_jina_key` (new here, `a67ac530`) decides whether the leaf is present with a raw string test on `"jinaApiKey"` (`live.py:277`), and `live_web_jina_key` hardcodes the same single spelling twice (`schema.py:2615, 2618`). But `Base.model_config` is `alias_generator=to_camel, populate_by_name=True` (`schema.py:24`), so a file may spell it `jina_api_key`, and `WebToolsConfig.vendor_key("jina")` reads that fine.

Same rotation, both spellings, through the real readers:

```
                              loader sees   web_jina_key   live_vendor_key
CONTROL  jinaApiKey  (camel)  'sk-new'      'sk-new'       'sk-new'      ok
SUBJECT  jina_api_key (snake) 'sk-new'      None           'sk-old'      <- stale boot value
```

The subject's config is valid and the new key is in it -- the loader proves that on the same line -- but the live lane answers with the value the process booted with. `web_fetch` defaults to Jina, so this is the default path: a refused key arms the pause, the user rotates the key exactly as the message says, and the next call sends the old one and re-arms.

The rule is the module's own, in two places:

- `CONTEXT.md:1384-1385` (**Live preference**): "The module is the roster -- every key this lane serves has a named reader there (both file spellings)".
- `live.py:569-572`, the `_block_pin` docstring right beside it: "the config models accept both by alias, so a reader that knew only one would answer None for half the configs on disk."

`web_search_key` does not have the problem because its leaf lives inside a *section* (`tools.web.search`) that pydantic validates, so the alias does the work. The Jina leaf is a scalar on `tools.web`, which is why the presence test had to be written by hand -- and why it needs both spellings written by hand too.

### 2. "select another vendor" is not true in two of the three places the text is emitted

`refusal_text` is new here (absent at `a40e911d`) and tells the reader:

> set a working key at `<slot>` **or select another vendor under `tools.web.<kind>.provider` (both are read from the config file without a restart)**

The key half holds everywhere. The vendor half holds for `web_search` and `web_fetch` in the main loop, whose providers are properties over the file. It does not hold here:

**(a) the sub-agent lane.** All three tools get a **frozen provider string** -- `raven_loop.py:476`, `:483`, `:490-491` -- while getting a live key reader. The main loop passes callables (`wiring.py:923, 935, 953`). Same edit, same process:

```
CONTROL  main loop   web_fetch provider: tavily -> firecrawl
SUBJECT  sub-agent   web_fetch provider: tavily -> tavily
```

The comment beside the live reader (`raven_loop.py:465-468`) says "Read on every call, as the main loop's are", which is exactly right about the key and is the sentence a reader takes as covering the provider beside it.

**(b) `image_search`, in the main loop.** Its vendor comes from `image_search_vendor(self.web_search_provider, self._web_key)` (`wiring.py:935`), and `_web_key` is the **boot** reader, so a vendor selected *and keyed* in the same edit tests as keyless and falls back to Serper:

```
CONTROL  web_search   offered: False -> True   key 'sk-added-later'
SUBJECT  image_search vendor : serper -> serper        key ''
```

after adding `tools.web.providers.tavily.apiKey` with `tools.web.search.provider = tavily`. `ImageSearchTool` emits `refusal_text(..., kind="search")`, so it tells the user to do the one thing it cannot follow; the pause stays armed on the refused pair and re-arms on each cooldown, so `image_search` stays dead for the life of the process while `web_search`, given the identical edit, moves on the next call. A thorough user who also deletes the dead Serper key ends up with `image_search` on Serper with no key at all, answering "Serper API key not configured" and naming a vendor they just stopped using.

The thing that makes this cheap to fix: **`wiring.py` already asks this exact question with the live reader, 600 lines away.** `_lane_searches_images` builds `key_for` around `self._live_vendor_key(vendor)` (`wiring.py:334-339`) and hands *that* to `image_search_vendor`. Line 935 hands it `self._web_key` instead. One definition of "does this vendor have a key, for choosing the picture vendor" already exists and is live; this is the second reader of it.

The comment two lines above line 935 says "Both halves follow the selection: the picture vendor is derived from it, so freezing either one pins the pair to the boot choice" -- and line 935 freezes the other half.

To be fair about (b): the *mechanism* predates this PR. I ran the same probe at `a40e911d` and it fails identically. What is new here is the pause that makes the failure sticky and the sentence that turns it into a broken promise -- before this PR nothing was promised and nothing stayed refused.

Either half closes both: make the vendor live where the key already is, or word the remedy per lane.

### Things I attacked and could not break

- **The 403 asymmetry is right, and right for the whole set.** All five `FETCH_PROVIDERS` (jina, anysearch, tavily, exa, firecrawl) relay a URL, so a 403 speaks of the page; all five `IMAGE_SEARCH_VENDORS` (serper, serpapi, brave, tavily, firecrawl) answer a query, so a 403 speaks of the key. Tavily and Firecrawl are in both lists under different endpoints -- which is why splitting by tool rather than by vendor is the correct axis.
- **The pause's identity.** On the real record: a keyless request never arms it; rotating the key away lifts it and rotating back does not restore it; one slot means a second vendor's refusal drops the first's (one extra request, which is all the docstring claims); two tools never share one.
- **`_resolve` really is the only place the pair is read in `execute`** -- the request, the envelope's `extractor`, the refusal record and the refusal text all take the `vendor` it returned, never `self.provider`.
- **Seats.** `raven/config/live.py` imports nothing above it; `lint-imports` 10 kept, 0 broken.

---

`GATES OK head=40893f7f tip=08be2a25 base=a40e911d ruff/lint-imports/commit/large/lang rc=0`; `test_agent_tools_web_providers` `test_agent_loop_web_tools` `test_config_live` `test_config_schema` `test_cli_agent_loop_wiring` `test_security_web_ssrf` `test_subagent_mode_wiring` -> 267 passed. CI on this head is green: the three checks showing `cancelled` were each superseded by a `success` re-run 45s later.

0xKT (approved):

PR #573, opened 2026-09-21
feat(*): deck template picker and full-frame deck preview for the web ui

Tchen-data:
## Summary

Two deck features for the web UI, both landing on the refactor tree.

**Deck template picker.** The composer gets a "Deck template" button beside the attach control. It opens a sheet of the templates the deck engine ships, shown by their cover pages; opening one pages through its slides, and "Use this template" attaches it to the turn exactly as an uploaded file would be. That attachment is what opens the host's route to the template engine (the route reads the attached .pptx, not the tier, see the companion backend change on main). Without the engine installed the sheet says so.

Three console methods back it: `deck.templates.list`, `deck.templates.pages`, `deck.templates.pick`. `pick` copies the template into the agent's uploads and answers in the `fs.upload` shape so the frontend treats it as one more attachment. The sheet lives in the composer domain (`cp-tpl-*` classes, domain stylesheet, `gui.tpl.*` strings).

**Delivered deck preview, ported from the pre-refactor deck bar branch.** A .pptx opened in the workspace now fills the whole pane under a single bar: the browser viewer's own toolbar is hidden, the frame takes every pixel to the bottom edge, and the bar carries Download and Show in Finder / Explorer / file manager (worded per the gateway's platform; a reveal the host cannot perform surfaces as a toast with the failure sentence). The frame URL carries the delivery version so a re-delivered deck under the same path is refetched. The transcript tile of a delivered deck shows its first page (`render=thumb` on the files route, PNG through the LibreOffice path; `to_pdf` grew a format parameter). The file pane opens at 960 (was 720) and the workspace column may grow to 1200 (was 760).

Class names from the original branch were moved under the workspace namespace (`workspace-fill`, `workspace-deck-acts`) to satisfy the class-namespace gate.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `ui-web`: `npx tsc --noEmit -p tsconfig.json` -> no errors; `npx vitest run src/features/desk src/chrome src/features/workspace src/features/transcript src/features/composer src/state` -> 1229 passed, 0 failed; `npx vitest run scripts/gates` -> 160 passed (class namespace, i18n keys, offline coverage, css one owner, fixture shape); `npm run gen:check` -> 192 methods in sync; `npx eslint .` -> 0 errors (3 pre-existing exhaustive-deps warnings in files this PR does not touch).
- `ui-tui`: `npm run lint:i18n` -> generated catalogue up to date.
- Python: `uv run --frozen pytest tests/test_rpc_console.py tests/test_rpc_files.py -q` -> 156 passed.
- Manual: a full gateway built from this tree (Playwright, full Chromium): picker opens, covers render, paging works, "Use this template" attaches the .pptx; a delivered deck opens full-frame with the one bar; Download serves the file; the tile shows page one. Reveal was exercised on a headless host and produced the toast `reveal failed: ... 'xdg-open'`; the darwin `open -R` branch was verified on macOS on the original deck bar branch.
- End to end: a high-tier deck request with a picked template routed to the template engine and delivered on that template; the same request without a template stayed on Raven-Design.

## Risk

- New RPC methods are additive; the `render=thumb` query is additive on the files route and falls back to the existing behaviour when absent.
- The wider default pane (960 / 1200) is the one user-visible layout change outside deck panes; it only affects the first open and the drag ceiling, saved widths are kept.
- Rollback: revert the squash commit; no schema or storage migration is involved.

## Related Issues

N/A


raven/rpc/deck_templates.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def _draw_in_background(template: Template) -> bool:
+    """Start drawing this template's cover unless one is already on its way; True while one is."""
+    if template.name in _failed or not _rasteriser_available():
+        return False
+    task = _drawing.get(template.name)
+    if task is not None and not task.done():
+        return True
+    task = asyncio.ensure_future(cover_for(template))
```
gloryfromca, line 227: Blocking: cold template listing launches all ten LibreOffice conversions concurrently.

`listing()` walks all ten bundled templates and every cache miss reaches this `ensure_future` before any conversion finishes. On a fresh cache I measured 10 simultaneous cover tasks; each then enters `pdf_preview` and starts its own `soffice` process/profile. The engine's own concurrency contract budgets one core and a few hundred MB per conversion and caps normal fan-out at 8 because that is already a couple of GB. Thus an ordinary first click can start more than 2 GB of rendering work at once and stall or OOM the gateway. Please put these starts behind a bounded queue/semaphore (or otherwise enforce a comparable cap) instead of starting one task per template.

gloryfromca (commented):
Blocking: cap the cold-cache deck cover render fan-out before merge.

I found one blocker inline. Otherwise the picker attachment path, sheet identity and session behavior, RPC models/schema/fixtures, full-frame preview versioning, delivery thumbnail fallback, older-gateway and missing-engine behavior, i18n, and the web architecture constraints are consistent.

I covered the project rules and context docs, the full diff, relevant callers and history, backward compatibility, test changes for weakening, and the architecture boundaries. Verification: `uv run pytest tests/test_rpc_console.py tests/test_rpc_files.py tests/test_utils_office.py -x` (164 passed); `npm test` in `ui-web` (187 files, 2561 tests passed); type-check, generated-client check, source-language gate, large-file gate, and `git diff --check` passed. ESLint had 0 errors and 5 pre-existing warnings in untouched files.

gloryfromca (commented):
Blocking: bound the cold-cache cover rendering fan-out tracked in the open thread.

The new `a0d2e2dcbeae` delta only adds the optional PyMuPDF deptry exceptions and updates the two boot goldens. I reviewed that delta and it is sound: `uv run --all-extras deptry raven` reports no dependency issues, and the web build reports both boot snapshots matching (254 and 255 nodes). The focused Python suite still passes 164 tests.

The prior inline finding remains reachable and unchanged because `raven/rpc/deck_templates.py` was not modified, so this revision is still blocked on that one item.

raven/rpc/deck_templates.py
Tchen-data: Fixed in a97f5f1b0. The cover renders now pass through a semaphore of three (`COVER_RENDERS_AT_ONCE`), so a cold listing of ten templates starts three conversions and the rest queue behind them; the picker still answers at once with what is on disk. A test drives ten misses through a blocked `pdf_for` and asserts the peak in-flight count is three.

gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed. I independently drove ten cold cover requests against a blocked renderer and observed a peak of exactly three in flight; the other seven waited, and all completed after release. The focused RPC/file/office suite also passes all 165 tests. This settles my finding.

gloryfromca (commented):

The semaphore fix settles the only open finding: ten cold cover requests now peak at three conversions while the picker continues answering immediately. I replied to the author and resolved my thread.

For this revision I covered the project rules and context, the fix diff and its caller path, relevant history, backward compatibility, test-strength changes, and the architecture boundary. Verification: the independent concurrency reproduction observed a peak of 3; `uv run pytest tests/test_rpc_console.py tests/test_rpc_files.py tests/test_utils_office.py -x` passed all 165 tests; the source-language gate, large-file gate, and `git diff --check` passed.

gloryfromca (commented):

The force-push rebased the already-reviewed feature onto the updated target; the only substantive new branch commit adds focused gallery tests and does not change production behavior. Those tests accurately exercise the previously uncovered discovery, rasterization, cache, failure, and duplicate-start paths. The settled concurrency thread remains resolved.

I covered the current target diff, the new test commit and surrounding implementation, relevant history, backward compatibility, test-strength changes, the repository rules/context, and the existing architecture boundaries. Verification: `uv run pytest tests/test_rpc_deck_templates.py tests/test_rpc_console.py tests/test_rpc_files.py tests/test_utils_office.py -x` passed all 176 tests; `uv run --all-extras deptry raven`, the source-language gate, large-file gate, and `git diff --check` passed.

arelchan (approved):

PR #574, opened 2026-09-21
feat(*): open the deck engine route on an attached template, not on a tier

Tchen-data:
## Summary

The route from Raven-Design to the Raven-PPT template engine used to open on the max tier alone. The same request then came back in three unrelated styles across the tiers, and the top one was built on a template the user never asked for. This PR moves the decision to the one thing that actually calls for the engine: a .pptx the user handed over.

- **Route on an attached template.** `SubagentRouteConfig` gains `needsFile` (a suffix), a third independent condition beside `needs` and `minTier`. Raven-Design's route to Raven-PPT now declares `needsFile: ".pptx"` and drops `minTier: "max"`. Every tier designs its deck on Raven-Design; the engine is reached only when a .pptx was handed over, either a bundled template picked in the web UI (companion PR #573) or one the user attached.
- **Handed over means attached in this turn.** A new `turn_attachments` ContextVar is frozen beside `turn_tier` on the turn path. A direct chat's media all count; a spawn or DAG node counts an attachment only when its task text names that attachment's path or file name. Task text alone never opens the route: a deck brief spells the deck's destination path, the format in words, or a stray file in the directory exactly like a template, and the first text-only scan sent a no-template run into the engine on its destination path.
- **Tiers change effort, not product.** medium asks for low reasoning effort, max for xhigh (the highest word a model may know; models that do not know it fall back to their default), and the baseline inherits the host's setting as before.
- **Design lane retries after streamed output.** `llmRetryAfterOutput: true` on the design lane's defaults, matching the ppt lane. A max run died at its thirtieth iteration to an upstream idle timeout after the body had already streamed to the host; the lane now retries instead of failing the turn.
- **Skill: use what the user handed over first.** Documents and data are the material, pictures are the pictures, and an attached .pptx is the engine's work, routed before the skill is read.

Docs: `CONTEXT.md` admission paragraph, `agents/README.md`, regenerated `schemas/subagent.schema.json`.

Also removed: the engine's internal decision log (`plugins-dist/ppt-engine/docs/ppt-raven-design.md`), a single-language working journal for the team building the deck lanes that has no reader in a public repository. The maintainer keeps it outside the tree; the two code comments that cited it by path now state their reason in place.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run --frozen pytest tests/test_subagent_routing_backend.py tests/test_agents_design_launcher.py tests/test_config_schema.py tests/test_subagent_registry.py -q` -> 130 passed (after rebase onto main).
- `ruff check` / `ruff format --check` on the changed Python files -> clean.
- End to end on a full gateway: a high-tier request with a picked template logged `classified for 'Raven-PPT'; routing there` and the engine delivered a 20-page deck on that template in about 60 minutes; the same request without a template logged `no .pptx the user attached was handed over; running here` and stayed on Raven-Design. A brief that only spelled the deck's destination path stayed on Raven-Design.
- Effort labels probed against `z-ai/glm-5.3-flash` over OpenRouter: `low`, `high`, `xhigh` all accepted; `max` spent the whole output window on reasoning, which is why the top tier is xhigh.

## Risk

- User-visible: medium and high tier deck requests no longer differ in product, only in effort; a deck on a template now requires attaching one. A bound direct-chat handle to Raven-Design stays on Raven-Design even with a .pptx attached (the route is evaluated at dispatch), a known boundary.
- `needsFile` is additive; routes that do not declare it are dispatched exactly as before.
- Rollback: revert the squash commit; no schema migration or storage change.

## Related Issues

N/A


raven/agent/subagent/backends/routing.py
```diff
@@ -79,15 +82,57 @@
...
+    if given:
+        return tuple(given)
+    handed: list[str] = []
+    for item in turn_attachments_in_force():
+        path = _path_of(item)
+        if not path:
+            continue
+        # The path or its name, ended the way a file name ends: ``brand.pptx``
+        # in ``brand.pptx.bak`` is a different file.
+        named = re.compile(re.escape(Path(path).name) + r"(?![\w.])")
```
gloryfromca: Blocking: make attachment-name matching reject suffixes of different filenames.

This regex constrains only the character after the basename, so it also matches the basename as the suffix of a different filename. I reproduced this with turn attachment `/uploads/template.pptx` and task `Ignore the attachment; save the new deck as /out/new_template.pptx`: `handed_files()` returns `/uploads/template.pptx`. For a deck request, that opens Raven-PPT even though the task did not hand the template over, defeating the documented ignore-attachment case. Please require an exact path/name boundary on the left as well and cover the prefixed-output-name case with a regression test.

gloryfromca (commented):
Blocking: make attachment-name matching reject suffixes of different filenames.

I found one blocking defect, marked inline.

I covered the full diff; AGENTS.md, CONTEXT-MAP.md, CONTEXT.md, and agents/README.md rules; spawn, DAG, and direct-chat callers; the prior tier/credential routing history; backward-compatible defaults and signatures; test changes (none were weakened to manufacture a pass); and the runtime/subagent architecture boundary.

Verification:
- `uv run pytest tests/test_subagent_routing_backend.py -q`: 55 passed.
- After `uv sync --frozen --all-extras`, the broader affected suite (routing, design launcher, schemas, manager, direct chat, DAG runner, and turn loop): 662 passed. The first attempt before syncing extras stopped on two missing-design-engine fixture failures; both passed after the declared extras were installed.
- Ruff check and format check on all changed Python files: passed.
- Source-language, large-file, commit-message, and `git diff --check` checks: passed.

raven/agent/subagent/backends/routing.py
Tchen-data: Fixed in 62aaa318f. The name is now bounded on the left as well as the right: `(?<![\w.-])name(?![\w.-])`, so `brand.pptx` in `new_brand.pptx`, `my-brand.pptx` or `brand.pptx.bak` is a different file and does not hand the attachment over. The negative briefs in `test_a_pptx_in_the_text_that_is_not_an_attachment_does_not_open_it` gain your reproduction (ignore the attachment, save as a prefixed name) and a hyphenated neighbour; both stay on Raven-Design with the template attached.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at `62aaa318f9dd`: the original `new_brand.pptx` reproduction and the added hyphenated neighbour no longer hand over the attached `brand.pptx`, while exact path/name handoffs still do. The affected suite passes 662/662. This resolves my finding.

gloryfromca (commented):

The follow-up commit fixes the reported left-boundary false positive and adds focused negative regressions without weakening existing coverage. Exact absolute-path, basename-only, and punctuated attachment handoffs still work. I found no new findings.

I rechecked the full diff and delta; AGENTS.md and the relevant CONTEXT/agent documentation; spawn, DAG, and direct-chat callers; routing history; backward compatibility; test changes; and the runtime/subagent architecture boundary.

Verification:
- Affected suite: 662 passed.
- Direct boundary probes: reported underscore and hyphen false positives rejected; valid path/name handoffs accepted.
- Ruff check and format check: passed.
- Source-language, large-file, commit-message, and `git diff --check` checks: passed.

gloryfromca (commented):

The attachment-boundary fix remains correct, and the latest cleanup commit only removes the internal design journal and makes its two former references self-contained. No dangling references remain and I found no new findings.

I reviewed the complete latest diff and the new delta; AGENTS.md and relevant CONTEXT/agent documentation; callers and routing history; backward compatibility; test changes; and the runtime/subagent architecture boundary.

Verification:
- Latest-head targeted suite: 117 passed.
- Prior full affected suite on the functional fix: 662 passed.
- Direct attachment-boundary probes: passed.
- Ruff, source-language, large-file, commit-message, and `git diff --check` checks: passed.

raven/agent/subagent/backends/routing.py
0xKT, line 1: **Blocking:** a task that ends its sentence right after the attachment's name closes the deck route. `brand.pptx.` fails the right bound, so the template the user attached is not handed over.

`handed_files` bounds the name on the right with `(?![\w.-])` (raven/agent/subagent/backends/routing.py:115). `.` is inside that class, so a task whose sentence ends with the file name plus an ASCII full stop reads as NOT naming the attachment, `handed` comes back empty, and `_target_open` (routing.py:258-261) refuses with "no .pptx the user attached was handed over".

REPRODUCED, on head a8685c86 merged onto tip 0ae68805 in an isolated worktree, driven through the real `RoutingBackend.run()` with the shipped manifests (not through `handed_files` directly). One attachment, `/home/u/.raven/tmp/tui/brand.pptx`, frozen by `turn_attachments`, no `media` kwarg:

    Raven-PPT      <- "Build the deck on the attached template /home/u/.../brand.pptx"
    Raven-Design   <- "Build the deck on the attached template /home/u/.../brand.pptx."

CONTROLS that do reach the engine on the same tree in the same run, so the setup is not what failed: a comma after the path, the path mid-sentence, `;`, `!`, `?`, backticks, double quotes, parentheses, end-of-line, and the FULL-WIDTH stop. Only the ASCII full stop closes it -- an asymmetry no reader would predict, and the one mark an English brief is likeliest to carry. `brand.pptx...`, `brand.pptx-based` and `brand.pptx_copy` close it too.

THE RULE IS THIS PR'S OWN, in four places, and none of them carves out punctuation:
* CONTEXT.md:376-378 (added here): "read off the files the user attached -- a direct chat's media, or an attachment of the turn that the spawn task names -- and never off the task text alone".
* agents/README.md:61-63 (added here): `"needsFile": ".pptx"` ... "an attachment of the turn that the spawn task names".
* `handed_files`' own docstring, routing.py:96-102.
* tests/test_subagent_routing_backend.py:598-600, the promise stated outright: "in whichever of the two texts it wrote and **however it punctuates around it**". The eight spellings that follow sample the full-width stop and the comma; the ASCII stop is the one that is missing.

INTRODUCED HERE, and for one tier it is a regression. `handed_files` and `needsFile` do not exist at the merge base db7862ae (`git grep handed_files db7862ae` is empty). The right bound was `(?![\w.])` at 40b18f5f and became `(?<![\w.-])...(?![\w.-])` at 62aaa318; `.` is in the class in both, so the miss arrived with the feature. The regression half: at db7862ae the route declared `"minTier": "max"` and nothing else, so a max-tier turn reached Raven-PPT whatever the text said. Same user, same attachment, same brief -- engine before, no engine after.

PRODUCTION REACHES IT, and the spawn lane is the only lane where it matters, because it is the lane that passes no `media`: raven/agent/subagent/manager.py:1055 is the sole writer of `kwargs["media"]` and it is `chat`, not `spawn`. spawn_tool -> manager.spawn -> `asyncio.create_task` (manager.py:901, which copies the turn's ContextVars set at raven/agent/loop/turn_path.py:2182) -> manager.py:1628 `backend.run(...)` with no media -> routing.py:382-387 `handed=handed_files(task, authored_task, media=kwargs.get("media") or ())` -> the regex -> the gate. dag_runner.py:1657-1672 reaches the same seam the same way.

NOBODY CATCHES IT DOWNSTREAM, and this PR is what closes the second door: with the gate shut, the lane is handed agents/raven-design/route-raven-ppt.md, which says "design it yourself, from your own design skills, **rather than from a packaged deck template**"; and the SKILL.md line added by this PR says an attached `.pptx` "is the engine's work, not this skill's: the host routes it there before this skill is read". The engine never sees the file and the design lane is told not to use it. The failure is silent -- the only trace is an info log.

SUGGESTED FIX, measured rather than guessed: keep the left bound (it is sound -- I tried a name preceded by `-`, by `.`, a quoted name, a Windows path and a name beginning with `.`, and every one of them held) and let the right bound admit a dot that is not part of a longer name:

    named = re.compile(r"(?<![\w.-])" + re.escape(Path(path).name) + r"(?![\w-])(?!\.[\w-])")

Checked on 22 spellings: the 11 that must match (including the two failures above and every one of this PR's positive cases) and the 11 that must not (every negative brief in tests/test_subagent_routing_backend.py:601-615, plus `brand.pptx.bak`, `brand.pptx.bak, keep it`, `brand.pptxx` and `brand.pptx-old`). 0 wrong, against 2 wrong for the expression as written.

ONE MORE THING THE READER SHOULD SEE, not filed as a separate defect because the base was no better here: the scan is wrong in the other direction too. "Ignore brand.pptx, design a fresh deck from scratch", "Do NOT use brand.pptx - it is last year's template" and "brand.pptx is the wrong file, build without a template" all OPEN the route, because naming a file to refuse it is still naming it. The suite samples `new_brand.pptx` and a brief that names no file at all, so neither direction of this is covered. It matters here only because it kills the argument that the miss above is a safe, conservative failure: it is not conservative, it is unpredictable.

MACHINE GATES on head a8685c86 / tip 0ae68805: GATES OK base=db7862ae ruff/lint-imports/commit/large/lang rc=0. `pyt.sh tests/test_subagent_routing_backend.py tests/test_agents_design_launcher.py tests/test_ppt_engine_quiet_and_tier.py` -> 117 passed; a wider run over every touched test file plus the ContextVar neighbours -> 347 passed; lint-imports "10 kept, 0 broken" including "the kernel stands alone".

WHAT HELD, so the reader knows where I looked and found nothing: the left bound (five spellings, all sound); every call site of `pick()` (exactly one, routing.py:382, and it computes `handed`, so no route is silently closed by the `()` default); `needsFile` soundness (schema.py:1846-1848 refuses a suffix without a leading dot, a bare `.`, and a full file name; `endswith` is case-folded on both sides); backward compatibility (`needsFile` defaults to `""` in schema.py, schemas/subagent.schema.json and registry.py, both spellings read, an existing `minTier`-only route behaves exactly as before); the ContextVar (`turn_attachments` is entered in exactly one place, turn_path.py:2182, on the same `with` as the already-trusted `turn_tier`, with proper token reset); and the deleted doc's two path citations, both cleaned up in the same commit.

0xKT:
**Not a blocker**, two observations from the same pass as the thread on routing.py. Neither holds up acceptance; both are things a later reader would have to re-derive.

(1) Max mode's user-facing text promises the top reasoning rung; the overlay declares the second one. This PR rewrites agents/raven-design/run.py:47 to "400 tool iterations at the highest reasoning effort the model offers" and repeats the claim in the source comment at run.py:191-192, while agents/raven-design/modes/max.json declares `"reasoningEffort": "xhigh"`. `xhigh` is not the top of this repo's ladder: raven/contracts/llm_provider.py:57 is `REASONING_EFFORT_LADDER = ("max", "xhigh", "high", "medium", "low", "minimal")`, and both agents/raven-code/modes/max.json and agents/raven-oncall/modes/max.json declare `"max"` for their own Max tier. The text is not internal prose: raven/config/product_render.py:422-448 copies `description` verbatim into the rendered `acp.modes` entry and lifts `reasoningEffort` onto the same entry, so the label and the rung a user gets sit side by side in what the mode picker shows. Either declaration could be the right one -- the commit message's "the highest word a model may know" suggests `xhigh` may be deliberate, for a provider that does not take `max` -- but then the label should say what it actually buys rather than "the highest the model offers". At the base the overlay named no effort at all and the label said "the host's full reasoning effort", which was accurate; both halves changed here.

(2) The deleted design record is still cited 22 times, by number. a8685c86 removes plugins-dist/ppt-engine/docs/ppt-raven-design.md, which defined D24-D64 (its own line 3: "from D24 on, the ppt engine's design decisions are recorded here"; D1-D23 live in the retired fork tree and are unaffected). The commit cleans up exactly the two citations that spelled the file PATH -- plugins-dist/ppt-engine/raven_ppt/profiles/registry.py:23 and tests/test_ppt_engine_quiet_and_tier.py:3 -- and leaves every citation that names only a decision NUMBER. Four such numbers are still cited (D35, D38, D41, D52), 22 times across 13 files, of which 6 files are shipped engine source:

    plugins-dist/ppt-engine/raven_ppt/profiles/registry.py:83            (D52)
    plugins-dist/ppt-engine/raven_ppt/services/gates/house_style.py:20   (D52)
    plugins-dist/ppt-engine/raven_ppt/services/gates/registry.py:179,383,403  (D52)
    plugins-dist/ppt-engine/raven_ppt/services/gates/registry.py:435     (D35)
    plugins-dist/ppt-engine/raven_ppt/services/measure/adherence.py:334  (D35)
    plugins-dist/ppt-engine/raven_ppt/services/measure/layout.py:230     (D52)
    plugins-dist/ppt-engine/raven_ppt/services/template/compose.py:902   (D41)

plus 13 more in tests/_ppt_engine_skill_claims.py, tests/test_agents_ppt_launcher.py, tests/test_ppt_engine_gates_house_style.py, tests/test_ppt_engine_gates_registry.py, tests/test_ppt_engine_measure_adherence.py, tests/test_ppt_engine_plugin.py and tests/test_ppt_engine_template.py. At the merge base every one of them resolved to a heading in the repository; at this head none of them does. Nothing breaks -- these are comments and docstrings -- but "(D52)" as the whole justification for a gate's severity is only useful while D52 can be read. `grep -rn 'ppt-raven-design'` finds the two that were fixed; `grep -rnP '\bD[0-9]+\b'` finds the twenty-two that were not. Whichever way you go -- keep the record, or strip the numbers where they are load-bearing and inline the reason -- the choice is worth making once rather than leaving the citations to rot.

For the record on the third one I checked and did not file: the ppt-engine's own `_PROVENANCE` comment (plugins-dist/ppt-engine/raven_ppt/profiles/registry.py:23) was rewritten in this PR to carry the reason inline instead of pointing at "docs/ppt-raven-design.md D2" -- which is exactly the right treatment, applied to one of the two readers in front of the author.

raven/agent/subagent/backends/routing.py
Tchen-data: Fixed in 88028db6f, with your expression: the right bound is now `(?![\w-])(?!\.[\w-])`, so a dot closes the name unless it opens a longer one. The positive spellings gain the ASCII full stop at the end of the sentence, an ellipsis, `!` and a quoted name followed by `;`; the negatives gain `brand.pptx-based`, `brand.pptx_copy`, `brand.pptxx` and `TEMPLATE.bak` mid-sentence. 100 tests pass across the routing and launcher files.

On the other direction ("Ignore brand.pptx"): agreed that naming a file to refuse it still names it, and this PR does not try to read intent from the sentence. The attachment gate is deliberately a presence test, not a parse; the lane that receives the template still decides what to do with it from the brief. Left as is, and noted in the test docstring's terms rather than promised.

Tchen-data:
(1) Fixed in 88028db6f: the Max label now says "at xhigh reasoning effort", and the source comment says why that rung and not `max` (on the providers this lane runs against, `max` spends the whole output window thinking). The overlay was deliberate; the label was not.

(2) Agreed on making the choice once. The record leaves the repository by the maintainer's decision, so the 22 citations of D35/D38/D41/D52 will be replaced by the reason inline where they are load-bearing (the gate severities and the adapt removal) and dropped where they are not. That is a comment-only sweep across six engine files and seven test files, and it will land as its own chore change rather than widen this one.

gloryfromca (commented):

I independently verified the latest delta. Sentence-final punctuation after an attached filename now opens the route, longer filename suffixes and extensions remain rejected, and the Max label accurately describes the configured `xhigh` effort. The new cases extend coverage without weakening existing assertions. I found no new findings.

I rechecked the complete diff and latest delta; AGENTS.md and relevant CONTEXT/agent documentation; spawn, DAG, and direct-chat callers; routing history; backward compatibility; test changes; and the runtime/subagent architecture boundary. The other reviewer’s thread remains theirs to resolve.

Verification:
- Affected suite: 662 passed.
- Ruff check and format check: passed.
- Source-language, large-file, commit-message, and `git diff --check` checks: passed.

raven/agent/subagent/backends/routing.py
0xKT: **Re-verified fixed** at 88028db6, closing this thread.

Re-ran the same probe on 88028db6 merged onto tip cf3c95fd (merge HEAD f3e651cc, isolated worktree), through the real `RoutingBackend.run()`. The one-character pair that opened this thread now agrees: with and without the ASCII full stop both route to Raven-PPT. All four negative controls still refuse -- `new_brand.pptx`, `my-brand.pptx`, `brand.pptx.bak`, and "any *.pptx will do".

Mutation-checked both halves of the fix, because a passing suite on a fix is the weakest evidence there is:
* reverting the right bound to `(?![\w.-])` -> `test_a_turns_attachment_opens_the_route_when_the_task_names_it` FAILED;
* dropping only the `(?!\.[\w-])` half, so `brand.pptx.bak` matches again -> `test_a_pptx_in_the_text_that_is_not_an_attachment_does_not_open_it` FAILED.

Two mutations, two different tests, each biting the half it belongs to. The tree was restored after each (`git status --porcelain` empty).

I also checked the fix cannot have opened a new hole rather than just closing this one: the new bound is a strict widening of the old in exactly one direction. The old expression refused `[\w.-]`; the new refuses `[\w-]` plus `\.[\w-]`. The difference is precisely "a dot not followed by a word character or hyphen" -- the end of a sentence. So everything the old bound admitted the new one still admits (no new false negative), and the only new admission is the case this thread was about.

Gates on the new head: `GATES OK head=88028db6 tip=cf3c95fd base=db7862ae ruff/lint-imports/commit/large/lang rc=0`; 159 tests pass across the routing, launcher, attachments, registry, product-render and ppt-tier files.

On the other direction: your answer is the right one and I am not reopening it. A presence test that does not parse intent is a defensible contract as long as it is stated as one, and the test docstring now describes what the scan does rather than promising what it means. It was in my own stated-not-filed list for the same reason -- the base was no better there, so nothing regressed.

The second observation from the plain comment (the 22 surviving `D<n>` citations of the deleted design record) is untouched by this push and stays non-blocking; it does not hold up acceptance.

0xKT (approved):

Issue #505, opened 2026-09-18
pic


ZuyiZhou:
<img width="2000" height="682" alt="Image" src="https://github.com/user-attachments/assets/8fa50c96-3d2d-4107-8a8c-7d041d1ac8b4" />
<img width="1000" height="1896" alt="Image" src="https://github.com/user-attachments/assets/379328ea-e3ee-430f-865d-68ed90dae69f" />
<img width="1528" height="520" alt="Image" src="https://github.com/user-attachments/assets/c8e9ac77-0377-487e-9d07-fb4fe51a0257" />
<img width="1000" height="1896" alt="Image" src="https://github.com/user-attachments/assets/3c5a2a9a-d6d7-40ad-b3dc-70e9f5c095df" />
<img width="1548" height="520" alt="Image" src="https://github.com/user-attachments/assets/032fdf6e-8b65-49a0-92ba-2ea3066a7451" />
<img width="2000" height="3766" alt="Image" src="https://github.com/user-attachments/assets/9ab75945-81ba-4437-aec0-9ef6a633ed55" />
<img width="1512" height="516" alt="Image" src="https://github.com/user-attachments/assets/4b606bae-667e-4f2f-ae7d-2880b47ddbf6" />
<img width="2000" height="3766" alt="Image" src="https://github.com/user-attachments/assets/b3dfce6c-366d-4e31-9a57-7c98df054618" />

ZuyiZhou:
<img width="2000" height="668" alt="Image" src="https://github.com/user-attachments/assets/92ff2c83-3b0f-49eb-9995-c1ad4c298452" />
<img width="2000" height="1082" alt="Image" src="https://github.com/user-attachments/assets/1a17eb72-485c-4b52-afcf-0bd7e73387a4" />
<img width="2000" height="668" alt="Image" src="https://github.com/user-attachments/assets/e842677a-8b71-4f1b-ae02-0162a440a1e8" />
<img width="2000" height="795" alt="Image" src="https://github.com/user-attachments/assets/8a27261f-b90f-4e4b-9f11-53d880070501" />

ZuyiZhou:
<img width="2000" height="1332" alt="Image" src="https://github.com/user-attachments/assets/75beba36-a112-46f2-949d-015856dae588" />

ZuyiZhou:
[...]

PR #576, opened 2026-09-21
docs: add webui screenshots to the launch webui section

LivXue:
## Summary

Two commits. The first fills the Launch WebUI section of both READMEs, which
carried three `Screenshot placeholder` lines and no images, with two figures
each: the new task page and the subagents page. The second lands the Raven mark
those figures show.

On the figures:

- Two, not three. The third placeholder asked for a task graph, which cannot be
  captured from stored data: the web DAG panel is driven by live `dag.*` events
  and its disk restore is gated on a localStorage note that only a live event
  writes, so showing one would have required a real `run_subagent_dag` turn.
  That slot is dropped rather than filled with something its caption does not
  match.
- They come from an isolated instance with its own `RAVEN_HOME` and an empty
  session store, so no conversation titles, memory entries or workspace file
  names appear. The A2A block was removed from its config, and the config
  carries no channels and no cron, so the throwaway instance had no outward
  side effects. Its home, which held a copy of real provider keys, was deleted
  after the run.
- Each README gets its own language, captured with `language` set to `en` and
  to `zh`. Following the convention already in these files, `alt` text stays
  English in both while the caption follows the prose language.

On the mark:

- The new file is a dark bird on an antique-gold halo. Against the dark theme's
  surface that ink measures 1.07:1, and no tone filter can lift it because it is
  not one colour to invert, so it carries an opaque plate rather than the
  `prefers-color-scheme` rule the previous file used. Ink against plate measures
  18.29:1 on both themes.
- The plate is a card, so the mark fills its frame the way `miromind.svg`
  already does. Keyed by file rather than by `data-agent`, which carries the
  preset and raven's own rows have none.
- The halo stroke goes from 1.6 to 3.2. Against a 64 unit canvas the old width
  lands under one device pixel at the two smaller slots the mark renders in,
  0.45 at 18px and 0.65 at 26px, so the ring washed out wherever the roster
  draws it.

Reviewer note: the figures were captured before the stroke went to 3.2, so the
halo in them is thinner than what this branch now renders. Everything else in
them is current. Re-capturing would mean a second upload for a difference of
half a device pixel, so they are left as they are.

Assets are hosted as a comment on issue #505 rather than committed, per the
repository asset rule.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
.venv/bin/python -m pytest tests/test_ui_agent_marks.py -q -p no:randomly
    -> 7 passed

.venv/bin/python -m pytest tests/test_scope_canon.py tests/test_docs_site.py \
    tests/test_docker_runtime.py -q -p no:randomly
    -> 27 passed

npx --prefix ui-web vitest run --root ui-web \
    scripts/agent-mark-css.test.mjs src/shell/agent-mark.test.tsx
    -> 2 files, 18 passed

make check-large-files        -> exit 0
make check-source-language    -> exit 0
make check-commits            -> exit 0
```

The gates were run over a committed range rather than an empty diff, and the
mark suite was re-run afterwards because `make check-commits` can rebuild the
virtualenv under it.

The mark change was rendered and measured rather than eyeballed: at 1440px the
mark is 28px, and the two candidate files differ only in a sub-pixel halo and a
plate that is near the tile colour in the light theme, so the two are not
visually separable at full size. The renders were compared by pixel signature
against standalone renders of each file.

Every uploaded asset was fetched back with a GET and compared by sha256 against
the local render; all four are byte identical. A HEAD request is not a valid
check here, because user-attachments URLs answer 403 to HEAD and 200 to GET.

Both README sections were rendered through the GitHub markdown API to confirm
the figures resolve, the alt text survives and each file references the assets
for its own language.

## Risk

Two user-visible changes. Both READMEs render two figures in Launch WebUI where
three placeholder lines used to be. The Raven mark in the subagents roster and
in every other slot it draws in is a different image, larger in its frame, and
no longer answers the theme from inside its own file.

Rollback is a revert of either commit independently; they touch disjoint files.
The hosted assets can stay either way.

Backward compatibility is not applicable: no interface, config key or stored
value changes.

## Related Issues

#505


gloryfromca (commented):

Reviewed the full diff against `github/main`, the surrounding bilingual README sections, relevant README history, and the applicable AGENTS.md and CONTEXT terminology rules. The four new attachment URLs each return HTTP 200 as valid 1440x900 PNGs; visual inspection confirms the English and Chinese captures match their captions and expose no user data. GitHub's Markdown renderer preserves the image links.

Verification:
- `PYTHONPATH=. uv run --frozen --python 3.13 --extra dev python scripts/check_large_files.py github/main...HEAD` passed.
- `PYTHONPATH=. uv run --frozen --python 3.13 --extra dev python scripts/check_source_language.py github/main...HEAD` passed.
- `uv run pytest tests/test_scope_canon.py tests/test_docs_site.py tests/test_docker_runtime.py -q -p no:randomly` passed: 27 tests in 5.99s.

This is documentation-only: no callers, runtime architecture, dependency contracts, or backward-compatibility paths change, and no tests are modified or weakened. The repository asset rule is respected because the screenshots remain externally hosted. (`make` is unavailable in this worktree, so I ran the exact checked-in scripts behind the required Make targets.)

Handsome-wzw (dismissed):

gloryfromca (commented):

Reviewed the new `e34c9ec99ca3..e514c0ebbc46` delta: the Raven mark replacement, its full-frame CSS treatment, and the updated contract test. I checked the affected asset-path callers, theme rules, existing size overrides, prior mark behavior, and the applicable source-language, test-naming, and UI asset rules. The literal copied asset path preserves the new CSS selector in production, and the opaque plate remains clipped by the existing tile at each supported size.

The test change preserves the prior dark-theme legibility contract while asserting the new plate mechanism and retaining the no-tone requirement; it does not weaken coverage to obtain a pass. This visual-only change does not alter runtime protocols, architecture boundaries, dependencies, or backward-compatible interfaces.

- `uv run pytest tests/test_ui_agent_marks.py -q -p no:randomly`: 7 passed.
- Focused Vitest run for agent-mark CSS/markup and both affected pages: 4 files, 199 tests passed.
- `npm run build`: passed.
- The large-file and source-language policy scripts passed for `github/main...HEAD`.

gloryfromca (commented):

Reviewed the one-line delta from the previously reviewed revision. It only changes the Python quoting used to spell the same `fill="none"` assertion, so behavior and coverage are identical. Callers, history, compatibility, architecture boundaries, repository rules, and the prior test-strength assessment are unaffected.

Verification: `uv run pytest tests/test_ui_agent_marks.py -q -p no:randomly` passed (7 tests), `uv run ruff format --check tests/test_ui_agent_marks.py` passed, and `git diff --check` passed.

gloryfromca (commented):

Reviewed the revision delta. This head only rewrites the latest commit message from `style(tests)` to `chore(tests)`; `773c99256ce6` and `c8e9d1b4c505` have the identical tree `28b312113b841631255e6f85463abca69714c32b`, and the inter-revision diff is empty. The prior code, caller, compatibility, architecture, repository-rule, and test-strength assessment therefore remains unchanged, as do its passing verification results.

Handsome-wzw (approved):

PR #578, opened 2026-09-21
feat(browser): give the model the shared browser, and a desktop MCP server to reach

TongLi31:
## Summary

The browser existed on every side except the model's: `raven/browser` drives a
Chromium the panel streams over CDP, and `browser.*` RPC gives the reader hands
on it, but `raven/agent/tools/` had no browser tool at all -- a model asked to
open a page had nothing to call. This registers eight `browser_*` tools over
that same driver, with no MCP hop, so the page the model drives is the page the
panel shows and a human can take the keyboard mid-task.

Desktop control reuses `macos-mcp` through the MCP client that was already
there; the only new code for it is a market catalog entry.

What the tools add beyond the driver:

- **Reading back.** Every acting call returns the page as it is afterwards
  (url, title, the actionable refs, the visible text), so the next call can be
  aimed without a separate read. `browser_screenshot` is the one that costs an
  image and is asked for on purpose; it goes out as an image block.
- **Who is acting.** Each call names an owner -- the sub-agent run in flight,
  else the conversation -- and the driver binds that owner to a tab. Two agents
  work in two tabs, neither can activate or close the other's, and a popup a
  click opens becomes the opener's tab. A reader's own clicks carry no owner:
  they are stamped, and the owner's next readback says the user touched the
  browser, once.
- **Permissions.** Reading and navigating default to allow, on the `web_fetch`
  reasoning (opening a page is a GET the agent could already make, and
  `raven.browser.policy` refuses what a GET must not reach). Clicking, typing
  and pressing ask, keyed by site rather than by ref, so one approval covers a
  site for the conversation.
- **Config.** `tools.browser.headfulOnAgentUse` pops the shared browser out
  into a real window on the model's first use. Off by default: a server has no
  desktop to pop into.

The tool surface grew, so the `reserved_tools` tripwire moved: descriptions
were trimmed first (~1270 to ~1090 tokens) and the bound raised from 6000 to
7400 against a measured 7115, with the trade written beside it. The schemas are
only present where the browser extra is installed; without playwright the tools
report themselves unconfigured and never reach the request.

No ui-web changes are in this PR. The web UI side of this work (a browser pane
in the desk, a per-tab miniature, docking) is written against surfaces that
`refactor/ui_web_architecture` removes, so it waits for that branch to land.

## Type

- [x] Feature

## Verification

- `uv run pytest tests/ -q` -> 23755 passed, 121 skipped, 1 failed. The failure
  is `test_cli_onboard_commands::test_clearing_does_not_disturb_a_key_the_user_exported_themselves`,
  which reads the developer's own shell environment and fails the same way on
  an unmodified `main` in this checkout.
- `uv run pytest tests/integration/test_browser_tools_real_web.py -q` -> 6
  passed against a real Chromium and a local form: fill and submit, a
  screenshot arriving as an image block, two owners in two tabs with neither
  able to take the other's, a read leaving the front tab where the last actor
  put it, a `target=_blank` popup landing its opener on the new tab, and a
  scripted model driving `AgentLoop`'s default tools end to end -- the gate
  asking once for the site, the grant covering the click that follows, and the
  panel's `browser.frame` returning the page the model left.
- `uv run ruff check raven tests`, `uv run ruff format --check raven tests`,
  `uv run lint-imports`, `make check-source-language`, `make check-large-files`
  -> clean.
- Desktop side exercised by hand through raven's own MCP client, over stdio and
  over streamable-http with a bearer token: launch Calculator, read its
  accessibility tree, click 2 + 3 =, read the result back. Screenshots arrive
  as image blocks.

Not verified: the desktop flow was driven by a script rather than by a model in
the loop; sub-agents that run in their own process (the ACP and cli lanes) have
their own `Browser` singleton and are not given these tools; Windows and Linux
desktop executors were not chosen.

## Risk

Notes. The acting verbs sit on the ask tier by default and a grant is scoped to
a site and a conversation, so an approved click on one host does not carry to
another. Navigation goes through the existing outbound policy. The desktop MCP
server runs locally under `uvx` and needs macOS Accessibility and Screen
Recording grants, which the operating system asks for itself; the catalog entry
is opt-in like every other market entry. Nothing here changes an existing
default: without the browser extra the tools are absent, and with it the only
new always-on cost is the tool schema.

## Related Issues

N/A


gloryfromca (commented):
Blocking: replace recyclable run ownership, honor browser availability in sub-agents, scope console readback by page, and fix Linux integration setup.

I reviewed `github/main...HEAD`, the browser RPC callers and tool-registry/permission lifecycle, relevant history, configuration and catalog changes, and the surrounding tests. I also checked AGENTS.md and CONTEXT-MAP.md/CONTEXT.md, backward compatibility of the existing owner-less RPC calls, architecture boundaries, and whether the token-budget assertion was merely weakened; its raised bound is quantified and tied to the deliberately larger tool surface.

Verification:
- `uv run --frozen --python 3.12 --all-extras pytest tests/test_browser_tools.py tests/test_browser_driver.py tests/test_browser_policy.py tests/test_agent_loop_token_budget.py tests/test_agents_design_launcher.py tests/test_agents_ppt_launcher.py tests/test_market.py -x`: 327 passed.
- Targeted Ruff check and format check: passed.
- `scripts/check_large_files.py github/main..HEAD` and `scripts/check_source_language.py github/main..HEAD`: passed (invoked directly because `make` is unavailable here).
- `uv run pytest tests/integration/test_browser_tools_real_web.py -x` after the documented install: 4 failures before xdist stopped, because the fixture redirected Linux to the macOS cache path. With that path overridden, Chromium was found but this host lacks `libatk-1.0.so.0`, so the real-browser behavior could not be completed on this machine.

raven/agent/tools/browser.py
```diff
@@ -0,0 +1,641 @@
...
+    Checked in that order because a sub-agent's task inherits the parent turn's
+    context variables, conversation id included; the run is the finer identity
+    and the one that must not share a tab with its parent.
+    """
+    from raven.agent.subagent import activity
+    from raven.token_wise import usage_context
+
+    run = activity.current()
+    if run is not None:
+        return f"run:{id(run):x}"
```
gloryfromca: Blocking: use a collision-safe identity for each sub-agent run.

`id(run)` is reusable as soon as `activity.collecting()` exits, while the driver keeps the resulting `run:<id>` binding alive for 600 seconds. In a three-iteration CPython reproduction, the second run's ID was immediately reused by the third. A later, unrelated run can therefore inherit the prior run's browser tab and page contents, potentially across parent conversations. The owner key needs lifetime uniqueness, or the binding must be explicitly released when collection ends.

raven/agent/subagent/backends/raven_loop.py
```diff
@@ -390,6 +390,14 @@ def allowed(name: str) -> bool:
...
                 WebFetchTool(api_key=self._web_key(fetch_provider), proxy=self.web_proxy, provider=fetch_provider)
             )
+        # The same browser the parent drives, in a tab of this run's own: the
+        # tools name the run in flight as their owner, so two sub-agents
+        # browsing at once are two tabs, never one page typed into twice.
+        from raven.agent.tools.browser import browser_tools
+
+        for tool in browser_tools():
+            if allowed(tool.name):
+                tools.register(tool)
```
gloryfromca, line 425: Blocking: withhold unavailable browser tools from raven-loop sub-agents.

`configured()` is only captured on `ToolSpec`; `ToolRegistry.get_definitions()` does not consult it. The main `AgentLoop` adds failed predicates to its withheld source, but this standalone sub-agent registry only installs `mcp_source.disabled_tools()` and then registers every browser tool. With `Browser.probe()` returning false, I reproduced all eight specs reporting unconfigured while all eight still appeared in `get_definitions()`. Thus a normal source install without the optional browser extra advertises tools to in-process sub-agents that can only fail, contrary to the documented withholding behavior.

raven/browser/driver.py
```diff
@@ -535,29 +715,30 @@ async def go(self, direction: str) -> dict[str, Any]:
...
+    async def snapshot(self, *, owner: str | None = None) -> dict[str, Any]:
         """The page as things a caller can act on, with stable refs."""
-        page = await self._ensure()
+        page = await self._page_for(owner, act=False)
         try:
             data = await page.evaluate(_SNAPSHOT_JS, MAX_REFS)
         except Exception as exc:
-            return await self._state(error=str(exc))
+            return await self._state(error=str(exc), page=page)
         data["console"] = list(self._s.console[-20:])
```
gloryfromca: Blocking: keep snapshot console data scoped to the selected owner page.

The snapshot now selects a page by owner, but it still attaches `_s.console`, which `_on_console` populates from every wired tab. I reproduced an owner-B snapshot of `b.test` returning an error placed in the shared buffer by owner A. Because every browser-tool readback renders these warnings/errors as the current page's console, concurrent agents receive unrelated page state and can act on it; page-specific readback needs page-specific console storage or filtering.

tests/integration/test_browser_tools_real_web.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def log_message(self, *args: object) -> None:
+        pass
+
+
+@pytest.fixture(autouse=True)
+def _browsers_from_the_real_home(monkeypatch: pytest.MonkeyPatch) -> None:
+    if not os.environ.get("PLAYWRIGHT_BROWSERS_PATH"):
+        real_home = pwd.getpwuid(os.getuid()).pw_dir
+        monkeypatch.setenv("PLAYWRIGHT_BROWSERS_PATH", os.path.join(real_home, "Library", "Caches", "ms-playwright"))
```
gloryfromca: Blocking: make the browser integration fixture use Playwright's platform cache.

On Linux, the documented `uv run playwright install chromium` installs under `~/.cache/ms-playwright`, but this autouse fixture overwrites an unset `PLAYWRIGHT_BROWSERS_PATH` with the macOS-only `~/Library/Caches/ms-playwright`. I ran the documented install and then this file: four tests failed before xdist stopped, all because Chromium was reported missing and the suggested reinstall command writes to the other directory again. Preserve Playwright's default or choose the cache path by platform so the new suite is runnable on Linux.

raven/agent/tools/browser.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed by giving every `RunActivity` a UUID-backed `uid` and deriving the owner from that stable run identity. The regression coverage also checks that the owner no longer exposes the Python object address.

raven/agent/subagent/backends/raven_loop.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed by composing MCP-disabled tools with every registered spec whose `configured()` predicate fails. The new test toggles the browser probe and confirms the standalone raven-loop registry withholds and restores the browser definitions.

raven/browser/driver.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed by storing console records on the wired page and reading only the selected page’s buffer. The new two-owner regression test confirms one owner’s console output is absent from the other owner’s snapshot.

tests/integration/test_browser_tools_real_web.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed with platform-specific Playwright cache paths. On this Linux host the integration test now finds Chromium in `~/.cache/ms-playwright` and reaches browser launch; it stops only because the host lacks Chromium runtime libraries such as `libatk-1.0.so.0`, not because of the fixture path.

gloryfromca (commented):

The new revision fixes all four prior blockers: delegated runs now use lifetime-safe UUID identities, unavailable browser tools are withheld from raven-loop sub-agents, console records are page-local, and the integration fixture follows Playwright’s platform cache. I found no new correctness issue.

I covered the full `github/main...HEAD` diff, the fix delta and commit history, affected callers and registry/driver behavior, backward compatibility, test changes (including checking that tests were not weakened), the browser/sub-agent architecture, and the applicable AGENTS.md and CONTEXT-MAP/CONTEXT.md rules.

Verification:
- 403 targeted unit tests passed.
- Ruff check and format check passed on all changed Python files.
- Source-language, large-file, and `git diff --check` gates passed.
- The real-browser integration test now locates Chromium correctly on Linux and reaches launch, but this host cannot complete it because Chromium runtime libraries including `libatk-1.0.so.0` are absent.

The four settled review threads have been replied to and resolved.

gloryfromca (commented):

This revision only removes the annotation that the type checker rejects on the page attribute; the assignment and per-page console behavior are unchanged. I found no new correctness issue.

I rechecked the complete change against its reviewed parent, affected callers, history, backward compatibility, test integrity, browser/sub-agent architecture, and the applicable AGENTS.md and CONTEXT rules.

Verification:
- `ty check` passed across the repository Python type-check targets.
- 89 affected browser and sub-agent tests passed.
- Ruff check and format, source-language, large-file, and diff checks passed.

0xKT:
**Not a blocker** on any of these -- I am accepting the PR. Four observations, in the order I would fix them, plus the one thing I went looking for hardest and did not find.

(1) `browser_tabs` is auto-approved, and its `close` action can shut the tab the person is watching -- and with the last tab, the browser.

`browser_tabs` is in `DEFAULT_ALLOW_TOOLS` (rules.py:85), whose stated criterion is that the worst case is reading. Its schema offers `close` (browser.py, the `action` enum `["list","new","activate","close"]`), and `driver.tab_close` guards only against closing ANOTHER OWNER's tab: `if holder is not None and holder != owner` (driver.py:626-628). A tab with no owner is, in this PR's own vocabulary, the reader's -- so it falls through the guard and closes, with no prompt. `tab_close`'s own docstring says what follows: "closing the last one closes the browser (native)". The listing even advertises the reader's tab as free (`held=False`).

The sharp version: the model's own `HANDOFF_NOTE` (browser.py:43-46) asks the person to finish a login or a CAPTCHA in the panel, and the same auto-approved tool can close that tab under them. Moving `close` alone behind the per-site approval the acting verbs already use would cost one entry in `BROWSER_SITE_KEYED_TOOLS`, or a holder check that treats an unowned tab as the reader's rather than as free.

(2) A read that has to open a tab moves what the panel shows -- which `_page_for`, the docs, and two tests all say it must not.

`_page_for(owner, act=False)` promises "a read from either side must not move what the panel shows" (driver.py:496-497). When the read belongs to an owner with no tab and the active tab is held by someone else, it calls `context.new_page()` (driver.py:517); that fires the "page" event registered at :303, and `_adopt` sets `self._s.page = page` (:456) and re-streams (:461) unconditionally, because the read path never assigned `_s.page` first. The act path is safe for the opposite reason: `_focus` assigns before its first await, so `_adopt`'s guard is already true.

This is the delegation case the feature exists for: the parent navigates, then a sub-agent's first call is `browser_snapshot` -- the tool whose own description says to call it whenever the page may have moved. The panel jumps to a blank tab because an agent READ. Both tests that encode the rule (test_browser_driver.py:728, test_browser_tools_real_web.py:190) pre-bind the owner and stub the context, so neither reaches the branch.

(3) CONTEXT.md and docs/browser-and-desktop.md disagree about the touch note, and CONTEXT.md is the wrong one.

CONTEXT.md:1664 defines the new term: calls with no owner "are stamped as a touch, which the tools report to the model once." The flag is cleared only on an ACTING readback -- `self._mark(owner)` sits under `if acted` (browser.py:203-204) and `BrowserSnapshotTool` never marks -- so every snapshot after a reader touch repeats the note indefinitely. docs/browser-and-desktop.md:53-55 describes the real behaviour ("until the owner acts again"). Per AGENTS.md section 6 a coined term has to be verifiable against the code; here the doc is right and the CONTEXT.md sentence is not.

(4) Nit: two places decide which browser verbs are site-keyed, and nothing holds them equal. The frozenset `BROWSER_SITE_KEYED_TOOLS` (builtin.py:134) and the `_ActingTool` hierarchy in browser.py (:225, with click/type/press at :392/:441/:480). The PR spots the same hazard one field down and says so out loud -- "spelled here as well as there so neither side can rename it alone" (builtin.py:136-137) -- but that care covers the parameter NAME, not the tool SET. A fourth acting verb added to `_ActingTool` without the frozenset entry keys per call instead of per site, and nothing goes red.

(5) Wording, in the comment that justifies the new permission tier. rules.py:54-58 says the reading verbs are allowed "on the web_fetch reasoning: opening a page is a GET the agent could make anyway". The agent's own GET refuses the destinations the browser allows. Measured on this head, same URL both ways:

    http://127.0.0.1:<port>/admin      web_fetch guard: Blocked (private)   browser guard: allowed
    http://10.0.0.5/internal           web_fetch guard: Blocked (private)   browser guard: allowed
    http://169.254.169.254/...         web_fetch guard: Blocked             browser guard: REFUSED (link-local)
    file:///etc/passwd                 --                                   browser guard: REFUSED (scheme)
    https://example.com                web_fetch guard: allowed             browser guard: allowed

(`raven.security.network.validate_url_target` vs `raven.browser.policy.navigation_refusal`; the last row is the control that keeps the first two honest.)

I want to be explicit that I chased this as a blocking finding and it does not stand, because the reasoning matters more than the verdict. The stance is not an oversight: policy.py:88-94 states it and argues it -- loopback and private "deliberately NOT refused ... a reader pointing the page at their own dev server is the ordinary case, and breaking that to close a hole that link-local already closes would trade a real feature for no gain" -- policy.py:41-46 documents the deployment lever that flips it, and policy.py:3-6 already scoped the policy to an agent caller reading instructions off a page it visited. None of those three files is touched by this PR: they came in on 2026-08-30 under "feat(browser): a deployment can refuse loopback and private targets", a separately reviewed decision by another author. So the destination policy is settled repo policy and not this PR's to answer for.

What IS this PR's is the sentence. "A GET the agent could make anyway" is the one claim in that comment that the tree contradicts, and it is load-bearing precisely because it is the reason given for skipping the prompt. Say what is true instead -- that the browser's destination policy is deliberately wider than the egress fetch's, and that `RAVEN_BROWSER_BLOCK_PRIVATE` is the lever -- and a reader who later wonders why these five verbs skip the prompt gets the real answer rather than one that dissolves on inspection.

WHAT I CHECKED AND FOUND SOUND, so the shape of the search is visible. `file://`, `data:` and `blob:` are refused by scheme, and 169.254.169.254 by address -- I ran both. The profile is raven's own (`~/.raven/browser-profile`), not the user's Chrome, so no pre-existing session of theirs is exposed. The desktop MCP server is stdio, so it opens no port; its catalog entry pins `macos-mcp==0.4.6`, and it is the ONLY one of the sixteen stdio entries in that catalogue that pins a version at all -- the other fifteen float on `@latest` or nothing, so this entry is stricter than the convention rather than looser. `risk_tier: 2` is the highest tier the catalogue actually uses (25 entries at 1, 16 at 2, alongside playwright and chrome-devtools), and both `risk_tier` and `publisher.verified` are read by code (market/catalog.py:118-120) rather than being decorative.

GATES on head 28f18a9c / tip 87f39286: `GATES OK base=cf3c95fd ruff/lint-imports/commit/large/lang rc=0`. `pyt.sh tests/test_browser_driver.py tests/test_browser_tools.py tests/test_agent_loop_token_budget.py tests/test_subagent_workdir.py tests/test_agents_design_launcher.py tests/test_agents_ppt_launcher.py` -> 188 passed.

0xKT (approved):

LivXue:

Scope: I reviewed only the `raven/agent/**` portion of this PR - 4 of 25 changed files (`tools/browser.py`, `loop/wiring.py`, `subagent/activity.py`, `subagent/backends/raven_loop.py`). The remaining 21 files were read as context but are not covered by this verdict. This landed after the merge, so it carries no inline notes: the findings below are records against code now on `main` at 608cedf18, and the two that want action are tracked in #600.

What I examined: the in-scope diff in full (`tools/browser.py` is a new file, so every line of it); the ownership contract it depends on in `raven/browser/driver.py`; the withholding parity between `_withheld_here` and the main loop's `_unconfigured_tool_names`; the permission surface in `raven/permissions/`; the five product configs; and the rounds already on this PR.

Confirmed findings:

- `_BrowserTool._acted` (browser.py:179) is never pruned, so every delegated run leaves a permanent entry; the driver's owner map it mirrors is reaped on a timer.
- `_ActingTool.cast_params` (browser.py:231) can write into the permission parameters a site the call will not act on, and the grant that banks covers later real actions on that site.
- `browser_tabs`' description (browser.py:551) does not mention that `close` now refuses an unheld tab.

Outside the reviewed scope, so recorded here rather than inline:

- `_adopt` (driver.py:456) now returns before `self._wire(page)`, which the previous code called first. An opener-less page arriving while `spawning` is non-zero is therefore never wired - no console capture, not adopted, still counted against `MAX_TABS`. Code-reading only: I could not establish on this host whether Playwright reports a null opener for `rel="noopener"` popups, which is the premise, so this is a question rather than a finding.
- The `BrowserBusyError` text at driver.py:531 says "close one before opening another", but an agent can no longer close an unheld tab, and tabs left behind by reaped owners are all unheld. The recovery is activate-then-close.
- `test_site_keyed_permission_set_matches_the_acting_tool_hierarchy` uses `_ActingTool.__subclasses__()`, which returns direct subclasses only, so a fourth acting verb added as a subclass of an existing one would escape the equality that test exists to hold.

Verification, all at fe3b6278f, with the head materialised via `git archive` and module resolution confirmed by printing `__file__`:

- `pytest tests/test_browser_tools.py tests/test_browser_driver.py tests/test_agent_loop_token_budget.py tests/test_subagent_workdir.py`: 99 passed.
- `lint-imports` run from the head tree: 851 files, 5856 dependencies, 10 contracts kept, 0 broken - including "cargo does not import the loop shell". (Run from the main checkout with the head only on `PYTHONPATH` it silently analyses the wrong tree: 850 files, identical to main.)
- `ruff check` and `ruff format --check` on the four in-scope files: clean, 4 already formatted.
- Tool-surface cost, measured with the same `estimate_prompt_tokens` the budget uses: `reserved_tools` 5946 at the merge base against 7115 at this head, so +1169; `reserved_system` unchanged (1207-1222 across repeated runs at both revisions, the spread being workspace path length). The eight schemas measured on their own come to 1170. The raised assertion in test_agent_loop_token_budget.py is a real measurement rather than a weakened bound.

Checked and deliberately not reported: `page._raven_spawned` assignment is safe, since no class in `Page.__mro__` defines `__slots__`; `"session:default"` cannot collapse two conversations onto one owner, because `usage_context.bind(cid, ...)` at turn_path.py:2866 wraps the whole turn body including tool dispatch; `_withheld_here` cannot recurse through the source it is installed as, because `names()` is unfiltered; all five `agents/*/config.json` opt in, none missed; no other `ToolRegistry` builder needed the same edit, as `raven/acp/methods.py` builds an MCP-holder registry and ACP turns route through the main loop. The four blockers from gloryfromca and the five observations from 0xKT are all answered at this head, and none of them is re-filed here.

PR #555, opened 2026-09-20
docs: rebuild the showcase layout and refresh the comparison board

ZuyiZhou:
## Summary

Rebuilds the Showcase layout in both READMEs, and replaces one image.

**Why the old layout did not hold.** The section had grown to nine cases
across three tables of mixed shape. A heading that wrapped to two lines
pushed its column's images down. Paired task graphs of unequal height left
the artifacts below them starting at different heights. And GitHub's
stylesheet stripes every second row while a case spanned three, so the
pattern drifted until some headings sat on white and others on the grey.

**What it is now.** Each case spans three rows - heading, task graph,
artifact - so cells in a row share a top edge and a long heading cannot
cascade into the images. Each pair gets its own table, which restarts the
stripe and puts the usual gap between pairs. The game run keeps the opening
slot and now takes the full width instead of leaving half a row empty.
Image widths are set per pair so both sides render to one height, and the
narrower image is centred. The divider sentence that separated the old
tables is gone, since the headings already say it.

**A new comparison board.** The Frameworks artifact is replaced with a
redrawn plate: each framework carries its vendor mark, and the
orchestration style became a three-way tag - explicit graph or canvas,
declared task flow, dynamic at runtime - where the old plate split the six
in two. Its alt text said "versus" and named two categories; it names three
now. The render is 2000x1547 against the old 2000x1332, so its width is set
to 86 percent to stay level with the sweep chart beside it.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `make check-large-files` -> exit 0
- `make check-source-language` -> exit 0
- The replacement board was fetched anonymously from its attachment URL:
  HTTP 200, 2000x1547, and its md5 matched the local source file byte for
  byte.
- Rendered the section through GitHub's own markdown endpoint under the
  published markdown stylesheet and read back the computed row backgrounds.
  All five heading rows now report one background; before the split they
  reported two.
- Measured every image box in that render. All four pairs match on top edge,
  and within 2px on height:

  | pair | left | right |
  | --- | --- | --- |
  | Song / Greece | 379x129 and 379x718 | 379x129 and 379x718 |
  | Pop / Abstract | 379x127 and 379x713 | 371x127 and 379x713 |
  | Frameworks / Sweep | 379x126 and 326x252 | 379x126 and 379x252 |
  | Poster / Explainer | 314x131 and 379x348 | 379x131 and 367x347 |

[...]

PR #579, opened 2026-09-21
docs: replace the sweep chart with the run's dashboard

ZuyiZhou:
## Summary

The parameter sweep case shipped a two-panel matplotlib plot. The same run
also produced a dashboard, and it carries what the plot could only assert:
the sixteen measured cells as a grid with recall and latency in each, the
best cell ringed at top_k 10 and chunk_size 1024, and a footnote naming the
single warm-up sample behind the 256/top_k=3 outlier. This swaps the plot
for the dashboard and rewrites the alt text to describe that grid rather
than the two lines.

**Language.** The dashboard was authored in Chinese. Both front pages share
one image set, and the comparison board beside this cell is already English,
as is the alt text on both pages, so what lands is an English render built
the same way that board was. The Chinese original is left untouched on disk
and nothing in the repository refers to it.

**Widths.** The new render is 2000x1568 against the board's 2000x1547, so
the two sit almost exactly on one aspect. The board goes back to full width
and the dashboard takes 99 percent; the board only needed 86 percent against
the old plot's 2000x1332.

## Type

- [ ] Fix
- [ ] Feature
- [x] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
git diff --check origin/main...HEAD
  clean

COMMIT_RANGE=origin/main...HEAD make check-large-files
  exit 0

PYTHONPATH=. uv run --frozen --python 3.12 --extra dev python scripts/check_source_language.py origin/main...HEAD
  exit 0

make check-commits
  exit 0

PR_TITLE="docs: replace the sweep chart with the run's dashboard" make check-pr-title
  exit 0

git merge-tree --write-tree HEAD origin/main
  clean
```

The attachment was fetched anonymously before it was referenced: HTTP 200,
2000x1568, md5 fcd434b3000cbcda9339e98c22bf974b, matching the local render
byte for byte.

Rendered the section through GitHub's markdown endpoint under the published
markdown stylesheet and measured every image box. All eight pairs share a
top edge and match within 2px on height; the Frameworks pair is now 379x293
against 375x294, where before this change it was 326x252 against 375x294.

Both front pages were compared after the edit: each carries the identical
set of attachment ids, and neither still references the old plot.

No test suite is relevant to a change that edits two markdown files. The
python lint job was not run locally for the same reason; CI runs it on the
head.

## Risk

Documentation only, in both READMEs. One attachment reference is replaced
and one width attribute is restored; no file is committed to the repository.
The old plot remains reachable at its own attachment URL, so a revert of the
commit restores it with no other action.

## Related Issues

Follows #555.


gloryfromca (commented):

I reviewed the complete `github/main...HEAD` diff, the surrounding bilingual showcase markup, the history of this section, and the current PR record. I also covered the repository rules in AGENTS.md/CLAUDE.md and CONTEXT-MAP.md, compatibility and consumers, test integrity, and architecture constraints. This is a documentation-only replacement, so there are no runtime callers, backward-compatibility effects, weakened tests, or architecture-boundary changes.

The new GitHub attachment resolves as a 2000x1568 PNG (538,929 bytes). I inspected it directly: its 16-cell grid, broadly flat latency, and highlighted `top_k = 10`, `chunk_size = 1024` best cell agree with the alt text. Its dimensions closely match the neighboring 2000x1547 board, so the 100%/99% sizing keeps the row balanced.

Verification passed:
- `git diff --check github/main...HEAD`
- `scripts/check_large_files.py github/main..HEAD`
- `scripts/check_source_language.py github/main..HEAD`
- `scripts/check_commit_messages.py github/main..HEAD`
- `scripts/check_pr_title.py`

The repository's GitHub checks were passing except for one unit-test shard still pending when reviewed; the docs build and repository-file checks had passed. The full local test suite was not rerun because the change is limited to external image links and README markup.

0xKT:
**Not a blocker.** Reviewed by hand. The swap is clean and both front pages
change identically -- I diffed the two hunks against each other and they are
byte-for-byte the same, 2 removed and 2 added in each file, with nothing else
touched. All three asset URLs resolve (HTTP 200, image/png), the old plot has no
remaining reference, and the new one is referenced only from the two READMEs.

The width arithmetic in the commit message checks out. Measured from the PNG
headers: the dashboard is 2000x1568, the comparison board 2000x1547, the old
plot 2000x1332. So the old `width="86%"` on the board is 1332/1547 = 86.1%, and
at 100%/99% the new pair stands level. Exactly as described.

I also opened the image rather than trusting the alt text, and every claim in it
holds: a sixteen-cell grid with recall and latency in each cell, the best cell
ringed at top_k 10 / chunk_size 1024, the 256/top_k=3 outlier footnoted as one
68.4 ms warm-up sample, and the whole render in English.

One thing the image says that its own data does not support:

**"BEST CELL ... recall 1.000, latency 44.3 ms (lowest in the grid)".** The grid
beside it reads 43.2 ms at chunk_size 128 / top_k=5, which is lower. 44.3 ms is
the lowest among the recall-1.000 cells (46.9 / 45.8 / 46.1 / 44.3), which is
almost certainly what was meant -- the cell CHOICE is right, it is the
parenthetical that overreaches. "lowest at recall 1.000" would be true.

The same subset shows up a second time: the colour scale under the grid runs
"44 ms" to "57 ms", and 43.2 ms is below its low end, so the scale was built
from the same set the "lowest" claim was.

Small, and the substantive conclusions above it are all correct -- recall is set
by top_k, latency is broadly flat, and the ringed cell is the right one to
recommend. But it is the front page, the whole point of the swap is that the
dashboard "carries what the plot could only assert", and a reader who reads the
grid will find the one sentence that does not survive it.

0xKT (approved):

PR #585, opened 2026-09-21
fix(*): keep archiving, fix the usage counts, and start conversations in smart mode

gloryfromca:
## Summary

Three reports from acceptance testing, one branch. All three were reproduced on a
real host before anything was changed, and the fixes were walked through the same
host afterwards.

**Archiving did not hold, and the archive page did not show it.** Three separate
causes:

- `session.archive` persisted by saving the whole session, which rewrites the
  metadata record from that process's copy of it. A page and a terminal over one
  home are two managers over one file: the other one had loaded the conversation
  before the archive, so its copy carried no `archived` key, and its next save of
  anything spoke for a flag it had never seen. The verb now appends the one key,
  the way the auto-archive pass already did. `append_metadata_patch` reports
  whether the patch reached a file, which is what lets the verb answer `pending`
  truthfully.
- The rail read that `pending` as a success, so a conversation with no transcript
  yet left the rail although nothing had reached the disk.
- The archive page asked `session.list` for fewer channels than the rail lists, so
  an archived scheduled run was invisible there and unreachable here.

**The usage page reported no data.** It read its range two ways. Calls, cost and
the per-model table came from the daily telemetry files, dated per day; the tool
table was topped up by reading session transcripts, and a transcript counted as
in-range when its file mtime was. A conversation touched recently therefore handed
the range every tool call it had ever made while its model calls, dated per day,
stayed outside it -- thousands of tool calls beside no model calls at all, which
reads as a page that failed to load. Telemetry has carried a row per tool call for
a while, so the transcript fallback only ever covered older conversations; dropping
it leaves one source and one reading of the range.

**A conversation now starts in the smart permission mode.** `ask` stopped the agent
on every mutation, and a prompt answered by reflex is not a gate. Smart is not the
weaker setting it sounds like: builtin denials and user deny rules hold in every
mode, the reviewer speaks only for the ask tier, and a reviewer that cannot run
leaves the call at the same prompt `ask` would have shown.

**Shared by the first two:** opening the settings dialog reloads the config
snapshot, and neither the archive list nor the usage totals come from it. Nothing
dropped them, so a dialog opened once held its first answers for the life of the
page. Both are now cleared on open, which their own lazy loads already key off.

**What this does not close.** `save` rewrites the whole metadata record from one
process's copy of it, and this branch changes who can do that rather than whether
it can be done. Archiving no longer erases what another writer added; five other
verbs still can (pin, rename, the per-conversation model and permission mode, the
ACP usage owner), and so does the turn path, which saves on every message and is
the one a user actually hits. So a conversation archived from one client and then
typed into from another still comes back.

Closing that needs `save` to stop speaking for keys it never touched, and removal
is currently expressed by omission -- unpin, `title_auto` and `output_limit_turn_at`
each delete a key by not writing it, so a merge on save would resurrect them. They
have to become explicit false values first. That work is a round of its own rather
than a fourth topic on this branch.

The ACP hosting pins the ask tier rather than inheriting the new default: the
person there is in an editor watching an agent work on their checkout, and the
prompt is how they see each write before it lands. The four other products carry no
permissions block and were written as things raven dispatches rather than things a
person drives, so they follow trunk.

## Type

- [x] Fix
- [x] Feature

## Verification

Backend, after the rebase:

```
uv run pytest tests/test_rpc_session.py tests/test_rpc_settings.py \
  tests/test_rpc_config.py tests/test_config_live.py \
  tests/test_rpc_contract_shapes.py tests/test_session_manager.py -q
# 498 passed

uv run ruff check raven tests          # All checks passed
uv run ruff format --check raven tests # 1636 files already formatted
make check-source-language             # pass
make check-large-files                 # pass
```

Frontends, after the rebase:

```
cd ui-web && npx vitest run            # 189 files, 2592 passed
cd ui-web && npx tsc --noEmit          # clean
cd ui-tui && npx vitest run            # 144 files, 2074 passed
cd ui-tui && npm run type-check        # clean
```

Every new or changed case was checked against the base revision with the fix
removed, by restoring the file from HEAD rather than stashing, and the files were
compared byte for byte afterwards:

| Case | On base |
|---|---|
| `test_archiving_keeps_a_key_another_writer_added` | fails: the foreign `pinned` key is gone after the archive |
| `test_usage_counts_a_tool_call_on_the_day_it_was_recorded` | fails: a transcript touched today adds two tool calls the range never saw |
| `opening again clears what the pages fetched for themselves` | fails: both fields survive the reopen |
| `says it failed when the call answers that nothing was persisted` | fails: `pending` is read as a success |
| `archived lists the archived sessions of every channel the rail shows` | fails: the call carries no channels |

Real host, `raven serve` on a temporary `RAVEN_HOME` with real provider credentials,
driven through the built page:

| Check | Before | After |
|---|---|---|
| Archive a conversation from the rail, then reopen Settings > Archive | the list still shows the previous answer | the conversation is at the top of the list |
| Telemetry rows on disk vs the usage page's call count, dialog reopened without a page reload | 13 on disk, 12 on the page | 13 and 13 |
| Tool table against the `tool_call` rows for the range | inflated by whole transcripts | `exec` 4, the number of rows |
| A fresh page with no `permissions` key in the config and no stored pick | chip reads "ask" | chip reads "smart" |
| `mkdir -p /tmp/...` through `exec` on that host | approval dialog, directory not created | reviewed and allowed, directory created |

The last row is the pair that makes the check able to fail: the same command was
run with `permissions.mode` set to `ask`, and it stopped at the prompt.

## Risk

The permission default is the user-visible change: an install that has never
written `permissions.mode` moves from being asked about every mutation to having
the reviewer speak for the ask tier. Denials and deny rules are unaffected, and a
reviewer that cannot run still leaves the call at the prompt. Anyone who wants the
old behaviour writes `permissions.mode: "ask"`, and rolling the default back is the
one literal in `PermissionsConfig.mode` plus the three places that repeat it where
they cannot read it.

Usage numbers will change for anyone whose home has conversations older than the
day telemetry started carrying tool rows: their tool counts drop to what the range
actually saw. That is the point of the change, but it will look like a loss.

`append_metadata_patch` gained a return value. The existing caller ignores it, so
nothing else moves.

## Related Issues

N/A


raven/rpc/methods/session.py
```diff
@@ -839,19 +839,23 @@ async def session_archive(
...
-        return {"archived": archived, "session_key": session_key, "pending": False}
-    return {"archived": archived, "session_key": session_key, "pending": True}
+    #
+    # One appended key rather than a saved session, the same way the auto
+    # archive pass writes it: ``save`` rewrites the whole metadata record from
+    # this process's copy of it, so a second client holding the conversation
+    # from before the archive dropped the flag on its next save of anything --
+    # a title, a model -- and the conversation came back.
+    try:
+        persisted = mgr.append_metadata_patch(session_key, {"archived": archived})
```
gloryfromca, line 852: Blocking: archive persistence must survive a later save from a manager that loaded the session before the archive.

This append preserves keys that were already on disk when the archive call runs, but it does not fix the order described immediately above it. With two `SessionManager` instances, load the session through manager B, archive through manager A, then add a message and call `manager_b.save(stale_session)`: the unchanged `save()` appends B's whole stale metadata record, and last-record-wins loading removes `archived` again. I reproduced exactly that sequence; the metadata first contained `archived: true` and then lost the key after B's save. The new test exercises only the reverse order (foreign patch first, archive second), so it cannot catch the reported reappearance path. The archive flag and ordinary saves need a persistence scheme that preserves intervening metadata in this order too.

raven/config/schema.py
```diff
@@ -1268,9 +1268,16 @@ class PermissionsConfig(Base):
...
+    ``smart`` out of the box. ``ask`` stopped the agent on every mutation of a
+    conversation, which a reader answers by reflex rather than by reading, and
+    a prompt answered by reflex is not a gate. Smart is not the weaker setting
+    it sounds like: builtin denials and user deny rules hold in every mode, the
+    reviewer speaks only for the ask tier, and a reviewer that cannot run
+    leaves the call at the same prompt ``ask`` would have shown.
     """
 
-    mode: Literal["ask", "smart", "full"] = "ask"
+    mode: Literal["ask", "smart", "full"] = "smart"
```
gloryfromca, line 1280: Blocking: keep ACP editor sessions in ask mode while changing ordinary new conversations to smart mode.

This schema-wide default also changes the config rendered by `agents/raven-code` when it has no explicit permissions block. That hosting deliberately retains `ask`: Raven-dispatched sub-agents have an approval responder, while a person using the agent from an editor must still receive the mutation prompt. The existing contract test `test_the_acp_render_leaves_the_ask_tier_alone` now gets `smart` and fails locally and in GitHub's unit shard, so this is a behavior regression rather than a stale expectation. Preserve the hosting-specific `ask` default while making the normal product conversation default `smart`.

gloryfromca (commented):
Blocking: archive state can still be erased by a stale writer, and ACP editor sessions lose required ask-mode prompts.

I reviewed the complete target diff plus the affected archive/save callers, usage telemetry path, permission-mode consumers, commit history, and the Runtime/Web/TUI architecture and naming constraints. I also checked that tests were not weakened to manufacture green results and considered backward compatibility.

Verification:
- `uv run pytest tests/test_rpc_session.py tests/test_rpc_config.py tests/test_config_live.py -x` - 267 passed.
- `PYTHONPATH=plugins-dist/everos-memory uv run pytest tests/test_rpc_settings.py -x` - 94 passed.
- Web focused tests - 40 passed; web type-check passed.
- TUI permission tests - 7 passed; TUI type-check passed.
- The dedicated ACP default-mode test fails locally, matching the failing GitHub unit shard.
- A direct two-manager reproduction shows `archived: true` immediately after the patch, then no `archived` key after the stale manager saves.

No other blocking or nonblocking findings survived refutation.

raven/config/schema.py
gloryfromca: Blocking: archive persistence still has to survive a stale manager save.

This ACP-specific finding is withdrawn. The smart-mode path does not remove the human approval channel: only a reviewer allow bypasses it, while an escalation or any reviewer failure produces `NeedsApproval` and reaches the interactive responder. Given the changes explicit product-wide smart default, the renamed assertion is consistent with that behavior rather than a test weakened around a regression.

gloryfromca (commented):
Blocking: archive persistence must still survive a later save from a manager that loaded the session before it was archived.

Reviewed the delta from `c0139daf1ac4`, the full target diff, the affected permission-gate and session-manager callers, commit history, backward-compatibility implications, and the Runtime/Web/TUI architecture and repository rules. The changed ACP test was checked against the actual smart-mode waterfall rather than accepted merely because it is green: reviewer escalation and reviewer failure still reach the human responder, so that earlier finding has been withdrawn and its thread resolved.

The archive path is unchanged. The two-manager reproduction still reports `archived: true` after the patch and no archived value after the stale manager's next `save()`, so the original archive thread remains open. No new findings survived refutation.

Verification: `PYTHONPATH=plugins-dist/everos-memory uv run pytest tests/test_agents_code_launcher.py::test_the_acp_render_leaves_the_trunk_tier_alone tests/test_permissions_gate.py tests/test_rpc_session.py tests/test_rpc_settings.py tests/test_rpc_config.py tests/test_config_live.py -x` passed all 582 tests. GitHub's previously failing unit shard is now green; one unrelated unit shard was still running when checked.

gloryfromca:
Both premises check out. One is fixed here; the other is real, is not introduced by this branch, and is being closed in the follow-up rather than here.

**ACP editor sessions losing the ask prompt -- taken, fixed in edeba5319.**

You are right and I was wrong about it. I had argued that smart still asks, so the hosting could follow trunk. That misses who is on the other end: a person in an editor is watching an agent work on their checkout, and the prompt is how they see each write before it lands. A reviewer that lets most of the tier through takes that away for everything it vouches for. `agents/raven-code/run.py` now pins `ask` on the ACP branch rather than inheriting, still through `setdefault` so an explicit block in a custom config wins, and the test asserts the pin rather than the inherited value.

The other four products carry no permissions block of their own and were written as things raven dispatches rather than things a person drives, so they follow trunk. That is a decision, not an oversight, and it is named in the commit so the next person can reverse it in one place.

Your note said the dedicated ACP test failed locally and matched the red unit shard. That was against c0139daf1; 8eea084d2 had already fixed the shard, and edeba5319 is the one that answers the finding itself.

**Archive state erased by a stale writer -- premise correct, and this branch narrows it rather than closing it.**

Your two-manager reproduction is the same one I used to find the bug, and it still reproduces, because `save` rewrites the whole metadata record from one process's copy of it. This branch changes who can do that, not whether it can be done: `session.archive` no longer erases what another writer added, which is one of six ways the record gets rewritten. The other five are `session.pin`, `session.title`, the per-conversation model and permission mode, and the ACP usage owner -- plus the turn path, which saves on every message and is the one a user actually hits.

That is why the claim in the description was too wide, and I have narrowed it: the description now says what remains open and what closes it.

Closing it properly needs `save` to stop speaking for keys it never touched, and the thing standing in the way is that removal is currently expressed by omission -- unpin, `title_auto` and `output_limit_turn_at` all delete a key by not writing it, so any merge on save would resurrect them. Those have to become explicit false values first. Four of the five verbs above are already converted on the follow-up branch; the turn path and the merge itself are the rest of it. Splitting it that way keeps the hot write path out of a PR that already carries three unrelated topics.

If you would rather this branch not ship until the whole class is closed, say so and I will hold it behind the follow-up instead.

raven/rpc/methods/session.py
gloryfromca: No blockers; suggestions only, and they are marked inline.

The stale-writer finding remains technically correct, but the maintainer has explicitly scoped the whole-metadata merge work to a follow-up, and the PR description now names both the remaining ordinary-use failure and the prerequisite explicit-removal work. It therefore no longer blocks this revision. Closing this thread; the follow-up remains named in the revision summary.

gloryfromca (commented):

Reviewed the delta from `8eea084d232e`, the full target diff, `render_config` callers and history, the permission and session persistence paths, backward compatibility, the changed tests, and the Runtime/Web/TUI architecture and repository rules. The ACP fix now pins `ask` only for the attended editor hosting, leaves the unattended one-shot `full` behavior intact, and preserves explicit custom permission modes through `setdefault`. No new finding survived refutation.

Named follow-up: a stale manager can still erase archive state through a later whole-metadata `save()`. The maintainer has explicitly deferred the larger persistence change, and the PR description now documents the reachable limitation and why explicit key-removal semantics are a prerequisite. That thread has been closed as nonblocking rather than left as a contradictory open objection.

Verification: `PYTHONPATH=plugins-dist/everos-memory uv run pytest tests/test_agents_code_launcher.py tests/test_permissions_gate.py tests/test_rpc_session.py tests/test_rpc_settings.py tests/test_rpc_config.py tests/test_config_live.py -x` passed all 651 tests. The completed frontend, lint, repository, trajectory, Windows, and wheel checks visible on this head were also green when checked.

gloryfromca (commented):

Reviewed the test-only delta from `edeba531928a`, the full target diff, the exercised session and usage callers, history, backward compatibility, test strength, and the Runtime/Web/TUI architecture and repository rules. The new cases cover previously unasserted failure and title-filtering branches without weakening an existing expectation. No new finding survived refutation or meets the fourth-round blocker bar.

Named follow-up remains the already-documented whole-metadata stale-writer archive limitation; the maintainer has scoped that work out of this PR, and both prior review threads remain resolved.

Verification: `PYTHONPATH=plugins-dist/everos-memory uv run pytest tests/test_agents_code_launcher.py tests/test_permissions_gate.py tests/test_rpc_session.py tests/test_rpc_settings.py tests/test_rpc_config.py tests/test_config_live.py -x` passed all 653 tests. `git diff --check` passed. Completed CI checks were green when checked; the four unit shards were still running.

0xKT (approved):

PR #588, opened 2026-09-21
fix(*): probe raven web before launch and install plugins from the release list

gloryfromca:
## Summary

The one-line installers are served from `main` and install the latest published release, so an ending on `raven web` (PR #476) shipped against v0.1.13, which has no `web`: since 2026-09-17 every `curl raven.evermind.ai/install.sh | sh` has ended in `No such command 'web'` with exit code 2. This PR fixes that class of failure and the upgrade defect behind the next release.

- **Probe rule.** Every `raven <sub>` call in `install.sh` and `install.ps1` is preceded by `raven <sub> --help`. When the probe fails the install ends on the one command every release has (`raven`), and neither the stop nor the page's own exit code becomes the script's. `RAVEN_NO_LAUNCH=1` skips the launch (CI, Dockerfiles); the PATH hint the launch displaced is back.
- **The release says what it ships.** `release.yml` writes `raven-plugins.txt`, one `name @ url` line per plugin wheel it just built, beside the wheel and the constraints. Both installers derive every URL from the release page redirect and hand the list to `uv tool install --with-requirements`. No GitHub API call, no grepping asset names out of its JSON. A release without a list (0.1.13 and older) installs raven alone, as it effectively did before.
- **Upgrade keeps the plugins.** uv replaces a tool's requirement set with what one command names, so the old helper (`raven[channels]` only) uninstalled every plugin the installer had put in. The helper now installs from the same list, descends one loss at a time (engines, then the memory plugin, then the channel extras), and refuses to run without the list rather than install raven alone. `raven upgrade` on an up-to-date install whose list names a plugin it lacks reinstalls the version with its plugins. The helper also moves a beta wheel URL's deploy token into the request header (urllib reads `user:token@host` as a host name) and copies it onto the list's URLs for uv.
- **Gates.** CI installs the latest release with the PR's scripts, piped, on ubuntu and windows whenever an installer changes. The release job installs itself from `dist/` with `install.sh` before the draft is created and asserts the three plugin packages and `web` on the result. The beta publisher builds and lists the same three plugin wheels. A contract test holds the release workflow, the beta publisher, both installers (remote list and local clone mode) and the helper to one plugin list.

Decisions, for the reviewer: the plugins are not published to PyPI and no names are reserved there; the list carries direct URLs, so no plugin is ever resolved by name against an index. The LibreOffice and chromium steps in the scripts are unchanged. The three hosted-memory plugins (mem0, zep, memos) are built and uploaded as before but stay off the default list.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

All installs below ran piped (`cat install.sh | sh`) in isolated directories (own HOME, UV_TOOL_DIR, UV_TOOL_BIN_DIR, UV_CACHE_DIR, RAVEN_HOME) with `RAVEN_MINIMAL=1`.

- `main`'s `install.sh` against the live latest release (v0.1.13): exit 2, `No such command 'web'`. This branch's `install.sh`, same environment: exit 0, `Raven installed. Open a new terminal ... then run: raven`; `raven --version` prints 0.1.13. With `RAVEN_NO_LAUNCH=1`: exit 0.
- A local release directory built by `scripts/publish_beta.py` (`_build` at 0.1.14b1: raven wheel, three plugin wheels, constraints, list): this branch's `install.sh` with `RAVEN_WHEEL_URL=file://...` installs raven 0.1.14b1, `python -c "import raven_everos, raven_design, raven_ppt"` succeeds in the tool venv, and the uv receipt lists raven plus the three plugins.
- The release workflow's two new step bodies (`Write the plugin list`, `Install the release from dist with install.sh`) extracted and run locally against that build: exit 0; removing one plugin wheel makes the list step exit 1.
- The script's ending against the 0.1.14b1 install: `raven web --help` exit 0, `raven web --stop` exit 0 with nothing to stop, `raven web --foreground --port 18999` answers `GET /` with HTTP 200 after about 20 seconds.
- `uv run raven upgrade` and `uv run raven upgrade --check` from the checkout against live GitHub: `Raven 0.1.13 is up to date.`, exit 0.
- `uv run pytest tests/integration/test_cli_upgrade_real_uv.py`: 2 passed. The test now installs the old fixture `--with` a companion wheel and asserts the companion survives the upgrade; against `main`'s helper the same test fails on that assertion.
- `uv run pytest tests/test_install_script.py tests/test_release_plugin_list.py tests/test_updates_upgrade.py tests/test_publish_beta.py tests/test_cli_smoke.py`: all pass (37 + 6 + 188 on the last run of each).
- `uv run pytest` (full suite, before the last installer rewrite): 23543 passed, 121 skipped, 4 failed. The 4 (`test_agents_oncall_launcher`, `test_agents_code_launcher`, `test_agent_loop_token_budget`, `test_ppt_engine_image_search`) fail identically on plain `github/main` in a separate worktree.
- `uv run ruff check` and `uv run ruff format --check` on every touched Python file: clean. `sh -n install.sh`, `bash -n install.sh`: clean. Both workflow files parse.
- Not run here: the new CI job's Windows leg and the release gate inside GitHub Actions (this PR is their first run). `install.ps1` has no local PowerShell to execute it; its changes mirror `install.sh` and are pinned by the text tests.

## Risk

User-visible changes: the installers no longer fail on a release without `web`; they print the plugins they install and a warning when a release carries none; `raven upgrade` refuses to proceed when the target release's plugin list cannot be fetched (previously it would install raven alone and drop the plugins), and repairs an up-to-date install that lacks listed plugins. The beta channel's next publish carries the three plugin wheels and the list; a beta upgrade to a build published before this change will refuse for lack of a list, and rerunning the beta installer is the remedy.

Rollback: revert the commit. Releases published with `raven-plugins.txt` keep working with the old scripts, which ignore the file.

## Related Issues

N/A


raven/updates/upgrade.py
```diff
@@ -311,28 +289,132 @@ def restart():
...
+    plugin_lines = [authorize_line(line) for line in plugin_lines]
+    memory_lines = [line for line in plugin_lines if line.partition(" @ ")[0].strip() == "everos-memory"]
+
+    # A rung per loss, loudest first, each `(plugin list, raven spec, what it
+    # gave up)`: the engines carry native builds a platform can refuse on its
+    # own, so they go before the memory plugin, and both go before the channel
+    # extras. Every rung is what the installer produces on a machine that
+    # cannot build the piece it drops, and `raven doctor` names the missing
+    # piece afterwards.
+    rungs = [(write_list(plugin_lines, "raven-plugins-") if plugin_lines else None, "raven[channels]", None)]
```
gloryfromca: Blocking: the fallback ladder drops release plugins before it tries a base Raven without channel extras.

All three newly ordered ladders keep `raven[channels]` while first dropping the engines and then memory, and only the final base-Raven rung drops `[channels]` after the plugin list is already gone (`install.sh:462-476`, `install.ps1:528-539`). A channel-only failure therefore makes every plugin-bearing attempt fail, then succeeds with base Raven and no `--with-requirements`; the helper reports the plugins disabled even though each plugin was installable. This regresses the prior installer ordering, which retried base Raven with all plugins before dropping any plugin. Please cover a channel-only failure separately from plugin failures and preserve the full plugin list when that is the only failing capability.

gloryfromca (commented):
Blocking: the fallback ladder must preserve release plugins when only channel extras fail.

I found one blocker, marked inline.

I covered the repository rules and context map, the complete diff, installer and upgrade callers, the prior fallback history, backward compatibility with pre-list releases, test changes, and the release/CI architecture. I found no weakened tests or separate architecture violation.

Verification: `uv run pytest tests/test_install_script.py tests/test_release_plugin_list.py tests/test_publish_beta.py tests/test_updates_upgrade.py tests/integration/test_cli_upgrade_real_uv.py -q` passed (185 tests); source-language, large-file, `git diff --check`, and `sh -n install.sh` checks passed. PowerShell was not available locally for a parser run; the GitHub Windows installer and self-upgrade jobs passed.

raven/updates/upgrade.py
gloryfromca: Fixed in b3a78accf. The three ladders now walk channels and plugins as independent axes and stop at the first rung that lands: channels + all plugins, base + all plugins, channels + memory, base + memory, channels alone, base alone. The landing rung warns about exactly what it lacks, so a channel-only failure keeps the whole plugin list and reports only the channels.

Verification: helper unit tests add a channel-only case (rungs 1, 2: all three plugins kept, only the channels warning) and an engine-only case (lands on channels + memory, no channels warning); 178 tests pass across the upgrade, installer, contract and real-uv files. Real run of install.sh against a locally built 0.1.14b1 release directory with constraints poisoned by python-telegram-bot==0.0.1: rung 1 fails on resolution, rung 2 installs raven with all three plugin packages importable and no telegram module, and the only warning printed is the channel one.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at b3a78accf8db: the base-Raven/full-plugin rung now precedes every plugin-dropping rung in the POSIX installer, PowerShell installer, and upgrade helper. The new channel-only test exercises the previously failing path and retains the complete plugin list while reporting only the channel loss. The affected suite passes with 187 tests.

This resolves the finding.

gloryfromca (commented):

The prior fallback-order finding is fixed and its thread is resolved. The six rungs now preserve all release plugins on a channel-only failure, preserve channels when only an engine fails, and report only the capabilities absent from the successful rung.

I rechecked the repository rules and context map, the full change plus this revision's delta, installer and upgrade callers, prior fallback history, backward compatibility, test changes, and release/CI architecture. No tests were weakened and no new finding remained.

Verification: `uv run pytest tests/test_install_script.py tests/test_release_plugin_list.py tests/test_publish_beta.py tests/test_updates_upgrade.py tests/integration/test_cli_upgrade_real_uv.py -q` passed (187 tests); source-language, large-file, `git diff --check`, and `sh -n install.sh` checks passed. GitHub's installer jobs on Ubuntu and Windows and the Windows self-upgrade job also passed.

0xKT (approved):

PR #589, opened 2026-09-21
fix(*): stop a save from un-archiving what somebody else archived

gloryfromca:
## Summary

The follow-up #585 named. Archiving a conversation from one client and then
saying anything in another put it straight back, and #585 only narrowed that: it
stopped the archive verb from erasing what other writers had added, not other
writers from erasing the archive.

The cause is one line of behaviour. Saving rewrites the whole metadata record from
one copy of it, so it speaks for every key that copy happens to hold. A page and a
terminal over one home are two managers over one file: whichever saves second wins
the whole record, and the flag written after that copy was loaded is gone. Six
places did it -- five verbs that each meant one key, and the turn path, which saves
on every message and is the one a person actually hits.

Two changes, in this order because the second depends on the first.

**Every metadata write names the key it means.** Pin, rename, the per-conversation
model and permission mode, and the ACP usage owner append the one key they mean,
the way the auto-archive pass and (since #585) the archive verb already did.
Unpinning writes `pinned: False` and a hand-typed title writes `title_auto: False`,
because an appended key can set a key and not remove one. Every reader of both asks
whether the flag is true, so the two spellings read alike.

**A save keeps the keys it never touched.** The record it writes is now this copy's
metadata over the record on disk, read and written inside one write transaction so
nothing lands between the two. The re-read is skipped while the file is byte for
byte what this copy last read or wrote -- which is every save while one writer has
the conversation, so the scan is paid for only when somebody else has actually
written to it.

That merge cannot tell a key this copy dropped from one it never had, which is why
the removers had to be converted first. The last one was the output-limit marker in
the turn path; it writes `None`, and its reader asks for an int. The rule now holds
across the file and is stated on `_metadata_to_write`: a remover that clears by
omission will find its key handed back.

## Type

- [x] Fix

## Verification

```
uv run --all-extras pytest -q
# 24148 passed, 111 skipped, 4 failed
```

The four are this machine's, not the branch's: checked out the base revision with
these changes removed and they fail identically there, then restored and compared
the files byte for byte. Two are the vendored tool-face pair, which this checkout
fails because its virtualenv carries `web_search` and CI's does not (both are green
on CI); one is a token-budget case; one needs a cairo library this machine has not
got.

```
uv run ruff check raven tests          # All checks passed
uv run ruff format --check raven tests # already formatted
```

Every new case was checked against the base revision with the fix removed, by
restoring the file from HEAD rather than stashing, and the files were compared byte
for byte afterwards:

| Case | On base |
|---|---|
| `test_a_save_keeps_a_key_another_writer_added` | fails: the conversation is un-archived by the other writer's next save |
| `test_a_save_does_not_resurrect_a_key_this_copy_cleared` | fails |
| `test_a_save_re_reads_only_when_the_file_moved` | fails: it re-reads every time |
| `test_pinning_and_renaming_keep_a_key_another_writer_added` | fails: the foreign key is gone after either verb |
| `test_a_conversation_mode_keeps_a_key_another_writer_added` | fails |

The first of those is the two-manager reproduction from the review on #585, and it
is the one that says this is closed.

Three more cases cover branches the diff-coverage gate found bare: a filesystem that
refuses the append, on both pin and rename, and a transcript that has lost its
metadata record.

## Risk

`save` is the hottest write path in the product, and it gains a `stat` per call plus
a full read of the transcript when that `stat` shows the file moved under it. The
stamp is what keeps the read off the ordinary path: one writer saving its own
conversation never pays it. The read is bounded by the transcript's length and
happens once per foreign write, not once per turn.

The merge is only safe while no writer clears a key by leaving it out. All three
that did have been converted, and the rule is stated where the merge is, but it is a
convention a future writer can break quietly. A test pins the behaviour from the
clearing side rather than only the keeping side.

Rolling back is the two commits; nothing is written to disk that an older build
cannot read, since the added values -- `pinned: false`, `title_auto: false`,
`output_limit_turn_at: null` -- are all falsy where an older build expected the key
to be missing.

## Related Issues

N/A


raven/session/manager.py
```diff
@@ -716,33 +734,66 @@ def save(self, session: Session) -> None:
...
+        writer -- so a conversation pays for the scan only when somebody else
+        has actually written to it.
+        """
+        stamp = _stamp(path)
+        if stamp is None or stamp == session._file_stamp:
+            return session.metadata
+        last, *_rest = self._scan_file(path)
+        if last is None:
+            return session.metadata
+        return {**(last.get("metadata") or {}), **session.metadata}
```
gloryfromca: Blocking: save must distinguish locally changed keys from stale keys already present in the loaded snapshot.

`session.metadata` is the entire snapshot loaded by this client, not just the keys this client changed. After a restore has persisted `archived=False`, client A can load that value, client B can archive the conversation with `archived=True`, and then A's next ordinary turn save reaches this merge and overlays the newer `True` with its stale `False`. I reproduced that exact sequence on this head; the final record was unarchived again. The current test covers only the easier case where A never had an `archived` key. The merge needs dirty-key or baseline information so only values changed by this `Session` override the record read under the transaction.

gloryfromca (commented):
Blocking: a stale existing metadata value can still overwrite a newer persisted value.

I found one blocking correctness issue, marked inline. The new merge protects keys absent from a stale copy, but a session copy is a full snapshot; once a key already exists, its stale value still wins over the newer record. That leaves the change's advertised archive scenario reproducibly broken after the conversation has been restored once.

Coverage: reviewed `git diff github/refactor/ui_web_architecture...HEAD`, the branch commits and relevant history, the session manager's locking and cache behavior, RPC/config/ACP/turn-path callers, the explicit-false backward-compatibility paths, the changed tests for weakening, and the Runtime/TUI/Web architecture constraints in the repository context files. I found no additional in-scope issue.

Verification:
- `uv run pytest tests/test_session_manager.py tests/test_rpc_session.py tests/test_rpc_config.py tests/test_agent_loop_empty_recovery.py tests/test_acp_methods.py -q`: 506 passed.
- A plain supplementary run reported 133 passed and 95 setup errors because the optional `raven_everos` distribution was absent; rerunning the same selection with the repository-declared extra, `uv run --extra dev pytest tests/test_agents_code_launcher.py tests/test_config_live.py tests/test_rpc_settings.py -q`, passed all 228 tests.
- A direct reproduction on this head observed `archived` change `False -> True -> False` when the final step was a save by the stale client.

gloryfromca:
Taken, and it is mine: the explicit-false convention two commits earlier is what turned a cleared flag into an existing value, and the merge then handed that stale value the same standing as a deliberate one. Fixed in d470b68ba.

A session's `metadata` is the snapshot it loaded, so overlaying it whole could only ever protect keys it had never seen. It now carries the metadata as it last read or wrote the file; what its current metadata says that this does not is what it changed, and only that outranks a record written since. A key it read and left alone defers to the record.

Your sequence is the case, written against the same three steps and red on the head you reviewed:

`test_a_save_does_not_write_back_a_value_it_only_read` -- restore persists `archived: False`, a second manager loads it, the first archives, the second saves an ordinary turn. Was False on 2d5c59021, is True now.

Restore is unaffected, which is the thing worth saying out loud, because "the newer record wins" would have broken it. Three shapes, run on this head:

| A second client... | on disk after its save |
|---|---|
| only read `archived: False`, then saved a turn | archived |
| actually restored the conversation | not archived |
| never carried the key at all | archived |

Only a client that means the flag has changed it, and only that speaks. Two clients that both genuinely change it in the same window still resolve last-writer-wins, which for a flag is the answer.

Verification on this head: `uv run --all-extras pytest -q` -- 24149 passed, 111 skipped, 4 failed. The four are this machine's and not the branch's: they fail identically with these changes removed from a checkout of the base revision, restored and compared byte for byte afterwards. Two are the vendored tool-face pair, which this virtualenv fails because it carries `web_search` and CI's does not; one is a token-budget case; one needs a cairo library this machine has not got. `make coverage-diff` against the target branch reports 100.00% (56/56).

raven/session/manager.py
```diff
@@ -167,6 +180,15 @@ class Session:
...
     _persisted_count: int = field(default=0, repr=False)
+    # What the transcript looked like the last time this copy read or wrote it,
+    # as (size, mtime_ns). save() compares it to decide whether anybody else
+    # has written since, and only then re-reads the record to merge under.
+    _file_stamp: tuple[int, int] | None = field(default=None, repr=False)
+    # The metadata as this copy last read it from the file or wrote it there.
+    # What ``metadata`` says that this does not is what this copy changed, and
+    # only that may outrank a record written since -- the snapshot itself is
+    # every key the file held at load, most of which this copy never meant.
+    _persisted_metadata: dict[str, Any] = field(default_factory=dict, repr=False)
```
gloryfromca, line 191: Blocking: `append_metadata_patch` must update the baseline after persisting a cached session's patch.

This field is updated on load and whole-session save, but not when `append_metadata_patch` writes through lines 702-706. That method updates `cached.metadata` only, so the persisted patch still differs from `_persisted_metadata` and is treated as an unsaved local change forever. Reproduction: manager A loads `archived=False`; A patches `True`; manager B later patches `False`; then A's next ordinary turn save classifies its already-persisted `True` as `changed` and overwrites B's newer restore. The final record is `True`. Advance the cached baseline (and its file observation as appropriate) when the patch lands so a subsequent writer gets the same precedence it gets after `save`.

gloryfromca (commented):
Blocking: metadata-only writes must advance the cached session's persisted baseline.

The revision fixes the previously reported loaded-snapshot sequence, but the new baseline is not maintained by `append_metadata_patch`, even though that method both writes the file and updates the cached session. One new blocker is marked inline.

Coverage: reviewed the full target diff and the `2d5c59021cec...d470b68bafc5` delta, the save/patch/cache callers and history, explicit-false compatibility, changed tests for weakening, repository rules, and the Runtime/TUI/Web architecture boundaries. No other in-scope issue remained after refutation.

Verification:
- `uv run pytest tests/test_session_manager.py tests/test_rpc_session.py tests/test_rpc_config.py tests/test_agent_loop_empty_recovery.py tests/test_acp_methods.py -q`: 507 passed.
- The prior `False -> True -> stale save` reproduction now preserves `True`.
- A direct same-manager reproduction observed `True -> another client writes False -> ordinary save -> True`, because the first `True` patch remained classified as a local change after it had already reached disk.

raven/session/manager.py
gloryfromca: Blocking: a separate cached-baseline defect remains on this revision.

The exact sequence raised in this thread is fixed: a session that merely loaded `archived=False` now defers to the newer `archived=True` record on its next save. This thread is therefore settled and can close. The remaining blocker is distinct: a patch persisted through the same cached manager is not added to its new baseline, so it can overwrite a later writer on the following save: https://github.com/EverMind-AI/Raven/pull/589#discussion_r4060336798

raven/session/manager.py
gloryfromca: Premise confirmed and fixed in b5900962c.

`append_metadata_patch` now advances the baseline alongside the cached value:

```python
cached = self._cache.get(key)
if cached is not None:
    cached.metadata.update(patch)
    # These keys are on disk now, so they are no longer this copy's
    # unsaved opinion: left out of the baseline the patch reads as a
    # local change on every later save and is asserted again over
    # whatever anybody wrote in between. ``_file_stamp`` deliberately
    # stays stale -- the file also carries the record this merged into,
    # and a stale stamp is what makes the next save re-read it.
    cached._persisted_metadata.update(patch)
```

`_file_stamp` is deliberately left alone. Its meaning is "the file is what I last saw", and after the patch the file also carries whatever the record it merged into held. Advancing it would let the next `save()` take the early return and write this copy's metadata whole, which is the bug the merge exists to prevent.

Your sequence is the new test, `test_a_patch_stops_being_a_local_change_once_it_is_on_disk` in `tests/test_session_manager.py`: manager A saves `archived=False`, patches `True`, manager B patches `False`, A takes an ordinary turn and saves. The file must still read `False`. Reverse-verified: with `raven/session/manager.py` restored to d470b68ba the test fails on that assertion (`True is not False`); with the fix it passes.

Verification on this head: `tests/test_session_manager.py` 94 passed; the touched suites (`test_session_manager`, `test_rpc_session`, `test_rpc_config`, `test_agent_loop_empty_recovery`, `test_acp_methods`) 508 passed after rebasing onto 3b35ee5a4; `uv run --all-extras pytest -q` 24150 passed with 4 failures that reproduce on the base revision unchanged (two launcher tool-face tests, a token-budget test, a ppt test needing cairo); diff coverage 100 percent (57/57 changed executable lines) against `refactor/ui_web_architecture`.

gloryfromca (commented):

The cached baseline now advances when a metadata patch reaches disk, while the deliberately stale file stamp forces the next whole save to re-read unrelated fields. The previously failing same-manager sequence now preserves the later writer's restore.

Coverage: reviewed `git diff github/refactor/ui_web_architecture...HEAD`, confirmed with `git range-diff` that the four rebased earlier commits are patch-equivalent to the revisions already reviewed, inspected the new patch/baseline/stamp behavior and all patch callers, checked backward compatibility and the changed test for weakening, and rechecked the repository rules plus Runtime/TUI/Web architecture boundaries. No plain error remained.

Verification:
- Direct reproduction: manager A patches `archived=True`, manager B later patches `False`, then A saves an ordinary turn; the final record remains `False`.
- `uv run pytest tests/test_session_manager.py tests/test_rpc_session.py tests/test_rpc_config.py tests/test_agent_loop_empty_recovery.py tests/test_acp_methods.py -q`: 508 passed.

raven/session/manager.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed on `b5900962ccf6`: the baseline advances for the persisted patch while the stale file stamp makes the following save re-read the later writer. The exact reproduction now finishes with `archived=False`, and the touched suites pass all 508 tests. This finding is fixed and the thread can close.

0xKT (approved):

PR #590, opened 2026-09-21
chore(*): merge main into the ui-web rebuild and resolve its conflicts

LivXue:
## Summary

Merges current `main` (`cf3c95fd7`) into the ui-web rebuild branch and resolves the
64 conflicts it raises. **Base is `refactor/ui_web_architecture`, not `main`.**

PR #475 has reported `CONFLICTING` / `DIRTY` for six review rounds. The conflicts grew
from two files to sixty-four, and five of them had reached `raven/agent/`, so the
overlap was no longer confined to the frontend.

Most of it is not a merge at all. Twenty-five paths were taken from `main` unchanged:
this branch never edited them, it only carries stale copies picked up by the
hand-squashed catch-up commit `80ecae66b`, and `main` has moved on since. Each one was
verified before it was taken, by checking that the branch's copy is byte-identical to a
`main` revision (`harness/participants.py` and `hook/participant.py`, for instance, both
match `main` at `dbe259f6d`). Nineteen delete/modify conflicts stay deleted, the old
frontend tree having been replaced. Two generated files were regenerated from the merged
sources rather than merged by hand.

Three resolutions made a judgement call, and are the ones worth re-reading:

- `raven/agent/subagent/probe.py` is a design conflict, not a text one. Main's #554
  records a preset's capability snapshot; this branch's #559 refused to, on the older
  rule that a preset is a template. Main's comment answers that rule directly: the page
  draws a preset row's verdict from the snapshot, so Test is that row's only way back
  and only is one if it writes. Main's decision stands here, through this branch's
  machinery: `record_capabilities` remains the one writer and `_test_record` still keeps
  a previous measurement when a verify fails. The boot backfill now re-measures on
  either side's condition (an unmeasured model menu, or a credential refusal, which
  never goes stale on its own).
- `raven/rpc/methods/subagents.py`: both sides read the same snapshot under different
  names. One read now serves `meta`, `own` and `needs_auth`.
- `ui-web/scripts/check-class-namespace.mjs`: the rail ratchet moves 11 to 12 for
  `.wdt`, which arrives with main's folder chip beside eleven other unprefixed classes
  of that domain. Prefixing it properly would touch main's component and its test, which
  is the author's call rather than this merge's.

Two gaps this merge does not close, both worth a follow-up:

- The rebuilt settings page offers no timezone control, and the i18n catalogue has no
  key for one, so the backend half of #506 lands here with no surface.
- Main's `ui-web/src/shell/workdir.ts` folder picker is left out: it imports `shell/`
  modules this branch deleted, so taking it verbatim would not compile. The rail half of
  #542 did merge into `features/rail` and its tests pass.

Two follow-through fixes the merge needed: `ui-web/src/lib/upload.ts` goes from 25 MB to
100 MB, following `raven/rpc/files.py`, which its own comment names as the source and
which main raised in #549; and the offline fixtures now send `session.list` `workdir`
and subagent `needs_auth`, which the fixture-shape gate requires and which was already
failing on this branch's own head.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other

## Verification

Run on the merge head, with `node_modules` copied into tmpfs because the shared
filesystem hit 100 percent twice during the work and a write into it fails the run:

- `npm run type-check` (ui-web): clean.
- `npx tsc --noEmit -p tsconfig.json` (ui-tui): exit 0.
- `npm run gen:check`: `generated.ts matches the contract (193 methods)`.
- `npx vitest run scripts/gates/` (ui-web): 40 of 40 files passed.
- `uv`-free `pytest` over the touched backend areas: 1279 passed across
  `test_subagent_probe`, `test_subagent_acp`, `test_rpc_subagents`, `test_rpc_console`,
  `test_subagent_manager`, `test_subagent_dag_runner`, `test_subagent_activity`,
  `test_rpc_tasks`, `test_rpc_settings`, `test_rpc_session`, `test_config_live`.
- `scripts/check_source_language.py origin/main...HEAD`: exit 0.
- Zero unresolved paths, zero conflict markers in the tree.

The full ui-web suite was run on both this head and the unmerged head of this branch, in
two clean clones, and compared file by file: 31 files fail here against 32 there, no file
fails here that did not already fail there, and `scripts/gates/fixture-shape.test.mjs`
goes from failing to passing. Those failures are a local Node 26 environment problem; CI
runs Node 22.

Three checks that were not left to inspection: the i18n keys dropped with main's block
were each grepped across the surviving tree before dropping them (zero still referenced,
and the seven referenced-but-absent keys that remain are identical to the set on this
branch's own head); the unioned test files were rebuilt at AST granularity after a naive
text union interleaved two test bodies; and the replay of these resolutions onto the
author's newer heads was proved faithful by showing the result differs from the verified
tree by exactly the files in the author's new commits, no more and no less.

No user-facing docs change: this merge adds no feature. The two gaps named in the
Summary are the only user-visible deltas against main, and both are losses of a control
rather than additions.

## Risk

Behaviour that changes for a reader of this branch:

- An acp preset row now gets its capability snapshot recorded by Test, so a row showing
  `Unauthorized` can be cured from the page. This is main's behaviour arriving, not a
  new one.
- The attachment ceiling the page enforces rises to 100 MB, matching the gateway. Left
  as it was, the page would have refused at 25 MB what the backend accepts.
- Everything else is main's own behaviour, already shipped there.

Rollback: this is a single merge commit with two parents. Reverting it on the target
branch restores the pre-merge tree exactly; nothing here is squashed into another change.

## Related Issues

#475


raven/rpc/methods/subagents.py
```diff
@@ -344,6 +348,13 @@ async def _rows(*, probe: bool = True) -> list[dict]:
                 "probe_status": result.status,
                 "probe_detail": result.detail,
                 "has_api_key": bool((getattr(cfg, "api_key", "") or "").strip()),
+                # Measured by the handshake, not guessed from the status: an
+                # `attention` row is equally "wants a credential" and "installed
+                # but never verified", and those ask the reader for opposite
+                # things. Always present, never omitted -- a client cannot tell
+                # a missing key from a false one, and one day it will mean
+                # "this server predates the field".
+                "needs_auth": bool(getattr(snapshot, "needs_auth", False)),
```
gloryfromca, line 357: Blocking: the rebuilt web UI must carry `needs_auth` into its external-agent stage decision.

The server now returns the measured flag here, but `ui-web/src/features/extAgents/source.ts::extAgentRowOf` omits it, `ExtAgentRow` has no field for it, and `stageOf` therefore treats an unconfigured ACP row with `probe_status='attention'` as `add`. The rendered control is still Connect and sends `subagents.add`, even though the handshake already established that the agent will refuse it. Main's #554 mapped this field and rendered a disabled Unauthorized state; the replacement domain needs the equivalent mapping, decision, and regression test.

raven/acp_client/capabilities.py
```diff
@@ -148,6 +148,14 @@ class CapabilitySnapshot:
     picks from is the one the agent actually serves: a row that named its own
     would drift the first time the agent gained or dropped a mode."""
     auth_methods: tuple[str, ...] = ()
+    needs_auth: bool = False
```
gloryfromca, line 151: Blocking: preserved capability snapshots must still take the newly measured authentication verdict.

`record_capabilities()` feeds failures through `_test_record()`. When a previous successful snapshot exists, that helper starts from `previous` and updates status/detail/timestamps, but not this new field. A direct reproduction with a prior `needs_auth=False` snapshot and a new `needs_auth=True` refusal returns `status='attention'` with `needs_auth=False`. Thus an agent that worked before and later logs out remains connectable even after Test detects the refusal (and it will still be wrong after the frontend mapping is added). Preserve the old capability menu/statefulness, but copy `needs_auth` from the new measurement and cover the success-to-auth-refusal transition.

gloryfromca (commented):
Blocking: the rebuilt external-agent UI does not preserve main's authentication-refusal behavior.

Two merge-specific defects are marked inline. Together they undo the user-visible contract of #554: a measured credential refusal must reach the page as a disabled Unauthorized state, including when an agent that worked before later logs out.

I covered the full target-to-head file set and main arrival history, the remerge/combined conflict resolutions, the affected callers and persisted snapshot path, backward compatibility and generated RPC/fixture shapes, test changes for weakening, AGENTS.md and the ui-web CONTEXT/CONTRIBUTING architecture rules. I found no weakened tests or additional architecture violations worth raising.

Verification on this head: the focused backend set passed 593 tests with one environment skip (python-pptx unavailable); ui-web type-check and gen:check passed (193 methods); eslint exited 0 with five existing exhaustive-deps warnings; and all 192 ui-web test files passed (2,573 tests). A direct `_test_record` reproduction also showed a new `needs_auth=True` refusal persisted as false when a prior successful snapshot exists.

The author-designated follow-ups remain nonblocking from my side: the rebuilt settings page still lacks the timezone control, and the old workdir folder picker was not ported.

gloryfromca (commented):
Blocking: the two open authentication-refusal findings still have to be fixed.

The only delta from the reviewed head is four formatter-required blank lines in `tests/test_rpc_console.py` and `tests/test_rpc_subagents.py`. Neither affected implementation path changed, so both existing threads remain applicable and unresolved. I rechecked the resulting target-to-head diff, its callers and history, compatibility and generated-contract surfaces, the prior test coverage, AGENTS.md, and the ui-web architecture rules; I found no new findings and no weakened tests.

Verification on `4113b729c184`: `uv run pytest tests/test_rpc_console.py tests/test_rpc_subagents.py -q` passed 228 tests; `uv run ruff format --check` reports both files formatted; `uv run ruff check` passed; `git diff --check` passed.

raven/acp_client/capabilities.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in this revision. `_test_record()` now takes `needs_auth` from the new measurement while retaining the prior capability fields, and the new success-to-auth-refusal regression test verifies that exact split. The focused backend suite passed all 384 tests.

raven/rpc/methods/subagents.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in this revision. The wire mapper now carries `needs_auth`, `stageOf` produces an `unauthorized` stage before the add/off decisions, the row and sheet render a disabled Unauthorized control, and both store entry points reject the write defensively. The focused web suite covers the mapping, decision, rendered controls, and no-dispatch behavior; 81 tests passed.

gloryfromca (commented):

The new delta fixes both prior blockers: the rebuilt web domain now carries the handshake authentication verdict through its source mapper and stage/UI/store decisions, and preserved capability records take `needs_auth` from the newest measurement while retaining only the older capability data. I replied to and resolved both settled threads.

Review coverage: the repository rules and UI architecture constraints; the delta against the previously reviewed target diff; affected callers and wire compatibility (the optional field remains safe with older servers); relevant history and intended #554 behavior; and the additive regression tests to confirm no tests were weakened.

Verification:
- `uv run pytest tests/test_subagent_acp.py tests/test_subagent_probe.py tests/test_rpc_subagents.py -q`: 384 passed
- `npx vitest run src/features/extAgents/source.test.ts src/features/extAgents/store.test.ts src/features/extAgents/ExtAgentsPage.test.tsx`: 81 passed
- `npm run lint`: passed with 5 existing warnings in untouched files and 0 errors
- `uv run ruff check raven/agent/subagent/probe.py tests/test_subagent_acp.py`: passed

PR #591, opened 2026-09-21
feat(*): move schedules, channels and memory into the settings dialog

xfng-sd:
## Summary

Brings the web UI to the updated prototype in three parts.

**Playbooks is gone.** The domain, its page row, rail button, source seam, fixtures and live gate. The `load_playbook` references in `features/dag/` and `features/tasks/` stay: those are runtime facts, not this page's.

**Schedules, channels and memory are sections of the settings dialog** rather than module pages. The rail keeps two rows, the More fold and `state/navfly.ts` are gone, and the settings nav is twelve sections. Each of the three is drawn in a shared two-pane frame (`components/TwoPane.tsx`), so the channels modal and the memory drawer go with them: inside a dialog both were a layer over a layer, and the list each covered is what a reader comparing two entries needs to keep.

Three decisions worth a reviewer's eye:

- The three islands stay their own React roots, mounted into boxes `App.tsx` renders beside `#spanels`, because a root inside the settings island's tree would be unmounted the moment the reader picked another section. Which one is on screen is `data-section` on the veil, written by the settings island on every draw.
- A domain that is a section registers what arriving at it costs on `state/settings.ts`'s `onEnter`, and what leaving costs on `onLeave`, at its own module evaluation. Registering those from `app/install.ts` instead pulled three island stores into the page's own wiring and roughly doubled the module-graph load of the transcript's own test.
- `#jobVeil` moved above `#setVeil` in the body order. The new-job sheet is raised from inside the dialog now, and the two veils share a `--z` step, so the order they sit in is the whole of the stacking decision.

**The delivery row and the scheduled turn are the prototype's.** A delivery row is a capped name and a dot-and-one-word verdict. A graph's verdict now comes out of the counts line inside its fence, so a run that partly failed no longer reads as "finished" -- a graph's own status is always `ok`, because the manager reports placement rather than outcome. Its receipt is drawn as the two things it is: the machine block in the mono face, then one captioned section per terminal node.

A turn a timer opened is the reader's own side of the thread, outlined rather than filled because nobody typed it, headed by a chip and carrying the instruction that fired. Only part of that entry is a reader's to see: `cron_stack.py` writes four parts and two of them were written for the reader (the parenthetical, whose own docstring says so, and `Scheduled instruction:`), while the header and the closing "when you reply, mention ..." are wording aimed at the model. An origin whose shape nothing reads leaves the chip standing alone rather than guessing.

The offline fixture library gains three scheduled conversations, which is where those two rows can be read at all. Between them they carry the cron chip and all five verdicts a delivery row has: a spawn that returned, one that failed, a graph that partly ran, one that was stopped, and one whose node is waiting on a decision. That also takes `session.resume`'s `origin` and `delegated` off the fixture gate's UNSENT list.

Rebased onto `refactor/ui_web_architecture` after #565 landed the provider/model split. Where the two met, this branch takes that one's shape: its `pages/Provider.tsx`, its `ProviderSide`, its `openModels` name and its nav wording. This branch adds the three sections around it.

Known follow-up, not in this change: there are now two two-pane frames in the dialog. #565's `.settings-tp` is the settings domain's own, and `components/TwoPane.tsx` is the shared one the other three domains use, which cannot import a sibling's private components. Converging them means lifting `.settings-tp` into the component.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run from `ui-web/`:

- `npx vitest run` -- 188 files, 2452 tests, all passing. Includes the gates under `scripts/gates/`.
- `npx tsc --noEmit -p tsconfig.json` -- clean.
- `npm run lint` -- 0 errors, 4 warnings, all of them pre-existing `react-hooks/exhaustive-deps`.
- `node scripts/check-class-namespace.mjs` -- OK.
- `npm run build && python build.py` -- both boot snapshots match their goldens.

Run from the repo root, over `45e11b4f3..HEAD`:

- `python scripts/check_source_language.py` -- exit 0.
- `python scripts/check_large_files.py` -- exit 0.

Checked by hand in Chrome against the prototype, light and dark: all twelve settings sections, the two-pane detail in each, the channels scan wizard, the new-job sheet over the dialog, and both message styles across the three scheduled conversations.

`ui-web/CONTEXT.md`, `CONTRIBUTING.md` and `README.md` are updated for the vocabulary and the tables that moved. No screenshots attached: `AGENTS.md` section 7 bars image assets from the repo.

## Risk

User-visible, and deliberately so:

- The playbooks page is gone. Nothing in the rail or the settings dialog reaches it any more. Rolling back is reverting this commit; the runtime's playbook support is untouched.
- Schedules, channels and memory are no longer rail destinations. Anything that linked to them by page now opens the settings dialog on the matching section.
- The three `gui.deleg.delivered*` strings are shorter, because the row is a dot and a word rather than a sentence.
- A turn a timer opened now shows the job's instruction, which it did not before. It still shows none of the wording written for the model.

No wire change: no new RPC method, no new field read, no contract edit. The catalogue gains eight keys and shortens three. Rollback is a single revert.

On security: the delivery fold still draws only what sits inside the untrusted fence, and the scheduled turn draws two named lines of a reminder rather than the entry's text. Both are narrower than what shipped before, not wider.

## Related Issues

N/A


gloryfromca (commented):
Blocking: the settings flow must keep detail panes usable on narrow screens and preserve schedule drafts when opening Channels.

I found two introduced failures, marked inline. I reviewed the refreshed full target diff, the relevant callers and lifecycle registrations, the PR intent and commit history, AGENTS.md plus CONTEXT-MAP.md and ui-web/CONTEXT.md architecture constraints, backward compatibility and user flows, and the test changes for weakening. The declared playbooks removal and scheduled-turn rendering changes match the stated scope; I found no additional issue in the new TUI catalogue follow-up.

Verification: `ui-web` `npm test` passed 188 files and 2,452 tests; `npm run type-check` passed; `npm run lint` reported 0 errors and 4 warnings. `ui-tui` `npm test` passed 144 files and 2,061 tests with 13 skipped; `npm run type-check` and `npm run lint:i18n` passed; `npm run lint` reported 0 errors and 23 warnings. Installs reported unsupported-engine warnings under local Node 23.10.0; both suites nevertheless completed successfully.

ui-web/src/styles/page.css
```diff
@@ -3811,6 +3852,112 @@ details[open] summary { margin-bottom: 8px; }
...
+   The hosted panel does not scroll: the two columns inside it do, each on its
+   own, so a long channel list never carries the form on the right down with
+   it. */
+.setveil[data-section="channels"] .spanels[data-for="channels"],
+.setveil[data-section="cron"] .spanels[data-for="cron"],
+.setveil[data-section="memory"] .spanels[data-for="memory"] {
+  display: grid; grid-template-rows: minmax(0, 1fr); overflow: hidden;
+}
+.two-pane {
+  display: grid; grid-template-columns: 276px minmax(0, 1fr);
```
gloryfromca, line 3887: Blocking: make the new two-pane settings sections usable at narrow viewport widths.

The settings modal explicitly switches to a 96vw, one-column layout below 900px, but this child grid keeps a fixed 276px list track. On a 375px viewport, the hosted panel has about 320px after its padding, leaving roughly 44px for `.two-pane-main`; that pane itself has 40px of horizontal padding, so the channel, schedule, and memory details collapse to only a few pixels. A normal narrow-screen user can select a row but cannot use its detail controls.

ui-web/src/features/cron/CronPage.tsx
```diff
@@ -472,14 +425,16 @@ function JobForm({ draft }: { draft: CronDraft }): JSX.Element {
...
+          the channels section, because what a result can be delivered TO is
+          whichever channel is in service. */}
       <div className="ff">
         <label>{t('gui.job.deliver')}</label>
         <div className="sustate" style={{ marginTop: 0 }}>
           <span>{t(DELIVER[draft.deliver] || DELIVER.app!)}</span>
           <span className="x">{t('gui.job.deliver_global')}</span>
           <span className="a">
-            <button className="mini ghost" onClick={() => settingsDialog.open()}>
+            <button className="mini ghost" onClick={() => settingsDialog.openSection('channels')}>
```
gloryfromca, line 437: Blocking: preserve the unsaved schedule draft when this link opens Channels.

`JobForm` is also the body of `JobSheet` for new and duplicated schedules. This click changes sections, so `openSection` first runs `leaveSection`; the registered `closeCronSheet` handler then sets `sheet` to null. That unmounts and permanently drops every unsaved field with no prompt, meaning the normal attempt to configure delivery while composing a schedule loses the draft.

gloryfromca (commented):
Blocking: the two existing unresolved findings must be addressed.

I reviewed the `b9be81106c91` delta and found no new issue. It regenerates the TUI catalogue and corrects runtime-turn wording, but it does not touch either web path covered by the existing inline findings, so both remain outstanding. I covered the delta, its callers and origin enum, commit history, the applicable repository and TUI architecture rules, compatibility fallback behavior, and the changed tests for weakening.

Verification: `ui-tui` `npm test` passed 144 files and 2,061 tests with 13 skipped; `npm run type-check` and `npm run lint:i18n` passed.

ui-web/src/styles/page.css
xfng-sd: Fixed in 69658a3db.

The two-pane stacks inside the same `@media (max-width: 900px)` block the dialog already gives up its side rail in: one column, the list as a capped band (`max-height: 38vh`) above the detail, each scrolling on its own, and the main pane's padding tightened to 14/16.

Measured at a 390px viewport, channels section with a row open: the list is 332x296 and the detail 332x265, with the credential form 300px wide. Before the change that form was about four pixels.

One thing I did not touch: `.settings-tp` from #565 has the same fixed track (286px) and no narrow-screen rule, so the Model providers section still collapses the same way. It is that PR's markup rather than this one's, and the breakpoint behaviour there is a call for its author, so I have left it; say the word if you want both covered here.

ui-web/src/features/cron/CronPage.tsx
xfng-sd: Fixed in 69658a3db. You are right about the path: `openSection` spends the leave slot, the slot called `closeSheet`, and the draft went with it.

The slot parks the sheet now rather than closing it. `features/cron/store.ts` gains a `parked` flag: leaving sets it when a draft is up, `CronApp` draws the sheet only when it is not set, and `enter()` clears it. The draft stays in the store the whole time. The sheet has to leave the screen -- it is a veil at the body, so it would otherwise float over whichever section the reader went to -- but nothing is thrown away, and arriving back at Schedules puts it up with what was typed in it. Cancel still drops it for good.

Renamed the slot `parkCronSheet`, since `closeCronSheet` no longer describes it, and `state/settings.ts` now says outright that a leave slot may not throw work away.

Two cases in `CronPage.test.tsx`: one types a name, leaves the section, and asserts the sheet is off screen, the veil closed, the draft kept, and the field back to `half written` on return; the other asserts a cancelled sheet does not come back. Mutation-checked by pointing `park()` at `closeSheet()` -- the first goes red. Also walked it in the browser: type a name, press Change, land on Channels, pick Schedules, draft intact.

ui-web/src/styles/page.css
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed: the narrow breakpoint now stacks the list over a usable-width detail pane while retaining independent scrolling. The `.settings-tp` case is pre-existing in the target and does not undermine this fix. This finding is resolved.

ui-web/src/features/cron/CronPage.tsx
gloryfromca: No blockers; this can merge as far as I am concerned.

Confirmed: leaving Schedules now hides the sheet without clearing its draft, re-entering restores the form from that same draft, and explicit cancel still clears it. The added tests exercise both persistence and cancellation. This finding is resolved.

gloryfromca (commented):

The responsive stacking and parked-draft changes fix both prior findings; I replied to and resolved those threads. I found no new issue in the fix delta or its interaction with the updated target base. I reviewed the full target diff and fix delta, relevant callers and lifecycle paths, commit history, AGENTS.md plus CONTEXT-MAP.md and ui-web/CONTEXT.md constraints, backward compatibility, and test changes for weakening.

Verification: `ui-web` `npm test` passed 189 files and 2,459 tests; `npm run type-check` passed; `npm run lint` reported 0 errors and 4 warnings.

0xKT (approved):

PR #593, opened 2026-09-21
fix(*): the answer footer, the permission popover, and one ground under the panes

xfng-sd:
## Summary

Eight items on the conversation view, on top of `refactor/ui_web_architecture`.

- **The answer footer was invisible for good, not just at rest.** `.turn.ai > .ansfoot .acts` is four classes and the reveal it was paired with, `.turn:hover .acts`, is three, so the transparent rule outweighed the hover and the copy and branch buttons never came back. Measured on the running page: `opacity` stayed `0` with `.turn:hover` matching. The footer now fades as one element at the turn level, which is also the shape the reference design gives it. Its button hover colours already matched the design and now actually show.
- **The permission popover** takes the reference design's geometry and type: 268px wide, 7px/9px rows at an 8px radius, the name at 13.5px going from 500 to 600 when it is the one in force, the hint at 11.5px in `--faint`, and a 14px amber tick with a 2px stroke. The per-row shield and the filled band behind the chosen row are gone. `gui.perm.smart_h` and `gui.perm.full_h` are reworded to the design's sentences, in both languages, and the tui catalogue is synced.
- **The boot splash and the favicon** are the flat raven mark (`ui-web/src/assets/raven.png`), square so a browser tab cannot stretch it, with the bird 68px of the 86px box and the rest transparent. The splash mark was sized for the old 960px illustration; at 111px square it renders the bird at the size it did before without upscaling much past the file. The mark is a silhouette, so it inverts on the dark ground - down both theme paths, the way the page's own mono marks do - and `raven-dark.png` is the same drawing inverted, behind a `media` query on a second icon link, because the favicon has the same problem against a dark tab strip.
- **The back-to-bottom pill** centres on the transcript column rather than on the chat. With the anchored desk taking a strip off the chat's right edge, the pill sat half that strip, 162px at the default panel width, right of the text it marks.
- **The wash moves from `.chat` to `.main`,** so a desk pane shares the conversation's ground instead of standing on flat `--page` two pixels beside it. That seam ran the full height of the window whenever a pane was open. `--page` is the wash's own last layer, so nothing changes where no gradient stop reaches.
- **The inline rename field sizes to the name in it.** It opened at `min(420px, 52%)` however long the name was, so a two-word conversation got a field with most of it empty paper. `field-sizing: content` makes the input measure its own value; the cap stays the heading's own, so a rename cannot widen the header past where the title already sat, and a 7ch floor keeps a field you have just emptied from collapsing to a sliver.
- **The desk reserve leaves the scroll container alone.** Its three columns now take the same inset as `max-width: min(cap, 100% - R)` plus `translate: -R/2`, which is the same geometry case for case and does not relay the transcript out on every frame of the animation.

That last one is a partial answer to a reported render glitch in the message list when panes open and close: two cards of the same width 225px apart with a clean horizontal break between them, which is a stale raster band rather than two layouts. It did not reproduce here - hundreds of frames of pane and palette cycling at the reported window size and with a wider dragged panel, checking every frame that the turns agree on their offset and that the column tracks the composer, found no disagreement. A padding animation on a scroll container is the standing suspect, so this removes it; it is offered as a plausible fix, not a confirmed one.

Three judgement calls worth a reviewer's eye:

1. `gui.perm.full_h` first took the reference design's "no restrictions" wording, which review caught as a false security model - and it was: the waterfall in `raven/permissions/gate.py` returns builtin denials and user deny rules BEFORE it reads the mode, so `PermissionMode.FULL` bypasses the ask tier and nothing else, which is what `PermissionsConfig` documents. The hint now keeps the design's dangerous-operation warning and drops the claim about the standing denials.
2. The popover keeps its note ("takes effect from the next tool call"). The design's menu is generic - the effort chip opens the same one - and has no slot for it, but the sentence is real information.
3. The popover heading keeps the page's own `.lab` face rather than the design's plain sans, because `#tierPop` opens from the chip next door wearing the same heading and one of the two in a different face is worse than either.

`ui-web/src/assets/ravens/main-agent.webp` is now unreferenced, as the other four in that directory already were. Left in place rather than deleted on my own.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Run in `ui-web/` on the rebased branch:

- `npm test` - 189 files, 2462 cases, all passing. `scripts/gates/desk-reserve-css.test.mjs` is rewritten: it pinned the old `padding-right` on `div.scroll` by literal, and now pins the cap-and-slide pair, the composer's `36px + R`, the pill's centring, and that the scroll container itself carries no inset.
- `npm run type-check` - clean.
- `npm run lint` - 0 errors, 4 pre-existing `react-hooks/exhaustive-deps` warnings, none in the touched files.
- `npm run build` then `python ui-web/build.py` - page assembled, both boot goldens match (235 nodes each).

Measured in Chrome against the dev server, stub fixtures, rather than asserted:

- the answer footer reads `opacity: 0` at rest and `1` with the turn hovered, the button at `--text` on `--surface`;
- the transcript column, the composer card and the pill share one centre with the desk closed (986) and open (824, reserve 324);
- the clamped narrow-window case lands the column's left edge on the scroller's, which is what the old padding did;
- the popover measures 268px with the row metrics above and no shield;
- the rename field measures 135px for an eight-character name where it used to take 420, caps at 420 for a 200-character one, and holds its 56px floor when emptied, at a 1177px header;
- the splash mark renders 111x111 from an 86x86 file, and the favicon is served as `image/png`.
- the splash mark resolves to `invert(1) drop-shadow(...)` over a #1d1d1d ground down both the explicit-dark and system-dark paths and to the shadow alone down both light ones, and the dark icon link matches only under `prefers-color-scheme: dark`;

## Risk

All of it is the conversation view's paint. The two that could bite beyond it:

- A `translate` makes its element a containing block for positioned descendants. Nothing in the scroller's three columns depends on an outer one today - measured on the running page, one absolutely positioned element and it sits inside its own positioned parent - but a new `position: fixed` inside the transcript would be re-based by it. The comment beside the rule says so.
- The wash on `.main` is inherited by anything that later renders there without a ground of its own. The module pages are opaque fixed layers and are unaffected; dark theme resolves to its own flat gradient.

`gui.perm.full_h` is the one wording change a reader could act on, and item 1 above is the caveat.

Rollback is per item: each is one or two rules in `ui-web/src/styles/page.css`, except the popover markup (`ui-web/src/chrome/PermPopover.tsx`, `ui-web/src/state/perm.ts`), the two catalogue strings, and the new asset with the two references to it in `ui-web/src/page.html`.

## Related Issues

N/A


i18n/messages.json
```diff
@@ -3388,16 +3388,16 @@
       "zh": "完全访问"
     },
     "gui.perm.full_h": {
-      "en": "No prompts - but operations that the platform deems dangerous are still refused.",
-      "zh": "不弹审批；平台级高危操作仍会被拒绝"
+      "en": "No restrictions - it can run dangerous operations.",
```
gloryfromca: Blocking: the full-access copy must not claim that deny guards disappear.

`PermissionMode.FULL` bypasses only the ask tier: `PermissionGate.check()` still returns built-in and user-rule denials before reaching the full-mode branch, and `PermissionsConfig` documents that contract. This new sentence tells operators the opposite ("No restrictions"), so they can select or diagnose the mode under a false security model. Please preserve the dangerous-operation warning without saying those standing denials are removed, and keep the Chinese/generated entries in sync.

ui-web/src/page.html
```diff
@@ -23,7 +23,7 @@
      load; live mode keeps it up until the boot sequence has real data. -->
 <div id="splash" aria-hidden="true">
   <div class="sp-in">
-    <img class="sp-rv" src="assets/ravens/main-agent.webp" alt="" decoding="async">
+    <img class="sp-rv" src="assets/raven.png" alt="" width="86" height="86" decoding="async">
```
gloryfromca, line 34: Blocking: the splash mark must remain visible in dark mode.

The new PNG has 2,635 fully opaque black pixels forming the bird, while dark mode makes `#splash` use `--ink` (`#1d1d1d`); no theme rule changes this image. The body therefore renders at roughly 1.25:1 contrast and disappears, leaving mostly the white eyes and gray beak. This is reached by either system dark mode or the explicit dark preference. Please give the splash a theme-adaptive mark treatment; the favicon should also be checked if it continues to share this asset.

gloryfromca (commented):
Blocking: correct the full-access security copy and keep the splash mark visible in dark mode.

I reviewed the project rules (AGENTS.md/CLAUDE.md and the Web UI context), the complete target-to-head diff, the affected DOM and callers, permission-gate semantics, relevant history, backward compatibility, test changes, and architecture boundaries. The two concrete failures are marked inline. The desk-reserve assertions track the new geometry rather than weakening coverage, and the remaining changed paths fit the documented Web UI boundaries.

Verification: `npm test` passed 189 files / 2,462 tests; `npm run type-check` and `npm run build` passed; `npm run lint` passed with four pre-existing warnings outside touched files; `uv run python ui-web/build.py` passed both 235-node boot goldens; the source-language and large-file gates passed against the PR base.

gloryfromca (commented):
Blocking: the existing full-access copy and dark-splash findings still need to be resolved.

I reviewed the complete target-to-head diff and the delta from the previously reviewed revision. The only new code sizes the inline conversation-title editor with `field-sizing: content`; I traced it through the title replacement path and existing rename tests, checked its fallback and cap behavior, and found no additional plain error. I also rechecked the project rules (AGENTS.md/CLAUDE.md and the Web UI context), callers, relevant history, backward compatibility, test changes, and architecture boundaries.

The two existing inline findings are untouched, unanswered, and not deferred in the discussion record, so both remain blocking and their threads remain open.

Verification: `npm test` passed 189 files / 2,462 tests; `npm run type-check` and `npm run build` passed; `npm run lint` passed with four pre-existing warnings outside touched files; `uv run --frozen --python 3.12 python ui-web/build.py` passed both 235-node boot goldens; the source-language and large-file gates passed against the PR base.

i18n/messages.json
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in `2ba323f59197`: the English and Chinese hints now state that full mode suppresses prompts while deny rules still hold, and the generated TUI catalogue matches.

ui-web/src/page.html
gloryfromca: No blockers; this can merge as far as I am concerned.

Fixed in `2ba323f59197`: the splash now inverts under both explicit and system dark themes, and the favicon has a dark media candidate. I also verified `raven-dark.png` is an exact RGB inversion of the 86x86 light asset with identical alpha.

gloryfromca (commented):

I reviewed the full diff against the new `refactor/ui_web_architecture` target, the rebase range-diff, the fix commit, affected callers and permission semantics, relevant history, backward compatibility, test changes, and the architecture and repository rules in AGENTS.md/CLAUDE.md and the Web UI context. The two earlier commits are patch-identical after rebase, both prior findings are fixed, and I found no new plain error. I replied to and resolved both threads I opened.

Verification: `npm test` passed 189 files / 2,462 tests; `npm run type-check` and `npm run build` passed; `npm run lint` passed with four pre-existing warnings outside touched files; `uv run --frozen --python 3.12 python ui-web/build.py` passed both 235-node boot goldens; the source-language and large-file gates passed against the new PR base. I also verified the dark PNG is 86x86, preserves alpha exactly, and inverts every RGB channel from the light asset.

0xKT (approved):

PR #594, opened 2026-09-21
fix(ui-web): re-align the onboarding model step with the provider split

0xKT:
## Summary

The provider split (#565) moved model picking to one body-level picker surface and left the compact provider card to the onboarding wizard alone, but the wizard's first step was not re-aligned with it. This PR does that, keeping the step in the prototype's shape.

**Regression fix.** The picker now lives at the picker layer (46) outside the wizard, so opening it from a role pill inside the wizard (110) drew it behind the wizard. It is lifted the same way the confirm veil already is while the wizard is shown, and the layer gate pins the ladder (picker token below the wizard, base rule on the token, lifted rule present).

**Wizard add block, aligned with the prototype and the catalogue page.**
- The vendor select shows the catalogue page's four groups (direct, gateway, browser sign-in, local). One `groupOf` answers both the wizard's select and the catalogue's filter, in the catalogue's order (an aggregator first, whatever credential it takes), so the two cannot drift; labels are the catalogue's and no i18n key is added. A new test pins the buckets, that every vendor lands in exactly one, the connected-first order and the search needle.
- An aggregator gets the address row the catalogue's connection card already gives it, prefilled from its default base. The base is sent whenever the field is non-empty, as before.
- When a device flow lands, the add form that started it closes: pointing at a now-connected slug, it used to redraw itself silently for the next unconnected vendor.

**The settings catalogue's columns.** Found while trying the catalogue in the wizard, fixed where it lives: the provider catalogue sized its grid with a max-height, which a grid's fr row does not resolve against, so both columns grew to the vendor list's height and were clipped -- the list could not scroll past the first seventeen vendors and the "pick one" placeholder sat below the fold. A definite height bounds the row track, so the list and the detail pane scroll on their own; a gate reads the sheet and pins it.

Decisions for the reviewer:
- Local vendors keep their optional key row (the prototype has none). It matches the catalogue's connection card and costs nothing; drop it on request.
- Known debt, not addressed here: the wizard's add block duplicates about sixty lines of form rules from the catalogue's connection card. Folding them into one component is a follow-up.
- Two small inconsistencies surfaced by review, left for that same follow-up: the vendor tag on the catalogue's detail card still says "Local / self-hosted" where the filter and the wizard say "Local deployment" (identical in Chinese), and three address-row conditions carry an `endpoint` term that the kind helper already implies.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

```
cd ui-web && npm test          # see the PR checks for the counts; includes the layer and catalogue gates
cd ui-web && npm run type-check  # clean
cd ui-web && npm run lint        # 0 errors, 4 pre-existing warnings (same on the base)
cd ui-web && node scripts/check-class-namespace.mjs   # OK
```

Real host: a gateway from this branch under an isolated RAVEN_HOME, driven by Playwright on the first-run wizard. Before: the picker computed z-index 46 and the element under its centre belonged to the wizard. After: z-index 111 and the element under its centre is the picker's search field; the vendor select shows the four groups with the same counts the registry gives the catalogue page (direct 25, gateway 21, browser sign-in 4, local 5). In the settings dialog the catalogue's columns measure one screenful and the list scrolls on its own (they measured 2008px and were clipped before).

## Risk

UI only. The lift rule applies only while the wizard is open. The settings dialog changes in two places: its catalogue columns are bounded and scroll (they were clipped before), and its filter reads the shared grouping, which files every one of the registry's 55 vendors exactly where the old predicate did (recomputed over the registry). The store change touches only the OAuth poll's success branch. Rollback is a plain revert.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the repository rules and UI vocabulary, the complete eight-file diff, the affected onboarding/model-picker and provider/OAuth callers, relevant history and wire data, backward compatibility, architecture boundaries, and the test changes for weakening or lost coverage. The shared classifier preserves the catalogue semantics while fixing the wizard grouping, the OAuth close is scoped to the form that started that flow, and the picker lift remains bounded to visible onboarding.

Verification: focused suites passed (31 tests); full `ui-web` suite passed (190 files, 2467 tests); type-check passed; lint completed with 0 errors and 4 pre-existing warnings in untouched files; RPC generation check passed; Vite build, both boot snapshots, and built-page check passed.

0xKT:
Scope change since the last review, on the author's request after a real-machine pass: the wizard's first step is no longer the prototype's compact provider card but the settings dialog's own two pages, stacked -- the provider catalogue with its detail pane, then the model roles card. The title and description are updated accordingly (feat, not fix).

What changed on top of the reviewed head:

- `ModelStepBody` renders `pages/Provider` and `Roles`; the compact card, its add form, the `provAdd` slot, their styles and nine catalogue-only i18n keys are removed (`ui-tui` catalogue regenerated). The shape helpers stay in `providers/Providers.tsx` for the catalogue page and its detail pane.
- Two defects the wizard's first run exposed are fixed where they live: the catalogue's grid used a max-height, which its fr row does not resolve against, so both columns grew to the list's height and were clipped (no scrolling past the 17th vendor, "pick one" placeholder below the fold, in the dialog too); it now has a definite height. The detail pane's add-model pop (a body portal at layer 47) opened behind the wizard the way the picker did; it is lifted the same way and the layer gate pins both.
- The roles card's "connect a provider" button switches the dialog's tab; inside the wizard it now scrolls the catalogue above it into view instead of resetting what the reader had open.

Verification on this head: `cd ui-web && npm test` 190 files / 2470 tests, `npm run type-check` clean, `npm run lint` 0 errors (4 pre-existing warnings), `node scripts/check-class-namespace.mjs` OK; full Python suite and the TUI i18n check pass in preflight. Real machine (isolated RAVEN_HOME, Playwright): catalogue 55 rows in a column that scrolls on its own, placeholder visible until a row is clicked, connect from the detail pane, picker and add-model pop both compute z-index 111 above the wizard (110), roles-card button scrolls the catalogue to the top of the step.

0xKT:
Scope settled, on the author's request after trying both on a real machine: the wizard's first step keeps the prototype's compact provider card. The branch is back to the head you reviewed (123945df: picker lift, four-group vendor select with the shared `groupOf`, aggregator address row, OAuth close) plus one commit, and the settings-pages rebuild pushed earlier this evening is withdrawn. Force-pushed with lease; the title and description are back to the fix.

The one commit on top of the reviewed head fixes a defect the experiment surfaced in the settings dialog itself: the provider catalogue sized its grid with a max-height, which a grid's fr row does not resolve against, so both columns grew to the vendor list's height and were clipped -- the list could not scroll past the first seventeen vendors and the "pick one" placeholder sat below the fold. It now has a definite height and a bounded row track, so the list and the detail pane scroll on their own; a new gate (`scripts/gates/settings-catalogue-css.test.mjs`) reads the sheet and pins it (reverting the rule turns it red).

Verification on this head: `cd ui-web && npm test` 191 files / 2469 tests, `npm run type-check` clean, `npm run lint` 0 errors (4 pre-existing warnings), `node scripts/check-class-namespace.mjs` OK; full Python suite passes in preflight. Real machine (isolated RAVEN_HOME, Playwright): the wizard's vendor select shows the four groups (25 / 21 / 4 / 5), the picker opened from a role pill computes z-index 111 above the wizard, and in the settings dialog the catalogue's columns measure one screenful with the list scrolling on its own (2008px and clipped before).

gloryfromca (commented):

The withdrawn settings-pages experiment is absent: this revision is the previously reviewed compact-card implementation plus the catalogue sizing fix. I reviewed the repository rules and UI context, the complete PR diff and one-commit delta, the catalogue markup and dialog sizing callers, relevant history, backward compatibility, architecture boundaries, and the tests for lost or weakened coverage. The definite grid height and bounded row track make the existing list/detail overflow rules effective without changing the onboarding path.

Verification on this head: focused suites passed (5 files, 33 tests); full `ui-web` suite passed (191 files, 2469 tests); type-check passed; lint completed with 0 errors and 4 pre-existing warnings in untouched files; RPC generation, production build, both boot snapshots, and built-page checks passed. There are no review threads from me to resolve.

gloryfromca (approved):

PR #599, opened 2026-09-21
fix: run ui-web page assembly on the project's pinned Python

### LINKED ISSUE #526: fix(ui-web): build.py needs Python 3.10+ for one keyword, so the documented from-source build fails on a stock macOS python3
## Summary

`ui-web/build.py` writes the assembled page with `Path.write_text(out, encoding="utf-8", newline="")`. The `newline` keyword on `write_text` exists only from Python 3.10, and the script is otherwise plain standard-library code that compiles on 3.9. Every documented way of running it that does not go through `uv` -- `make build-ui` (`Makefile`, the `build-ui` target runs `python3 ui-web/build.py`), the `python3 ui-web/build.py` lines in `.github/workflows/ci.yml`, the usage line in the script's own docstring, and the self-hosting page's "from source" recipe (`make install-deps && make build-ui && uv run raven web`) -- picks up whatever `python3` is first on `PATH`. On a stock macOS that is Python 3.9.6, and the build dies after Vite has already succeeded.

`install.sh` and `install.ps1` are not affected: both run the script through `uv run --no-project python`, which is the project's 3.12.

## Steps to reproduce

1. On macOS with no other `python3` ahead of `/usr/bin/python3` (3.9.6), clone the repo and install Node and uv as the self-hosting page says.
2. `make install-deps && make build-ui`

## Expected behavior

`build-ui` assembles `ui-web/dist/index.html`. Either the script runs on any Python 3 (it uses nothing else that is version-specific), or the Makefile runs it on the interpreter the project already requires.

## Actual behavior

```
  File "ui-web/build.py", line 198, in main
    (dist / "index.html").write_text(out, encoding="utf-8", newline="")
TypeError: write_text() got an unexpected keyword argument 'newline'
```

The `newline=""` was added deliberately: a text write translates `\n` to `os.linesep`, which on Windows emits a CRLF page whose `<script>` payloads `check-page.mjs` can no longer extract. That intent is right; the keyword is just the one way to express it that ties the script to 3.10+.

Two remedies, both small; the first is the one that removes the dependency, the second only picks the right interpreter for one caller:

- **`ui-web/build.py`** (under the current `ui-web/` freeze, so filed rather than opened as a PR): write bytes instead of text, which has no newline translation on any platform and no version floor:

  ```python
  (dist / "index.html").write_bytes(out.encode("utf-8"))
  ```

  Measured on this branch: the page built by Python 3.9.6 and by 3.12 with this line have the same sha256, and it equals the sha256 of the page built by the current `write_text(newline="")` on 3.12; `check-page.mjs` passes on the result and the file holds zero CR bytes. The comment above the line should say why bytes: the script is run by whatever `python3` the caller has.

- **`Makefile`** `build-ui`: `uv run --frozen --python $(PYTHON_VERSION) python ui-web/build.py`, so the documented `make` path uses the project's interpreter like `install.sh` already does. This half does not touch `ui-web/` and can go now.

The self-hosting page's prerequisites do say Python 3.12, but `make build-ui` does not use it, which is what makes the recipe fail on a machine that satisfies the prerequisites.

## Environment

- Raven `main` at 04eb8a31 (2026-09-20)
- macOS 15, Apple Silicon; `/usr/bin/python3` = 3.9.6; repo venv Python 3.12

## Logs or screenshots

The traceback above is the whole output after `npm run build` succeeded.

insuffer:
## What

`make build-ui` assembled the served page with a bare `python3`, which
resolves to whatever interpreter is first on PATH. On a stock macOS
(no Command Line Tools ahead of `/usr/bin/python3` = 3.9.6) the
documented self-hosting recipe `make install-deps && make build-ui`
died after the vite build had already succeeded:

    File "ui-web/build.py", line 198, in main
        (dist / "index.html").write_text(out, encoding="utf-8", newline="")
    TypeError: write_text() got an unexpected keyword argument 'newline'

`Path.write_text(newline=...)` exists from Python 3.10, while pyproject
`requires-python` is `>=3.12` and every other Makefile target already
runs its Python through uv. `build-ui` was the one caller that bypassed
the pinned interpreter.

## Change

One line in the Makefile: run the script through
`uv run --frozen --python $(PYTHON_VERSION) python ui-web/build.py`,
the same form `install.sh` already uses for this script (and already
documents the rationale for). No source, test, or dependency changes.

`ui-web/build.py` itself is untouched: its `newline=""` write needs
3.10+, but that half of the fix sits under the ui-web freeze, so this PR
takes the Makefile half only, as the issue suggests.

CI keeps its bare `python3` (the ubuntu runners provide a working one
through setup-python / the runner image); this PR changes the documented
local path only.

## Verification

Reproduced the failure with Python 3.9.13 (the stock-macOS analogue):

    $ python3.9 ui-web/build.py
    TypeError: write_text() got an unexpected keyword argument 'newline'
    (exit 1)

Then ran the changed target's commands in order, on Windows with
uv 0.12.17 and Python 3.12.7:

    npm ci --prefix ui-web
    npm run --prefix ui-web build
    uv run --frozen --python 3.12 python ui-web/build.py    -> exit 0
    -> built dist/index.html (1,496,450 bytes)

Artifact checks:

    sha256 ui-web/dist/index.html = a1068871929266cfe746b5f321eb7faed2158785c8b19d4984783099b4c23181
      (byte-identical to the pre-change output on 3.12)
    node ui-web/scripts/check-page.mjs             -> OK (3 scripts parse)
    node ui-web/scripts/count-shared-globals.mjs   -> OK
    node ui-web/scripts/check-css.mjs              -> OK
    node ui-web/scripts/check-class-namespace.mjs  -> OK
    dist/index.html carries 0 CR bytes (LF-only artifact)

Repo gates on this commit:

    scripts/check_source_language.py HEAD~1..HEAD  -> pass
    scripts/check_commit_messages.py HEAD~1..HEAD  -> pass

Based on main at 87f392866e95b6828023cd0305f308831f4105f0.

Closes #526

AI disclosure: this PR was drafted with AI assistance (ZCode agent,
model step-5-preview); the reproduction, commands, and results above
were produced by it. The diff is one line; please review before merging.


insuffer:
Hi — first contribution here: a one-line fix so the ui-web page assembly runs on the project's pinned Python instead of whatever is first on PATH. The `Commit lint`, `CI`, and `Docs` workflow runs are waiting for fork approval — could a maintainer approve them so the checks can run? Thanks!

gloryfromca (approved):

PR #601, opened 2026-09-21
fix(ui-web): show what two settings writes actually changed

gloryfromca:
## Summary

Two settings-dialog writes whose effect the page did not show. Both were found
by an adversarial review of the archive and model fixes and then reproduced on a
real host.

**The permission chip could name a weaker tier than the conversation ran at**,
which is the one direction this control must never fail in. A mode picked before
there is a conversation is staged, and the first message applies it to the
session it mints. The chip's re-read did not know that: with no session the read
is default-scoped, so the gateway answers the configured default, and the chip
was painted from it unconditionally. Opening the dialog re-reads the chip, and
so does every settings write, so the pick was overwritten seconds after it was
made while the write that matters still went through. The draft keeps its own
pick now, which is what the model chip already does for the same case. A draft
that picked nothing still follows the default, and a conversation still follows
its own mode: both come from the session-scoped read the guard does not touch.

**Turning auto-archive on showed nothing.** The sweep runs inside session.list,
so writing the setting moves nothing by itself: the card went on saying nothing
was archived and the rail went on listing the stale conversations, until a
reload ran the sweep and a pile of them disappeared at once, with no moment
connecting that to the switch. The toggle now re-reads both lists, which is what
performs the sweep and then shows what it did. A refused write re-reads nothing,
so a failure cannot read as though it took.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Both driven on a real host (`raven web` on an isolated `RAVEN_HOME` built from a
real config), reading the result from the layer that owns it rather than from
the page: the mode a session was created on from `permissions_mode` in the last
metadata record of its transcript, and the archived state from `archived` in the
same place.

The permission chip, with `permissions` empty so the shipped default applies:

| Step | Chip before | Chip after |
| --- | --- | --- |
| New-task screen | smart | smart |
| Pick full access | full | full |
| Open the settings dialog | **smart** | full |
| Chip at the moment of sending the first message | **smart** | full |
| `permissions_mode` on the session that message minted | **full** | full |

The last two rows are the defect: the chip and the session disagreed, and the
chip was the weaker of the two.

Auto-archive, against two conversations backdated past the threshold with the
setting off to start with:

| Step | Before | After |
| --- | --- | --- |
| Flip the switch on | switch on, card still empty, rail still 5 rows | card lists both at once, rail drops to 3 |
| Reload the page | now they are gone from the rail | unchanged |

The third backdated conversation stays on the rail in both columns: it carries
an `archived` key from an earlier restore, which the sweep skips by design.

Commands:

- `npm test --prefix ui-web` -- 189 files, 2475 tests, all pass
- `npm run --prefix ui-web build && python3 ui-web/build.py && node
  ui-web/scripts/check-page.mjs` -- boot-snapshot OK (235 nodes match golden),
  page OK

Four cases were added, each with a control that must stay green through the same
removal:

- `a draft that picked a mode keeps it when the chip is re-read`, against
  `a draft with no pick of its own, and a conversation, both follow the read`
- `turning the switch on re-reads both lists, which is what runs the sweep`,
  against `a refused write re-reads nothing`

Each of the two first cases was run against its unfixed module and fails there
on the assertion it is named for.

## Risk

The permission change is one early return, reached only when there is no session
AND the draft staged a pick; what the gate enforces was already the staged
value, so it changes what the reader is told, not what runs. The auto-archive
change adds two reads after a write that succeeded. No server change, no stored
state change in either. Rollback is reverting the two commits, which are
independent of each other.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the target diff, the settings source/store callers, session registry and runtime staging flow, the backend `session.list` sweep, and the relevant history. I also checked AGENTS.md/CLAUDE.md, the Web UI context and layering rules, backward compatibility, and whether the new tests could pass without exercising the claimed behavior. The archived-list request itself runs the sweep before filtering, so the implemented ordering updates the card and then the rail; the staged permission guard preserves only an explicit draft pick while retaining default and session-scoped refreshes. I found no weakened tests or actionable defects.

Verification: `npm test --prefix ui-web -- --run src/features/settings/source.test.ts src/features/settings/pages/Archive.test.tsx` (21 passed); `npm test --prefix ui-web` (189 files, 2475 tests passed); `npm run type-check --prefix ui-web` (passed); `npm run lint --prefix ui-web` (0 errors, 4 warnings in untouched files).

0xKT (approved):

PR #602, opened 2026-09-21
fix(*): take the delete off the memory section

gloryfromca:
## Summary

The memory section offered a delete that no store behind it can honour, so this
takes the control off and drops the `memory.delete` method with it. Base is
`refactor/ui_web_architecture`, not `main`.

What the reader hit: deleting a know-how entry answered "this backend cannot
delete a agent_skill memory". That row could never have been deleted. The
adapter rebuilds the skill directory path from the row's name, and a directory
written before EverOS started sanitizing names keeps its spaces, so the derived
path matches nothing and delete reports False.

Why the path rule was not fixed instead:

- EverOS has no user-facing memory deletion. Its one destructive operation on a
  skill exists to reap the directory an update left behind when it renamed a
  skill. Retirement is a design decision EverOS states and defers: its extractor
  lowers the confidence of a skill it wants retired, and the store writes that
  back as an ordinary skill.
- Episodes were the one kind the adapter could act on, through the
  `deprecated_entries` frontmatter map EverOS itself uses, and that answered
  True. It still was not what the reader asked for: the section lists through
  `POST /api/v1/memory/get`, which does not filter `deprecated_by` the way
  `/search` does. A retired episode left recall and stayed on the list, so the
  delete was reported as done and the row came back on the next load.
- Profiles and cases answered False throughout, so two of the four kinds could
  only ever report the same failure.

Removed: the armed delete button and its two call sites, the memory source's
`remove` and the store's guard around it, the `memory.delete` method and its
contract entry (193 wire methods to 192), the five catalogue strings, the
offline fixture's parameters for the call, and the tests that drove all of it.
`memory.*` is now the read-only surface its registration comment already
claimed, so `register_memory_methods` no longer takes a loop factory.

`MemoryBackend.delete` and the EverOS adapter that implements it are left as
they are. Nothing calls them now; the backend contract test still holds the
adapter to answering a bool, so a backend that grows a real deletion has a
contract to meet when a UI for it is designed.

No doc, spec or screenshot in the tree names the removed control, so none
needed updating.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Commands and results:

- `uv run --frozen --all-extras pytest tests/test_rpc_memory.py tests/test_rpc_registration.py tests/test_rpc_schema_match.py tests/test_rpc_contract_shapes.py tests/test_i18n_boundary.py tests/test_memory_backend_protocol.py tests/test_memory_backend_contract.py tests/test_cli_tui_bootstrap.py -q` -> 506 passed, 0 skipped
- `npm test` (ui-web) -> 189 files, 2469 passed
- `npm run type-check` (ui-web) -> clean; `npm run lint` (ui-web) -> 0 errors (4 warnings, all pre-existing in connections / cron / subagents)
- `npm run gen:check` (ui-web) -> generated.ts matches the contract (192 methods); `npm run lint:rpc` and `npm run lint:i18n` (ui-tui) -> in sync
- `node scripts/check-css.mjs`, `node scripts/check-class-namespace.mjs` -> OK
- `make check-source-language`, `make check-large-files`, `scripts/check_commit_messages.py`, `commitlint` -> exit 0 each (checker exit codes, not a pipeline's)
- Isolated acceptance run: a gateway on its own port with its own `RAVEN_HOME`, reading the running EverOS. Settings -> Memory -> Know-how -> the OCR entry opens with no button on it; the profile view has none either; the gateway answers `memory.delete` with `-32601 method_not_found` while `memory.stats` still reports its 396 episodes.
- The assertion the removed tests carried was put back on the pick test (a kind switch drops the picked memory) and checked by no-op-ing `closeDetail`: red with the mutation, green without it.

## Risk

User-visible change: the memory section loses its delete control. Nothing else
on the section changes; listing, semantic search, the kind switch and the picked
memory all behave as before.

Wire change: `memory.delete` is gone from the contract, so a client that still
calls it gets `-32601`. The TUI never called it, and the page's only caller is
removed in this change.

Rollback: revert the commit. The backend implementation is untouched, so the
control returns in the state it was in.

## Related Issues

N/A


gloryfromca (commented):

The delete removal is complete across the memory island, source/store seam, Runtime registration and models, Wire Schema, fixtures, generated clients, i18n catalogue, and focused tests. Repository-wide reference checks found no stale `memory.delete` caller. The retained backend `delete` contract remains independently covered and is not coupled to this read-only RPC surface.

Review coverage: AGENTS.md/CLAUDE.md and the Runtime/Web UI CONTEXT vocabulary; the full target diff; surrounding callers and repository history; wire compatibility and generated artifacts; whether removed tests merely tracked removed behavior; and the Runtime-to-Web architecture boundary.

Verification: `uv run --frozen --python 3.12 --all-extras pytest tests/test_rpc_memory.py tests/test_rpc_contract_shapes.py -q` (61 passed); focused memory Vitest suite (9 passed); Web UI RPC generation, lint, and type checks passed (four pre-existing warnings outside the diff); TUI RPC generation, i18n generation, and type checks passed; focused Ruff check passed. GitHub's page checks, UI RPC contract, TUI checks, and all four Python unit shards are also green. A full local Web UI run was stopped after it remained active under unsupported Node 23; the focused affected suite and GitHub page checks completed successfully.

0xKT (approved):

PR #604, opened 2026-09-21
fix(ui-web): one provider list per scope, as the models already are

gloryfromca:
## Summary

The model picker showed a conversation's model under a vendor whose key does not
serve it.

`is_current` is the whole difference between the two `model.options` reads: the
backend marks the conversation's provider when the request names a session, and
`agents.defaults`' when it does not. Both answers landed in one module-level
array, so the default-scoped read -- the settings dialog's, repeated by every
settings write, and the agents page's -- overwrote the conversation's.

What the reader saw: a conversation running a Gemini model while the default is
a DeepSeek one, one visit to the settings dialog, then a click on the model
chip. The picker opened on DeepSeek's column with `gemini-3.5-flash` prepended
and ticked inside it, DeepSeek's count raised by that phantom row, and the
current dot on both vendors.

Split, the way the model pair one layer down already is (#596 separated
`defaultModelLive` from the chip's own value and left this array shared).
`providersLive` is the visible conversation's answer, read by the picker through
`modelSource`; `defaultProvidersLive` is the configured default's, read by the
settings snapshot, the onboarding model step, and the agents page, which loads
that list for itself.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

Driven on a real host (`raven web` on an isolated `RAVEN_HOME` built from a real
config, default `deepseek/deepseek-v4-pro`, the conversation switched to
`gemini/gemini-3.5-flash`):

| Picker opened after one visit to the settings dialog | Before | After |
| --- | --- | --- |
| Column it opens on | DeepSeek | Gemini |
| Where the tick sits | on a prepended `gemini-3.5-flash` row inside DeepSeek | on Gemini 3.5 Flash, in Gemini |
| DeepSeek's model count | 4 (three plus the phantom) | 3 |
| Vendors carrying the current dot | DeepSeek and Gemini | Gemini |

Without the settings visit the picker was correct in both columns, which is what
makes the shared array the cause.

Commands:

- `npm test --prefix ui-web` -- 191 files, 2486 tests, all pass
- `npm run type-check --prefix ui-web` -- clean, and it is what found the third
  reader: the agents page imports this export under an alias, so grep missed it
- `npm run --prefix ui-web build && python3 ui-web/build.py && node
  ui-web/scripts/check-page.mjs` -- boot-snapshot OK (235 nodes match golden),
  page OK

The new case, `a default-scoped read leaves the conversation rows alone`, was run
against a module made to share one array again and fails there on the assertion
it is named for.

## Risk

One module-level array becomes two, and every reader was enumerated and pointed
at the scope it means. No server change, no stored state change, no change to
what any read asks for. Rollback is reverting one commit.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the full target diff and the surrounding provider consumers. The new default-scoped list is read by settings, onboarding, and Raven-hosted agent model choices, while conversation switching and the composer continue to read the session-scoped list through `modelSource`. The renamed export has no stale callers, and the added test covers the cross-scope overwrite without weakening existing assertions.

Covered the repository rules and Web UI context, the diff, callers, relevant history, backward compatibility, test changes, and architecture boundaries.

Verification:
- `npm test --prefix ui-web`: 191 files, 2486 tests passed

0xKT (approved):

PR #605, opened 2026-09-21
fix(ui-web): a node is drawn where the layout put it, on WebKit too

Handsome-wzw:
## Summary

On Safari, the task board drew every node card of a graph on the first node's spot: a two-step run showed one box with both titles overprinted and an arrow pointing at empty canvas where the second step should have been. Chrome drew the same graph correctly.

The cause is how `DagGraph` placed a node: a `transform="translate(x y)"` on the node's `<g>`, with the card inside a `<foreignObject x=0 y=0>`. WebKit hit-tests a `foreignObject` where the layout put it but paints its HTML without the ancestor group's transform, so every card was painted at the SVG's origin. Removing the CSS transform on the board's viewport did not change anything; writing the offset onto the `foreignObject` itself did, and that is what this change does: the node's place goes on each shape (`rect`, mark, label box, or the caller's card) and the group carries no transform. Edges, layout, selection and keyboard handling are unchanged.

One file. The label branch used by the transcript card gets the same treatment, since it has the same shape and would fail the same way on a graph wide enough to notice.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `npx vitest run` in ui-web: 189 files, 2469 tests passed. New test in `DagGraph.test.tsx`: two dependent nodes carry no group transform, their rects sit at different places, each label box shares its rect's row and sits 31px inside it, and a caller's card takes the node's own x/y.
- `npm run type-check`, `npm run lint` (0 errors; the 4 warnings are the base branch's), `npm run build`, `python ui-web/build.py`, `check-page.mjs`, `check-css.mjs`: all pass.
- Real page, served by a gateway on this branch head, opened in Playwright's WebKit 26.5 (the Safari engine the report came from) and in Chromium: before the change WebKit painted both cards of a two-step DAG at the first node's position; after it both engines paint the two cards at their own rows with the arrow between them, and every `.nd` group has no `transform` attribute.

## Risk

Rendering only; no data, contract or behaviour change. Geometry is identical to before on engines that honoured the group transform, since the same numbers now sit on the shapes. Reverting the commit restores the previous markup.

## Related Issues

N/A


gloryfromca (commented):

The absolute coordinates are applied consistently to the default rect, status mark, label foreignObject, and caller-supplied card, while edges already use the same layout coordinates. I found no caller, interaction, or CSS dependency on the removed group transform. The regression test covers both renderer branches and multiple node positions.

Reviewed the complete target diff and surrounding DAG/task-board code, callers, CSS, and relevant history. I also checked AGENTS.md/CLAUDE.md and the Web UI context, backward compatibility, architecture boundaries, and that existing tests were extended rather than weakened.

Verification: `npm test` (189 files, 2469 tests passed); focused DAG and task-page tests (74 passed); `npm run type-check` passed; `npm run lint` completed with 0 errors and 4 pre-existing warnings in untouched files; `git diff --check` passed.

0xKT (approved):

PR #606, opened 2026-09-21
fix(ui-web): restore the prototype's own numbers across the page

arelchan:
## Summary

Style only, against `refactor/ui_web_architecture`. Every value here is the
design prototype's, restored where the port had drifted. No behaviour, no copy,
no data: the one `.ts` file is the anchored panel's arithmetic and the one
`.tsx` is how that panel is dismissed.

**The control standard, which had not been ported at all.** Anything you type
into is paper with a `--line` edge and turns amber with a glow on focus; one row
of controls shares a height and a radius; grey is left to what is not a control.
The blanket `input[type=text], input[type=password], select` rule sat on
`--surface` at radius 7 with no height, and outranked the settings sheet's own
rules, so one form carried five grounds, five heights and three radii. Measured
after: every control in the settings dialog is 32px.

**Switch, segmented control, small button**, to the prototype's geometry:

| | before | after |
|---|---|---|
| switch | 32x18, grey knob | 38x22, 16px paper knob with a shadow |
| segmented | joined buttons, monospaced 11.5px | grey track, 2px of air, pressed segment is a paper card, 13px |
| `.mini` | one size, 12.5px / radius 8 | 28/7/12.5 on module pages, 32/8/13 in the settings dialog -- the two tiers the prototype has |

**A module page is read on `--page`**, which is what that token is for. On
`--ink` the cards, which are `--ink` themselves, lost their ground and the page
read flat.

**The model panel hangs down from its field again, sized off the card.** It ran
132px out of a 520x500 sheet, because the placer knew only `.smodal` and fell
through to the window. Now the box decides the width (the field's left edge to
20px inside the card, floored at 320 and capped at 560) while the window decides
the height: down first, giving up height to the shelf it has, up only from a
shelf too shallow to read. Its own type came down from 15/14.5px to the
prototype's 14/13.5/13. It closes on a press outside it, which is what a menu
does, and the Cancel button beside the search box is gone.

**`raven.svg` is the mark the chrome draws** (`components/RavenMark.tsx`), not a
second bird. The file shipped here was different artwork carrying an opaque
white plate, while `AgentMark.tsx` and `scripts/gates/agent-mark-css.test.mjs`
both already described a file with a `prefers-color-scheme` rule. The gate that
pinned the plate now pins either mechanism, which is what its own prose says it
guards.

**Two dead values found on the way.** `--green` is used once, by the new-file
marker in a change list, and is defined nowhere -- that marker had no colour.
And the caret on a rail group heading, which the prototype shows on hover or
when the group is folded, had lost all three of its opacity rules and sat there
permanently.

Not in this PR, and not style: the roster prints registry keys
(`claude_code`, `hermes`) rather than brand names, `Raven-PPT` has no catalogue
entry so its line falls back to an English probe verdict, and the agent sheet's
"what it is good at" shows the dispatch prompt. Those are copy and data.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [x] Other -- style parity with the design prototype

## Verification

```
cd ui-web && npx vitest run              # 191 files, 2492 tests pass
cd ui-web && npx tsc --noEmit            # clean
cd ui-web && npx vite build              # ok
node ui-web/scripts/check-css.mjs        # OK
node ui-web/scripts/check-class-namespace.mjs  # OK
python3 ui-web/build.py                  # boot-snapshot OK (235 nodes match golden)
node ui-web/scripts/check-page.mjs       # OK, 2 scripts parse
uv run pytest tests/test_ui_agent_marks.py -q  # 7 passed
```

Read back in a real browser on the built page in both themes, with computed
styles measured rather than eyeballed, on: the agents page, the settings dialog
(usage, providers, tools), channels, scheduled work, the transcript and the task
graph. The model panel was measured open: left edge on the field's, width 478 of
an available 478, height shrunk to the shelf, opens downward, closes on an
outside press.

## Risk

Visual only, and wide: the blanket input rule and `.mini` reach every page, so
the risk is a control somewhere off this PR's read path looking different. The
sizes are the prototype's, so a difference is a page that had drifted rather
than a regression. `.mini.update` opts out of the new width floor explicitly.
Rollback is reverting the commit; nothing here is persisted or migrated.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the full `github/refactor/ui_web_architecture...HEAD` range and the style commit separately. I covered the repository rules and Web UI vocabulary, the diff and affected callers, the stacked import-progress/provider-scope changes, backward compatibility, test changes for weakening, and the UI architecture constraints. I also checked the current replacement head: it is tree-identical to the reviewed revision and only corrects the commit type from `style` to `fix`.

Verification: `uv run pytest tests/test_ui_agent_marks.py -q` (7 passed); `uv run pytest tests/test_importer_orchestrator.py tests/test_rpc_import_sync.py -q` (68 passed); targeted Vitest picker/agent/CSS gates (71 passed); full `npm test` (191 files, 2492 tests); `npm run type-check`; Vite build; `uv run python ui-web/build.py`; `check-css`; `check-class-namespace`; `check-page`; and `boot-snapshot` all passed. The full Vitest run emitted happy-dom localhost fetch/teardown noise but exited successfully with all tests passing.

gloryfromca (commented):

The delta from the reviewed `3a9f79e2513e` tree only removes the extra blank line at the end of `ui-web/src/lib/popover.ts`, clearing the earlier `diff --check` warning without changing behavior. I rechecked the full target diff and the affected picker path; the prior review's coverage of repository rules, callers/history, compatibility, test integrity, and Web UI architecture remains applicable.

Verification on this head: `git diff --check github/refactor/ui_web_architecture...HEAD`, `npm test -- --run src/components/ModelPicker.test.tsx` (12 passed), and `npm run type-check` all passed.

gloryfromca (approved):

PR #607, opened 2026-09-21
fix(ui-web): the strip says how many tasks run and opens the tasks tab

0xKT:
## Summary

The strip above the composer drew one chip per running task (name only, up to three, plus an overflow chip), each opening that task's own desk pane. The prototype -- the spec for this page -- draws one pill, "{n} running", with the task names as its hover title, and a click opens the desk's tasks tab, whose rows already name each run and say how far it has got. The strip is a reminder that background work is under way, not a second copy of the tasks list, so this redraws it as the prototype has it.

Three commits:

- `fix(ui-web)`: the strip becomes one `.tkrunhint` pill inside `.tkruns`, on the dock's column axis, with a `.tkrundot` running dot matching the task row's. It subscribes to the language store (the whole pill is a catalogue word; the old overflow chip never did and stayed in the served language after a pick). `gui.tasks.running_n` replaces `gui.tasks.overflow_more`; the TUI's generated table is regenerated. The old `.runs` / `.trun` rules leave `page.css` for the domain sheet, the class-namespace pins for tasks come down by two, and the CONTEXT.md "Task strip" term, the Dock comment and two dot comments are rewritten to match.
- `refactor(ui-web)`: the tasks store's `hover` key and the row's `.hl` class go. They existed so the strip's chip and the list row could light each other up; with one pill there is no other side, and a row lighting itself through a store round trip is what `.sarow:hover` already does. The row's class is a plain attribute again, so the class-namespace tool counts `.task` on the local list instead of the expression list: one pin up, one down, nothing added (the tool's own header documents this kind of move).
- `fix(ui-web)`, found by the browser regression for the pill: a spawn's pending frame files a row under the manager's `task_id`, and the running frame renames it onto the record id. A `tasks.list` read landing between the two brings the same spawn in under its record id, and the store's merge keeps the pending row beside it; the running frame then updates the record row and leaves the pending one standing until the next full read. On the served page the pill counted three tasks where two ran. The frame that names both ids now folds the pending row into the record row.

Targets `refactor/ui_web_architecture`, not `main`.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

All from the repo root of the branch checkout, against the merge base `refactor/ui_web_architecture` @ cbaa057d:

- `npm test --prefix ui-web` -- 191 files, 2494 tests passed (base: 2492; the two new cases are the strip's live repaint and the reducer's fold). Includes the gates: `check-class-namespace` OK at 260 unprefixed / 57 in expressions (base 261 / 59), `css-one-owner`, `i18n-keys`, `island-lang`.
- `npm run type-check --prefix ui-web` and `npm run lint --prefix ui-web` -- clean (the 4 `exhaustive-deps` warnings are the base's, in files this branch does not touch).
- `npm run build --prefix ui-web && python3 ui-web/build.py && node ui-web/scripts/check-page.mjs` -- both boot snapshots match their goldens (235 nodes), page check OK.
- `npm run gen:check --prefix ui-web` and `npm run --prefix ui-tui lint:i18n` -- the RPC client and the TUI's generated catalogue are current.
- `make lint-python`, `make test-python`, `pre-commit run --from-ref origin/refactor/ui_web_architecture --to-ref HEAD`, `scripts/check_commit_messages.py`, `scripts/check_source_language.py`, `scripts/check_large_files.py` -- all green (24487 python tests passed, 109 skipped); the TUI lane (`npm run lint && npm run lint:rpc && npm run type-check && npm test && npm run build` in `ui-tui/`) passes with the regenerated catalogue.
- Served page, real browser (playwright against a gateway started from this checkout on an isolated `RAVEN_HOME`): two slow `spawn`s in one conversation show one pill "2 running" (its Chinese counterpart in the zh build) whose title lists both names; with the desk shut, a click opens it on the tasks tab with the two rows and opens no pane; with the desk on the deliverables tab, a click brings the tasks tab forward; after `config.set language=en` and a reload the pill reads "2 running"; once both spawns settle the pill is gone. 6/6 steps pass on the three-commit build. The reducer fix in commit 3 came out of this run: on one attempt the pill briefly read 3 where 2 ran.

## Risk

User-visible: while tasks run, the composer shows one pill "{n} running" instead of up to three named chips; clicking it opens the desk's tasks tab rather than one task's pane (the pane is one more click away, on the row). The tasks tab, its running-count badge and the rows are unchanged. Rollback is reverting the two commits; nothing outside `ui-web/`, the catalogue and its generated TUI table is touched.

## Related Issues

N/A


gloryfromca (commented):

Reviewed the refreshed `github/refactor/ui_web_architecture...HEAD` diff and the relevant task store, live reducers, desk opener, language subscription, CSS, and history. I also checked the repository rules and Web UI architecture vocabulary/gates, backward compatibility with the target branch's two newer UI commits, and that the changed tests replace the old behavior with coverage for the new contract rather than weakening assertions.

Verification:
- Focused task suites: 113 tests passed.
- Full branch UI suite: 191 files, 2494 tests passed.
- Synthetic merge with the current target: 191 files, 2495 tests passed.
- Type check and generated-client check passed.
- ESLint passed with four pre-existing hook warnings in untouched files.
- Source-language gate and `git diff --check` passed.

gloryfromca (approved):

PR #608, opened 2026-09-21
feat(*): type the approval request and let no approval wait forever

gloryfromca:
## Summary

The approval prompt used to reach the page as one English sentence plus a truncated argument line, with a 30 s countdown the web never drew and a 35 s deadline that silently retired the sheet. This makes the request typed and lets the sheet be drawn from facts instead of prose:

- `approval.request` now carries `kind` (shell.exec / file.write / mcp.call / unknown), `family` (the shell command family that words the prompt), `origin` (the main agent or a named sub-agent) and `evidence` (the tool's own account of the call: the command and its directory, a path and the diff the write would make, an MCP tool and its input). Each tool describes itself through the new `Tool.approval_kind` / `Tool.approval_evidence`; the gate only reads them. `created_at`, `expires_at` and `action_digest` leave the frame: nothing read them.
- No pending approval can block forever, and none is lost silently. The broker serializes requests per conversation (a background sub-agent asking while the main agent's prompt is up waits its turn instead of replacing it), keeps a day-long floor instead of the 35 s ceiling, and answers `approval.pending` so a reloaded or reconnected page draws the open requests again; a request in another conversation lights the line above the composer. The TUI countdown and its expiry state are removed.
- The runtime declares the seven shell command families before it builds the loop, so the terminal, the page and the channels word a delete as a delete, the way the ACP editor already did.
- Web sheet redesign: title and why-line by family, evidence block by kind, three answers (deny, always with the suggested rule shown, allow once), a landed line after the answer, undo for a saved rule (`approval.revoke`), a sentence field after a refusal that goes on as the next message, and a line above the composer when another conversation is waiting.
- A rule that reached the config file gets no session grant beside it, so taking the rule back really does mean being asked again -- unless the user's own stricter rule outranks it (the strictest matching rule wins), in which case the session grant stays and the identical call is not asked twice.
- The landed line says so when the engine did not take the answer (`approval.respond` returned `ok: false` or the socket dropped); a digit typed into a field is text, not an answer.
- Recovery and undo are scoped to the surface that asked. `approval.pending` answers a connection with the conversations it owns plus the ones nobody owns, so a second surface cannot read an opaque id it was never shown and answer with it. The undo names the answer rather than the rule's text, and waits for the gate to report what its write actually did: measured, `approval.respond` answers the page before the gate persists, so a quick undo used to find nothing, report a failure, and let the rule land and stay.

Deviations from the design record (`.work_context/permission_mode/11`): `cwd` rides inside `evidence` for shell.exec rather than as a top-level field; the notification frame stays hand-typed on both front ends because the OpenRPC schema only covers methods.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- Rebased twice onto the target as it moved, most recently onto the tip two commits further on. The only conflict in either rebase was the generated `ui-web/src/rpc/generated.ts` (both sides added RPC methods), resolved by regenerating from the merged `rpc-schema/openrpc.json`. Everything below was re-run after the latest rebase.
- `uv run --frozen --python 3.12 --all-extras pytest -q` -- full suite. 4 failed, 24519 passed, 111 skipped in 5:02. The four are pre-existing local failures (oncall launcher tool face, token budget ceiling, code launcher tool face, ppt image search key notice); they fail on the base commit too and none is touched by this change.
- `cd ui-web && npx vitest run` -- 192 files / 2511 tests pass, including the 40 gate files; `npx tsc --noEmit -p tsconfig.json` clean; `npm run gen:check` in sync (194 methods); `node scripts/check-class-namespace.mjs` clean; `npm run build && python3 ui-web/build.py` green (both booted-DOM goldens match at 236 nodes).
- `cd ui-tui && npx vitest run` -- 143 files / 2068 tests pass; `npm run type-check`, `npm run lint:rpc`, `npm run lint:i18n` clean.
- `scripts/check_source_language.py` and `scripts/check_large_files.py` clean over this PR's range; `npx commitlint` over the same range exits 0.
- Two independent review passes (correctness, completeness) before the second real-host round; their findings are folded in, and the ones that changed behaviour are listed under Risk.
- Real host, two rounds: `raven serve` on an isolated `RAVEN_HOME` in ask mode with deepseek-v4-pro, driven with playwright-cli; the TUI from source in tmux against the same home. Delete (`rm proj/coverage.xml proj/coverage.json`): titled as a delete, allowed once, files gone, landed line leaves after 4 s. Edit (`edit_file notes.md`): diff shown, refused, note typed after the refusal arrived as the next message, file untouched. `glab mr list ...`: always-allow wrote `permissions.tools.exec["glab mr list *"]`, undo removed it. A pending approval in another conversation shows the waiting line on a new task; clicking it switches back. After a reload the sheet is gone and stop releases the turn.
- Round two: a pending approval survives a reload (the sheet is drawn again, or the waiting line appears when the conversation is not on disk yet); a main-agent request and a background sub-agent request two seconds apart arrived one after the other, never side by side (gateway log 22:59:50 / 22:59:52, second sheet only after the first was answered); the TUI prompt shows no countdown and is still standing 45 s later.
- The real-host runs found two things the unit tests could not: the dock passes pointer events through, so the waiting line needed `pointer-events: auto` before a real click reached it; and the waiting line filtered out a conversation whose very first turn was blocked, which is not on disk yet and so not in the list.
- Acceptance package with the case document, evidence and report page: `.work_context/approval_dialog/acceptance/` (local, not committed).

## Risk

- A pending approval no longer expires in 35 s; it waits for the person, up to a day. Reloading the page redraws it; another conversation's request is announced above the composer. Two requests in one conversation are answered one at a time, in order.
- Wire: `approval.request` gains four fields and drops three; the TUI reads none of the dropped ones any more. `approval.revoke` and `approval.pending` are new methods. ACP's `session/request_permission` is unchanged.
- A saved rule that a stricter user rule outranks no longer re-asks within the session (the session grant stays); before this change the same click would have re-asked on the next identical call.
- Contract tier version 28 -> 30 (`Tool` and `ApprovalResponder` shape, then `ApprovalOutcome` carrying the id of the request it answered).
- Rollback: revert the commits; no data migrates.

## Related Issues

N/A


gloryfromca (commented):
Blocking: approval recovery and rule undo can cross their intended authority boundaries.

I found two blocking defects, both marked inline. I reviewed the repository rules and canonical terminology, the full target diff, the approval callers and transport ownership model, the branch commit/history, backward compatibility of the responder contract, the changed tests for weakened assertions, and the Web UI architecture constraints.

Verification: `uv run pytest tests/test_approval_broker.py tests/test_rpc_approval.py tests/test_permissions_gate.py tests/test_filesystem_tools.py tests/test_shell_approval.py tests/test_rpc_spine.py tests/test_core_runtime_swap.py tests/test_config_update.py tests/test_acp_permissions.py tests/test_mcp_client.py -x` passed: 700 tests. The focused Web UI and TUI tests could not start because Node dependencies are not installed in this worktree (`vitest: not found`; `esbuild: not found`). `git diff --check github/refactor/ui_web_architecture...HEAD` passed.

raven/rpc/methods/approval.py
```diff
@@ -43,17 +52,46 @@ async def approval_respond(
...
+
+
+async def approval_pending(
+    params: dict[str, Any],
+    *,
+    approval_broker: "ApprovalBroker",
+) -> dict[str, list[dict[str, Any]]]:
+    """The requests still open -- one conversation's, or all of them."""
+    conversation_id = str(params.get("session_id") or params.get("conversation_id") or "")
+    return {"requests": approval_broker.pending(conversation_id or None)}
```
gloryfromca: Blocking: approval recovery must remain scoped to the connection that owns the conversation.

The page calls `approval.pending` with `{}` on every boot/reconnect, and this returns `broker.pending(None)`. On the shared `WsGateway` dispatcher that includes requests owned by other sockets. Each returned record contains both the opaque approval ID and conversation ID, so that socket can then call `approval.respond` and authorize another surface's command. This bypasses the `conversation_scoped` boundary used for the original request notification. Please filter recovery by connection ownership, or require a safe ownership re-claim before returning these requests.

ui-web/src/features/composer/approve.ts
```diff
@@ -128,111 +128,244 @@ export function open(
...
+     answer: the request had gone, or the socket dropped mid-call. The turn is
+     then still waiting, which the line has to say rather than "allowed". */
+  const undelivered = (): void => {
+    show({ text: t('gui.confirm.land.unsent') })
+    stay(LINGER_MS)
+  }
+  void Promise.resolve(sent).then((ok) => { if (ok === false) undelivered() }, undelivered)
+
+  if (choice === 'allow_always' && pattern) {
+    const onUndo = handlers.onRevoke
```
gloryfromca, line 338: Blocking: Undo must revoke exactly the grant produced by this answer, after that grant exists.

The Undo control is exposed immediately and revokes only by pattern, but `approval.respond` returns as soon as the broker future is resolved; `PermissionGate` persists the rule later. A quick Undo can therefore find nothing, after which the continuation writes the rule. Pattern-only identity also lets this remove a pre-existing allow rule when a broader strict `ask` rule caused the prompt and `_persist` merely kept the existing narrower allow. A server-issued grant/receipt (with response ordering through persistence) is needed so Undo cannot race the write or delete configuration this prompt did not create.

gloryfromca (commented):
Blocking: the two previously reported approval authority defects remain unchanged.

This revision only removes stale deadline references from comments and adjusts the related rail test comment. The pending-approval connection-scope issue and the pattern-only, unordered rule-revocation issue are untouched, so both original threads remain open. I found no additional defect in the new delta.

Coverage: repository rules and canonical terminology, the full target diff plus the revision delta, affected callers and transport ownership, branch history, responder backward compatibility, changed tests for weakened assertions, and the Web UI architecture constraints.

Verification passed: `git diff --check github/refactor/ui_web_architecture...HEAD`; 700 focused Python tests; 141 focused Web UI tests across 7 files; and 13 TUI approval round-trip tests. Node 23.10.0 produced engine warnings because the locked Vitest versions declare supported even-numbered Node releases, but the test processes completed successfully.

raven/rpc/methods/approval.py
gloryfromca: Confirmed, and fixed. The push side draws this boundary through `connection.conversation_scoped`; the pull side did not, and that asymmetry is the defect.

`approval_pending` now filters by connection ownership (`connection.owns_conversation`, the read side of `claim_conversation`). A connection is handed the requests for conversations it owns, and naming another socket's conversation in `session_id` does not lift that.

One deliberate inclusion: a request whose conversation nobody has claimed is still returned. Its original notification went out through `conversation_scoped`'s broadcast fallback, so every attached surface already saw it; hiding it on recovery would strand a question no other surface can answer either. The reason is in the docstring rather than implied.

Pinned by `tests/test_rpc_approval.py::test_approval_pending_hands_back_only_this_connection_s_own_requests`, which stands up two bound connections plus an unowned conversation and asserts the split both ways.

ui-web/src/features/composer/approve.ts
gloryfromca: Both halves confirmed, and the race is worse than I had judged it. I had reasoned that the gate's continuation is scheduled at `set_result` and therefore runs before the dispatcher writes the response. That reasoning was wrong, and a probe against the real gate and broker says so plainly:

```
ORDER: ['respond returned', 'persist']
VERDICT: RESPOND ANSWERS BEFORE THE WRITE
```

So an undo could reach the config before the rule was there, report a failure, and then the write would land and stay. Thank you for not letting the reasoning stand in for the measurement.

The fix is the receipt you described, in three parts:

- `ApprovalOutcome` carries `approval_id` (contract tier 30), so the gate knows which request it is reporting about.
- `PermissionGate._persist` now answers two questions rather than one: whether the rule is on disk (which still decides the session-grant fallback) and whether *this* call is the reason it is there. `allow_exec_pattern` already returned that `added` flag; it was being discarded. The gate hands both to the transport through `record_grant`, guarded by `getattr` because the ACP wire has no undo surface and no such method.
- `approval.revoke` takes `approval_id` instead of `pattern`, asks the broker what that grant wrote, and waits for the gate's word (`GRANT_RECEIPT_TIMEOUT_S`, 5s, a ceiling on the race rather than a poll). A receipt is read once, so a second undo cannot reach a rule a later prompt wrote under the same text, and a grant reported as `written=False` (the rule was already the reader's) revokes nothing.

Tests: `tests/test_rpc_approval.py::test_approval_revoke_takes_back_only_what_that_answer_wrote` covers the pre-existing rule and the double undo; `::test_approval_revoke_waits_for_the_engine_to_say_what_it_wrote` asserts the undo is still pending before the report arrives; `tests/test_permissions_gate.py::test_the_gate_tells_the_transport_whether_its_write_added_the_rule` covers the `added` flag both ways, and `::test_a_transport_with_no_undo_is_not_asked_for_one` covers the ACP shape and the allow-once case.

raven/rpc/methods/approval.py
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified in the implementation and regression test: pending recovery filters through `owns_conversation`, an explicit foreign session cannot cross the owner boundary, and the unowned fallback matches the existing broadcast path. This closes the original finding.

ui-web/src/features/composer/approve.ts
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified in the broker, gate, RPC method, and UI tests: the approval ID is carried into a one-shot receipt, revocation waits for persistence, and `written=false` protects a pre-existing rule. This closes both the ordering and identity parts of the original finding.

gloryfromca (commented):

Reviewed the full target diff and the ce431a5d6e49..2222e49bbe80 fix delta, surrounding broker/gate/RPC/UI callers, ownership and persistence history, backward compatibility across generated clients, and the repository rules in AGENTS.md and CONTEXT-MAP.md. I also checked that the new tests exercise the repaired boundaries rather than weakening existing expectations.

The two prior blockers are fixed and their threads are resolved: pending recovery is connection-owned, while revocation is keyed to the exact approval receipt and waits for the persistence result before removing only a rule created by that grant. The security-focused pass over all 14 changed files found no reportable issue.

Verification:
- `uv run pytest tests/test_approval_broker.py tests/test_rpc_approval.py tests/test_permissions_gate.py tests/test_contracts_two_tier_ledger.py tests/test_config_update.py tests/test_rpc_spine.py -x` — 385 passed
- `npm test -- --run src/features/composer/approve.test.ts src/state/session/pipeline.notify.test.ts` — 59 passed
- ui-web `npm run type-check` and `npm run gen:check` — passed
- ui-tui `npm run type-check` and `npm run lint:rpc` — passed
- `git diff --check github/refactor/ui_web_architecture...HEAD` — passed

gloryfromca (commented):

The branch was rebased onto the two latest target commits. A range-diff shows the three previously reviewed PR commits are patch-identical; the only new PR commit adds comments documenting the broker maps' process-lifetime retention and why a timed-out receipt is discarded. Those comments describe hidden lifecycle and ordering constraints and are consistent with the implementation.

Covered this round: AGENTS.md/CLAUDE.md and the runtime/web context vocabulary; the refreshed full target diff and revision delta; broker callers and commit history; RPC/generated-client compatibility; the existing regression assertions (none were weakened); and the connection-ownership plus exact-grant architecture boundaries. The two earlier threads remain resolved.

Verification:
- `uv run pytest tests/test_approval_broker.py tests/test_rpc_approval.py tests/test_permissions_gate.py -x` — 260 passed
- `npm test -- --run src/features/composer/approve.test.ts src/state/session/pipeline.notify.test.ts` — 59 passed
- ui-web `npm run type-check` and `npm run gen:check` — passed
- ui-tui `npm run type-check` and `npm run lint:rpc` — passed
- `git diff --check github/refactor/ui_web_architecture...HEAD` — passed

gloryfromca:
Superseded: the ui-web refactor branch this targeted has landed on main, so these five commits were replayed onto main and the work continues in a new PR against it. Nothing here is abandoned -- the new PR carries the same commits plus the fixes from this round's five-dimension review.

PR #609, opened 2026-09-21
feat(*): fold a large tool catalog by default, with the meta-pair reserved

gloryfromca:
## Summary

Make progressive tool disclosure safe to ship on, then ship it on.

`tools.toolSearch.enabled` now defaults to true. Below the compaction threshold
that costs a deploy nothing (the strategy drops `tool_search` from every request
while the catalog fits); above it, most schemas are withheld and reached through
`tool_search` / `tool_call`, so context stops scaling with tool count and an MCP
connect stops moving the prompt-cache prefix.

Two things had to land first, and both are in this PR:

1. **The meta-pair is reserved from `tools.disabledTools`.** Their absence from an
   array is how the fold reads "this request has no search route", and it answers
   by shipping every schema. So an off switch there did not slim a folded request,
   it unfolded it -- mid-turn, since the switch is read once per assembled array.
   A measured turn went from 12 tools to 61 between two model calls, discarding the
   whole cached prefix. They join the MCP resource and prompt meta-tools as names
   the loop owns for its lifetime; `tools.toolSearch.enabled` speaks for both.
2. **A folded request names the connected MCP servers whose schemas it withheld.**
   The runtime context named a server only while it awaited authorization or was
   still connecting, so a connected one read as nothing at all and a model that did
   not think to search reported a capability it held as missing. One line per
   server with its tool count, only while `tool_search` actually reaches the
   request.

Knock-on changes: the four packaged agents drop their now-inert disable rows, so
`tool_call` joins their faces and the ledger tests record it; each pins
`toolSearch.enabled` to false, which keeps a curated face off the fold. The
settings page stops drawing a switch that would write a preference nothing acts
on. The strategy also stops assuming the array's shape -- replay feeds recorded
arrays back through the same chain, and a corrupt recording used to end the turn
with a TypeError instead of reaching replay's own backstop.

## Type

- [ ] Fix
- [x] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest -q` (with SERPER_API_KEY unset): 24103 passed, 111 skipped.
- `uv run ruff check raven/ tests/`: all checks passed. `uv run ruff format --check`
  on every touched Python file: already formatted.
- `npm run type-check` and `npx vitest run src/features/plugins/capabilities-source.test.ts`
  in ui-web: clean, 6 passed.
- Both new guards were reverse-verified: with the reservation removed
  `test_disabled_tools_cannot_withhold_the_meta_tools` fails, and with the notice
  branch removed `test_a_folded_catalog_names_its_connected_servers` fails.
- Note for reviewers: the packaged-agent face tests read `SERPER_API_KEY` from the
  environment despite being hermetic, so `web_search` joins the face and they fail
  on a machine that has one. That predates this branch and is untouched here.

## Risk

User-visible: a deploy whose catalog passes `tools.toolSearch.compactionThreshold`
(50 by default) now sends most tool schemas only on demand. An existing config file
that spells `enabled: false` keeps today's behaviour with no action. Rollback is
that one setting, or `compactionThreshold` raised past the catalog size.

Also user-visible: `tools.disabledTools` entries naming `tool_search` or `tool_call`
stop taking effect and are reported once at startup, and the four packaged agents
now offer `tool_call`.

## Related Issues

N/A


gloryfromca (commented):

The default fold, reserved meta-pair, product opt-outs, and registry dispatch checks are internally consistent. One nonblocking inline note covers the folded MCP inventory notice using registered counts instead of the effective hidden set.

Covered: the full diff; AGENTS.md and CONTEXT-MAP.md constraints; tool registry, strategy, MCP lifecycle, runtime and direct-loop callers; commit history; backward compatibility and product-specific opt-outs; and whether the updated tests weakened prior assertions (they did not). There are no dependency or repository-asset changes.

Verification: all GitHub checks pass, including four unit shards, coverage, lint, page checks, and RPC contracts. A relevant local pytest selection passed 366 cases; its remaining optional raven-ppt face case could not load the uninstalled `raven_ppt` distribution, matching two environment-dependent launcher failures in the broader local attempt.

raven/agent/loop/mcp_glue.py
```diff
@@ -211,6 +220,12 @@ def _mcp_tool_notices(self) -> list[str]:
                     f"from this turn's definitions and register themselves when the handshake "
                     f"finishes, so they are available from a later turn without anyone acting."
                 )
+            elif folded and snap["connected"] and snap["tool_count"]:
```
gloryfromca: No blockers; suggestions only, and they are marked inline.

This condition uses the manager snapshot's `tool_count`, which counts every registered MCP wrapper rather than the tools actually hidden by this fold. With a catalog folded by other tools, a server whose tool is in `tools.disabledTools` gets a notice claiming `tool_search` can find it even though `ToolSearchController.search()` filters it out; a tool named in `toolSearch.alwaysVisible` is likewise described as absent while its schema is present. That gives the model false capability metadata and can send it through searches that cannot succeed. Please derive the notice from the server's currently offered schemas that the fold actually excludes, or distinguish the exposed/disabled cases.

gloryfromca: Confirmed and fixed in 333613e6a. Reproduced first: with the fold active and a
server registering three tools -- one plain, one in `tools.disabledTools`, one in
`toolSearch.alwaysVisible` -- the snapshot said three while `search()` returned
one and `tool_call` refused the disabled name.

The count is now the server's registered names that this turn's array carries and
the fold withholds (`get_definitions()` minus `visible_names()`), so a switched-off
tool and an always-visible one are both out of it, and a server with nothing
withheld gets no line at all. Pinned by
`test_the_count_is_what_the_fold_withholds_not_what_the_server_registered`, which
fails if the filter is removed.

gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at `333613e6a`: `get_definitions()` applies the current availability filters, subtracting `visible_names()` leaves only schemas the fold removes, and `names_from()` then scopes that set to the server named by the notice. Disabled and always-visible tools therefore no longer produce the false claim. The focused MCP/search suite passes all 118 tests, so this finding is resolved.

gloryfromca (commented):

The new delta correctly limits each connected-server notice to currently offered schemas that the fold actually removes. The prior finding is fixed and its thread is resolved; no new issue surfaced.

Covered on this revision: the full branch diff and new commit delta; AGENTS.md and CONTEXT-MAP.md constraints; MCP manager, registry, search controller, strategy, runtime-context callers, and relevant history; backward compatibility of disabled and always-visible tools; and the regression-test changes, which strengthen rather than weaken the prior assertions. No dependency or asset constraints are involved.

Verification: `uv run pytest tests/test_agent_loop_mcp_sync.py tests/test_agent_loop_tool_search.py tests/test_tool_search.py -x` passed 118 tests. All completed GitHub checks are green; unit shard 4 was still running when this review was posted.

gloryfromca:
Closing: main moved on (the ui-web rebuild landed), and this branch is based on the old tip. The same four commits are replayed onto the current main in a new PR; the only change is the settings-page edit, dropped because ext.list already reports both meta-tools as builtin there and the rebuilt page draws a fixed control for those.

gloryfromca:
Superseded by #616, which replays the same work onto the rebuilt main.

PR #610, opened 2026-09-21
fix(*): drop the playbook chip from the desk task row

0xKT:
## Summary

The desk's tasks list row drew `TaskRow.playbook` -- the playbook library name the backend recovered from the first node id's prefix -- as a grey monospaced chip between the task name and the error tag. Next to a task it reads as a node id (`ai-competitor-compare`), and the prototype's row never had it: a dot, the summary, the duration line and the error tag.

Three commits:

1. `fix(ui-web)`: remove the span, its `.tksrc` rule and the never-rendered `.tsrc` rule in `page.css` the chip's styling was borrowed from; the test that asserted on the chip becomes a guard that a playbook run's row shows the summary alone.
2. `refactor(*)`: with the chip gone the `playbook` field on `tasks.list` had no reader in either UI, while every call still listed the playbook library to derive it. The schema property, the pydantic field, the derivation (`_derive_playbook`, `_playbook_names`, the tag regex) and its four tests go; both generated clients are regenerated from the contract; the offline fixtures stop sending the field. A playbook run still tells itself apart on the wire by its node ids, which the executor prefixes with the playbook's name.
3. `test(ui-web)`: the guard asserts the name line holds a single element, so any second element beside the name fails it, whatever feeds it.

Which playbook a run came from stays readable on the `load_playbook` receipt in the conversation, and on the instance handle the task pane's work-order tab shows.

## Type

- [x] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [ ] Refactor
- [ ] Other

## Verification

- `uv run pytest tests/test_rpc_tasks.py tests/test_rpc_schema_match.py -q` -> 434 passed
- `uv run ruff check` and `uv run ruff format --check` on the three edited Python files -> clean
- `npm test --prefix ui-web` -> 191 files, 2492 passed (domain suites and the gate suite)
- `make lint-ui` -> generated.ts matches the contract (192 methods); eslint 0 errors (4 pre-existing warnings in connections / cron / subagents); tsc clean
- `npm run --prefix ui-web build`, `python3 ui-web/build.py`, `node ui-web/scripts/check-page.mjs`, `node ui-web/scripts/check-css.mjs`, `node ui-web/scripts/check-class-namespace.mjs` -> boot-snapshot 235/235, every check OK
- `npm run lint:rpc --prefix ui-tui`, `npm run type-check --prefix ui-tui`, `npm run lint --prefix ui-tui`, `npm test --prefix ui-tui`, `npm run build --prefix ui-tui` -> generated.ts in sync; tsc clean; 0 errors; 144 files, 2074 passed; built
- `uv run pre-commit run --files <the nine files of commit 2>` -> all hooks passed
- Mutation check on commit 1: with the chip line put back, the rewritten test fails (`expected 'Cross-check quotesnightly-checks' not to contain 'nightly-checks'`)
- Real gateway (`raven web` on this branch, the user's home): after commit 1 the row of the playbook-dispatched E2E run reads `Turn the incident note into an archived record... / 1m32s / 6 files` with no `.tksrc` element; after commit 2 `tasks.list` for that session answers rows without a `playbook` key and the row still draws without a chip

- [ ] User-facing docs or screenshots are updated when needed (the design note's field list drops `playbook`; no doc names the chip)

## Risk

User-visible: the playbook name no longer shows on the task list row. Wire: `tasks.list`'s `TaskRow` loses the optional `playbook` property; the only clients of that contract are the two UIs in this repo, both regenerated here, and the ACP surface does not expose `tasks.list`. Rollback: revert the two commits.

- [x] Security impact considered (display and a dropped read-only field)
- [x] Backward compatibility considered (see Wire above)

## Related Issues

N/A


gloryfromca (approved):

gloryfromca (commented):

Reviewed the exact three-commit PR diff from its recorded base, plus the wider requested local target diff and relevant task-row callers/history. I covered AGENTS.md, CLAUDE.md, CONTEXT-MAP.md and the Web UI task terminology; the RPC model/schema and both generated clients; the task list renderer and fixtures; backward compatibility of removing the unused optional field; and whether the replacement test was weakened. The removal is internally consistent, the old clients treated the field as optional, and the shape assertion now guards against any second element beside the task name.

Verification: `uv run pytest tests/test_rpc_tasks.py tests/test_rpc_schema_match.py -q` (434 passed); focused `TasksPage` and live reducer Vitest suites (91 passed); `npm run gen:check`; `npm run type-check`; and `node scripts/check-class-namespace.mjs` all passed. The initial frontend attempts could not start before locked dependencies were installed; after `npm ci`, the checks above passed.

PR #611, opened 2026-09-21
refactor(*): the dag card hands its graph to the desk task pane

0xKT:
## Summary

The orchestration card in the transcript (a `run_subagent_dag` or `load_playbook`
row, expanded) drew a graph of its own and, under it, a node panel whose "open the
run" link raised the old agents-panel record window. The desk's task pane already
draws the same graph and reads the same `dag.node` record, node by node, so the
card was a second renderer for work the desk owns. The prototype's card is the grid
alone, with the task cell as the door to the run. This PR makes the card that.

Two commits, the second deleting only what the first killed:

1. **The card hands its graph to the desk.** The card keeps its grid (task, scale,
   playbook, state, cost, run id, replanned-into) and nothing else. The task cell is
   a door -- `role=button`, a title, a standing arrow -- whenever the run has an
   id, and opens the run's task pane on the desk. `openDagRun` goes through the
   tasks seam: `openRun` when the tasks store holds the row, one `tasks.list` read
   when it does not, the tasks tab when neither answers (a branched conversation
   replaying its parent's delivered row lands there). The delivered row already
   went through `openDagRun` and follows. The node panel, the `openDagNode` seam,
   `features/dag/open.ts`, the 21 catalogue keys only the panel read and the
   `dag-renderer` gate go.
2. **The renderer's default box goes.** The desk's board is `DagGraph`'s only
   caller and hands it its own node card, so the default SVG box, the card
   surface, the label trimming, the mark table and the per-node clock had no
   reader left.

User-visible changes besides the door: the first failed node is no longer opened
unasked inside the card (the desk pane's failure banner names it, one click away);
a node's internal dependencies are read off the board's edges rather than a list;
the node id, which the deleted panel spelled out, rides on the desk node panel's
title.

Left for a follow-up, deliberately: `features/dag/mount.ts` and `store.ts` (the run
state `state/session/resume.ts` still writes and nothing reads) -- deleting them
means either a new row on `domain-shape`'s shrink-only exceptions or dissolving
`features/dag/` into its two remaining consumers, which is a decision of its own.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Docs
- [ ] CI / tooling
- [x] Refactor
- [ ] Other

## Verification

- `npm test --prefix ui-web` -- 188 files, 2464 tests, all pass (three suites
  fewer than the base: the two deleted gates and the deleted label suite)
- `node ui-web/scripts/check-class-namespace.mjs` -- OK; the transcript and dag
  pins lowered to what the tool prints
- `npm run --prefix ui-tui lint:i18n` -- generated catalogue up to date
- `npm run --prefix ui-web build && python3 ui-web/build.py` -- boot-snapshot OK
  (235 nodes match golden)

Driven on a real host (`raven serve` from this branch on an isolated `RAVEN_HOME`
built from a real config, deepseek as the default model, every channel off), with
a playwright driver against the served page:

| Step | Result |
| --- | --- |
| Expand the card of a running three-node graph | grid only, no graph, no node panel; task cell is `role=button` with a title and the arrow; a long title is cut and the arrow stays in view |
| Click the task cell | exactly one desk pane, `task:dag:<run_id>`, headed by the run's summary, showing the board |
| Click a node on the board | the desk's node panel with the node's record |
| Click the task cell again | no second pane |
| Click the delivered row after the run settles | the same pane |
| Reload, expand the restored card, click | the same pane |
| A graph with one node on a lane that cannot reach its model | the card's state row counts `1 failed`; the desk pane shows the failure banner before any node is picked; the banner opens the failed node |
| The board before and after commit 2 | screenshots identical (node cards, edges, lane frames) |

## Risk

Web UI only; no server or wire change. The transcript card loses its graph and
node panel on purpose; every fact they showed is on the desk's task pane, reached
from the card's task cell and from the delivered row. `features/dag/mount.ts` keeps
running unchanged. Rollback is reverting the squash commit.

## Related Issues

N/A


gloryfromca (approved):

ui-web/src/features/transcript/source.ts
```diff
@@ -72,31 +69,26 @@ export function renderHistory(messages: HistoryMessage[]): void {
...
-      /* With what the node is FOR, not only its id. The run holds every node's
-       summary and this row handed over the slug alone, so a pane opened from
-       the trail was headed by a name for the machine while the same node opened
-       from the sheet was headed by what it did. */
-      const n = d && d.nodes && d.nodes.get ? d.nodes.get(last) : null
-      dagOpenNode(id, { id: last, summary: (n && n.node_summary) || null })
-      return
-    }
+    ds('tasks').one('dag', id)
+      .then((row) => { if (row) openDeskTask(row); else openDeskTab('tasks') })
```
gloryfromca, line 85: No blockers; suggestions only, and they are marked inline.

Nonblocking: please guard this one-shot lookup against a conversation switch. `TasksSource.one()` addresses the conversation that is current when the call starts, but this callback opens its result unconditionally. If the reader clicks the DAG door and switches conversations while `tasks.list` is pending, the old conversation's task pane is added to the new conversation's desk. The tasks store's `refresh()` and `reconcile()` paths already capture the session and discard exactly this kind of stale answer; this path should do the same, ideally with a delayed-response regression test.

gloryfromca (commented):

The refactor has one task renderer after the transcript card hands off through the tasks seam, and the removed DAG renderer surface, CSS, catalogue keys, and tests have no remaining callers. The new summary/door behavior and the narrowed caller-supplied `DagGraph` contract are covered well.

I checked AGENTS.md/CLAUDE.md and the Web UI context, the full diff, surrounding callers, commit history, compatibility of the removed exports and i18n keys, and that the tests were replaced with state/door assertions rather than merely weakened. I found one nonblocking stale-response race in the fallback opener, marked inline.

Verification: `npm test` (188 files, 2,464 tests passed); `npm run type-check` (passed); `npm run lint` (passed with four existing hook warnings in untouched files). `git diff --check` passed.

PR #613, opened 2026-09-21
fix(*): settle the lint and contract drift the replay left behind

gloryfromca:
## Summary

CI went red on main after #612 landed. Four files, three causes, all of them
drift the replay left rather than anything the rebuilt branch decides:

- `probe.py` still imported `verify_agent` for the preset branch that the
  replay removed, once a preset started recording its capability snapshot like
  any other row. Ruff calls the import unused, and it is.
- `manager.py` and `test_subagent_probe.py` carry blank-line spacing that
  `ruff format` rewrites. The merge resolutions produced those two by hand.
- `ui-web/src/rpc/generated.ts` predated main's `cancelled` delegation status,
  so the committed bindings stopped matching `rpc-schema/openrpc.json`.
  Regenerated with `npm run gen`, which adds `cancelled` to the three
  delegated-status unions.

## Type

- [x] Fix

## Verification

- `make lint-python` -- exit 0 (ruff check and ruff format, 2034 files).
- `make lint-types` -- exit 0. `make lint-imports` -- 10 contracts kept.
  `make lint-deps` -- no dependency issues.
- `npm run gen:check` in ui-web -- generated.ts matches the contract, 192 methods.
- `npm run type-check` in ui-web -- exit 0. `npm run lint` -- 0 errors.
- `npm test` in ui-web -- 188 files, 2464 tests, all pass.
- `uv run pytest tests/test_subagent_acp.py tests/test_subagent_probe.py
  tests/test_subagent_manager.py` -- 465 passed.

The fourth red on that push, `check_source_language.py`, is not addressed here
and does not need to be: it fired on the four CJK lines in
`ui-web/src/rpc/fixtures/sessions.ts` that arrived with #591, which were added
lines relative to the pre-merge main. They are carried lines now, so the gate
passes on this range (`check_source_language.py 52eb09b25..HEAD` exits 0) and
on every push after it. Whether that text should be translated or covered by an
exemption zone under AGENTS.md section 1.3 is a separate call.

## Risk

`generated.ts` gains a status value the schema already declared; nothing reads
a narrower union. The other three changes are an unused import and whitespace.

Rollback: revert this commit; main returns to the state that fails the three
checks above.

## Related Issues

Follows #612.


0xKT (approved):

gloryfromca (commented):

Reviewed the complete four-file delta at `1972b8180e25`. The unused import removal is safe because `_test_acp` calls `record_capabilities`, which retains the local `verify_agent` import. The three generated status unions now match both the Python wire models and `rpc-schema/openrpc.json`; this is an additive compatibility correction, and `gen:check` confirms the checked-in client is deterministic. The remaining edits are formatter-only, and no tests were weakened.

Coverage included AGENTS.md/CLAUDE.md and the relevant context map, the diff and surrounding callers, the schema/model/generator history, backward compatibility, test integrity, and the UI-to-runtime wire-schema architecture boundary.

Independent verification: Python lint and format passed across 2034 files; the focused subagent suite passed (465 tests); `ty`, all 10 import contracts, `deptry`, and the source-language gate passed; web RPC generation, type-check, and lint passed (zero errors, four pre-existing warnings); and all 188 web test files passed (2464 tests). GitHub merged the PR at this same head while the review was in progress.

PR #614, opened 2026-09-21
refactor(plugins): serve the four EverOS memory roles from raven's own config

gloryfromca:
## Summary

EverOS's four memory roles -- llm, embedding, rerank and multimodal -- lived in
`~/.everos/raven/everos.toml`, a file raven wrote and EverOS read once at startup.
Three defects followed from that arrangement and none was fixable in place:

- a model changed on the settings page did not take effect until something
  unrelated restarted the service, with nothing on screen saying so;
- the rerank section needs a request shape (`deepinfra` / `vllm` / `dashscope`)
  that only the wizard's own vendor table knew, so reranking configured from the
  settings page posted to the wrong path against DeepInfra and SiliconFlow;
- an EverOS root the user manages is read-only to raven, so those users had no
  way to configure a role at all.

raven now records a **pin** per role -- a model and the vendor serving it, never a
credential -- and resolves the address and key from that vendor at the moment they
are needed. Rotating a key is one edit, and every role on that vendor follows.
The values reach EverOS as `EVEROS_*` variables, emitted on every spawn and bound
into raven's own process so the in-process multimodal tool reads the same models
the service does. `everos.toml` keeps its runtime knobs, which EverOS merges per
key; raven writes only `[api]` there now.

**A save applies itself.** The spawn precheck runs before anything is stopped, so a
restart that cannot succeed leaves the old server serving; a stop that does not
finish never falls through to the spawn, which would otherwise adopt the old
server and report success for a configuration that never took. The outcome
reaches every session waiting on it as `memory.health` -- success included, since
that frame is the only thing that clears the standing banner.

**Upgrades migrate themselves.** An install upgrading into this still has its roles
in the file. The migration runs on the first memory start: it names the vendor
from the address and the key, creates the provider row when raven carries none,
and leaves a role it cannot name unset with a notice rather than guessing. It is
idempotent and never edits `everos.toml`. Measured against the real shape on a
maintainer's machine, where two of the four roles point at DeepInfra, a vendor
raven carries no provider spec for.

Key decisions, recorded in `docs/specs/2026-09-21-everos-role-config-design.md`:

- the pin holds no credential, so nothing on the settings wire carries one in
  either direction;
- an operator who exports `EVEROS_<ROLE>__*` still outranks raven, and the slot
  says so rather than accepting a save that would never take effect;
- the rerank request shape comes from the vendor table where it is known, and is
  recorded on the role only for a self-hosted endpoint no table can answer for;
- the migration is scheduled from the plugin, not from raven's config migrations,
  because the host may know this plugin only through the plugin contract.

The last commit is unrelated to the rest: `make lint-python` fails on `main`
without it (one import bound twice in the same function, two files the formatter
would rewrite). It is carried here so this branch's CI can be green.

## Type

- [x] Refactor

## Verification

```
uv run --all-extras pytest -q
# 24517 passed, 4 failed, 111 skipped -- the 4 are pre-existing, confirmed by
# checking out the branch base in the same environment and re-running them there

uv run pytest <the 28 files that reference the changed modules> -q
# 1516 passed, 1 skipped

npm --prefix ui-web test          # 2496 passed, 191 files
npm --prefix ui-web run gen:check # generated.ts matches the contract (192 methods)
npm --prefix ui-tui run lint:rpc  # in sync
npm --prefix ui-tui run lint:i18n # catalogue up to date
uv run ruff check raven/ plugins-dist/ tests/        # All checks passed
uv run ruff format --check raven/ plugins-dist/ tests/  # 1835 files already formatted
make check-large-files            # exit 0
```

A real gateway was started on a fixture built from a maintainer's own config in
the pre-upgrade shape. The migration moved all four roles, created the
`deepinfra` provider row, resolved reranking to `/v1/inference` rather than the
chat endpoint, and left `everos.toml` byte-identical (sha256 unchanged). That run
is also what found the defect described in the next section.

The remaining real-host acceptance cases in
`docs/specs/2026-09-21-everos-role-config-acceptance.md` have not been run yet;
they are the next step and will be reported on this PR.

## Risk

User-visible changes:

- the four role slots on the settings page now show the vendor as stored rather
  than deriving it from an address, and offer every vendor that can serve the
  role;
- a save restarts the memory service, which takes a few seconds and reports its
  outcome in the standing banner;
- a role whose endpoint comes from exported `EVEROS_*` variables is read-only,
  and so are all four on a root the user manages;
- reranking on a self-hosted endpoint asks for its request shape once.

Rollback: revert the branch. The migration only ever adds to raven's config and
never edits `everos.toml`, so an install that has migrated still has its original
file and falls back to it.

One defect worth naming, because it was invisible to every unit test: an
embedding role with no width pinned emitted `EVEROS_EMBEDDING__DIMENSIONS=""`,
which EverOS's own settings model refuses as an integer -- the server exited
before serving anything. Fixed, and the case that catches it now builds EverOS's
real `Settings` from the emitted environment in a subprocess, which is the layer
that owns the fact.

## Related Issues

N/A


gloryfromca (commented):
Blocking: two settings controls have no successful backend path.

I reviewed the full `github/main...HEAD` change and the docs-only follow-up at the current head; that follow-up does not affect either finding. Coverage included the AGENTS/CLAUDE and CONTEXT rules, the diff and commit sequence, role migration and backward compatibility, provider resolution, process restart, RPC/schema and generated clients, web callers, plugin/host architecture boundaries, and whether tests were weakened. The focused tests are green, but the two inline cases fall between them: one test proves a self-hosted rerank provider is offered without selecting it, and no test asserts that a required role lacks a clear control.

Verification:
- `uv run pytest tests/test_everos_config.py tests/test_config_update_everos.py tests/test_config_update_providers.py tests/test_everos_backend.py tests/test_everos_server.py tests/test_rpc_settings.py tests/test_rpc_contract_shapes.py tests/test_plugin_boundary.py -q`: 639 passed.
- `npm test -- --run src/features/settings/providers/Roles.test.tsx src/features/settings/source.test.ts`: 33 passed.
- `npm run gen:check` and `npm run type-check`: passed.
- `npm run lint`: passed with four pre-existing warnings outside this diff.
- Source-language, large-file, and `git diff --check` gates passed for the current head.

The first Python attempt ran before the declared EverOS dev extra was installed and failed collection with missing `raven_everos`/`everos`; after `uv sync --frozen --extra dev --dev`, the same suite produced the passing result above.

ui-web/src/features/settings/providers/Roles.tsx
```diff
@@ -147,7 +160,7 @@ async function setRole(r: Role, model: string, provider: string, typed: boolean,
     return src.load()
   }
   if (r.keys) { await src.set(r.keys[0], model); return src.set(r.keys[1], provider) }
-  if (r.everos) return src.everosSet(r.everos, { model }, provider)
+  if (r.everos) return src.everosSet(r.everos, model, provider)
```
gloryfromca, line 178: Blocking: the self-hosted rerank selection cannot be saved.

`roleProviders` deliberately offers self-hosted rerank providers (the new test names Ollama), but this call never supplies a rerank protocol. The RPC refuses a first save for exactly those providers at `settings_everos_set` unless `protocol` is one of `deepinfra`, `vllm`, or `dashscope`; its fallback only helps when the same provider was already pinned. Thus an ordinary picker selection is toasted as an operation failure and cannot create the advertised configuration. Either collect/pass the request shape on this path or stop offering providers the web UI cannot configure.

ui-web/src/features/settings/providers/Roles.tsx
```diff
@@ -217,6 +230,18 @@ export function RolePill({ role }: { role: Role }): JSX.Element {
...
+  if (locked) {
+    return (
+      <span className="settings-mpill settings-dim" title={t(`gui.settings.roles.locked_${locked}`)}>
+        <span className="settings-id">{val ? val.model : t('gui.settings.roles.unset')}</span>
+        {val && <span className="settings-pv">{providerName(s.snap, val.provider)}</span>}
+      </span>
+    )
+  }
   const dim = !val
   const clearable = !!val && role.id !== 'chat'
```
gloryfromca: Blocking: the embedding clear control always fails.

This marks every populated non-chat role clearable, including `embedding`, and renders the clear button below. This revision also makes `embedding` a member of `REQUIRED_ROLES`, so `settings_everos_set` rejects that click with "embedding is required ... and cannot be cleared." A configured embedding slot therefore exposes a control whose only result is an error toast. Align the UI with the new required-role contract (or change that contract if clearing is intended).

gloryfromca (commented):
Blocking: the two existing settings-path defects still have to change.

This revision rebases the same substantive change onto the newer main; the only net tree delta from the previously reviewed head is an unrelated three-line comment removal in `raven/agent/subagent/probe.py`. Re-checking the full diff and both affected call chains shows the existing findings still stand unchanged: `Roles.tsx` still offers self-hosted rerank providers without passing the protocol that `settings_everosSet` requires, and it still renders a clear control for embedding while the RPC rejects clearing every `REQUIRED_ROLES` member. The existing inline threads remain the complete blockers, so no duplicate findings are added and neither thread is resolved.

Review coverage: repository rules and domain terms; full diff and replay history; UI callers through RPC validation and plugin writers; legacy migration and backward compatibility; generated wire clients; restart lifecycle and plugin/host architecture boundaries; and whether tests were weakened or merely missed these interactions.

Verification on this head:
- Focused Python suite: 639 passed.
- Relevant web suite: 33 passed.
- Web RPC generation and TypeScript checks: passed.
- Source-language, large-file, and diff checks: passed.

ui-web/src/features/settings/providers/Roles.tsx
gloryfromca: Confirmed and fixed in 8eff440. You are right that the picker offered a configuration the write refuses: the first save of a self-hosted rerank provider needs a request shape no vendor table can name, and this page has nowhere to ask for one.

Taking your second option, narrowly. `roleProviders` now drops a self-hosted vendor from the **rerank** slot unless that vendor is the one already pinned there. The wizard is where the shape is asked, so:

- a fresh pick of such a vendor is no longer offered, rather than offered and refused;
- one the wizard configured stays pickable, so its model remains editable -- the RPC fallback reads the shape recorded on the role for that case;
- llm, embedding and multimodal are unaffected, since none of them needs a shape.

Pinned by `the rerank slot offers a self-hosted box only once something knows its shape` in `Roles.test.tsx`, which asserts both halves. The test you flagged as proving only the offer now targets `memllm`, where the offer is the whole claim.

The dropdown that would let this page ask for the shape is recorded as deferred work; with this change nothing advertises a path it cannot complete.

ui-web/src/features/settings/providers/Roles.tsx
gloryfromca: Confirmed and fixed in 8eff440. The page was mirroring the required-role contract and had drifted from it -- `optional: true` in its own `ROLES` table meant "may be unset", which is not the same question as "may be cleared".

Rather than re-mirror the list, the server now sends it: `describe_roles()` returns `required`, `SettingsEverosResult` carries it, and `clearable` reads it. A mirrored contract is one that goes stale again the next time `REQUIRED_ROLES` changes.

Pinned by `a required role has no clear control` in `Roles.test.tsx`: no clear button on the memory LLM or embedding slots, and one on rerank, so the case fails in both directions.

ui-web/src/features/settings/providers/Roles.tsx
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at 8eff440: fresh local/custom rerank providers are no longer offered by this page, while an already pinned self-hosted provider remains editable. The focused role tests cover both states and pass. This resolves the original save-path objection.

ui-web/src/features/settings/providers/Roles.tsx
gloryfromca: No blockers; this can merge as far as I am concerned.

Verified at 8eff440: the server now publishes its required-role set and the clear control consults that contract, so llm and embedding are not clearable while rerank remains clearable. The focused UI test and RPC contract checks pass. This resolves the original control-path objection.

gloryfromca (commented):

The two earlier failures are fixed at this revision: a fresh self-hosted rerank choice is withheld until the wizard has recorded its request shape, while an existing pin stays editable; and clearability now follows the required-role contract returned by the server. Both reviewer-owned threads are resolved.

Coverage included the AGENTS/CLAUDE rules and the mapped Runtime, Web UI, and TUI context constraints; the 4e24395..8eff440 fix delta in the context of github/main...HEAD; affected RPC/UI callers and provider auth-shape mapping; commit history; backward compatibility of the defaulted wire field; confirmation that tests were added rather than weakened; and the shared RPC-schema boundary across both generated clients.

Verification: `uv run pytest tests/test_everos_config.py tests/test_config_update_everos.py tests/test_config_update_providers.py tests/test_everos_backend.py tests/test_everos_server.py tests/test_rpc_settings.py tests/test_rpc_contract_shapes.py tests/test_plugin_boundary.py -q` (639 passed); focused Vitest role/source suite (35 passed); `npm run gen:check`; `npm run type-check`; source-language and large-file gates; and `git diff --check github/main...HEAD`.

gloryfromca (commented):

This revision rebases the already reviewed five-commit series onto current main (`d50a13c1c`). `git range-diff` reports all five commits patch-equivalent, and the trees differ from the prior reviewed head only by the upstream test-stub repair now present on main. The two earlier findings remain fixed and their threads remain resolved.

Coverage rechecked the repository rules and mapped Runtime/Web/TUI constraints, `github/main...HEAD`, callers and history, wire backward compatibility, test changes, and the shared RPC-schema architecture boundary. No tests were weakened in the rebase.

Verification: focused EverOS/RPC/backend suite (639 passed); `tests/test_subagent_probe.py` (58 passed); focused web role/source suite (35 passed); generated RPC contract check; TypeScript type-check; source-language and large-file gates; and `git diff --check github/main...HEAD`.

gloryfromca (commented):

Reviewed the `27f6fa2..8b41c14` delta in the context of `github/main...HEAD`. The new commit adds direct coverage for the role migration, restart, validation, and failure-reporting paths. Its only production change removes an obsolete missing-provider branch: the picker has one successful return path, that path always carries the selected vendor as `provider`, and every back/cancel path exits before the read. The earlier reviewer-owned threads remain resolved.

Coverage included the AGENTS/CLAUDE rules and mapped Runtime/Web/TUI constraints, the diff and affected callers, commit history, backward compatibility, confirmation that assertions were added rather than weakened, and architecture boundaries. The RPC wire schema and both generated clients are unchanged.

Verification: `uv run pytest tests/test_cli_onboard_commands.py tests/test_everos_backend.py tests/test_everos_config.py tests/test_everos_server.py tests/test_rpc_settings.py -q` (752 passed); source-language and large-file gates; and `git diff --check github/main...HEAD`.

gloryfromca (commented):

The current head is tree-identical to the reviewed `9f04ecab` revision; only the commit subject was amended. The acceptance update now matches A21 and the implementation: quick successive saves queue serialized restart chains, the later chain rereads the final configuration, and a third PID is expected when that later chain is needed to apply the second save.

Coverage included the repository rules and mapped Runtime context, the one-line `github/main...HEAD` diff, restart callers and history, compatibility, confirmation that no tests were weakened, and the existing RPC/plugin architecture boundary.

Verification: `uv run pytest tests/test_rpc_settings.py::TestTheSaveRestartsTheService::test_two_saves_run_one_after_the_other_and_the_last_one_wins -q` (1 passed); source-language and large-file gates; and `git diff --check github/main...HEAD`.

gloryfromca (commented):

The T6.5 correction is directionally right: the page has no credential-state element, so that half could not be observed there, and the implementation/tests still enforce the RPC contract.

Named nonblocking follow-up: after removing that half from T6.5, A18's `api_key_set` behavior has no real-host observation. T6.4 is marked `contract`, and this document says contract cases do not constitute acceptance; because Roles.tsx ignores `api_key_set`, a live gateway could report the wrong value while T6.5 still passes. The later statement that A18 is accepted on a real host is therefore no longer true. Either observe `api_key_set` from the live socket in a real-host case or state that this half of A18 remains contract-only. This does not block the implementation or merge.

Coverage included the repository rules and Runtime/Web context, the one-line `github/main...HEAD` diff, the role-card caller, A18/T6.4/T6.5 mappings and history, compatibility, confirmation that no tests were weakened, and the RPC/UI boundary.

Verification: the two focused RPC credential-state tests passed; the Roles UI suite passed (20 tests); source-language and large-file gates passed; and `git diff --check github/main...HEAD` passed.

gloryfromca:
The premise checks out, and it was a hole I opened. Removing T6.5's second half left `api_key_set` with only T6.4, which this document itself says does not constitute acceptance; `Roles.tsx` never reads the field, so nothing on a real host was watching it.

Taking the first of your two options, in `111c7de6`: **T6.6**, which reads the flag where it is computed rather than where it is displayed.

Run on a real host just now -- a seeded `RAVEN_HOME` with four roles pinned on `openrouter`, a real gateway (pid 79138) and its everos child (79142), over `ws://127.0.0.1:18898/rpc`, the same socket the page uses:

```
settings.everos, key present:
  llm         provider='openrouter'  api_key_set=True
  embedding   provider='openrouter'  api_key_set=True
  rerank      provider='openrouter'  api_key_set=True
  multimodal  provider='openrouter'  api_key_set=True

model.disconnect {"slug": "openrouter"} -> {"disconnected": true}

settings.everos, same connection:
  llm         provider='openrouter'  api_key_set=False
  embedding   provider='openrouter'  api_key_set=False
  rerank      provider='openrouter'  api_key_set=False
  multimodal  provider='openrouter'  api_key_set=False
```

and the role blocks on disk are byte-identical across the two readings -- no edit to any role, which is what the flag has to mean if it means anything. `model.disconnect` is the page's own Disconnect action, so the trigger is a real user path rather than a file edit behind the server's back.

The same pair also covers A18's other clause: `rerank.provider` reads `openrouter` in both readings and never `deepinfra` / `vllm` / `dashscope`.

The coverage note is corrected too: A18's real-host case is now named as T6.6, with the reason spelled out -- T6.5 only covers the rendered provider name, and the page does not read `api_key_set`, so without T6.6 a running gateway could report it wrongly unobserved.

The probe is re-runnable by anyone: `probe_ws.py <RAVEN_HOME> [read|disconnect]`, kept beside the other acceptance evidence.

gloryfromca:

Verified at 111c7de6. T6.6 now observes both halves of A18 through the running gateway: `api_key_set` changes after the real `model.disconnect` path without a role edit, and rerank continues to report the vendor rather than an EverOS request protocol. The coverage note now accurately identifies T6.6 as A18's real-host case, so the previously named follow-up is closed.

gloryfromca (commented):

T6.6 closes the only follow-up from the prior revision. It observes A18 at the live RPC boundary before and after the page's own disconnect operation, verifies that the credential flag follows the provider rather than a role edit, and confirms that rerank's `provider` remains the vendor rather than the EverOS protocol. The coverage table now correctly names T6.6 as A18's real-host case.

Coverage included the repository rules and Runtime/Web context, the `33b6a6c..111c7de6` delta in `github/main...HEAD`, RPC/UI callers and history, compatibility, confirmation that no tests were weakened, and the architecture boundary. The author's supplied real-host evidence was checked against the documented trigger and observation point. The earlier reviewer-owned threads remain resolved.

Verification: the two focused RPC credential-state tests passed; source-language and large-file gates passed; and `git diff --check github/main...HEAD` passed.

plugins-dist/everos-memory/raven_everos/config.py
0xKT: **Blocking:** two consequences of one decision. `embedding` is deliberately exempt from the ownership guard and is declared un-clearable, but both rules are enforced at one door each, and the other doors reach the same operations. Measured on `111c7de6` merged onto `0e3d6e41`, each with a control and a base control.

The decision itself is sound and well argued -- `set_role`'s comment for the exemption, and `REQUIRED_ROLES`' docstring for the rule, both say exactly why. What follows is what the other callers do with them.

## 1. `raven onboard` -> embedding -> "Skip" wipes the endpoint every knowledge base embeds with

The wizard offers "Skip" for any role whose table says `optional`, and embedding's does (`onboard.py:381`). This PR changed that branch from `clear_everos_section(section)` to `clear_role(section)` (`onboard.py:926`).

```
base    clear_everos_section("embedding")  ->  del data["embedding"]; _write_atomic(everos.toml)
head    clear_role("embedding")            ->  set_embedding_endpoint({"model": "", "provider": ""})
                                               -> raven's OWN top-level `embedding` block
```

Driven on the shipped `clear_role`:

```
CONTROL  clear_role("rerank")     embedding block unchanged: {'model': 'text-embedding-3-large', 'provider': 'openai'}
SUBJECT  clear_role("embedding")  embedding block = None     role_pin('embedding') = None
                                  log: "config/update: embedding endpoint cleared"
```

That block is the one endpoint `raven/knowledge/_embedding.py` reads (`:195`), so after the click nothing can be indexed or answered -- and the CONTEXT.md entry this PR adds gives that as the reason the pin lives there at all.

The same PR says this must not be possible. `REQUIRED_ROLES` (`config.py:802`) includes `embedding`, its docstring reads "`embedding` is what every stored vector was written under. Neither should be one stray click away; both are still editable, just not erasable", and the RPC door enforces it (`console.py:1387-1388`, "embedding is required for EverOS memory and cannot be cleared"). `clear_role` never consults `REQUIRED_ROLES`, and the wizard keeps its own older answer.

Both answers are tested, in this PR, against each other:

```
tests/test_rpc_settings.py:523  test_clearing_a_required_role_is_refused        for section in ("llm", "embedding")
tests/test_cli_onboard_commands.py:5697  test_skipping_embedding_names_what_it_costs   asserts embedding's skip_note
```

For the record, the four roles disagree in exactly one place -- the other three line up, which is what says this is a gap and not a policy:

```
role         wizard `optional`   REQUIRED_ROLES says required   agree?
llm          False               True                           yes
embedding    True                True                           NO
rerank       True                False                          yes
multimodal   True                False                          yes
```

## 2. Saving the embedding role on a user-managed root stops the user's EverOS server

`settings_everos_set` ends `return _everos_applied(agent_loop_factory, cost)` for every role (`console.py:1474`). The three plugin-slice roles never get there on an unowned root, because `set_role` raises first. `embedding` is exempt from that guard by design (`config.py:915-927`, "Not gated on owning the EverOS root, and the other three are"), so it falls through to the restart:

```
role         set_role on an unowned root                clear_role
llm          REFUSED (EverosRootNotOwnedError)          REFUSED   <- CONTROL
rerank       REFUSED                                    REFUSED
multimodal   REFUSED                                    REFUSED
embedding    got past the ownership gate                REFUSED
```

`_everos_applied` -> `_restart_everos_for_config` (`console.py:1267`) -> `restart_for_config_change` (`server.py:954`) -> `stop_for_reload` (`server.py:935`) -> `stop_pid`, which SIGTERMs whatever `lock_holder` names for that root -- the user's own server. The spawn that would replace it is then refused by `_require_owned` inside `set_everos_api`, so the host is left with no memory service and a banner that only says "refusing to write [api]".

This one is the PR's own written rule, three times: `everos_owned`'s docstring (`config.py:216`) -- "`False` means read-only reuse: record the address, never touch the config, **never start or stop the process**"; the same file's header ("a root the user manages is read-only -- raven records its address and nothing else"); and the design spec added in this PR, C13: "On a user-managed root raven neither writes the role config nor starts or stops the server."

`_require_owned` is placed at the write primitives "so a new caller cannot opt out" (`config.py:895`). The restart is a new caller that is not a write primitive, so it opted out.

Neither of these is an argument against the exemption or against `REQUIRED_ROLES`. Both are one rule enforced at one door while another door reaches the same operation -- `clear_role` consulting `REQUIRED_ROLES`, and the restart asking `everos_owned` before it stops anything, would close them where the rules already live.

## Also, smaller

- The design spec this PR adds defines and uses **"role block"**, which the CONTEXT.md entry the same PR adds puts on its `_Avoid_` list (`docs/specs/...-design.md:14` against `CONTEXT.md` "role pin", _Avoid_: "role block" and "role section").
- The vendor table still spells the rerank protocol `rerank_provider` (`config.py:769`) -- the word the CONTEXT entry beside it says to avoid, and the collision it says already configured reranking against the wrong endpoint.
- A18 in the acceptance doc states a criterion that is false about the code: `provider` legitimately carries `deepinfra` and `dashscope`.

## What I attacked and could not break

- **"Two homes, one reader" holds.** Every read of either home is inside `config.py` (10 sites), and the write side converges on `set_role`. The knowledge base's own `load_raven_config().embedding` is a separate reader of the same block, but it predates this PR and is precisely what the split exists for.
- **"Never a credential" holds, and then some.** `set_role` writes `{model, provider}` (+ `protocol` for rerank) and nothing else; `set_embedding_endpoint` filters against an explicit `allowed = {"model", "provider", "dimensions"}`, so a key cannot be smuggled through the dict -- and on every write it pops `baseUrl` / `base_url` / `apiKey` / `api_key` out of the block, so a retired shape's secret is removed rather than left to read as current.

---

`GATES OK head=111c7de6 tip=0e3d6e41 base=d50a13c1 ruff/lint-imports/commit/large/lang rc=0`; `test_everos_config` `test_everos_backend` `test_everos_server` `test_everos_discover` `test_config_update_everos` `test_config_update_providers` `test_plugin_boundary` `test_rpc_contract_shapes` -> 573 passed, 1 skipped; `test_rpc_settings` `test_cli_onboard_commands` -> 441 passed.

0xKT: Two corrections to my own finding 2, both narrowing it toward what it actually is.

**Reachability, stated properly.** I wrote it as "saving the embedding role", which reads like a click. It is not one today: `Roles.tsx:145` returns `'foreign'` when `snap.everos.owned === false` and `:252-259` renders a non-interactive span, and ui-tui has no EverOS screen at all. The reachable caller is the registered RPC method itself -- `settings.everosSet`, an L4 entrance every console client shares. That is a narrower door than I implied, and I should have said so.

It is also the exact door your own acceptance criterion names. T5.1: "a forced RPC write is refused with a message naming the root path ... the everos pid is unchanged after both a save attempt and a gateway start." So the case is not hypothetical reachability -- it is the one the doc asks to be refused, and it is not.

**Where the fix belongs, more precisely than I put it.** I said "the restart should ask `everos_owned` before it stops anything". The better location is the precheck you already built for this. `restart_for_config_change` (`server.py:954`) deliberately orders precheck-then-stop, and C11 in the design spec says why: "The spawn precheck runs before the running server is stopped, so a restart that cannot succeed leaves the old server up." `precheck_spawn` (`server.py:919`) asks `_require_llm_configured` and `_inotify_gate` -- and never asks `everos_owned()`. Ownership is precisely the condition that refuses the spawn three frames later, at `set_everos_api` -> `set_everos_section` -> `_require_owned` (`config.py:590`).

So the restart that cannot succeed is the one whose precheck cannot see why, and that is the one that stops the old server. One question added to `precheck_spawn` closes it where your own design already says the check lives.

Nothing above changes the severity or finding 1.

0xKT (approved):

